PR #5, opened 2018-02-27
Support Narrow by PM and Stream.

amanagr:


model.py
```diff
@@ -12,13 +13,54 @@ class ZulipModel(object):
...
         self.anchor = 10000000000000000
         self.num_before = 50
         self.num_after = 0
         self.menu = [
             u'All messages',
             u'Private messages',
         ]
+
+        ''' 
+        Stores all the messages, type: Dict[str, Dict[Dict[str, Any]]]
```
amanagr: I am thinking of storing messages in this fashion to allow easy accessing of messages when narrowing to a stream/PM.
PMs will be stored by the sender's email id and stream messages by their stream name.
I am currently not supporting multi-recipients.( will do later )
Do you think this is ok? @showell

helper.py
```diff
@@ -12,3 +18,26 @@ def wrapper(*args, **kwargs):
         thread.daemon = True
         return thread.start()
     return wrapper
+
+def classify_message(new_messages: NMSGL, classified_messages: CMSG=None):
```
amanagr: Look at this for more details.

model.py
showell: Yeah, it's a good idea.  We eventually should do this for our web client, too.

timabbott: You should store stream messages by stream ID, not name.  Otherwise, you'll have to later go through and fix a bunch of bugs related to renamed streams.

(And of course, you want to use the list other user IDs for a PM threads, to support group PMs; I'd recommend at least naming your variables based on that concept to help simplify future related to  replacing "sender id", which regardless is wrong because it'll miscategorize the messages you yourself sent)

amanagr: I will do it in the next commit since I will have to rework a lot of functions to implement it.
Thanks for the awesome advice. :)
(I will make next PR for improving logic for all the functions.)

PR #11, opened 2018-03-01
Revert "Zulip-Terminal: Support Scrolling and use Stream ID to store …

amanagr:
…stream messages."

This reverts commit 1e6d8003a68046da45a58905fcdf1c31bb383197.


amanagr:
Just to improve commit message.(... in commit message doesn't looks nice.)
I will keep this in mind from later on.

PR #23, opened 2018-03-05
hotkeys: Add `c` to compose a new stream message.

Balaji2198:
`c` hotkey will be useful when the user wants to compose a message in the same stream but in a different topic.
c-hotkey


amanagr:
Thanks @Balaji2198 merged!

PR #36, opened 2018-03-19
Make code base compliant with PEP-8

Balaji2198:
This PR fixes the failures we had while running the `pytest linter`


amanagr:
Thanks @Balaji2198

PR #63, opened 2018-04-02
Use pipenv for setup and add Travis CI integration.

amanagr:


.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: python
+sudo: required
+python:
+  - "3.4"
+  - "3.5"
+  - "3.6"
+install:
+  - sudo rm -vf /etc/apt/sources.list.d/*
```
punchagan: Why do we need this?

.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: python
+sudo: required
```
punchagan: Do we really need `sudo`?

Pipfile
```diff
@@ -0,0 +1,25 @@
...
+"pytest-pep8" = "==1.0.6"
+pytest-mock = "==1.7.1"
+pytest-cov = "==2.5.1"
+pudb = "==2017.1.4"
+snakeviz = "==0.4.2"
+mypy = "==0.560"
+
+[requires]
+
+python_version = "3.5"
```
punchagan: I'll need to look up what this `python_version` means.  Do we not support Python 3.4 or 3.6? If so, why are the configured [here](https://github.com/zulip/zulip-terminal/pull/63/files#diff-354f30a63fb0907d4ad57269548329e3R4)?

.travis.yml
amanagr: This refers to an issue of travis CI https://github.com/travis-ci/travis-ci/issues/8607.
I don't think we need this if we remove sudo.

Pipfile
amanagr: It recommend in the pipenv docs https://docs.pipenv.org/basics/#general-recommendations-version-control

punchagan: This would mean that [other Python versions won't work](https://github.com/pypa/pipenv/issues/1050) (including `3.4` and `3.6`)

You could [remove](https://docs.pipenv.org/basics/#specifying-versions-of-python) the `python_version` requirement, but then I guess users will have to explicitly specify that they want to use Python3 (and not 2).

Pipfile.lock
```diff
@@ -0,0 +1,297 @@
+{
```
punchagan: I think this file can be [removed](https://docs.pipenv.org/basics/#general-recommendations-version-control)

amanagr:
Agreed and update! Thanks, @punchagan for your valuable feedback.

Issue #18, opened 2018-03-03
Add feature to read settings from a file before running.


amanagr:
Just like vim uses `.vimrc` to store its settings, a setting file for zulip-terminal called `.ztermrc` or similar could be used to read settings like font, theme, etc.

mkaito:
Please also read the `zuliprc` file from a sensible location.

amanagr:
Do mean automatically reading the `zuliprc`  if placed inside the zulip-terminal folder in case no config file is specified?

mkaito:
I just looked at the contents of `zuliprc`. I figure this can be a section in the config file.

You do not want to rely on anything being in the `zulip-terminal` folder, or even being able to write to it. This would make packaging the program extremely hard otherwise.

As for config file placement, look at the [XDG Spec](https://specifications.freedesktop.org/basedir-spec/latest/). Or [this extract](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support), which is probably more readable.

timabbott:
If we want to do something quick here before we have time to do the XDG thing, the default would be to have it look at `~/.zuliprc` with a ` --config` command-line option to specify an alternative path.

mkaito:
You can just check environment for `XDG_CONFIG_DIR`, which can not be set, in which case assume it's `~/.config`. To this value, you append `/zulip/terminal.ini` or something like that. That's all there is to the XDG spec (for this particular use case)

KurtovicUM:
hey, @bolognam and I are exploring the addition of this feature, we'll keep you updated with progress

PR #72, opened 2018-04-13
Add feature to read settings from a file before running #18

bolognam:
@KurtovicUM and I wrote `get_config_files()` to get the config filenames (`zulip.ini` and `zterm.ini`) with error checking for the case where `XDG_CONFIG_DIRS` isn't set and `~/.config` doesn't exist as well as the case where `/zulip/zulip.ini` isn't in `XDG_CONFIG_DIRS` or `~/.config`. If `zterm.ini` has settings, `parse_zterm()` will get the theme from the file and otherwise use the default theme. The `zulip.ini` filename and the theme are then passed into `Controller().main()` instead of using command-line arguments.

### DIFF
--- a/run.py
+++ b/run.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 import argparse
+import configparser
 import sys
+from os import environ, path
 
 from zulipterminal.core import Controller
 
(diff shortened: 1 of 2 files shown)


amanagr:
@bolognam and @KurtovicUM, Thanks for working on this. 
Can you update the README file for this new setting? Also you should add #18 to you commit message.
Am I right to say that the `zuliprc` file needs to preset in `/etc/xdg/zulip/` as `zulip.ini` file?
If yes, I don't think we want that.
Maybe @timabbott and @mkaito have better thoughts on this.
Also we would love to have a default `zuliprc` file path to `~/.zuliprc`.
Adding usage examples in README can be a good Idea.

run.py
```diff
@@ -44,11 +46,36 @@ def parse_args():
...
+
 def main():
     """
     Launch Zulip Terminal.
     """
     args = parse_args()
+    if args.config_file:
+        zuliprc_path = args.config_file
+    else:
+        zuliprc_path = '~/zuliprc'
```
mkaito: This should be `~/.zuliprc`, but preferrably `$XDG_CONFIG_HOME/zulip/config.ini`, or whatever is the appropriate file extension for the used format.

`$XDG_CONFIG_HOME` should be expected to be empty, in which case it defaults to `$HOME/.config`.

PR #77, opened 2018-05-12
Allow simpler pip install

TheLonelyGhost:
Using `Pipfile` (a la pipenv) is great, but when it comes to packaging for redistribution it doesn't work so well. This PR adds support for `pip install .` or even [`pipsi install`](https://github.com/mitsuhiko/pipsi) options.

While I was in there, I translated some of the codified stuff from `./tools/run-mypy` into `setup.cfg` so one could run `mypy zulipterminal` or `./tools/run-mypy` interchangeably.

Thoughts? Comments?


timabbott:
@TheLonelyGhost thanks for working on this!  Looks like a nice set of cleanups, though I'm not going to merge (since I don't work on this part of the project(); hopefully @amanagr will get a chance to look soon :)

TheLonelyGhost:
@amanagr Fair point. Docs are usually the hardest part IMHO, so I'd rather get feedback on the functional changes before writing up user documentation for it.

How does the latest readme look? I made some pretty significant changes...

Second thing to note is that I rebased with the latest merged changes, which made my TravisCI commit unnecessary. I can add it back in if you'd prefer, but I figured better to not touch it than to change it for its own sake. 🤷‍♂️

amanagr:
Thanks, @TheLonelyGhost. This looks great so far. I would love some review on this from @punchagan.

setup.cfg
```diff
@@ -0,0 +1,41 @@
...
+mypy_path = mypy_stubs
+
+[mypy-urwid,zulip,pudb]
+ignore_missing_imports = True
+
+[coverage:run]
+branch = True
+omit =
+    zulipterminal/cli/*
+#     a/b.py,
```
punchagan: These commented lines can be removed

punchagan:
Looks good to me, though I'm not very familiar with `pipsi`.  @amanagr if both the user install and dev install seem to work with the instructions in the README, I think this is good to be merged!

setup.cfg
amanagr: I think these can be used as examples.

amanagr:
Thanks, @TheLonelyGhost for this awesome contribution.

PR #81, opened 2018-05-15
hotkeys: Add `n` hotkey to read next unread topic.

amanagr:


zulipterminal/helper.py
```diff
@@ -231,10 +242,13 @@ def classify_unread_counts(unread_msg_counts: Dict[str, Any])\
 
     for stream in unread_msg_counts['streams']:
         count = len(stream['unread_message_ids'])
-        if not unread_counts.get(stream['stream_id']):
-            unread_counts[stream['stream_id']] = count
+        stream_id = stream['stream_id']
+        stream_topic = str(stream_id) + '_' + str(stream['topic'])
```
showell: In Python you can use tuples as dictionary keys, which is usually more robust than hand-made keys like foo+'_'+bar.

amanagr: Thanks, I will look into it. :)

PR #84, opened 2018-05-21
user_search: Add feature to search for a user using `w` hotkey.

amanagr:


zulipterminal/helper.py
```diff
@@ -275,3 +276,17 @@ def classify_unread_counts(unread_msg_counts: Dict[str, Any])\
...
+
+
+def match_user(user: Any, text: str) -> bool:
+    """
+    Matches if the user full name, last name or email matches
+    with user or not.
+    """
+    keywords = user.caption.split()
+    keywords.append(user.email)
+    keywords.append(user.caption)
```
punchagan: It may be worth adding a comment on why the full name is appended to the list, after starting with the name and splitting it.

zulipterminal/helper.py
```diff
@@ -275,3 +276,17 @@ def classify_unread_counts(unread_msg_counts: Dict[str, Any])\
...
 
     return unread_counts
+
+
+def match_user(user: Any, text: str) -> bool:
+    """
+    Matches if the user full name, last name or email matches
+    with user or not.
+    """
+    keywords = user.caption.split()
```
punchagan: The search is case sensitive.  It might be worth making this case insensitive. 
image

timabbott: We do everything case-insensitive in Zulip with names/stream names/topics/etc., so that's probably the right way to go here as well.  Might be worth doing a quick codebase audit for similar issues.

zulipterminal/helper.py
amanagr: the keywords contain for name `Aman kumar agrawal`
* Aman
* kumar
* agrawal
* aman kumar agrawal
* user_email

adding full_name helps in further narrowing down the right user. (eg: Aman kumar and aman singh)

punchagan: Yes, I understand that. But, it's not obvious when I looked at the code first. So, the last sentence from the comment above can go as a comment in the code, I think.

zulipterminal/helper.py
```diff
@@ -275,3 +276,19 @@ def classify_unread_counts(unread_msg_counts: Dict[str, Any])\
...
+    Matches if the user full name, last name or email matches
+    with `text` or not.
+    """
+    full_name = user.caption.lower()
+    keywords = full_name.split()
+    keywords.append(user.email.lower())
+    # adding full_name helps in further narrowing down the right user.
+    keywords.append(full_name)
+    for keyword in keywords:
+        if keyword.startswith(text):
```
punchagan: Wouldn't this also need a `text.lower()`?

zulipterminal/ui_tools/views.py
```diff
@@ -286,3 +292,64 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+        super(RightColumnView, self).__init__(self.users_view(),
+                                              header=search_box)
+
+    @async
+    def update_user_list(self, search_box: Any, new_text: str) -> None:
+        if not self.view.controller.editor_mode:
+            return
+        # wait for any previously started search to finish to avoid
+        # displaying wrong user list.
+        while self.searching:
```
amanagr: @punchagan do you think there is a better way we can do this?

punchagan: I think using a [lock](https://docs.python.org/3/library/threading.html#lock-objects) would work well here.  Also, you could try using regex search and see if it is noticeably faster than what you have implemented.

PR #86, opened 2018-05-22
ui: Fix minor bugs.

amanagr:


README.md
```diff
@@ -70,6 +70,7 @@ theme=default
 | Search Messages | `/` |
 
 Note: You can use `arrows`, `home`, `end`, `Page up` and `Page down` keys to move around in Zulip-Terminal.
+You can use `meta` with `j`, `k`, `l` and `h` keys to move around in the Compose Box.
```
punchagan: Curious about where you picked these shortcuts from, @amanagr. I would expect [readline like shortcuts](https://github.com/chzyer/readline/blob/master/doc/shortcut.md) to work, and would have to explicitly train my muscle memory to use these.

amanagr: actually, as mentioned [here](https://github.com/urwid/urwid/issues/140), there are some problems with using `ctrl` key. There are some workarounds mentioned there, so until we figure out an way to make readline shortcuts work, these can help.

amanagr:
@punchagan I am feeling great about the `urwid_readline` library. It has an autocomplete feature, which we can use to implement @mention suggestions and support narrow suggestions while searching. Can you take another look at this?

punchagan:
@amanagr It looks like the multiline support is broken with this change.  Can you look into that?

punchagan:
Looks good to me, now!

PR #89, opened 2018-05-29
Write Tests for Model Class.

amanagr:


tests/model/test_model.py
```diff
@@ -57,3 +59,80 @@ def test_init(self, model, initial_data, user_profile):
...
+            'narrow': json.dumps(model.narrow),
+        }
+        model.client.do_api_query.assert_called_once_with(
+            request, '/json/messages', method="GET")
+        assert model.index == index_all_messages
+        anchor = messages_successful_response['anchor']
+        assert model.index[str(model.narrow)] == anchor
+        assert model.update is True
+
+        # TEST FOR get_messages() with first_anchor=False
```
punchagan: I would split these into separate tests - if and when a test fails, it would be easier to figure out which, and start debugging.

tests/model/test_model.py
```diff
@@ -57,3 +59,80 @@ def test_init(self, model, initial_data, user_profile):
...
+        model.get_messages(first_anchor=False)
+        messages_successful_response['anchor'] = 0
+        self.client.do_api_query.return_value = messages_successful_response
+        # anchor should have remained the same
+        assert model.index[str(model.narrow)] == anchor
+
+        # TEST `query_range` > no of messagse received
```
punchagan: I think the comment incorrectly says `>`, when it should say `<`

tests/model/test_model.py
```diff
@@ -0,0 +1,59 @@
...
+                     return_value=[])
+        # NOTE: PATCH WHERE USED NOT WHERE DEFINED
+        self.classify_unread_counts = mocker.patch(
+            'zulipterminal.model.classify_unread_counts',
+            return_value=[])
+        mocker.patch('zulipterminal.core.Controller.client.get_profile',
+                     return_value=user_profile)
+        return Model(self.controller)
+
+    def test_init(self, model, initial_data, user_profile):
```
punchagan: I feel like there's not much value in this test, given all the other test we are running.

tests/model/test_model.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # Setup mocks before calling get_messages
+        self.client.register.return_value = initial_data
+        self.client.get_members.return_value = {
+            'members': initial_data['realm_users']}
+        model = Model(self.controller)
+        assert model.initial_data == initial_data
+
+        # Test if raises Exception
+        self.client.register.return_value = Exception()
```
punchagan: Can this be set as `side_effect`?

punchagan: Again, I'd probably want this to be in a different test

tests/model/test_model.py
```diff
@@ -0,0 +1,329 @@
...
+        another_msg.original_widget.message = index['messages'][2]
+        mocker.patch('zulipterminal.model.create_msg_box_list',
+                     return_value=[mock_msg])
+        model.update_reaction(response)
+        print(model.index['messages'][1]['reactions'])
+        update_emoji = model.index['messages'][1]['reactions'][1]['emoji_code']
+        assert update_emoji == response['emoji_code']
+        self.controller.loop.draw_screen.assert_called_once_with()
+
+        # Test removing of reaction.
```
punchagan: Better to split into separate test?

tests/model/test_model.py
amanagr: but it's recommended to have one, right?

amanagr:
Agreed and updated!

punchagan:
Feel free to go ahead and merge this, @amanagr

PR #94, opened 2018-06-05
 test: Add test for views.MessageView.

amanagr:


tests/ui/test_ui_tools.py
```diff
@@ -0,0 +1,179 @@
...
+        key = "down"
+        msg_view.new_loading = False
+        mocker.patch(VIEWS + ".MessageView.focus_position", return_value=0)
+        msg_view.log.next_position.return_value = 1
+        msg_view.keypress(size, key)
+        msg_view.log.next_position.assert_called_once_with(
+            msg_view.focus_position)
+        msg_view.set_focus.assert_called_with(1, 'above')
+
+        # Raise exception
```
punchagan: I think it would be better to split these into separate tests.

PR #100, opened 2018-06-14
ui: show help menu dialog.

amanagr:


amanagr:
@punchagan I will add tests for these if you approve the idea.

zulipterminal/ui_tools/utils.py
```diff
@@ -35,3 +35,54 @@ def create_msg_box_list(model: Any, messages: Union[None, Iterable[Any]]=None,
     if focus_msg is not None:
         model.index['pointer'][str(model.narrow)] = focus_msg
     return w_list
+
+
+HELP_TEXT = """\
```
punchagan: I wonder if there's a way to make this be populated dynamically -- currently, we have to maintain these keybinding in 3 places - the actual code, this help text and the README.  It might be easier to maintain if we are able to make the keybindings a little more "self documenting".

PR #108, opened 2018-06-24
model: Improve encapsulation of narrow

neiljp:
This introduces:
* `model.set_narrow`
* `model.get_focus_in_current_narrow`
* `model.set_focus_in_current_narrow`

It also migrates the `get_current_ids` method from `views.py` to `model.py`, since this is very model-specific code.

There are a few remaining accesses of `model.narrow`, but these are minor; I'd suggest a followup commit to rename `narrow` -> `_narrow` to clarify that this is intended as essentially a private variable.

Tests are also added and migrated, where applicable; other than seeking confirmation that this change seems reasonable, my concern is perhaps that I'm not using the existing test fixtures to the fullest extent, but perhaps this can be discussed further subsequently.

I've looked into further improving the encapsulation, of the index, but this is more challenging and it'd be good to get feedback on this first :)

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -89,6 +89,16 @@ class TestModel:
         model.set_focus_in_current_narrow(msg_id)
         assert model.index['pointer'][str(narrow)] == msg_id
 
+    @pytest.mark.parametrize('bad_args', [
+        dict(topic='some topic'),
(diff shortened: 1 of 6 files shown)


amanagr:
Looks great. Some of the tests can serve as standard examples for future code. Thanks Neil.

PR #109, opened 2018-06-27
Streams: Refactoring & pinned-stream implementation

neiljp:


tests/ui/test_ui.py
```diff
@@ -68,10 +69,17 @@ def test_left_column_view(self, view, mocker):
         pile = mocker.patch('zulipterminal.ui.urwid.Pile')
         return_value = view.left_column_view()
         menu_view.assert_called_once_with()
-        streams_view.assert_called_once_with()
+        expected_streams_view_calls = [
```
amanagr: wow

zulipterminal/ui.py
```diff
@@ -90,14 +92,18 @@ def streams_view(self) -> Any:
...
+        stream_view = StreamsView(streams_btn_list)
+        self.stream_w.extend(stream_view.log)  # For multiple calls
+        w = urwid.LineBox(stream_view, title=title)
         return w
 
     def left_column_view(self) -> Any:
         left_column_structure = [
             (4, self.menu_view()),
-            self.streams_view(),
+            self.streams_view(stream_list=self.pinned_streams,
```
amanagr: Since every user has only a few pinned streams, this could be :
```
        left_column_structure = [
            (4, self.menu_view()),
            ('weight', 1, self.streams_view(stream_list=self.pinned_streams,
                                            title="Pinned Streams")),
            ('weight', 5, self.streams_view(stream_list=self.unpinned_streams,
                                            title="Other Streams")),
        ]
```
Also, can we not display the `pinned streams` line_box when there are no pinned streams.

amanagr: You can refer to this [link](http://urwid.org/reference/widget.html#urwid.Pile) for more details.

neiljp: Mainly I had got to this stage and was fairly happy with the result in terms of it being moderately effective, leaving fine-tuning for a later iteration.

In addition:
* For relative length, I wanted maximum rows for the pinned streams of the number of pinned streams, but optimizing heavily for displaying those over the other streams - I assume that's the intent of the `weight` parameter
* I also planned to leave the no-pinned case for later, though I had considered resolving it.

tests/ui/test_ui.py
neiljp: Was this in reference to a particular line?

zulipterminal/ui.py
amanagr: Fixing rows for any widget could backfire unless you take into account the height of the terminal (which could get a lot messier). So, I would suggest following the `weight` approach I mentioned.

I believe for no-pinned case a simple if statement will do for now.
Thanks neil.

tests/ui/test_ui.py
amanagr: line 72-78. This is my first time seeing this approach.

zulipterminal/ui.py
neiljp: I wasn't looking to fix the number of rows explicitly, but more set a maximum, which I saw in the documentation but hadn't explored enough to implement at this point. So something like the weighting you suggest (prioritizing pinned), but never showing *more* than the number of streams there (I get that at the moment - there are gaps below my last pinned stream)

amanagr: Yeah, that sounds like a better idea.

neiljp:
@amanagr Updated against latest master and with an extra commit to hide if pinned or unpinned are empty.

amanagr:
I clearly understand the reason for this. I will fix this soon with a PR. Thank Neil.

amanagr:
@neiljp I would love to get this merged. Can you resolve the conflicts?

neiljp:
Closing, since this is superseded by #187, which is now merged.

PR #110, opened 2018-06-29
reduce startup time.

amanagr:


amanagr:
@punchagan I did some bench marking and this reduces startup time by 3s on average on a 300KB/s connection. Can you have a look?

zulipterminal/model.py
```diff
@@ -33,15 +34,36 @@ def __init__(self, controller: Any) -> None:
         self.stream_dict = {}  # type: Dict[int, Any]
         self.recipients = frozenset()  # type: FrozenSet[Any]
         self.index = None  # type: Any
-        self.get_messages(first_anchor=True)
-        self.initial_data = self.fetch_initial_data()
-        self.user_id = self.client.get_profile()['user_id']
+        self.user_id = -1  # type: int
+        self._update_user_id()
+        # Thread Processes to reduces start time.
```
punchagan: Would it be better to move all this threading code into the `_update_initial_data` method?

PR #111, opened 2018-07-01
view: Display current narrow at the bottom of the view.

amanagr:
This is first version of view.footer. I am thinking of including help text at the bottom of the view too.

### DIFF
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -102,7 +102,8 @@ class View(urwid.WidgetWrap):
             urwid.Divider(div_char=div_char),
         ])
 
-        w = urwid.Frame(self.body, title_bar, focus_part='body')
+        w = urwid.Frame(self.body, title_bar, footer=urwid.Text(""),
+                        focus_part='body')
         return w
 
     def keypress(self, size: Tuple[int, int], key: str) -> str:
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from typing import Any, List, Tuple
 import threading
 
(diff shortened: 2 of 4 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -122,11 +123,27 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
         key = super(MessageView, self).keypress(size, key)
         return key
 
+    def update_current_footer(self, message_view: Any) -> None:
+        if hasattr(self.model.controller, 'view'):
```
punchagan: It would be nice to just return if the `view is None` and do the footer computation otherwise.

zulipterminal/ui_tools/views.py
```diff
@@ -122,11 +123,27 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+            message_view.last_message = defaultdict(dict)
+            is_stream = message_view.message['type'] == 'stream'
+            if is_stream:
+                header = message_view.stream_view()
+            else:
+                header = message_view.private_view()
+            header.contents[1] = (urwid.Text("Press ? to view Help.",
+                                  align='right'),
+                                  header.options())
+            self.model.controller.view._w.set_footer(header)
```
punchagan: This reads weirdly - `set_footer(header)`

zulipterminal/ui.py
```diff
@@ -139,7 +139,8 @@ def main_window(self) -> Any:
             urwid.Divider(div_char=div_char),
         ])
 
-        w = urwid.Frame(self.body, title_bar, focus_part='body')
+        w = urwid.Frame(self.body, title_bar, footer=urwid.Text("DEMO TEXT"),
```
punchagan: Do we need to use `DEMO TEXT`? I think it's better to leave it as an empty string

PR #113, opened 2018-07-03
hotkeys: Show help menu on pressing `?`.

amanagr:
@punchagan this looks like a better idea. I would love some review from you on this.

### DIFF
--- a/zulipterminal/config.py
+++ b/zulipterminal/config.py
@@ -1,26 +1,110 @@
 KEY_BINDINGS = {
-    'k':
-        ('up', 'scroll up'),
-    'j':
-        ('down', 'scroll down'),
-    'h':
-        ('left', 'Go left'),
-    'l':
-        ('right', 'Go right'),
-    'K':
-        ('page up', 'scroll to top'),
-    'J':
-        ('page down', 'scroll to bottom'),
-    'G':
-        ('end', 'Go to last message in view'),
-    'x':
-        ('x', 'New private message'),
+    'PREVIOUS_MESSAGE': {
+        'keys': {'k', 'up'},
+        'help_text': 'Previous message',
+
+    },
+    'NEXT_MESSAGE': {
+        'keys': {'j', 'down'},
+        'help_text': 'Next message',
+    },
+    'GO_LEFT': {
+        'keys': {'h', 'left'},
+        'help_text': 'Go left',
+    },
+    'GO_RIGHT': {
+        'keys': {'l', 'right'},
+        'help_text': 'Go right',
+    },
+    'SCROLL_TO_TOP': {
+        'keys': {'K', 'page up'},
+        'help_text': 'Scroll to top',
+    },
+    'SCROLL_TO_BOTTOM': {
+        'keys': {'J', 'page down'},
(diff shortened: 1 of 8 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -339,3 +341,38 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+            commands = ""
+            other_key = pair[0]
+            help_text = pair[1]
+            commands += key
+            if key != other_key:
+                commands += ", " + other_key
+            self.log.append(
+                urwid.Columns([
+                    urwid.LineBox(
+                        urwid.Text(commands),
```
punchagan: I like it when the commands are aligned right, so that I can quickly scan what a key does.

zulipterminal/config.py
```diff
@@ -1,18 +1,42 @@
 KEY_BINDINGS = {
```
punchagan: Should we rename this file to `keyconfig`?  Or do we expect to add other config to this file?

zulipterminal/config.py
```diff
@@ -1,18 +1,42 @@
...
         ('right', 'Go right'),
     'K':
-        ('page up', 'scroll to top'),
+        ('page up', 'Scroll to top'),
     'J':
-        ('page down', 'scroll to bottom'),
+        ('page down', 'Scroll to bottom'),
     'G':
         ('end', 'Go to last message in view'),
+    'r':
```
punchagan: It would be worth changing the structure of `KEY_BINDINGS`. 

Something like 
```python
...
'NEXT_MESSAGE': {'keys': {'j', 'down'}, 'help_text': 'Next message'},
'REPLY': {'keys': {'r'}, 'help_text': 'Reply to a message'},
...
```
might be good. 

This would allow us to replace all the hard-coded checks for keys (like `key == "r"` to something like `key in KEY_BINDINGS['REPLY']['keys']`).  We could also add a helper class to add methods for these kind of checks, etc. 

Later, this would allow for making it easy for users to overwrite/add more keys for specific actions.

zulipterminal/config.py
punchagan: It would probably be nice to have a developer tool (+ linter that runs on CI?) that can automatically generate the README section for keybindings from this data-structure.

amanagr: I do have other config I plan to add to this file.

amanagr:
@punchagan I have resolved conflicts with this PR. We can generate README for hotkeys in different PR.

zulipterminal/ui.py
```diff
@@ -109,24 +109,41 @@ def main_window(self) -> Any:
...
             self.controller.editor_mode = True
             self.controller.editor = self.stream_w.search_box
             return key
-        else:
-            return super(View, self).keypress(size, get_key(key))
+        elif is_command_key('HELP', key):
+            # Show help menu
+            self.controller.show_help()
+            return key
+        elif is_command_key('PREVIOUS_MESSAGE', key):
```
punchagan: I think it would help to add a comment on why this replacement needs to be done.

PR #115, opened 2018-07-08
tests: Disable async mode during tests.

amanagr:


amanagr:
@neiljp I made some changes to your origin idea after discussing with @punchagan.
We cannot import `pytest` in `helper.py` since it is a part of the dev requirements and not installed normally.

neiljp:
The example with the pytest module was just to check if it worked ok. I tried with the environment variable but it didn't work with another approach I tried, so it's good to know it appears to work this way.

However, how do you know if it's working? (the title seems inaccurate too, FWIW)

punchagan:
> However, how do you know if it's working? 

Just adding a `print` in the section of the `async` wrapper that shouldn't be reachable in the testing mode should confirm that it's working.

PR #117, opened 2018-07-09
ui: Change emojis before `All messages` and `PM Button`.

amanagr:
Advance emoijs don't work on many platforms for urwid. So replace
them simply with a BULLET.


zulipterminal/ui_tools/buttons.py
```diff
@@ -26,7 +26,7 @@ def update_count(self, count: int) -> None:
 
     def widget(self, count: int) -> Any:
         return urwid.AttrMap(urwid.SelectableIcon(
-            [u' 🏡 ', self.caption,
+            [u'\N{BULLET} ', self.caption,
```
punchagan: Just adding a space before the bullet seems to make it look better. 

image

image

PR #121, opened 2018-07-09
Improve messages output (MessageBox)

neiljp:
This broadly improves the layout of the central window, at least to my eye, by:
* shifting the date in line with the author-name (sender)
* omitting repeated author-names, where people send messages without another author replying

This also saves vertical space, through the first change, which was one original motivation.

The first commit is less necessary now, since the time is in the `main_view` method, but I still think is a reasonable functionality to extract.

Please give this a try to see the change in layout.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 from time import ctime
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


neiljp:
Updated to ensure author names are shown after a new topic header.

amanagr:
Thanks @neiljp. merged.

PR #122, opened 2018-07-10
Add private recipient grouping

neiljp:
This does for private messages, what already exists for streams/topics, ie. combining conversation 'titles', as well as removing the user's name from the conversation to make it clearer.


amanagr:
Thanks @neiljp. merged!

PR #123, opened 2018-07-11
debugging: Write sys.stdout data in /tmp/debug.log.

amanagr:
Add instructions and tips for debugging in README.


amanagr:
Thanks @punchagan for the idea.
@neiljp I would love some feedback from you on this.

neiljp:
@amanagr See chat.zulip.org, since you merged this already.

PR #125, opened 2018-07-12
ui: Clear screen for any distortions after loop draws screen.

amanagr:
This corrects many of the distortions in the which appear
when ZT updates widgets. There are still some distortions left
with user search which need to be investigated further.


punchagan:
Isn't the distortion problem because of calling the [redraw from threads](https://github.com/urwid/urwid/issues/165)?  Can we try to move all the redraws to the main thread and see if this helps?

amanagr:
Duplicate #130.

PR #130, opened 2018-07-13
ui: Remove distortions in ui by running draw_screen in main thread.

amanagr:
Running MainLoop.draw_screen in a  threaded process distorts the
screen. Setting up pipe method to always run draw_screen in
main thread removes all distortions.


amanagr:
Thanks @punchagan for this.

PR #134, opened 2018-07-18
Display messages by converting html to markup.

amanagr:


amanagr:
@punchagan this takes care of some of the frequently user markdown syntax. I will have to change the pointer style before merging this as there are some overlapping color issues which this PR brings in.

PR #137, opened 2018-07-21
Add initial support for reactions via '+' key for thumbs_up.

neiljp:
This first stab at 'thumbs_up' works, but as with other cases of new messages being added to the message-view, doesn't quite update the view properly each time in ZT.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -111,6 +111,30 @@ class Model:
             current_ids = self.index['search']
         return current_ids.copy()
 
+    @async
+    def react_to_message(self,
+                         message: Dict[str, Any],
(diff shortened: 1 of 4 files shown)


zulipterminal/model.py
```diff
@@ -111,6 +111,29 @@ def get_message_ids_in_current_narrow(self) -> Set[int]:
             current_ids = self.index['search']
         return current_ids.copy()
 
+    def react_to_message(self,
```
amanagr: We should make this method `async` so that the user can still interact with the ui when request is being sent.

neiljp: Done.

PR #139, opened 2018-07-21
boxes: Remove excess spacing in message display for wide screens.

neiljp:
These figures are a little too embedded for my liking, but for now they make the wider-screen behavior much less cramped, while retaining more spacing than before.


amanagr:
The emoji's are disoriented now. They could use `width=('relative', 90), left=25,` too.

Messages were compressed to allow users to read the whole message without moving their head/eye.

neiljp:
Perhaps the solution is to have a configuration option/setting, since different people will prefer different options? However, what is the use of the wider window, if not to use the space? With this change, you can just make the window narrower and you can achieve what you say :)

Agreed about the emojis, but I'll hold off on that until the above is decided.

PR #140, opened 2018-07-21
boxes: Show topic changes by bar background & surrounding spaces.

neiljp:
This spaces out the topic slightly from the stream and the dividing '>' character, adds a space after the topic, and then styles the 'bar'.

This is intended to bring back some of the topic separation/heading styling, but maintaining the stream/topic styling distinction.


amanagr:
Looks Great! Thanks @neiljp

PR #152, opened 2018-10-24
Mark private streams

neiljp:
This PR shows private streams with a 'P' (rather than '#'), much like a lock is shown in the zulip web-app.

Current tests pass with the first commit; the second commit extends them to include a private-stream test-case and so improves the test coverage.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -214,6 +214,22 @@ def initial_data():
             'email_address': '',
             'color': '#c2c2c2',
             'in_home_view': True
+        }, {
+            # This is a private stream;
+            # only description/stream_id/invite_only/name/color vary from above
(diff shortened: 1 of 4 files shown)


amanagr:
Great! Thanks @neiljp

timabbott:
I wonder if there's a unicode lock-icon type thing we could use to make this even nicer.

neiljp:
@timabbott I did look into this, but while it displayed ok in the browser it didn't when I pasted it into the source and/or ran it. Something like this: https://www.fileformat.info/info/unicode/char/1f512/index.htm

timabbott:
Yeah, many unicode characters don't actually display nicely in the terminal; sounds like this is one of them.

PR #153, opened 2018-10-24
bugfix: Elide divs/spans with no class in soup2markup.

neiljp:
This caused a crash; this commit checks for `element.attrs` in the
conditionals before getting the class.

See https://chat.zulip.org/#narrow/stream/9-issues/subject/Un-classed.20DIV.20in.20rendered.20content.3F/near/655208

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -212,7 +212,9 @@ class MessageBox(urwid.Pile):
             return ''
 
     def soup2markup(self, soup: Any) -> List[Any]:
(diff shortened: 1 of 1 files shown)


neiljp:
Just pushed an extra bugfix I found, which may have been triggered by the first fix or just due to an empty message. The triggering message contained an emoji, which `soup2markup` removed and left an empty list which urwid failed on.

The (updated) bugfix as it stands is to initialize the list to `['']`, which imposes an extra workload but minimal extra storage; an alternative would be to just check if markup is empty and return `['']`.

PR #154, opened 2018-10-24
Reorder shortcut keys (& fix a few typos)

neiljp:
* Reordered the shortcut key listing in the help (via `config.py`) to group similar commands;
* Collected together some typos I remembered from browsing the code.

The first one is small but important - if I scan the help and find 'new [something] message' I would expect to find 'new [otherthing] message' nearby, as well as how to reply, or edit messages, and the same for other aspects.


amanagr:
Great! Thanks @neiljp

PR #155, opened 2018-10-28
Fix path to debug.log

FoolDresden:
The docs previously stated that the debug.log was generated in /tmp/debug.log however the default place for the output was ./debug.log with no file being generated in /tmp/debug.log. Correcting that.


amanagr:
Thanks @FoolDresden

PR #157, opened 2018-11-01
Fix issue with internal (cross-realm) bots

neiljp:
This fixes the bug for me, and adds some cross-realm bot data as currently in czo.

I tried to generalize the tests to cover the case in question, but there appear to be issues with pytest or the way the code/tests work, which I couldn't bypass - and it seems important to get this fix in :)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -186,8 +186,51 @@ def initial_data():
             'user_id': 6085,
             'is_active': True
         }],
-        'cross_realm_bots': [
-        ],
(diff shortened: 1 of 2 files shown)


neiljp:
I should clarify, that since I couldn't add an explicit test for this case, the manual testing I've looked at has been with `Notification Bot` rather than the `Welcome Bot`, but I replicated the issue with the former and it should work with the latter AFAICT. I'd really like an explicit automated test!

amanagr:
Looks great @neiljp

PR #158, opened 2018-11-04
Refactoring: encapsulate Model.get_messages parameters

neiljp:
This simplifies the use of `get_messages` by moving `num_before`, `num_after`, and `anchor` to be parameters, rather than being set in the model first. 

These changes should be straightforward to understand and merge.

Tests run locally, though possibly mypy may complain.
**update** mypy issues fixed using `TypedDict`, and also upgraded mypy.


amanagr:
Great! Thanks @neiljp

PR #159, opened 2018-11-10
Add extra MessageBox tests

neiljp:
These increase overall coverage by 0.5% and are a precursor to other commits which refactor the MessageBox code. One small code fix is also included, uncovered by adding the tests.


neiljp:
Fixed tests to use date code using local date via `MessageBox` function rather than hardcoded date/time.

amanagr:
Great! Thanks @neiljp

PR #160, opened 2018-11-12
Add tests & extra function for config.

neiljp:
Originally motivated by broadening the `keypress` tests, which are currently hardcoded to specific keys rather than the `KEY_BINDINGS` in `config.py`, to iterate through listed possible keys. Then noticed this file was untested, thought of refactoring, but wanted to add (the minimal!) tests first :)

It turned out easier to write the `is_command_key` tests in terms of `keys_for_command`, hence the test infrastructure commit is currently separate reflecting my reordering.

PLEASE CHECK: The docstring for `is_command_key` currently indicates it returns "a key if mapped or the key otherwise". I've not intended to change this behavior, but it seems not to describe current master? I can incorporate this if relevant.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -19,3 +19,16 @@ def test_keys_for_command(command):
 def test_keys_for_command_invalid_command(command):
     with pytest.raises(config.InvalidCommand):
         config.keys_for_command(command)
+
+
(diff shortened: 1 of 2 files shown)


amanagr:
This looks better! Thanks!

PR #161, opened 2018-11-12
Show starred messages & refactor MessageBox.main_view()

neiljp:
* Shows stars applied to messages in other clients
* Amends and adds tests related to the star-viewing code
* Refactor code to aid readability using additional dicts, `any` and comprehensions

Old tests amended, more tests added (coverage still at 65.7%).


amanagr:
Looks great @neiljp

PR #162, opened 2018-11-14
Bugfix: Update & access index correctly for multiple PM recipients

neiljp:
I fixed this while refactoring, but backported it since this seems to cause a crash on master.

I've added/amended tests/fixtures to include a multiple-recipient case, and if the commits are reordered (or the bugfix is removed) then the additional test indeed fails.

The bugfix commit itself is small and appears to work fine, and the `core.py` code is now more consistent with similar functions, however:
* it doesn't set `button.user_id` (it doesn't seem an issue?)
* was the different layout/branching of the old code important?


amanagr:
Looks great @neiljp! Thanks

PR #164, opened 2018-11-16
Server and user starred-message interactions

neiljp:
This is a followup to initial work in #161, which makes starred messages pretty usable in ZT:

* If star status changes on the server, ZT now notices and updates the view
* Users can now press '*', and toggle the star status of a message

Tests are included.

NOTE: Setting 'ctrl s' as the key for the star status doesn't appear to work for me locally; this may be an issue with urwid-readline, but I've not debugged further.


neiljp:
@amanagr I don't understand why these tests are failing; did we actually move to python 3.6+ though? If so, should we actually test against 3.4 and 3.5 on travis?

PR #169, opened 2018-11-25
UI: Add marker to left of message.

### LINKED ISSUE #98: Add a box around message in focus.
Currently, when a message is in focus, it's colour is changed to red. We can add a `LineBox` widget around the message in focus to make it visually appealing.

neiljp:
The marker:
* is always present
* is colored according to read/unread/active status
* emphasizes how messages are separate when content_headers are shared
* indicates message status, when text cannot be colored
  (eg. just bold/italic text, or just a user-mention).

I was going to implement this as a column of text, but getting the height correct seemed complex - particularly when a LineBox can be used, as suggested in #98. I experimented with fuller LineBox's, but decided upon the left-side since it's like the webapp, takes up less space, and looks good :)

This could be considered to resolve #98, but I'm not sure if the intent there was to make a distinction on shape alone, in case of a no-colored approach. Otherwise, this fixes #98, and perhaps a separate issue should track potential color-absent use, if that becomes desirable.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -343,9 +343,16 @@ class MessageBox(urwid.Pile):
         # Content
         soup = BeautifulSoup(self.message['content'], 'lxml')
         content = (None, self.soup2markup(soup))
-        content = urwid.Padding(urwid.Text(content),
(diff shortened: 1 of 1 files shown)


amanagr:
Great!

PR #170, opened 2018-11-26
Message markup: Add listitems & placeholders; includes tests & refactoring

neiljp:
The main changes to the UI here are:
* adding bullet points to `<li>` tags (this makes eg. the commit stream read a lot better)
* marking where some message components exist, but are not rendered (`[{} NOT RENDERED]`)
* the previous point includes some changes which may cause messages to output differently; there was no test previously, so this isn't completely obvious, but eg. table content may have actually been output, but just as a jumble of text.

On a per-commit basis:
* Add test for core `soup2markup` functionality
* Mark list items with a `*`
* Avoid needing to process high-level tags by passing the `<body>` to `soup2markup`
* Improve existing tests with more realistic content (div -> p)
* Add explicit support for many more tags, rather than just looking inside each unknown one
* Extract handling of unrendered simple tags and classes of div into single blocks (2 commits)


amanagr:
Wow! amazing work. I think there is some merge conflict. I think this can merged after resolving the conflict.

neiljp:
Conflicts should be resolved; also added a minor extraction of `unrendered_template` to the last commit.

PR #173, opened 2018-11-28
README: Reorder & add more 'shields' (pypi/python/mypy).

neiljp:
This adds some more shields and places the chat first.


amanagr:
Thanks! Looks great.

PR #182, opened 2018-12-03
Bugfix: Prevent leaking exceptions when Ctrl+C is pressed during authorisation

vonknio:
If the `zuliprc` file is missing and the user is prompted for their credentials, pressing `Ctrl+C` (especially after a couple of failed attempts to authorise) will result in a traceback of uncaught exceptions. This is fixed by moving the call to `parse_zuliprc` into a try block.

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -112,19 +112,19 @@ def main() -> None:
     """
 
     args = parse_args()
-    if args.config_file:
-        zuliprc_path = args.config_file
-    else:
-        zuliprc_path = '~/zuliprc'
-
-    zterm = parse_zuliprc(zuliprc_path)
 
     if args.profile:
         import cProfile
         prof = cProfile.Profile()
         prof.enable()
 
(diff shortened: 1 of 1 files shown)


neiljp:
@vonknio I can see how this would resolve the issue, so we could merge this as-is; however, you mentioned on chat.zulip.org about the recursive calls, so perhaps that is the root cause of the uncaught exceptions? I can reproduce the issue you refer to, so am happy to discuss this.

It would also be great if you could review the guidelines for contributing (https://github.com/zulip/zulip-terminal#contributor-guidelines), in particular here with regard to the commit message.

vonknio:
@neiljp The recursive calls will indeed make the exception stack trace longer in this situation, but it can appear without them. Pressing `Ctrl+C` as soon as the prompt is shown still illustrates the problem, but no recursive call has been made at that point.
I do agree that recursive calls are worth eliminating, but they aren't the sole cause of the problem here.

neiljp:
@vonknio Thanks!

PR #183, opened 2018-12-05
Refactor index fixtures

neiljp:
These commits:
* extract the `empty_index` fixture, and assert equivalence of slight variations of it to current `index_*` fixtures (which passes tests locally)
* modifies those `index_*` fixtures to then use the simpler variation forms.

This reduces the duplication in the `index_*` fixtures and makes them much easier to use.

This PR is also a precursor to my upcoming starred-UI work, which uses `empty_index` to produce an `index_all_starred` fixture and is used in other tests.


amanagr:
Great!

PR #186, opened 2018-12-08
Add 'M' to indicate muted streams in place of absent count

neiljp:
This at least shows that muted streams are muted, whereas previously this was not obvious.

In the webapp the counts are currently still shown, but faded out; whether we duplicate that is not clear, but this at least shows that the streams are muted.


amanagr:
LGTM

PR #187, opened 2018-12-10
Pinned message UI/backend & refactor of stream setup in model

neiljp:
This is an improved replacement for #109.


amanagr:
Superb Work!

PR #188, opened 2018-12-10
Add version.py & use in core & setup.py.

neiljp:
While exploring the code to check the agent being reported, I noticed the version was hard-coded and doesn't match `setup.py` (or pypi).

This PR pulls this from a central file for both, and given 0.2.1 is on pypi, I set the version to `0.2.1+git`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ import sys
 import codecs
 from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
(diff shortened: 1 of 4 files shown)


neiljp:
I agree about doing a release soonish, as per on #zulip-terminal, and probably v0.3, but I thought that renumbering this could be part of the process, so this PR would make it easier to centralize that whenever it happens.

amanagr:
Sure!

PR #189, opened 2018-12-10
setup: Specify python requirements for pip>=9.

neiljp:
For a sufficiently up to date pip, we can specify python when installing using pip; given we test against 3.4-3.7 now, I set this to `>=3.4, <3.8`.


amanagr:
Looks good!

neiljp:
@amanagr I specifically manually tested this with my local python2.7 install in Debian, and it wouldn't install, which is a good data point :)

PR #190, opened 2018-12-14
Disable ctrl-z and ctrl-s (enabling star key use)

### LINKED ISSUE #166: Investigate why keys such as 'ctrl s' do not appear to work in config.py.
In #164 I was hoping to use 'ctrl s' as the mapping for toggling starred messages, as with the zulip webapp. However, this does not currently seem to work; for now I've put an alternative key in, but this may impact other cases where we wish to use keys with modifiers ('ctrl' or otherwise), so should probably be investigated.

My guess was that this might be related to urwid-readline interfering, but I've not confirmed this.

neiljp:
It seems reasonable to disable both of these since:
- we don't currently respond appropriately to ^Z
- flow control is probably not relevant at this point (we could disable ^Q as well as ^S?)

We could potentially enable/disable the latter two for use as shortcut keys, and amend the signal code in `core.py` at run-time, but for now it seems reasonable to just disable it.

Not excluding ^Q should be safe, since that resumes the flow after a ^S.

The second commit fixes #166, so [un]starring keys will then incorporate the webapp (but retain `*`).

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -165,5 +165,6 @@ class TestController:
         controller.view.palette = {
             'default': 'theme_properties'
         }
+        mock_tsk = mocker.patch('zulipterminal.ui.Screen.tty_signal_keys')
         controller.main()
         assert mock_loop.call_count == 1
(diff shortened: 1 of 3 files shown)


amanagr:
Looks great.

PR #191, opened 2018-12-15
Travis: Add darwin on build matrix.

rht:
~This may or may not cache the pipenv files.~ Found the pipenv cache location, as per https://pipenv.readthedocs.io/en/latest/diagnose/. Added via 3rd commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,15 @@ matrix:
     - python: 3.7
       dist: xenial
       sudo: true
+    - os: osx
+      cache:
+      - $HOME/Library/Caches/pip
+      osx_image: xcode10
+      sudo: required
+      language: generic
+      install:
+      - pip install pipenv==11.10
+      - pipenv install --three --dev Pipfile
 install:
   - pip install pipenv==11.10
   - pipenv install --dev Pipfile


rht:
The second commit's main purpose was to test the caching. It should have been put in another PR otherwise.

rht:
Test time (first commit vs fourth commit):
- 3.4: 2 min 50 sec -> 2 min
- 3.5: 2 min 53 sec -> 2 min 16 sec
- 3.6: 2 min 30 sec -> 1 min 59 sec
- 3.7: 2 min 25 sec -> 1 min 58 sec
- darwin: N/A -> 4 min 12 sec

First commit: https://travis-ci.org/zulip/zulip-terminal/builds/468274674?utm_source=github_status&utm_medium=notification
Fourth commit: https://travis-ci.org/zulip/zulip-terminal/builds/468278356

neiljp:
@rht This is testing against python3.4; that seems like a good baseline, but was that intentional?

rht:
The vm uses homebrew-provided python, which is 3.7 (https://travis-ci.org/zulip/zulip-terminal/jobs/469486796#L140). I'd say for macOS, people tend to use the latest version provided by homebrew if possible. The `Python: 3.4` might be due to the first of the `python` list or something.

neiljp:
Right, yes, the 3.4 does originate from that I think, which was why I wanted to check that. OTOH if the code works in 3.4 then it should work in 3.7? :)

amanagr:
I think testing py3.4 on OSX will do for now.

amanagr:
Thanks @rht, merged!

rht:
Note: the Travis env var is misleading, the test happens on 3.7 as provided by either Homebrew of XCode.

amanagr:
I opened a bug report on Travis Community for this.

PR #192, opened 2018-12-15
[WIP] Travis: Add caching, limit git depth, name jobs, reduce repeated tests 

neiljp:


rht:
The cache has yet to take effect: https://travis-ci.org/zulip/zulip-terminal/jobs/468316993#L458. When it does, it should shave off ~30s.

.travis.yml
```diff
@@ -1,20 +1,24 @@
...
-  - "3.5"
-  - "3.6"
+cache: pip
 matrix:
   include:
+    - python: 3.4
+      name: "Python 3.4 pytest (default linux)"
+    - python: 3.5
+      name: "Python 3.5 pytest & mypy type-checking (default linux)"
+      env: RUN_MYPY=yes
```
rht: I found that the Travis trusty image has 3.6 by default, but has to download other versions for 3.4 and 3.5. I think the overall test time would have been reduced if Mypy test is piled on 3.6.

rht: https://travis-ci.org/zulip/zulip-terminal/jobs/468316993#L415 it is not logged how long it takes to download python-3.4.tar.bz2. It is possibly ~5s.

amanagr: Yup, I seems like a good idea to run mypy on 3.6 instead of 3.5.

neiljp:
Rebased & updated, including shifting the mypy to 3.6 and merging the install block to be common across each test section.

@rht How does the cache code look now?

.travis.yml
```diff
@@ -1,11 +1,21 @@
 language: python
-python:
-  - "3.4"
-  - "3.5"
-  - "3.6"
+cache: pip
+git:
+  depth: 20
+cache:
```
rht: There should additionally be `pip: true` in `cache:` so that pipenv itself doesn't have to be redownloaded each time.

amanagr:
I don't have a lot of experience with travis but this definitely looks like an improvement.

.travis.yml
neiljp: The new version has this update.

.travis.yml
neiljp: Is that since we should combine the `cache:` entry, and when we do it needs `pip: true`?

Perhaps like this?
https://docs.travis-ci.com/user/caching/#enabling-multiple-caching-features

neiljp: @rht Is this what you meant? ^

rht: Yes, to be exact:
```
cache:
  pip: true
  directories:
    - $HOME/.cache/pipenv
```
Otherwise I'm not sure how yaml handles a field redefinition.

neiljp:
@rht I pushed an update which resolves the original cache issue; does this interact properly with the OSX cache directive, do you think?

rht:
I can't recall whether the `pip: true` captures darwin pip cache directories or not. Regardless of whether this is the case, the PR is LGTM, and is better merged soon to speed up CI.

neiljp:
Merged after the positive feedback from @amanagr, @rht and also @punchagan.

PR #195, opened 2018-12-15
config: Switch the key for stream narrow and topic narrow.

rht:
This is in order to be consistent with the webapp.
A screenshot of the webapp key map exists at
https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Narrowing.20behavior


neiljp:
As discussed on #zulip-terminal, simple as expected, tested and works :)
Thanks for the report and fix @rht :+1:

PR #196, opened 2018-12-15
deps: Install typing only on < 3.5.

rht:
This has been fixed in master though I couldn't find the reason why. It would be safer to not install `typing` on newer version.


neiljp:
Is this also an issue in the main Zulip repo?

amanagr:
Thanks @rht

PR #198, opened 2018-12-16
requirements: Upgrade urwid_readline to 0.8.

### LINKED ISSUE #177: Test and upgrade uriwdn_readline library to version 0.8
Urwid_readline was recently upgrade to 0.8. Test that it works properly and upgrade the version used in zulip-terminal.

amanagr:
Fixes #177

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -8,7 +8,7 @@ typing = {version="==3.6.4", markers="python_version < '3.5'"}
 urwid = "==2.0.1"
 zulip = "==0.4.7"
 emoji = "==0.5.0"
-urwid-readline = "==0.7"
+urwid-readline = "==0.8"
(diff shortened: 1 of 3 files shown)


neiljp:
@amanagr There are no breaking changes on this upgrade?

amanagr:
It fixes some issues with the arrow keys. It's better to have it, than not.

neiljp:
Sure, just checking if there were any consequences.

Also, I just added a PR for urwid-readline (rr-/urwid_readline#7), since I noticed there was a completion option which we could use, but it requires the completion key be "tab", and in ZT as it stands this is used to cycle between parts of the compose window.

amanagr:
Great! we should definitely work in that direction.

PR #199, opened 2018-12-16
ui: Add 16:default,256:gruvbox colorscheme.

rht:
Proof-of-concept of using 256-color mode with gruvbox theme (https://github.com/morhetz/gruvbox/blob/master/colors/gruvbox.vim).

Before:
2018-12-16-075049_1920x1080_scrot

After:
2018-12-16-080025_1920x1080_scrot

Still hardcoded. I should make 2 preparatory commits to: enable/disable the 256-color on terminals that support 256-color, conf to easily modify the color instead of at ui.py.

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -2,6 +2,21 @@ from typing import Dict, List, Tuple, Optional
 
 ThemeSpec = List[Tuple[Optional[str], ...]]
 
+# Colors used in gruvbox-256
+# See https://github.com/morhetz/gruvbox/blob/master/colors/gruvbox.vim
+BLACK = 'h234'  # dark0_hard
+WHITE = 'h246'  # light4_256
+WHITEBOLD = '%s, bold' % WHITE
+DARKBLUE = 'h24'  # faded_blue
+DARKRED = 'h88'  # faded_red
+LIGHTBLUE = 'h109'  # bright_blue
+YELLOW = 'h172'  # neutral_yellow
+YELLOWBOLD = '%s, bold' % YELLOW
+LIGHTGREEN = 'h142'  # bright_green
+LIGHTRED = 'h167'  # bright_red
+LIGHTREDBOLD = '%s, bold' % LIGHTRED
+GRAY = 'h244'  # gray_244
+
 THEMES = {
     'default': [
         (None,           'white',           'black'),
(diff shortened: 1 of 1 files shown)


rht:
Wait, I should add a new pallete instead of modifying default.

neiljp:
@rht Having another scheme option is great, if you can fix the PEP8 failures then this looks fine to me, given it's essentially just a bunch of declarations :)

rht:
Rebased + fixed the trailing whitespace lint error.

neiljp:
@rht I just tried this out manually from the PR by setting the theme in my zuliprc, and it shows all the channels as white; is that intentional? That wasn't visible in your example above, so is something not working correctly?

neiljp:
@rht are the colors not working due to my terminal settings, perhaps? Is a fallback not working?

Also, what do you think about changing the name to something simpler, as it's quite long at the moment? You could add a comment in the theme file if more notes are necessary?

neiljp:
@rht Also see #204 which might be of interest.

rht:
Huh, I just tested on `urxvt` and `kitty`, both display the color correctly. The fallback is exactly the default version. `urwid` must have detect your terminal as 256, but fails to translate the color code somehow.
2018-12-18-184539_385x865_scrot

rht:
> Also, what do you think about changing the name to something simpler, as it's quite long at the moment? 

It is the shortest descriptive name of how the theme operates at 16 and 256 color that I can think of. Maybe I should just shorten it to `gruvbox`, and figure out the approximate 16 mode mapping of the colorscheme.

rht:
I shortened it to `gruvbox`.

neiljp:
screenshot-20181218-11 35 16-cropped
This is what I see with the previous version of the branch.

neiljp:
@rht I think I know the issue, and I plan to integrate it with my recent PR shortly!

neiljp:
This almost works, but the styling doesn't quite look right:
screenshot-20181218-12 09 55-cropped

neiljp:
With the latest change, the styling looks almost right, but my urwid color knowledge isn't fully developed yet, so any advice on adjusting the `view.palette.append` lines in `StreamButton` beyond the second commit of #204 would be appreciated @amanagr @rht 
Currently I have:
screenshot-20181218-12 32 40-cropped

rht:
^ does that mean urwid has behavior that varies across systems, or that there was a bug introduced when the pallete is being moved to a new file? I am sure I had rebased this branch.

neiljp:
@rht In your local branch, check your `buttons.py` file for `view.palette.append` lines; they are the colors that are accessed/set for the streams (which could be set elsewhere in future, perhaps). In my local rebased master branch, they set the `default` theme only, which means there should be no colors in your theme for these, unless you have other changes left over from when you modified the default theme?

rht:
I did not touch `buttons.py` at my dirty local branch.

My changes:
```
diff --git a/Pipfile b/Pipfile
index 914dc01..3de6363 100644
--- a/Pipfile
+++ b/Pipfile
@@ -4,13 +4,11 @@ url = "https://pypi.python.org/simple"
 name = "pypi"
 
 [packages]
-typing = {version="==3.6.4", markers="python_version < '3.5'"}
 urwid = "==2.0.1"
 zulip = "==0.4.7"
 emoji = "==0.5.0"
 urwid-readline = "==0.8"
 beautifulsoup4 = "==4.6.0"
-lxml = "==4.2.3"
 mypy_extensions = ">=0.4"
 
 [dev-packages]
diff --git a/requirements.txt b/requirements.txt
index 0f5d59f..cb2b2d8 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-typing==3.6.4; python_version < '3.5'
 urwid==2.0.1
 zulip==0.4.7
 pytest==3.4.2
@@ -11,4 +10,3 @@ mypy==0.641
 emoji==0.5.0
 urwid_readline==0.8
 beautifulsoup4==4.6.0
-lxml==4.2.3
diff --git a/setup.py b/setup.py
index 00184e8..dd5fab2 100644
--- a/setup.py
+++ b/setup.py
@@ -81,13 +81,11 @@ setup(
     },
     tests_require=testing_deps,
     install_requires=[
-        "typing==3.6.4; python_version < '3.5'",
         'urwid==2.0.1',
         'zulip==0.4.7',
         'emoji==0.5.0',
         'urwid_readline==0.8',
         'beautifulsoup4==4.6.0',
-        'lxml==4.2.3',
         'mypy_extensions>=0.4',
     ],
 )
diff --git a/zulipterminal/ui_tools/boxes.py b/zulipterminal/ui_tools/boxes.py
index cc6a24c..22d978e 100644
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -174,8 +174,8 @@ class MessageBox(urwid.Pile):
         bar_color = self.model.stream_dict[self.stream_id]['color']
         bar_color = 's' + bar_color[:2] + bar_color[3] + bar_color[5]
         stream_title_markup = ('bar', [
-            (bar_color, '{} >'.format(self.caption)),
-            ('title', ' {} '.format(self.title))
+            (bar_color, '{} '.format(self.caption)),
+            ('title', ' {} '.format(self.title))
         ])
         stream_title = urwid.Text(stream_title_markup)
         header = urwid.AttrWrap(stream_title, 'bar')
```
which should be mostly harmless.

rht:
Found that 24-bit (true color / absolute color) is incoming to master: https://github.com/urwid/urwid/pull/284.

neiljp:
@rht Is this PR dependent upon that support, or do you feel this is good to review/merge as-is?

This still leaves the issue of why the spaces are differently colored, and without my patch on #204?

amanagr:
@rht #204 is merged. Let me know when you fell this is ready for review.

rht:
I rebased, will retest for the last time in an hour.

rht:
Tested, works fine. Aside: I found the startup time has become >10s again.

amanagr:
Great!

PR #200, opened 2018-12-16
Add Starred messages key (f) and button

neiljp:
This completes my push to add starred message support; this was working some time ago, but is now sufficiently polished and tested to be a PR :)

I find this useful not just for marking messages as starred/unstarred (via `*` or `ctrl s` to toggle), but now also for showing those messages with just a key, and occasionally via the starred-message button.

Given the duplication in StarredButton, the last commit also is quite a simplification of similar buttons.

Also @amanagr `MenuButton` isn't used, correct? Could we remove that in a different PR?


amanagr:
This looks amazing! Thanks @neiljp

PR #202, opened 2018-12-16
Centralize shortcut keys and themes into new directory

neiljp:
This should make it easier to find the mostly-declarative 'code' for default shortcut keys and themes, which was previously split between `config.py` (keys) and `ui.py` (themes) and is now in `config/keys.py` and `config/themes.py` respectively in this PR.

Making it easier should improve in-tree customization by users/contributors.

We could also add an issue for customization via the config file, which would override the defaults?


amanagr:
Seems to a nice restructure to me. It appears there are some conflicts you should resolve.

neiljp:
@amanagr Extracted the `test_config.py` move into a separate commit to make things cleaner (separate from changing the file to pass tests in earlier commit) and also to rename to `test_keys.py` too.

amanagr:
Great! Thanks @neiljp

PR #204, opened 2018-12-18
run/core: Validate & indicate source of themes early in run, not core.

neiljp:
This PR validates theme selection before reaching the `Controller`, and includes much more error handling.

This was also motivated by theme specification on the command-line not being recognized - this PR shows where themes are being specified, both in the case of failure and success :)

I'm aware the code is all in the try block - this can be refactored out for cleanliness, but this feature seemed useful to have ASAP!

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -7,6 +7,7 @@ from typing import Dict, Any
 from os import path, remove
 
 from zulipterminal.core import Controller
+from zulipterminal.config.themes import THEMES
 
 
 def parse_args() -> argparse.Namespace:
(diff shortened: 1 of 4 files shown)


neiljp:
Requesting review from @amanagr and @rht, since Rein is working with a theme atm and may find this branch/PR useful :)

rht:
ACK; validation of CLI arguments should be localized to be at the CLI wrapper instead of the core loop.

neiljp:
Just pushed a second commit, which now means the `Controller` only has access to the current/active theme (palette). This simplifies some redirection, and makes it obvious that an issue with #199 is that existing code was setting the stream colors only in the 'default' theme/palette.
(@rht this was the issue I found)

amanagr:
@neiljp Looks great!

PR #205, opened 2018-12-19
CI: Add testing on Windows.

rht:
While Travis also has beta support for Windows, I have tested it to be slower than appveyor. Needs to be enabled at https://www.appveyor.com/.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,23 @@
+version: "{build}"
+os: Windows Server 2012 R2
+cache:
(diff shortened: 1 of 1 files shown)


neiljp:
This would be great :) @rht interestingly-named branch :)

amanagr:
ZT is not supported on windows. I don't think we need this.

amanagr:
No, it doesn't work with cygwin. I have tried it.

rht:
Alright, if you say so.

Issue #181, opened 2018-12-03
Test installation *without* dev packages?


neiljp:
From what I can tell, all the current CI tests are there to run the actual tests. As such, they install all the dependencies for testing, and don't test without these. Perhaps we need to just depend upon user reports to resolve this, but it seems there should be a better way - running headless and look for exceptions re missing packages, perhaps?

timabbott:
It'd also be reasonable to just add another suite called `production` that only installs the production dependencies.

neiljp:
@timabbott I think the problem only arises when actually running the code, but I think we're agreeing on the general form of the solution :)

neiljp:
29fa557 should ensure that installation without dev packages is reasonable, so adjusting the title of this issue to clarify this issue.

PR #206, opened 2018-12-19
[BUGFIX] Export the zulipterminal/config to site-packages.

rht:
The directory is ignored by pip without an __init__.py.


neiljp:
Is this summary suitable, to fit with the style and other messages in the log?
  `bugfix: Ensure pip installs config/ folder by adding __init__.py.`

rht:
Amended.

neiljp:
Small but important fix - thanks @rht for identifying and reporting! This could be identified in the future by something like #181, perhaps.

PR #207, opened 2018-12-19
[WIP (tests/locking?)] Update read and starred messages methods

neiljp:
This is intended to add read-flag updates sent from the server, so if messages are read in the webapp and ZT is running, the intention is that the unread counts and status in ZT should update at the same time. It also extends the capability of the similar starred feature to cover where multiple message updates occur at once, in line with the read-flag version.

I would appreciate any testing of this PR :)

CAVEATS:
* On reading many messages quickly (in the webapp), the numbers can become out of sync with ZT (is this due to some kind of locking issue? multiple messages sent from the server which aren't responded to?)
* Some stream totals are occasionally wrong on loading?
* I'd like to add some tests for the new method, but wanted to look for feedback first.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -383,6 +383,35 @@ class Model:
             self.index['messages'][message_id] = message
             self.update_rendered_view(message_id)
 
+    def update_read_status(self, event: Dict[str, Any]) -> None:
+        assert not event['all']  # FIXME Should handle
+        indexed_messages = set(self.index['messages'])
+        messages_to_mark_read = set(event['messages'])
+
+        for message_id in messages_to_mark_read & indexed_messages:
+            msg = self.index['messages'][message_id]
+            if event['operation'] == 'add':
+                if 'read' not in msg['flags']:
+                    msg['flags'].append('read')
+            elif event['operation'] == 'remove':
+                if 'read' in msg['flags']:
+                    msg['flags'].remove('read')
+            else:
+                raise RuntimeError(event, msg['flags'])
+
+            self.index['messages'][message_id] = msg
+            self.update_rendered_view(message_id)
+
+        # set_count depends on message being in the index
+        for message_id in messages_to_mark_read - indexed_messages:
+            self.get_messages(num_after=0, num_before=0, anchor=message_id)
+
+        assert all(msg_id in self.index['messages']
+                   for msg_id in messages_to_mark_read)
+
+        set_count(event['messages'], self.controller,
+                  -1*len(messages_to_mark_read))
+
     def update_star_status(self, event: Dict[str, Any]) -> None:
         assert len(event['messages']) == 1  # FIXME: Can be multiple?
         message_id = event['messages'][0]
@@ -462,6 +491,7 @@ class Model:
                     if hasattr(self.controller, 'view'):
                         self.controller.view.handle_typing_event(event)
                 elif event['type'] == 'update_message_flags':
-                    # TODO: Should also support 'read' flag changes?
                     if event['flag'] == 'starred':
                         self.update_star_status(event)
+                    elif event['flag'] == 'read':
+                        self.update_read_status(event)
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -413,10 +413,11 @@ class Model:
                   -1*len(messages_to_mark_read))
 
     def update_star_status(self, event: Dict[str, Any]) -> None:
-        assert len(event['messages']) == 1  # FIXME: Can be multiple?
-        message_id = event['messages'][0]
+        assert not event['all']  # FIXME Should handle
+        indexed_messages = set(self.index['messages'])
+        messages_to_mark_starred = set(event['messages'])
 
-        if self.index['messages'][message_id] != {}:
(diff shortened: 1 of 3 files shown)


rht:
For current master, reading messages on ZT directly update webapp's unread count. I have confirmed that it is not the case when the causal order is the other way around.

neiljp:
@rht Do you find the numbers match, or are there certain ones which become out of sync?

I'm fairly convinced that this PR would be an improvement over master, but there are definitely some further updates to make it match completely, which can perhaps be filed in a separate issue?

rht:
On this branch, I tried reading on webapp 2 messages, then toggled the stream list display with `q` then `esc`. Didn't see any update even after 20s+.

neiljp:
@rht Do you have a comparison with master and this branch? It is difficult to duplicate without running a server, perhaps? Also, I've found it may work better when I had the current autohide code reverted, so I'm not sure if this branch plus the one I pushed recently for better autohide behavior (particularly if autohide is disabled using it) may work better, though of course we want it to work with autohide too!

zulipterminal/model.py
```diff
@@ -383,11 +383,41 @@ def update_reaction(self, response: Dict[str, Any]) -> None:
...
+                    msg['flags'].remove('read')
+            else:
+                raise RuntimeError(event, msg['flags'])
+
+            self.index['messages'][message_id] = msg
+            self.update_rendered_view(message_id)
+
+        # set_count depends on message being in the index
+        for message_id in messages_to_mark_read - indexed_messages:
+            self.get_messages(num_after=0, num_before=0, anchor=message_id)
```
timabbott: Generally, it's not great to do potentially dozens/hundreds of API queries to the server like this; for one thing, you'll hit server rate limits.  I think the correct thing to do here for marking messages as unread would be potentially to follow the webapp's system of having ultra-thin object for unread count tracking only for messages you know exist and are unread but don't have a copy of.  

The other option would be to add a bulk version of this query (`get_messages_by_id`, e.g.), but that seems wrong since the client doesn't actually need a heavy version of those messages.  Note that it's very possible to mark 1000+ messages as read all at once.

amanagr: @neiljp we are already tracking unread counts for all the messages (received or not in zt) via `model.unread_counts` . We can update the new counts in `model.unread_counts` and update them accordings via `set_count` in zt. I also don't think we need to do these many queries to the server.

rht: Is this the error caused by the rate limit or something else?
```
  File "/home/rht/code/venv/lib/python3.7/site-packages/zulip_term-0.2.1+git-py3.7.egg/zulipterminal/cli/run.py", line 146, in main
    Controller(zuliprc_path, THEMES[theme_to_use[0]]).main()
  File "/home/rht/code/venv/lib/python3.7/site-packages/zulip_term-0.2.1+git-py3.7.egg/zulipterminal/core.py", line 36, in __init__
    self.view = View(self)
  File "/home/rht/code/venv/lib/python3.7/site-packages/zulip_term-0.2.1+git-py3.7.egg/zulipterminal/ui.py", line 35, in __init__
    super(View, self).__init__(self.main_window())
  File "/home/rht/code/venv/lib/python3.7/site-packages/zulip_term-0.2.1+git-py3.7.egg/zulipterminal/ui.py", line 90, in main_window
    self.center_column = self.message_view()
  File "/home/rht/code/venv/lib/python3.7/site-packages/zulip_term-0.2.1+git-py3.7.egg/zulipterminal/ui.py", line 43, in message_view
    self.search_box)
  File "/home/rht/code/venv/lib/python3.7/site-packages/zulip_term-0.2.1+git-py3.7.egg/zulipterminal/ui_tools/views.py", line 278, in __init__
    msg_list = MessageView(model)
  File "/home/rht/code/venv/lib/python3.7/site-packages/zulip_term-0.2.1+git-py3.7.egg/zulipterminal/ui_tools/views.py", line 35, in __init__
    self.set_focus(self.focus_msg)
  File "/home/rht/code/venv/lib/python3.7/site-packages/urwid/listbox.py", line 581, in set_focus
    self._body.set_focus(position)
  File "/home/rht/code/venv/lib/python3.7/site-packages/urwid/listbox.py", line 238, in set_focus
    self.focus = position
  File "/home/rht/code/venv/lib/python3.7/site-packages/urwid/monitored_list.py", line 165, in _set_focus
    raise IndexError('focus index is out of range: %s' % (index,))
IndexError: focus index is out of range: 10000000000000000
Zulip Terminal has crashed!
You can ask for help at:
https://chat.zulip.org/#narrow/stream/206-zulip-terminal
```

amanagr: @rht this is different issue which is fixed on master.

neiljp: @amanagr @timabbott This was intended as a v1 towards this approach, and I agree we can work on a simplifier approach which only stores the unread flags. I'll look at that for 0.3.1+ I think now?

amanagr: I think creating an issue for it so that it doesn't lose track would be nice.  (Also commenting it in the code itself would be great if are doing it in a later commit.)

PR #211, opened 2018-12-21
mypy: Stricter checks, exclude some files & add Index TypedDict

neiljp:


amanagr:
The code looks much clearer now. Thanks @neiljp !

neiljp:
@amanagr The followup is on the way, maybe late today for me.

PR #213, opened 2018-12-22
Refactoring: indexing

neiljp:
This started as a follow-up to #211 - hence the `initial_index` commit, which avoids a need for `Optional[Index]` and lots of asserts - as well as being based on some older index-encapsulation work I had in another branch.

Subsequent commits are to:
- simplify `get_messages` (and avoid an error case) by returning `None`
- remove an unnecessary internal `index` being tracked in a view
- require the index parameter in `index_parameters` to be an `Index` (since it's initialized already)
- simplify (hopefully!) `set_narrow` via use of `locals()`, and a `dict` for the names/narrows
- simplify core (`Controller`) narrowing methods by using the `model.get_message_ids_in_current_narrow` method
- remove an unnecessary reset of the search index in `boxes.py`.

This works towards simplifying index access, behavior and encapsulation.

This does not explicitly add new features, but improves typing, some edge cases, and uses existing functions rather than digging into data structures directly, which should all make the code easier to read. For example, the narrowing functions are much clearer now IMO, and could possibly be simplified further based on some work I did in another branch locally.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,12 @@
 from collections import defaultdict
+from copy import deepcopy
 from typing import Any, Dict
 
 import pytest
 
 from zulipterminal.ui_tools.boxes import MessageBox
 from zulipterminal.ui_tools.buttons import StreamButton, UserButton
+from zulipterminal.helper import initial_index as helper_initial_index
 
 
 @pytest.fixture(autouse=True)
@@ -381,6 +383,11 @@ def initial_data():
         'last_event_id': -1,
         'muted_topics': [],
     }
+
+
+@pytest.fixture
+def initial_index():
+    return deepcopy(helper_initial_index)
 
 
 @pytest.fixture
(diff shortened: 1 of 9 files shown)


amanagr:
@neiljp I have gone through the code line by line, but I don't really know if there are any faults in this refactoring, we can merge this and fix bugs we find along the way. Overall, looks like a good refactoring to me.

neiljp:
@amanagr The idea is that this can be read on a per-commit basis and be reasonable to read/check :)

That said, I generally work on the principle that each commit should pass tests individually, too, not just per branch. Some of the changes (eg. the `set_narrow` adjustment) were adjusted from my initial attempts, in order to pass the tests - when they're in place, they're really helpful for testing :)

Moving to using the pre-existing functions should be good too, since it's on a like for like basis.

amanagr:
That sounds great to me.

PR #214, opened 2018-12-23
Improved autohide behavior & zuliprc configuration option

neiljp:
The new autohide (the first commit) involves the following changes:
- Users list returns to the right side
- Initial view defaults to showing streams & messages, focused on home
- GO_LEFT/GO_RIGHT keys move into Streams/Users lists respectively
- Streams and Users lists autohide when moving away from them
- Two View methods are used to show/hide Streams & Users (left/right panels)
- View parameter added to MiddleColumnView __init__ for show/hide calls

The second commit then adds similar option behavior as recently added for themes, but to enable or disable the autohide behavior.

Try this out and let me know what you think :) It feels a lot more natural to me, but then again I wrote it!

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -23,9 +23,10 @@ class TestController:
     def controller(self, mocker) -> None:
         self.config_file = 'path/to/zuliprc'
         self.theme = 'default'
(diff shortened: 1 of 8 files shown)


amanagr:
@neiljp looks wonderful to me. merged!

PR #215, opened 2018-12-26
Link to CHANGELOG from README & add further entries for 0.3.0 release

neiljp:
This seems correct, but please check if this makes sense for those who have contributed :)

We should only merge this just before releasing, as it updates CHANGELOG to switch 'since last update' to '0.3.0'.


amanagr:
Merged! Thanks @neiljp ! I will be doing a release soon.

PR #218, opened 2018-12-27
[WIP] views: Do not render inactive users.

rht:
Will add the config to enable/disable this later. I haven't figured a nonintrusive way to wire the config to view, since the config dict is used only at run.py. (would preferably want not rendering inactive to be the default tho)

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -418,6 +418,8 @@ class RightColumnView(urwid.Frame):
         for user in self.view.users:
             unread_count = self.view.model.unread_counts.get(user['user_id'],
                                                              0)
+            if unread_count <= 0:
+                continue
             self.users_btn_list.append(
                 UserButton(
                     user,


rht:
I once tested the speedup for czo to be 78%, but should be less pronounced for small-to-mid realms.

amanagr:
@rht I feel we should work towards a `buddy list` like one the mobile app implements. Perhaps we should discuss this a little on czo.

rht:
This is blocked on me: I haven't fully tested the fix that removes the redundant users list fetching. So far I have timed things by manually placing `time()` in several places. If there is a more systematic method to test line by line, it would have been done much earlier.

However, I recalled that this PR still considerably speeds up the rendering of the whole screen when I tested on top of master with the removing-the-redundant-users-list-fetching fix.

neiljp:
@rht The fix in master doesn't fetch a reduced number of users - I'm not sure that's possible, exactly; rather, it doesn't show 'inactive' users. See `get_all_users` in `model.py`.

rht:
@neiljp what I had in mind was this commit: https://github.com/zulip/zulip-terminal/commit/7ea4475d3323ecbc8c4c114b8f0dcd7204068108.
I didn't know that https://github.com/zulip/zulip-terminal/commit/d7fc64e6097c4d7801402556889cd26e27020455 had since made such that inactive are not rendered -- I have not benchmarked this.

rht:
This PR is definitely obsolete since https://github.com/zulip/zulip-terminal/commit/d7fc64e6097c4d7801402556889cd26e27020455.

PR #226, opened 2018-12-29
model: Remove unnecessary fetching of realm_users data twice.

amanagr:
"realm_user" was added to the initial register call to server but
client.get_members call wasn't removed which basically fetches the
data already included in "realm_user".


amanagr:
@rht this might be the cause of slow startup time you were experiencing.

rht:
Cool, I will check.

neiljp:
@amanagr @rht I doubt this will make much difference, as this is simply removing an extra IO which is running in parallel with a larger one? Unless the extra 'accounting' around the IO is really that large, that is.

PR #229, opened 2018-12-30
Centralize client.register calls into Model & combine

neiljp:
Currently there is a `fetch_event_types` register call in `Model`, and an `event_types` register call in `Controller` (in core). This PR brings the latter into the `Model`, and then combines them to use a single `register` call to the API. The method moved across is maintained for re-use in the case of the server queue being garbage-collected, with the `event_types` data extracted into the top of the `Model` itself.

This was discussed in this topic:
https://chat.zulip.org/#narrow/stream/206-zulip-terminal/subject/Fetching.20realm_users/near/677096

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -24,8 +24,6 @@ class TestController:
         self.config_file = 'path/to/zuliprc'
         self.theme = 'default'
         self.autohide = True  # FIXME Add tests for no-autohide
-        mocker.patch('zulipterminal.core.Controller.'
-                     'register_initial_desired_events')
         return Controller(self.config_file, self.theme, self.autohide)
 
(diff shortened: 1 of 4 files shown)


timabbott:
Looks reasonable.  I'd also encourage moving the per-event processing logic into its own function/file separate from the network protocol stuff; that was a very helpful change for readability in the webapp.

neiljp:
@timabbott I've been thinking of moving the python-zulip-api & other endpoint code into a separate class, since it's not ideal to have it together for testing, if nothing else - was that the change you meant?

amanagr:
@neiljp seems like a nice idea to move the api stuff into a separate class. I should have done it in the first place. 😸

neiljp:
@amanagr Good to hear this sounds like a good idea :)

PR #230, opened 2018-12-30
view/model : Set preferences and user status for users panel as done by webapp.

amanagr:
This commit introduces `offline` and `inactive` users whose
difference are specifiend in `get_all_users` function.

Also, it modifies `users_list` to only display `offline`, `active`
and `online` users by default.

When the user searches for keyword then `inactive` users are also
included in the search.


amanagr:
@neiljp I would love some feedback on this. I am working on tests for this but should be good enough for you to run locally and give some feedback.

amanagr:
Yeah! I will get to it soon.

amanagr:
@neiljp I am thinking of doing the color highlight thing in a different PR since this PR has a different purpose.

neiljp:
@amanagr This looks to work fine overall :) Did you change it much from the original?

It is difficult to review (retrospectively, since you merged it) since it affects a lot of files in ways that feels unrelated. For example, the `user_id` fixture could have been a separate commit (and potentially improved the existing tests/fixtures which use it). I try to keep each commit minimal and focused. It's tempting but I try not to move lines around just to tidy up, when mainly changing other code. I will go back and isolate features into smaller commits, if I do a lot of prototyping work in one large commit.

amanagr:
@neiljp thanks for the suggestions. I will keep them in mind. 😄 

I don't think I changed the original code from your last review (and hence I thought you have already given it a pass).

I was thinking of breaking the commits but then it was hard to make each commit pass all tests. Regardless, I will try to split them from now on!

neiljp:
@amanagr Essentially, I can't do a final review of the code, when it's not in it's final form, whether potentially split into smaller commits, or without additional code (including test code). I thought you were going to split it into smaller commits, and didn't know if you would change it after adding the tests, so couldn't really review it in it's final state, at that earlier point ;)

Thinking about how the tests pass for smaller commits can help with how the commits can be split, or the order of them. For example, the `user_id` fixture is independent of the rest of the code, so could easily be extracted into an earlier commit, and doing that then searching for use of that number could make it a good standalone commit - ie. migrate uses of a number to a named user_id for the user of the app (logged in user). Similarly, I might have placed the model adjustments first, and tested them (both in one commit), then added the UI stuff in another commit.

amanagr:
It seems a lot easy after you put it that way. 😅
(I have surely got into the bad habit of requesting reviews before getting PR into its final form. )

neiljp:
There's no harm in asking for an early (preliminary) review/feedback, but it's just that - preliminary :)

Given users can always push an update to a PR, then it can totally invalidate the earlier review, because anything could have changed ;) Hopefully it hasn't too-signifcantly, but if you want a re-review then just ask :)
(hopefully I've followed this advice, too; there are cases where eg I've split the commit up where it is simple enough, though arguably that could be re-reviewed to ensure it still looks right)

amanagr:
Sure. Thanks!

neiljp:
@amanagr Another point to add - you could have included an update to the CHANGELOG since this is a very visible change to the UI for the next release :)

amanagr:
Oops! I will do it in a separate commit now.

Issue #225, opened 2018-12-28
UI Improvements


amanagr:
Left sidebar:
- [x] could be wider on wider screens. Ideally stream names that fit on web should be fully visible on ZT left sidebar as well
- [x] stream name overflow should indent or I think I would just do: `very very long stream na...`
- [x] I would glue the unread count to the right edge of the sidebar, and make it more visible.
- [x] Blinking cursor is confusing (the fact that you get a blinking cursor) -- urwid doesn't provide option to turn off blinking cursor or hide cursor, could use more deep diving.

- [x] Blue highlight makes the stream name hard to read on some terminal profiles
- [ ] I think it's a bit confusing that hitting "Enter" on a stream name removes the left sidebar

Help menu
- [x] `?` menu: too wide, and the break in `_____    _____` is confusing. I would also reorder this a bit; we can chat separately.
- [ ] This is a big project but I think super high valuable: make the default suggestion context-dependent. 
- [x] I would make the default suggestion highlight `n` and `r` instead of `@`

Message feed:
- [ ] Grey highlight for message in focus is not great; it makes the message too unreadable
- [ ] For replying to a message, I would consider making the top of the compose box something like `zulip-terminal > **feedback**`

neiljp:
@amanagr I think we've only covered the 'glue unread to right edge' task so far?

If you consider the blinking cursor issue fixed, then could you amend the docs to specify that this is expected to be resolved in the terminal app/system end? I think this should still be able to controlled, and is commented to be a missing feature in urwid as per urwid/urwid#170.

amanagr:
Ofcourse, we could do more diving into the blinking cursor part. Also, I think the left sidebar is wide enough.

neiljp:
Blinking cursor & blue highlight contrast should be solved by the recent merge of #281.

neiljp:
"I would make the default suggestion highlight n and r instead of @"! is now tracked in #290, after clarification on #zulip-terminal on chat.zulip.org

omi10859:
> I think it's a bit confusing that hitting "Enter" on a stream name removes the left sidebar 

It is confusing.
Should I make PR to change the behaviour?

neiljp:
> > I think it's a bit confusing that hitting "Enter" on a stream name removes the left sidebar
> 
> It is confusing.
> Should I make PR to change the behaviour?

[...]

PR #231, opened 2018-12-31
UI: Right-align unread-counts, for current fixed-width left-column-view.

neiljp:
This goes some way to supporting #225, though only for a fixed-width left-panel (leftcolumnview).

Most of the effort here is passing through the width; in an ideal world we might have a `Column` widget instead, though we can actually squeeze more space out by compacting the total width, in principle.


neiljp:
Pushed after splitting the commits and adding a second CHANGELOG entry for the extra commit.

neiljp:
@amanagr Do you mean the terminal application? I cannot do that here (in the application at least, stty maybe?), and I'm not sure we want it turned off in all cases, eg. when entering text?

amanagr:
Most of the terminals provide this preference, well this [link](https://unix.stackexchange.com/questions/3759/how-to-stop-cursor-from-blinking) provides some interesting alternatives if you want to try out.

PR #234, opened 2019-01-02
Simplify model event handling

neiljp:
This PR was motivated after examining how I might extract the server-handling logic as discussed previously, and having the event-to-action mappings separate already feels cleaner, as well as supporting that goal if we choose to pursue it.

This PR is focused on the model, simplifying the incoming (from server) event logic dispatch, and making it clearer which methods can be used as event-handlers based on their type annotations.

This includes bringing the typing-handler code into the Model, while leaving the view-specific code in a separate method in the View.

Tests should be maintained, and coverage slightly up.


amanagr:
@neiljp This looks awesome. Merged!

PR #235, opened 2019-01-02
Right panel count alignment and tests

neiljp:
Main feature: adjust styling in user-list to align unread counts and indent slightly, in line with left column.

Additionally:
* some renaming/removal of internal variables in ui.py
* tests for `show_left_panel` and `show_right_panel`


amanagr:
Looks Good!

PR #240, opened 2019-01-04
Improve error handling in ZT.

amanagr:


amanagr:
@neiljp what do you think about the third commit?

zulipterminal/model.py
```diff
@@ -282,8 +282,8 @@ def _update_initial_data(self) -> None:
                 )
             except zulip.ZulipError as e:
                 raise ServerConnectionFailure(e)
-            except Exception:
-                print("Invalid API key")
+            if response['result'] == 'error':
+                print("\n\033[91m" + response['msg'] + "\033[0m")
```
neiljp: I didn't change it when I added the other exception case, since this error handling was already present, but we should probably handle the error in core.py or run.py, with the `ServerConnectionFailure` perhaps?

It feels like a bit of a 'code smell' to have `urwid` referenced here.

zulipterminal/model.py
```diff
@@ -281,9 +281,11 @@ def _update_initial_data(self) -> None:
                     apply_markdown=True,
                 )
             except zulip.ZulipError as e:
+                self.controller.loading = False
```
neiljp: It would be simpler to handle this in the `Controller` rather than the `Model`, after we've raised an exception; feel free to chat if you want to discuss this.

zulipterminal/core.py
```diff
@@ -24,6 +24,10 @@ class Controller:
 
     def __init__(self, config_file: str, theme: ThemeSpec,
                  autohide: bool) -> None:
+
+        # Is False when ZT's UI has been loaded without any errors.
+        self.loading = True  # type: bool
```
neiljp: I would make this variable name more specific - not only does naming something this simply make it difficult to know where it might be referenced, but using eg. `self.ui_loaded = False` or `self.ui_loaded_successfully` (note the inverse case) probably makes the comments unnecessary?

tests/model/test_model.py
```diff
@@ -3,6 +3,7 @@
 from typing import Any
 
 import pytest
+import urwid
```
neiljp: This use of urwid in the Model makes me question the `ExitMainLoop` being handled here, as I said in the comment on the previous commit. Also, does the addition of this test belong with changes made to add `loading` variable? It doesn't seem obviously connected.

zulipterminal/cli/run.py
```diff
@@ -174,11 +174,14 @@ def main() -> None:
     except Exception as e:
         if args.debug:
             # A unexpected exception occurred, open the debugger in debug mode
-            import pudb
-            pudb.post_mortem()
 
-        sys.stdout.flush()
-        traceback.print_exc(file=sys.stderr)
+            sys.stdout.flush()
```
neiljp: Should this line go with the spinning cursor code?

zulipterminal/cli/run.py
```diff
@@ -174,11 +174,14 @@ def main() -> None:
...
+            traceback.print_exc(file=sys.stderr)
+            run_debugger = input("Run Debugger? (y/n): ")
```
neiljp: I've not really used the debugger much; this now *only* shows the traceback if debug mode is enabled?

Showing the traceback was discussed as ok previously, I think? We can discuss this, but what do you think to leaving `-d` automatically go into the debugger, but tracebacks being logged to files (maybe with dates in their name)? We could add some text to request a description, which could be put in the top of the traceback file, and then ask for that file to be sent/posted to czo or some email that would post to a czo stream?

amanagr: Yup, this definitely seems like a thing to do. 👍

zulipterminal/core.py
amanagr: Yeah, makes sense

zulipterminal/model.py
amanagr: The `client` (i.e. python-zulip-api) doesn't crash unless there's no internet from my past experience. So, I think we can let this pass.

zulipterminal/cli/run.py
neiljp: To clarify, my thoughts would be to isolate the tracebacks, but offer the user a chance to write what they were just doing, as a way to annotate the issue - and that shouldn't depend on debug mode being enabled. I think it would be useful to keep `debug.log` as a general debugging tool, separate from crash-reporting?

amanagr:
@neiljp I have updated this PR you can have a look.

zulipterminal/cli/run.py
neiljp: You're going to address in a future version?

zulipterminal/core.py
```diff
@@ -262,6 +262,15 @@ def _finalize_show(self, w_list: List[Any]) -> None:
...
 
+    def exit_with(self, exception_to_raise: Any,
+                  error: Any=None) -> None:
+        if exception_to_raise is None:
+            raise urwid.ExitMainLoop()
+        elif error is None:
+            raise exception_to_raise()
+        else:
+            raise exception_to_raise(error)
+
```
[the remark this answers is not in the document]
amanagr: Since every time an we have to close ZT when an exception is raised we also need to set `self.ui_loading = False`. This function helps us not repeat `self.ui_loading = False` everywhere. Also, it stops us from using `urwid` in model.py which you were strongly against.

zulipterminal/cli/run.py
amanagr: I think we should leave it there to print whatever the message is in the buffer before printing traceback?

zulipterminal/core.py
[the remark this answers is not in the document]
amanagr: I think the spinning cursor is better suited in controller since it marks the beginning of ZT really and not UI.

As for the `ui_loading` I think we can do that.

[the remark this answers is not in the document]
amanagr: @neiljp  I think perhaps this is what you meant by catching the exception in `core.py` right?

zulipterminal/cli/run.py
```diff
@@ -185,11 +185,14 @@ def main(options: Optional[List[str]]=None) -> None:
     except Exception as e:
         if args.debug:
             # A unexpected exception occurred, open the debugger in debug mode
-            import pudb
-            pudb.post_mortem()
 
-        sys.stdout.flush()
-        traceback.print_exc(file=sys.stderr)
```
neiljp: With this change, if debug is *not* enabled then an unexpected exception will cause what appears to be a regular exit to a user, I think - which seems rather undesirable? Could you add something like an `in_color('red', "zulip-term exited unexpectedly")`, perhaps with reference to the final log-file name we decide upon, if one was created?

zulipterminal/core.py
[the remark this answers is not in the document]
neiljp: Hmm, I agree that the spinning cursor is not part of the UI and that the controller handles things that aren't UI-related, but it also writes to the terminal rather than the UI, and to me that seems within the remit of `run.py` - opinions may differ :)

I'm still not convinced what we gain by using `exit_with`; I would split that out and just leave the `ui_loading` changes in this commit, and put the rest into another commit/branch for now? We can discuss on czo about it?

For the "catching the exception in `core.py`", my recollection was that a solution might be to wrap some code in  `Controller.__init__` inside a `try`, and then on noticing an exception adjust `ui_loading` and re-raise the exception so that `run.py` can handle it.

zulipterminal/cli/run.py
```diff
@@ -181,11 +185,13 @@ def main(options: Optional[List[str]]=None) -> None:
     except ServerConnectionFailure as e:
         print(in_color('red',
                        "\nError connecting to Zulip server: {}.".format(e)))
+        logging.exception(e)
```
neiljp: We've handled the exception here - why do we need to log it?

zulipterminal/cli/run.py
```diff
@@ -181,11 +185,13 @@ def main(options: Optional[List[str]]=None) -> None:
...
         sys.exit(1)
     except Exception as e:
+        logging.exception(e)
```
neiljp: This looks good, to resolve the issue I pointed at in the first commit. I would probably have moved this commit to before the first, or possibly squashed the two together

Also, if each traceback will append to this file, should we have a separator to distinguish between tracebacks from different events?

zulipterminal/cli/run.py
```diff
@@ -5,11 +5,15 @@
 import tempfile
 from typing import Dict, Any, List, Optional
 from os import path, remove
+import logging
 
 from zulipterminal.core import Controller
 from zulipterminal.model import ServerConnectionFailure
 from zulipterminal.config.themes import THEMES
 
+LOG_FILENAME = 'debug.log'
```
neiljp: We can improve upon this further in future, but for now what do you think about having this be something more specific such as `zulip-terminal-tracebacks.log` instead? The file will be placed in the current directory, which could be anywhere, so `debug.log` would not be as clear to end-users on its own? Also `debug.log` is used for actual debugging during development so keeping them distinct could be useful?

zulipterminal/cli/run.py
amanagr: I think a full traceback of even a handled exception could be useful. right?

amanagr:
@neiljp I have updated this PR with the error handling code aside. I will most probably open a PR for `ui_loading` with the unread_count code.

Thanks for the detailed reviews in this PR.

PR #242, opened 2019-01-06
Limit captions depending on length

### LINKED ISSUE #237: Users list: Names that are too long should be truncated to end with '...'
On webapp:
2019-01-03-233137_348x50_scrot

Current ZT:
2019-01-03-233255_322x63_scrot

neiljp:
This resolves #237, the second bullet of #225, and also applies the same logic to the Buttons derived from `TopButton`.

I've introduced more test code for each first, then extended them when adding in the truncation logic.

There is also one ui adjustment commit which I've included - it removes the last hard-coded width=25 for the left- or right-hand panels, which could cause strange behavior if these were *not* 25 and autohide was enabled.


neiljp:
@rht You reported the original issue; does this PR branch look good for you now? You can also explore adjusting `View.LEFT_WIDTH` or `View.RIGHT_WIDTH` and see the effects further.

rht:
`LEFT_WIDTH` set to 10:
2019-01-06-230302_146x225_scrot
Right sidebar with long name with default `RIGHT_WIDTH`:
2019-01-06-230410_318x94_scrot

neiljp:
@rht The small left-width likely affected the top-left buttons too?

rht:
Right, they are all truncated with '..'.

amanagr:
@neiljp this looks great! :)

PR #244, opened 2019-01-07
Add preliminary CLI tests

neiljp:
The CLI (run.py) code is currently completely untested. I initially started with some 'black box' testing by running it on the command-line using Popen, but that severely limits the ability to mock. Instead, I've slightly adjusted the function signature of `main` and `parse_args` to explicitly pass in either specific options (for testing) or `sys.argv`, enabling things to broadly still just work with `setup.py`.

The latter two commits then add tests for -h/--help and connection failure respectively.

The main intent here is to enable further (and indicate examples through existing) tests to be added based on these.

This does reduce the code coverage, since it adds `cli/*` back into the coverage being checked, though not by much.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,6 @@ ignore_missing_imports = True
 [coverage:run]
 branch = True
 omit =
-    zulipterminal/cli/*
 #     a/b.py,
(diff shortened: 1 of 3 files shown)


amanagr:
Looks good!

PR #245, opened 2019-01-08
Refactor Buttons by extending general features into TopButton & inheriting 

neiljp:
This continues some of the previous simplification of inheriting from `TopButton`, and extends it for `StreamButton` and `UserButton`. This dramatically reduces the duplicated widget code, which was clear when adding the 'truncated text' code, at the slight code of adding some parameters for the 'prefix character' (bullet, '#', 'P', etc, which can be colored), accounting for muted counts, and overall coloring (as per UserButton). I can see this being useful to easily specify eg. different icons for home/starred etc in future if go choose to go that way, though the motivation at this point is to simplify.

**NOTE** The last commit I am a little less sure about, and would appreciate feedback on. It ensures that a click and pressing `ENTER` do the same thing; currently a click on a button does not autohide, it just narrows, leaving any side panel still visible. The difference in behavior can be strange, but if users are clicking then the left/right panels now, with autohide as standard, then they may expect the window to disappear anyhow?

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1159,48 +1159,73 @@ class TestMessageBox:
 
 
 class TestTopButton:
+    @pytest.mark.parametrize('prefix', [
+        None, '\N{BULLET}', '-', ('blue', 'o'),
+    ])
     @pytest.mark.parametrize('width, count, short_text', [
         (8, 0, 'c..'),
         (9, 0, 'ca..'),
+        (9, -1, 'c..'),
         (9, 1, 'c..'),
         (10, 0, 'cap..'),
+        (10, -1, 'ca..'),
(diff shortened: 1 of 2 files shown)


neiljp:
(of course, one outstanding issue not explicitly in here, is that `TopButton` may no longer be the best of names, other than it is 'top' in the button hierarchy!)

amanagr:
@neiljp This looks fine to me other that the last commit. I think the current behaviour is more preferable, but I don't really know. I will go with it for now. Perhaps in the future when we get more feedback by our users can tune it accordingly. Also, matching the `enter` and `click` behaviour could solve some unknown bugs since we don't explicitly handle the `click` cases.

neiljp:
@amanagr As I said, I was unsure about this; we can consider it in the future? I have opened #248 to track this.

PR #250, opened 2019-01-09
Starred update generalization

neiljp:
This partially covers the work started in the not-merged #207, in smaller blocks, making it easier to cover the read-message-flags case from this improved point.

This:
* generalizes the method (including name) to deal with all message-flag updates
* generalizes the method internals to work with a general flag-to-change text
* extends the handling to cope with multiple message ids
* ensures we fail fast if we receive an (unhandled as yet) request to flag 'all' messages


amanagr:
Looks like a good first step in this direction!

PR #251, opened 2019-01-12
Test theme completeness

neiljp:
This work arose out of observing users use the older/incomplete light and blue themes. Hence
* the first commit checks whether themes are complete (have all styles) for the CI; two are currently marked as incomplete, which causes them to be expected failures (and so OK, for now)
* the addition of `complete_and_incomplete_themes` (2nd commit) allows `run.py` to determine complete themes at runtime, and consequently give feedback (a WARNING) on potential quality issues which might be seen

Other than those explicit changes, this includes quite a few other changes, and tests, including
* tests for `in_color` (introduced recently)
* extension of `in_color` to other colors (including yellow, subsequently used for the added WARNING
* add a simple `all_themes` method to `themes.py`; this is minimal, but helps mocking when testing
* some test refactoring

Hopefully this is rather straightforward, and should improve the user experience :)

As and when the existing and/or new themes are being developed, we can amend the parameters so that tests are not expected to fail, of course ;)

### DIFF
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -41,17 +41,21 @@ def test_main_help(capsys, options):
     assert captured.err == ""
 
 
-def test_valid_zuliprc_but_no_connection(capsys, mocker, tmpdir,
+@pytest.fixture
+def minimal_zuliprc(tmpdir):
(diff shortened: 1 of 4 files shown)


amanagr:
@neiljp this was a much need  PR. Thanks!

Issue #247, opened 2019-01-09
Crash creating new stream message in streams with no previous conversation.


sumanthvrao:
The `c` Hot Key works in creating a new stream message works in streams which has previous conversations.  However the terminal crashes when used in streams which do not have any previous conversation. Expecting the Key to create a new message with stream set to current one and topic blank. 

Steps to reproduce bug in development environment (newest version) :
1. logged in as Iago@zulip.com
2. Use `c` in to create new messages in streams with no convo like #sales, #support 
3. ZT crashes with the following error message 
```
File "/home/sumanthvrao/.local/share/virtualenvs/zulip-terminal-sRIAC_7x/lib/python3.6/site-packages/urwid/container.py", line 1429, in _set_focus_position
    raise IndexError("No Pile child widget at position %s" % (position,))
IndexError: No Pile child widget at position 0
Zulip Terminal has crashed!
```
A workaround which avoids the error is to change `Stream` in message header (to #sales for example) when `c` is used to create new messages in streams with previous conversation.

amanagr:
@sumanthvrao thanks for reporting this.

PR #252, opened 2019-01-12
view: Fix support for HotKeys in new streams.

sumanthvrao:
c HotKey used to crash zt when used in streams
with no previous messages. c now calls stream_box_view
while opening footer and sets name to stream name.

Fixes: #247.

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -135,24 +135,29 @@ class MessageView(urwid.ListBox):
                 return key
 
         elif is_command_key('SCROLL_TO_TOP', key) and not self.old_loading:
-            if self.focus_position == 0:
+            if self.focus is not None and \
+               self.focus_position == 0:
                 return self.keypress(size, 'up')
             else:
                 return super(MessageView, self).keypress(size, 'page up')
 
         elif is_command_key('SCROLL_TO_BOTTOM', key) and not self.old_loading:
-            if self.focus_position == len(self.log) - 1:
+            if self.focus is not None and \
+               self.focus_position == len(self.log) - 1:
                 return self.keypress(size, 'down')
             else:
                 return super(MessageView, self).keypress(size, 'page down')
 
         elif is_command_key('THUMBS_UP', key):
-            self.model.react_to_message(self.focus.original_widget.message,
-                                        reaction_to_toggle='thumbs_up')
+            if self.focus is not None:
+                self.model.react_to_message(
+                    self.focus.original_widget.message,
+                    reaction_to_toggle='thumbs_up')
 
         elif is_command_key('TOGGLE_STAR_STATUS', key):
-            message = self.focus.original_widget.message
-            self.model.toggle_message_star_status(message)
+            if self.focus is not None:
+                message = self.focus.original_widget.message
+                self.model.toggle_message_star_status(message)
 
         key = super(MessageView, self).keypress(size, key)
         return key
@@ -331,20 +336,28 @@ class MiddleColumnView(urwid.Frame):
 
         elif is_command_key('REPLY_MESSAGE', key):
             self.body.keypress(size, 'enter')
-            self.set_focus('footer')
-            self.footer.focus_position = 1
+            if self.footer.focus is not None:
+                self.set_focus('footer')
+                self.footer.focus_position = 1
             return key
 
         elif is_command_key('STREAM_MESSAGE', key):
             self.body.keypress(size, 'c')
+            # For new streams with no previous conversation.
+            if self.footer.focus is None:
+                stream_id = self.model.stream_id
+                stream_dict = self.model.stream_dict
+                self.footer.stream_box_view(
+                    caption=stream_dict[stream_id]['name'])
             self.set_focus('footer')
             self.footer.focus_position = 0
             return key
 
         elif is_command_key('REPLY_AUTHOR', key):
             self.body.keypress(size, 'R')
-            self.set_focus('footer')
-            self.footer.focus_position = 1
+            if self.footer.focus is not None:
+                self.set_focus('footer')
+                self.footer.focus_position = 1
             return key
 
         elif is_command_key('NEXT_UNREAD_TOPIC', key):


zulipterminal/ui_tools/views.py
```diff
@@ -337,6 +337,11 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
 
         elif is_command_key('STREAM_MESSAGE', key):
             self.body.keypress(size, 'c')
+            # For new streams with no previous conversation.
+            if self.footer.focus is None:
+                self.footer.stream_box_view(
+                    caption=self.model.stream_dict
+                    [int(self.model.stream_id)]['name'])
```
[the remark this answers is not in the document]
sumanthvrao: True. its not needed.

zulipterminal/ui_tools/views.py
```diff
@@ -337,6 +337,11 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
 
         elif is_command_key('STREAM_MESSAGE', key):
             self.body.keypress(size, 'c')
+            # For new streams with no previous conversation.
+            if self.footer.focus is None:
+                self.footer.stream_box_view(
+                    caption=self.model.stream_dict
```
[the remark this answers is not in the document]
sumanthvrao: Something like this?
`current_stream_id = self.model.stream_id`
followed by caption using : 
`caption = self.model.stream_dict[current_stream_id]['name']`

zulipterminal/ui_tools/views.py
```diff
@@ -331,18 +331,28 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
```
neiljp: You seem to have an extra line, which I'm guessing you don't need, at the bottom of the commit text - was this from a squashed commit?

zulipterminal/ui_tools/views.py
```diff
@@ -331,18 +331,28 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+            if self.footer.focus is None:
+                return key
             self.set_focus('footer')
             self.footer.focus_position = 1
             return key
 
         elif is_command_key('STREAM_MESSAGE', key):
             self.body.keypress(size, 'c')
+            stream_id = self.model.stream_id
+            stream_dict = self.model.stream_dict
```
neiljp: These two lines are only used in the conditional (`if` statement), so can be moved into there.

zulipterminal/ui_tools/views.py
sumanthvrao: Yup this was from a squashed commit.

zulipterminal/ui_tools/views.py
```diff
@@ -331,18 +335,28 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
 
         elif is_command_key('REPLY_MESSAGE', key):
             self.body.keypress(size, 'enter')
+            if self.footer.focus is None:
+                return key
             self.set_focus('footer')
             self.footer.focus_position = 1
             return key
```
neiljp: Here the key is returned at the bottom and in the conditional; how about placing the other code into an opposite conditional and having the common return statement at the bottom?

zulipterminal/ui_tools/views.py
```diff
@@ -331,18 +335,28 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
             self.footer.focus_position = 0
             return key
 
         elif is_command_key('REPLY_AUTHOR', key):
             self.body.keypress(size, 'R')
+            if self.footer.focus is None:
+                return key
             self.set_focus('footer')
             self.footer.focus_position = 1
             return key
```
neiljp: Same point here as above.

zulipterminal/ui_tools/views.py
```diff
@@ -141,17 +141,21 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+            if self.focus is not None:
+                self.model.react_to_message(
+                    self.focus.original_widget.message,
+                    reaction_to_toggle='thumbs_up')
 
         elif is_command_key('TOGGLE_STAR_STATUS', key):
-            message = self.focus.original_widget.message
+            if self.focus is not None:
+                message = self.focus.original_widget.message
             self.model.toggle_message_star_status(message)
```
neiljp: I'm not sure if your local branch is up to date with this one, but this doesn't fix the issue. I was alerted by reading this, since message is not defined if `self.focus` is `None`, but this also fails when I try to star the message when running it (as one might expect).

Do you see the problem?

zulipterminal/ui_tools/views.py
sumanthvrao: Yes done.

zulipterminal/ui_tools/views.py
sumanthvrao: @neiljp . I am afraid i don't understand. This seems to work fine in my branch.
The normal error which causes the crash is:
```
File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/ui_tools/views.py", line 158, in keypress
    message = self.focus.original_widget.message
AttributeError: 'NoneType' object has no attribute 'original_widget'
```
Doesn't this mean that the message doesn't exist and self.focus object is None?

sumanthvrao: Oh i am sorry! I had somehow not indented the 
`self.model.toggle_message_start_status` line due to which it was failing.`
This seems to work when i tried some manual testing. Is this not the approach desired?

neiljp:
Just tested this manually and it seems to work fine for me against chat.zulip.org. Given the crashing nature of this issue/PR, I'm going to merge now.

Thanks @sumanthvrao for this work, after initially reporting the issue, particularly as it turned out to be a little bigger issue than originally expected!

For future reference, when there is one commit that fixes/adds a specific bug/feature, feel free to include a change to the CHANGELOG in the same commit, or at least the same PR :) In this case we can get to that later.

sumanthvrao:
@neiljp thanks for the merge with master. Duly noted :) will make the changes to CHANGELOG from further.

PR #253, opened 2019-01-12
 Delete the client queue before shutdown

### LINKED ISSUE #241: Make sure to delete the client queue during shutdown (C-c).
The shutdown process should include deregistering the client queue. See: https://chat.zulip.org/api/delete-queue.

realsdx:
This deletes the queue(deregister) before shutdown(^C) of the ZT.
Fixes #241

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -3,6 +3,7 @@ from typing import Any, List
 import os
 import sys
 import time
+import signal
 
 import urwid
 import zulip
@@ -263,6 +264,14 @@ class Controller:
         if focus_position >= 0 and focus_position < len(w_list):
             self.model.msg_list.set_focus(focus_position)
 
+    def deregister_client(self) -> None:
+        queue_id = self.model.queue_id
+        self.client.deregister(queue_id, 1.0)
+
+    def exit_handler(self, signum: int, frame: Any) -> None:
+        self.deregister_client()
+        sys.exit(0)
+
     def main(self) -> None:
         screen = Screen()
         screen.set_terminal_properties(colors=256)
@@ -270,6 +279,9 @@ class Controller:
                                    self.theme,
                                    screen=screen)
         self.update_pipe = self.loop.watch_pipe(self.draw_screen)
+
+        # Register new ^C handler
+        signal.signal(signal.SIGINT, self.exit_handler)
 
         try:
             # TODO: Enable resuming? (in which case, remove ^Z below)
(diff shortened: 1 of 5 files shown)


zulipterminal/core.py
```diff
@@ -262,6 +263,17 @@ def _finalize_show(self, w_list: List[Any]) -> None:
         if focus_position >= 0 and focus_position < len(w_list):
             self.model.msg_list.set_focus(focus_position)
 
+    def deregister_client(self) -> None:
+        try:
+            queue_id = self.model.queue_id
+            self.client.deregister(queue_id)
+        except Exception as e:
```
amanagr: @realsdx we don't need to a try-except statement here since we are already handling general exception already in `run.py`. you can simply deregister from the queue here.

realsdx: Okay.

amanagr:
@realsdx You should squash (combine) the last two commits.

zulipterminal/core.py
```diff
@@ -3,6 +3,7 @@
 import os
```
neiljp: You probably want to place the #241 reference in some text below the commit summary, as something like "Fixes #241", which also has the benefit that it will close that issue when this commit is merged :)

zulipterminal/core.py
```diff
@@ -264,11 +264,8 @@ def _finalize_show(self, w_list: List[Any]) -> None:
             self.model.msg_list.set_focus(focus_position)
 
     def deregister_client(self) -> None:
-        try:
-            queue_id = self.model.queue_id
-            self.client.deregister(queue_id)
-        except Exception as e:
-            print(e)
+        queue_id = self.model.queue_id
+        self.client.deregister(queue_id)
```
neiljp: As per @amanagr 's comment, this should definitely be squashed into/with the previous commit.

timabbott:
@neiljp if you need/want a python-zulip-api release as part of your work on this, ping @eeshangarg and he can make one happen.

neiljp:
@realsdx Absolutely no problem :)

neiljp:
@realsdx Do you want another review of the latest changes? Force-pushing doesn't alert subscribers to the PR, and we don't always know if you're ready even if they've been updated, so post a review request in #zulip-terminal on chat.zulip.org, or a comment on github :)

zulipterminal/core.py
neiljp: @realsdx This is optional, but I noticed you didn't have this, so just a reminder in case you'd like to.

neiljp:
Merged! Thanks @realsdx - looking forward to work on tests for this :)

PR #255, opened 2019-01-13
BUGFIX: boxes: Avoid crash in soup2markup with empty content/soup.

neiljp:
In the middle of working on another branch, I noticed that going to the `test here` stream on chat.zulip.org caused a crash, due to an empty message content - which is of course rather rare. Possibly even the web client will not send empty messages by default?

In any case, this PR avoids a crash, adds a test case, and amends the CHANGELOG ready for the next release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - Move intrusive flashing cursor in left panel to far left side
 - Show user status as `active` (Green), `idle` (Yellow) or `offline` (White) using different colors.
 
+### Important bugfixes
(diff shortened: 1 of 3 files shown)


amanagr:
@neiljp It's weird that we have this error. Even deleted messages are sent as "deleted" by the server. Also, sending empty messages is forbidden.

neiljp:
@amanagr Agreed - the webapp won't allow sending them, but I think in principle there's not something stopping them. I think maybe it was @eeshangarg who sent one to 'test here'.

Issue #126, opened 2018-07-12
Generate hotkeys section/doc using script and check with CI


amanagr:
All the keybindings used by zulip-terminal are stored in `zulipterminal/config.py`. We can generate README section of the hotkeys by travic CI.

We can create a different section for hotkeys that are imported from `urwid` and `urwid_readline` libraries.

punchagan:
There should be a tool that developers can run locally, to generate this section of the README.  TravisCI should only check that any commit doesn't miss out on doing that, when keyconfig has changed in anyway. Some sort of a test can be run that compares the output of README when this script is run, and the README committed to the repo.

neiljp:
I just thought this myself when opening #165 for some recent changes, and #164 for starred messages :)

neiljp:
I believe we've discussed moving the hotkeys from README into docs/, so that might be a good first step, updating all the links, followed by the automatic synchronization via a script and then checking that with a GitHub Action (since we don't use Travis now).

neiljp:
This is a similar issue to #892.

mkp6781:
@zulipbot add "PR needs review"

PR #262, opened 2019-01-14
Improve/reorder help for shortcut keys & remove QUIT_HELP

neiljp:
@rishig How does this look like, reordering/naming, based on what you said before?

We can discuss further on #z-t, but I was reading it recently and these where some more obvious improvements (at least to my eye):
* reordered/clearer/consistent help text
* remove help-specific 'exiting'

### DIFF
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -1,6 +1,14 @@
 from typing import Set
 
 KEY_BINDINGS = {
+    'HELP': {
+        'keys': {'?'},
+        'help_text': 'Display help menu',
+    },
+    'QUIT_HELP': {
+        'keys': {'q', 'esc'},
+        'help_text': 'Quit help menu',
+    },
     'GO_BACK': {
         'keys': {'esc'},
         'help_text': 'Go Back',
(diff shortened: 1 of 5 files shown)


zulipterminal/config/keys.py
```diff
@@ -35,27 +43,27 @@
...
     },
     'STREAM_MESSAGE': {
         'keys': {'c'},
-        'help_text': 'New stream message',
+        'help_text': 'New message to a stream',
     },
     'PRIVATE_MESSAGE': {
         'keys': {'x'},
-        'help_text': 'New private message',
+        'help_text': 'New message to a person',
```
rishig: "New message to a person or group of people"?

rishig:
lgtm!

zulipterminal/config/keys.py
neiljp: @rishig Good catch - I agree about this and have changed it in the PR. I wanted to check this in the code before changing, but it has exposed the way in which we handle PMs, which I think might need more work too (emails vs names).

neiljp:
@rishig Updated.

In addition to the text already & responding to the review comment, I tweaked a few others and sync'd with the `README.md` keylist (which is a little tedious to do manually). There was a disconnect between `s` and `S`, which pushes the importance of doing the synchronized automatically, perhaps!

Seeking feedback from @amanagr too.

amanagr:
@neiljp changes are definitely an improvement. For synchronisation between Readme and keys.py, we could use travis. #126 is already present for this.

neiljp:
Force-pushed to resolve conflicts.

PR #263, opened 2019-01-15
keys: Add `ctrl+d` as keyboard shortcut for sending a message.

amanagr:


zulipterminal/config/keys.py
```diff
@@ -62,7 +62,7 @@
         'help_text': 'Toggle focus box in compose box'
     },
     'SEND_MESSAGE': {
-        'keys': {'meta enter'},
+        'keys': {'meta enter', 'ctrl d'},
```
neiljp: As per #262, you probably want to update `README.md` too :)

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Next Release
 
+### Interactivity improvements
+- Send message by pressing <kbd>ctrl</kbd>+<kbd>d</kbd> along with <kbd>meta</kbd>+<kbd>enter</kbd>.
```
neiljp: Perhaps simpler, like the following?
`Add additional shortcut key for sending messages (<kbd>ctrl</kbd>+<kbd>d</kbd>)`

We should probably pick a consistent term to use, ie. 'hot keys` vs `keyboard shortcuts` vs `shortcut keys`, etc

PR #265, opened 2019-01-16
[WIP] Fit help text to column sizes.

neiljp:
@amanagr @rishig Interested to see what you think of this; the widths are not calculated as tidily as I would prefer, but appear to work fine.

There's something to be said for the current 'full screen' help menu, in that it isolates the background app 'noise' from the help menu, but perhaps this just highlights the importance of emphasizing the 'modal' help menu with brighter colors or something like that?


amanagr:
For terminal width <75 the Shortcut keys disappear. The help text should wrap in it's own column for smaller widths. You can look at https://github.com/louipc/turses/blob/master/turses/ui.py#L868 for some inspiration.

neiljp:
@amanagr Thanks for the sample, it's useful to see more `urwid` code.

That said, while I agree that disappearing help text is a regression, the updated commit I just posted should resolve this by only fixing the width of the key text. In addition, there is the outstanding issue that the horizontal dividers split - the second commit I just added fixes this, though merges the lines together, though this was one of @rishig's original suggestions, I think?

amanagr:
Looks good!

rishig:
Sorry for the delay; was travelling and then sick for the last few days. Could you post a screenshot?

neiljp:
@rishig This is what it looks like for the current branch for a range of widths; you can see the last one has a little of a broken title, but that is a *very* narrow screen! (~40?)

screenshot-wide
screenshot-maybenormal
screenshot-narrow
screenshot-verynarrow

neiljp:
I should emphasize, that the line-breaking behavior with multi-line help-items is much better with this styling, as the bottom lines pushed themselves up to the text if they were shorter, on master.

rishig:
cool; this looks great to me!

neiljp:
@rishig Great, will merge soon subject to any further comments in the short term - this resolves the first point of the 'help' section in #225 that you raised?
(I'm not sure what the third point means, but can discuss on #z-t)

PR #267, opened 2019-01-16
Allow 'P' and 'f' keys to work from side panels

neiljp:
Currently these keys only work from the central panel, but I feel these should be narrowing keys which work from anywhere in the app.

I moved the relevant `Controller` calls from the `MessageView` and made it change focus to the messages.

CHANGELOG also amended.


amanagr:
Thanks @neiljp for the fix!

PR #268, opened 2019-01-18
Improve fixture consistency

neiljp:
This started as one commit to improve the consistency between fixtures, and the followup to clarify a test, and then a follow-up commit which makes things clearer and simpler in terms of knowing the active user - and potentially testing with other user details later, even.

I would particularly like review of the first commit, as I'd like the intent of the tests to be confirmed in terms of the test ids (descriptions) that I added.


amanagr:
Yeah, we should try to make our fixtures more robust, consistent and to make them cover different cases. This area could use some good work BTW.

PR #269, opened 2019-01-20
Improve Installation notes for users & developers

neiljp:
I've attempted to improve/update the notes again, including broadening the details of virtual environments and   adding a native pip (as opposed to pipenv) install, to give people an alternative option.

Re that last point, I'm happy to move the pip notes to after pipenv, if that would be preferred.

I'm still interested in adding tox support so we can run multiple environments locally like on Travis, but that would be yet another alternative :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,15 +15,12 @@ Please see the [CHANGELOG](CHANGELOG.md) for released & recent changes.
 
 ## Installation & Running
 
-We recommend installing `zulip-term` in a python virtual environment, something like the following:
-```
-[sudo] pip3 install virtualenv
-virtualenv /tmp/zt/
-. /tmp/zt/bin/activate
-pip3 install zulip-term
-```
+We recommend installing `zulip-term` in a new python virtual environment (venv); with the required python 3.4+, the following should work on most systems:
+1. `python3 -m venv zulip-terminal-venv` (creates a venv named `zulip-terminal-venv` in the current directory)
+2. `source zulip-terminal-venv/bin/activate` (activates the venv; this assumes a bash-like shell)
+3. `pip3 install zulip-term` (downloads and installs the latest zulip-terminal release from PyPI)
 
-Zulip Terminal installs as `zulip-term`, so just run:
+Zulip Terminal installs as `zulip-term`, so you can then run:
 ```
 $ zulip-term
 ```
@@ -35,6 +32,10 @@ Loading with:
    autohide setting 'autohide' specified with no config.
 Welcome to Zulip.
 ```
+
+### NOTE: Running in subsequent/different sessions
+
+If you open a different terminal window (or log-off/restart your computer), you'll need to run **step 2** of the installation again before running `zulip-term`, since that activates that virtual environment. You can read more about virtual environments in the [Python 3 library venv documentation](https://docs.python.org/3/library/venv.html).
 
 ### NOTE: The zuliprc file
 
--- a/README.md
+++ b/README.md
@@ -181,32 +181,18 @@ $ pip3 install -e .[dev]
 $ zulip-term
 ```
 
-### Running tests
+## Development tasks
 
-* To run all tests (including the linter):
-```
-pipenv run pytest
-```
+Once you have a development environment set up, you might find the following useful, depending upon your type of environment:
 
-* To generate coverage report for tests:
-```
-pipenv run pytest --cov-report html:cov_html --cov=./
-```
+| Task | Pip | Pipenv |
+|-|-|-|
+| Run all tests (including linter) | `pytest` | `pipenv run pytest` |
+| Build test coverage report | `pytest --cov-report html:cov_html --cov=./` | `pipenv run pytest --cov-report html:cov_html --cov=./` |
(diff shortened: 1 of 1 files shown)


rht:
Alternatively, https://docs.snapcraft.io/python-apps/6741
`snap install zulip-term` along the line of `snap install zulip-desktop`, which bypasses the debate of whether to recommend making a separate virtualenv or not, or which virtualenv library to use.

rht:
It would be informative to add the quick installation step without virtualenv.

README.md
```diff
@@ -15,15 +15,12 @@ Please see the [CHANGELOG](CHANGELOG.md) for released & recent changes.
...
-We recommend installing `zulip-term` in a python virtual environment, something like the following:
-```
-[sudo] pip3 install virtualenv
-virtualenv /tmp/zt/
-. /tmp/zt/bin/activate
-pip3 install zulip-term
-```
+We recommend installing `zulip-term` in a new python virtual environment (venv); with the required python 3.4+, the following should work on most systems:
+1. `python3 -m venv zulip-terminal-venv` (creates a venv named `zulip-terminal-venv` in the current directory)
+2. `. zulip-terminal-venv/bin/activate` (activates the venv)
```
vishal-vardhan: here i think it should be `source zulip-terminal-venv/bin/activate`.

README.md
```diff
@@ -15,15 +15,12 @@ Please see the [CHANGELOG](CHANGELOG.md) for released & recent changes.
...
+3. `pip3 install zulip-term` (downloads and installs the latest zulip-terminal release from PyPI)
 
-Zulip Terminal installs as `zulip-term`, so just run:
+Zulip Terminal installs as `zulip-term`, so you can then run:
 ```
 $ zulip-term
```
vishal-vardhan: Line 23: `(zulip-terminal-venv) $ zulip-term`

README.md
amanagr: No. @vishal-vardhan It is correct.

README.md
amanagr: I think it serves the purpose and the change you requested is not required.

README.md
rht: `.` is POSIX-compliance, `source` is for bash-only (and bash-like shells). Nowadays it is fine to use `source` -- one minor advantage: it stands out more than `.` because it could happen that people might forget to select all the way to `.`.

vishal-vardhan: @amanagr  @rht sorry about that i forgot.

neiljp: @rht Is `source` sufficiently cross-shell/platform now? I can see the visual preference, but obviously cross-platform behavior would be preferable.

rht: Digging at zulip/zulip for a tie-breaker, but can't find any usage of it. IIRC it was recently replaced by `./executable.sh` instead of `. executable.sh`. Then found that https://virtualenv.pypa.io/en/latest/userguide/#activate-script uses `source` by default, but describes the caveats of using it in a paragraph afterward. TIL that `bin/activate` is only for bash-like shells. `csh`/`fish` users must source from a different file. Ergo, since saying `bin/activate` itself already assumes that users are on bash-like shells, there isn't any additional assumption to make when saying `source`. tl; dr: use `source`?

neiljp:
@rht I think other install methods probably belong in separate issues/PRs, whether for conda, snap, or others.

I'm interested in the possibility of using tox, but that's yet another area :)

README.md
neiljp: @rht @amanagr So what do you think about just having '`source zulip-terminal-venv/bin/activate` (activates the venv, assuming a bash shell)' ?

rht: Yes, that's a safe assumption for people using a terminal client on laptops nowadays. Can't imagine anyone would run ZT on TempleOS, but if someone does, s/he probably knows enough to use `.` and a `venv/bin/activate.shellname`.

amanagr: I am fine with using `source`.

neiljp: OK; updating it now.

neiljp:
I've just pushed an update ready to be re-reviewed. Some of the changes include:
* uses `source` instead of `.`
* places pipenv before pip
* changes to use pip3 rather than pip in dev setup (I think that caught someone out)
* extra commit extracts common setup (clone, cd) above pipenv/pip, and 'running' into the tasks section

Let me know what you think :)

neiljp:
I think we should make a separate issue to track other options, possibly separate ones for regular installs and development installs?

README.md
```diff
@@ -135,57 +146,40 @@ $ printf '\nexport PATH="%s:$PATH"\n' '${HOME}/.local/bin' | tee -a ~/.bashrc
...
-$ git clone git@github.com:zulip/zulip-terminal.git
-```
-
-3. Install dev requirements
+2. Install zulip-term, with the development requirements
 ```
-$ cd zulip-terminal
 $ pipenv --three
 $ pipenv install --dev
 $ pipenv run python setup.py develop
```
punchagan: Should we replace this with something like `pipenv run pip3 install -e .[dev]`?

neiljp: Assuming I understood the conclusion of the discussion, merging this as-is so that we have an improvement to the documentation present now, which we can incrementally improve upon.

PR #275, opened 2019-01-22
Centralize client access within model

neiljp:
These commits centralize the use of the zulip client into the `Model`, rather than being scattered through the code. I'm not fully convinced that the `Model` is the 'right' place for them, but the rest are there right now and it feels logical to have them all together, and know where to look for them!

In summary, this PR:
* moves the `update_flag` function into being a `Model` method
* extracts code from `WriteBox` into two further `Model` methods
* adds various attributes to the `Model` from the initial data, and the `server_url` from the client
* uses `user_id` from the `initial_data`, rather than obtaining it separately

After previously adding the ThreadPoolExecutor for doing the initial data fetch in parallel, the last change actually shrinks this back down to one worker thread. For now I feel it's reasonable to leave this infrastructure in place.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -80,7 +80,11 @@ class Model:
         self.stream_id = -1
         self.recipients = frozenset()  # type: FrozenSet[Any]
         self.index = initial_index
+
         self.user_id = -1  # type: int
+        self.user_email = ""
+        self.user_full_name = ""
+
         self.initial_data = {}  # type: Dict[str, Any]
 
         # Register to the queue before initializing further so that we don't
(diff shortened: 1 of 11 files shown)


zulipterminal/model.py
```diff
@@ -339,7 +331,7 @@ def _update_initial_data(self) -> None:
         results = {name: future.result()  # type: ignore
                    for name, future in futures.items()}
         if all(results.values()):
-            self.user_id = results['user_id']
+            self.user_id = self.initial_data['user_id']
```
amanagr: oh! nice catch!

punchagan:
:+1: The changes look good to me too!

PR #276, opened 2019-01-23
Indicate server connectivity via heartbeat

neiljp:
This PR uses the automatic heartbeat event from the server to show the connectivity status in the bottom-right of the application.

Note that although the window says "Last server contact", this only updates on heartbeat events; we can change this to be more expressive/accurate, eg. trigger on every server event, or specifically say 'heartbeat' or similar.

I'm not sure that we handle reconnecting properly, but this gives an indication of whether things should be working, visually, which is perhaps not always obvious.

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -49,6 +49,24 @@ class TestView:
         assert view.users_view == right_view()
         assert return_value == line_box()
 
+    def test_update_last_connection_time(self, view, monkeypatch,
+                                         day=23, month=1, year=2019,
+                                         hour=12, minute=29, sec=35):
+        import datetime
+
+        class dt:
+            def now():
+                return datetime.datetime(year, month, day, hour, minute, sec)
+        monkeypatch.setattr('zulipterminal.ui.datetime', dt)
+
+        view.update_last_connection_time()
+
+        (view._w.footer[1].set_text.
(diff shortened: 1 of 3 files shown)


amanagr:
* "Last server contact: time (date)" format is too long.  There is also text overflow in some long help text messages.
* Just having a heartbeat reply from server doesn't mean things are going well.
if we want to show user if he is connected properly to the server or not, we can use GREEN and RED DOT at the top before user name to indicate if he is connected or not(which we are perhaps already doing in USER PANEL??).

neiljp:
@amanagr I understand there are concerns regarding this PR, and didn't expect it to be an easy merge :)
I was mostly interested if we could use heartbeat events to give extra info on connectivity, both as a v1 user feature, and for debugging.

I am open to suggestions on a shorter format for the 'server contact' details.

I think it could be confusing to a user to indicate server connection/status through user status:
- if the client is badly connected to the server, it should not be a small indicator next to the user showing this
- if the client is badly connected, lots of things will be potentially out of date, including the presence list

amanagr:
If we only want to show how recently things were updated, we can use `updated 5s ago`. This should server the purpose right?

Then change it to `Connection Error` if things go wrong?

neiljp:
I left it as just the time, as it wasn't clear to me at this point what 'go wrong' means, though I appreciate that 'updated' is simpler, if not necessarily any clearer.

Issue #254, opened 2019-01-13
Add new styles to older incomplete themes


neiljp:
Current master has two complete themes (`default`, `gruvbox`) and two older themes (`light`, `blue`).

The latter do not have all the latest style names listed, so do not render well, and after merging #251 then using them will give a warning since lacking the required styles tends to lead to minimal and unexpected rendering.

The older themes need the newer required styles adding, and ideally updating to highlight the features possible with the new styles.

This is a refocusing of #6 to improve theme quality.

ishanrai05:
@zulipbot  claim

zulipbot:
Welcome to Zulip, @ishanrai05! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

ishanrai05:
@neiljp @amanagr 
Hey guys... I was wondering if the only file that needs to be updated is `config/themes.py`  and the list from which I need to pick up is styles names is `required_styles` in `themes,.py` and then add colors for the remaining style names in `light, blue` ? Also why are there two colors listed against a feature for instance `('selected',     'white',        'dark blue')` ?

Also any recommendation of colors that should be used? :)

amanagr:
Hi @ishanrai05 , you are correct that you only need to update `config/themes.py` file.  For the colors to be used for incomplete themes, I would suggest you to get a feel for the themes are really trying to achieve (inversion of colours/ make it look blue-ish) and add the colors accordingly.

Thanks for working on this issue!

neiljp:
@ishanrai05 I merged #277; this issue didn't close automatically, so I've done so manually. You're welcome to work on improving the themes next, if you'd like.

PR #277, opened 2019-01-24
themes: Add new styles to incomplete themes.

ishanrai05:
This PR adds new styles and completes the light theme and the blue theme.

`test-theme` has also been updated to include both light and blue themes in list of completed themes.

**Screenshots:**
Light Theme:
sslight

Blue Theme:
ssblue

solves issue #254


amanagr:
Thanks @ishanrai05 for working on this! Looks fine to me, but I would wait for @neiljp to have his final thoughts on this.

rht:
> but I'm not sure whether github will complain that the first commit already closed the issue

Why not let it be and see what happens? (Though what is most likely to happen is that the next commits will be ignored in favor of the first one.)

IMO the PR as it is, is already fine to merge. As in, better have a complete the first, then fix the color synergy.

amanagr:
@neiljp sounds reasonable to me. Also, I don't think there is a problem with merging with two separate commits closing the issue.

neiljp:
@ishanrai05 To clarify, if you can tidy the commit text that would be sufficient to merge the commits, though we will want to look at the styling soon after.

ishanrai05:
I'll make the required changes asap. :) thank you for reviewing this. @neiljp

amanagr:
LGTM now :)

PR #285, opened 2019-01-29
cli: Add functionality to check version from command line. Fixes #280

### LINKED ISSUE #280: Add functionality to check version from the command line
`zulip-term --version` should print out the version of the terminal client.

traintraveler:
Print the version of the terminal client, also prints out the git short hash if on a git version.
Fixes #280

### DIFF
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -1,6 +1,7 @@
 import pytest
 from zulipterminal.cli.run import main, in_color, THEMES
 from zulipterminal.model import ServerConnectionFailure
+from zulipterminal.version import ZT_VERSION
 
 
 @pytest.mark.parametrize('color, code', [
@@ -31,7 +32,8 @@ def test_main_help(capsys, options):
         '-h, --help',
         '-d, --debug',
         '--profile',
-        '--config-file CONFIG_FILE, -c CONFIG_FILE'
+        '--config-file CONFIG_FILE, -c CONFIG_FILE',
+        '-v, --version'
     }
     optional_argument_lines = {line[2:] for line in lines
                                if len(line) > 2 and line[2] == '-'}
@@ -107,5 +109,20 @@ def test_warning_regarding_incomplete_theme(capsys, mocker, monkeypatch,
             format(server_connection_error)),
     ]
     assert lines == expected_lines
+
+    assert captured.err == ""
+
+
+@pytest.mark.parametrize('options', ['-v', '--version'])
+def test_zt_version(capsys, options):
+    with pytest.raises(SystemExit) as e:
+        main([options])
+        assert str(e.value) == "0"
+
+    captured = capsys.readouterr()
+
+    lines = captured.out.strip('\n')
+    expected = 'Zulip Terminal ' + ZT_VERSION
+    assert lines == expected
 
     assert captured.err == ""
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -14,6 +14,7 @@ from zulipterminal.model import ServerConnectionFailure
 from zulipterminal.config.themes import (
     THEMES, all_themes, complete_and_incomplete_themes
 )
+from zulipterminal.version import ZT_VERSION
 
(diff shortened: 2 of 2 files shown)


zulipterminal/cli/run.py
```diff
@@ -152,6 +158,31 @@ def main(options: Optional[List[str]]=None) -> None:
         prof = cProfile.Profile()
         prof.enable()
 
+    if args.version:
+        msg = 'zulip-terminal ' + ZT_VERSION
```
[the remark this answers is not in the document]
traintraveler: Yes, you're right. Sorry for that. I've trimmed the commit.

traintraveler:
Yes, I'll do it.

neiljp:
We'll have to decide if this does fix #280 fully in its current form, too; @rht ?

rht:
IMO the git hash thing can be done in another PR. Found a simple library for it if @traintraveler wants to try it: https://pypi.org/project/setuptools-git-version/. 

Though I'm sure in `npm` this can be done much more easily. For some reason setup.py in most projects is vanilla most of the time.

neiljp:
@rht Great, thanks for the resource! I agree; we can leave the git hash details to another PR, against issue #288.

traintraveler:
@rht, Thanks for the resource, I'll give it a try.

neiljp:
@traintraveler To be clear, if you wish to squash these commits (though may need some tidying) then we'd be happy to push the release version first, and then add the githash code as a second PR building on that.

traintraveler:
@neiljp  Please review the changes, I haven't added the test, should the test be something like `test_import_zt_version`

zulipterminal/cli/run.py
```diff
@@ -152,6 +158,11 @@ def main(options: Optional[List[str]]=None) -> None:
...
+        print(msg)
```
neiljp: Since we use both 'zulip-term' and 'zulip-terminal' names for the command/project, it might be simpler to just use 'version: ' here, since users must know the command they're running :) Now that you're not using msg more than once, I'd just combine the two lines together, too.

amanagr: @neiljp I think "Zulip Terminal ZT_VERSION" should be better. Would it also be good to include the path from where zulip-term is running? I should help users easy differentiate if they are running the dev version or released one.

traintraveler:
@neiljp, Added test, removed the commit trail from squash.

zulipterminal/cli/run.py
traintraveler: @amanagr , Wouldn't it be better to add the include path with #288 ?

tests/cli/test_run.py
```diff
@@ -109,3 +111,13 @@ def test_warning_regarding_incomplete_theme(capsys, mocker, monkeypatch,
     assert lines == expected_lines
 
     assert captured.err == ""
+
+
+@pytest.mark.parametrize('options', ['--version'])
```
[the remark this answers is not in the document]
amanagr: yeah both `-v` and `--version` should work in the same way so we can include `-v` here too.

zulipterminal/cli/run.py
amanagr: Hmm... maybe but I am not sure since git already means it's the dev version. Well I don't think it fits well with this PR as well.

neiljp:
@traintraveler Are you ok making the suggested changes? (check exception value, add `-v` as alternative to `--version`)

traintraveler:
Yep, been a bit busy the past few days.

neiljp:
I posted the following in the issue rather than here, by mistake; @rht has raised the point of whether to include these entries in commits, but until we're agreed on that, I'll leave it here too:

This looks great! My only thought is whether you'd like to include an extra line in CHANGELOG.md in the same commit (perhaps in the 'interactivity' section?) regarding the addition of this change?

Otherwise, I'm happy for this to be merged.

traintraveler:
@neiljp I kind of agree with @rht on this. Seems better to be updated in bursts.

neiljp:
Well, let's merge and get things in for now :+1:

PR #287, opened 2019-01-31
WIP: Barely working zforms support.

aero31aero:
This is just a proof of concept. I'm not sure where to add what code and I essentially wrote a lot of unneeded functions to export things from the message for now. I'll clean this up in the next few commits.

The current interface is to press a number key from 1 to 9 to select one of the choices available in a message:

image

This creates an obvious restriction of maximum 9 choices per message. I wonder if a popup menu to choose would be a better UI?

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -305,6 +305,11 @@ def classify_unread_counts(model: Any) -> Dict[str, Any]:
     return unread_counts
 
 
+def check_submessage_syntax(submessage: object):
+    if submessage:
+        return True  # Trust server's code checks.
+    return False
+
 def match_user(user: Any, text: str) -> bool:
     """
     Matches if the user full name, last name or email matches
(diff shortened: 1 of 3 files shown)


aero31aero:
I'll do a little bit of a self review on this:

- [ ] We should probably abstract the message sending functionality in `MessageBox.respond_to_submessage(key)` and avoid having to expose 2 functions from MessageBox.

- [ ] We should create a separate `submessages.py` file that does validation AND identification of various (message)widgets and also create (urwid)widgets.

- [ ] Along with the above, add 2 functions for the zform choices widgets: `get_widget_to_render()` and `send_response(key)`, both being used inside `MessageBox` functions.

- [ ] Currently we always call `.send_stream_message`. Figure out how to call `.client.send_message` directly or add an `if` check for PMs.

- [ ] Consider adding navigation like `<next option> <previous option> <select option>` if there are more than 9 options.

- [ ] Unit tests.

neiljp:
This seemed to crash for me, with an `IndexError` on L458 of boxes.py? I'm not sure if anyone else has tried it?

neiljp:
@aero31aero Just checking if you're wanting some kind of review on this, or plan to work on it more first?

PR #289, opened 2019-02-06
Fix quoting original message, not rendered version 

aditya-hari:
Fixed the issue of quoting resulting in quoting the rendered version rather than the original message.  
Fixes issue-239. 
 
QUOTE_TEXT used to get quote from self.message 
It now gets raw_content from the model.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -472,7 +472,7 @@ class MessageBox(urwid.Pile):
             self.model.controller.view.middle_column.set_focus('footer')
         elif is_command_key('QUOTE_REPLY', key):
             self.keypress(size, 'enter')
-            quote = '```quote\n' + self.message['content'] + '\n```\n'
+            quote = '```quote\n' + self.model.client.get_raw_message(self.message['id'])['raw_content'] + '\n```\n'
             self.model.controller.view.write_box.msg_write_box.set_edit_text(
                 quote)
             self.model.controller.view.write_box.msg_write_box.set_edit_pos(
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -472,7 +472,7 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
             self.model.controller.view.middle_column.set_focus('footer')
         elif is_command_key('QUOTE_REPLY', key):
             self.keypress(size, 'enter')
-            quote = '```quote\n' + self.message['content'] + '\n```\n'
+            quote = '```quote\n' + self.model.client.get_raw_message(self.message['id'])['raw_content'] + '\n```\n'
```
neiljp: This looks good for a first version and seems to work well :+1: 

You will notice that there is a red 'x' next to this PR - this is since when running pytest, it failed due to the PEP8 check. You can run pytest (and run-mypy) locally before pushing, which I'd encourage in future before pushing to a github branch; travis will do this too, but it is often slower than on your own machine. You'll see that the error is due to a long line here.

neiljp:
@aditya-hari I forgot to mention, that you could also add a statement to `CHANGELOG.md`, perhaps in the 'important bugfixes' section? Included in the same commit.

zulipterminal/ui_tools/boxes.py
aditya-hari: I haven't been able to get pytest to run. That's probably a consequence of not doing the installation of the development version properly. "pipenv install --dev" results in an error for me. No idea how to resolve it.

PR #292, opened 2019-02-08
Fix quoting original message, not rendered version

### LINKED ISSUE #239: Quote original message content, not rendered version (html)
If quoting an image, and maybe other items, zulip-terminal quotes the html, rather than the unrendered message, which leaves quite a mess if sent!

We should be able to request the original content, if we don't have it, and store/send it.

This should probably be a request to the Model (possibly via controller), before quoting the correct text.

aditya-hari:
Fixes #239
Updates CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Avoid crash in rare case of empty message content
 - Set terminal locale to `utf-8` by default which removes issues with rendering double width characters.
 - Avoid crash on receiving multiple starred-message events
+- Fix quoting original message, not rendered version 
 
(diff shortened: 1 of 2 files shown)


aditya-hari:
Sorry it took this long. Took me far longer than it should have to figure out the workflow.

neiljp:
Thanks for working on this @aditya-hari :+1:

This looks great now; for future reference you should have been able to makes these changes on your original branch and push them here, which should then have appeared on the original PR. Let us know if you need help with this in future, on #zulip-terminal or one of the other streams.

PR #293, opened 2019-02-09
Unread counts: Introduce UnreadCounts TypedDict & avoid mixed indices.

neiljp:
This improves the situation regarding unread-counts, so that if you receive a message from a user who is in the right-panel, a count appears next to their name properly. It should also slightly improve stream unread counts, if there was a conflict present, which I'm not sure there often would be.

The second commit avoids the issue where a message sent by yourself would increase the unread count, which I'm fairly sure was also a bug.

See the commits for a little more detail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Make stream icons bold and correct background color
 
 ### Important bugfixes
+- Fix bug potentially mixing unread counts for messages from users & streams
 - Exit cleanly if cannot connect to zulip server
 - Avoid crash in rare case of empty message content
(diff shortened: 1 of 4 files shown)


sumanthvrao:
@neiljp Hi. This worked pretty well when i tested it manually.
I tested it when narrow was set to current stream/PM's and event when it wasn't.
The count gets updated fine. I did get a traceback once when I tried scrolling up with `Key Error:127` but i was not able to reproduce it again. Its mostly a separate issue unrelated to this. :)

amanagr:
Thanks @sumanthvrao for testing it. Please let us know if you can reproduce the issue.

Issue #297, opened 2019-02-12
Compact the help view


matrixise:
Because the HelpView takes the complete size of my screen (this one is a 25"), I prefer to use a small interface for the shortcuts.

neiljp:
Just to clarify, do you mean a small screen, or rather a certain text resolution?

I have a much smaller (laptop) screen, and with the extra lines currently present I can see all but one of the help lines. However, if the font was larger (and so there were fewer text lines) I could see that being an issue, but then we scroll anyway?

Could you describe the problem a little more?

I do think that removing the lines could be good, as long as we have something similar to address the matching by-eye of the actions to the keys, but I'd appreciate knowing your motivation in this case :)

matrixise:
Hi @neiljp 
Here is a screenshot with the help.
![zulip-term](https://user-images.githubusercontent.com/38737/52778279-f6065e80-3045-11e9-9874-c43986293473.png)

neiljp:
@matrixise This is not a match for the current development release - the menu is a lot more compact, at least horizontally. The current help looks more like the screenshots in #265 (merged ~18 days ago).

neiljp:
@matrixise If you're willing to update your development copy, I'd appreciate if you'd consider whether we've reached a suitable level of improvement to close this issue, given the merger of your PR and #301.

matrixise:
@neiljp I will do it asap (maybe tomorrow)

neiljp:
@matrixise Just checking in with you regarding this issue.

neiljp:
@matrixise I'm closing this for now since we've not heard from you and the help menu is dramatically changed.

If you have other suggestions/ideas, please open another issue or discuss on #**zulip-terminal** on chat.zulip.org.

PR #298, opened 2019-02-12
Compact the HelpView.

matrixise:
Define the height attribute (default to the number of key bindings) and
remove the bottom line character.

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -84,6 +84,7 @@ class Controller:
         self.loop.draw_screen()
 
     def show_help(self) -> None:
+        cols, rows = self.loop.screen.get_cols_rows()
         help_view = HelpView(self)
         self.loop.widget = urwid.Overlay(
             urwid.LineBox(help_view,
@@ -92,7 +93,7 @@ class Controller:
             align='center',
             width=help_view.width+2,  # +2 from LineBox
             valign='middle',
-            height=('relative', 100)
+            height=rows//2
         )
 
     def exit_help(self) -> None:
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -580,7 +580,7 @@ class HelpView(urwid.ListBox):
                     (max_widths[1], urwid.Text(", ".join(binding['keys'])))
                               ], dividechars=2),
                 tlcorner='', brcorner='', trcorner='', blcorner='',
-                rline=' ', lline=' ', bline='-', tline='',
+                rline=' ', lline=' ', bline='', tline='',
              )
              for binding in KEY_BINDINGS.values()])
 


matrixise:
https://github.com/zulip/zulip-terminal/issues/297

amanagr:
@neiljp this look like a good idea to me. What do you think?

matrixise:
with a 25" screen, when your terminal is full, you see a table with a lot of dashed lines :/ I prefer a small box with the essential.

zulipterminal/core.py
```diff
@@ -85,14 +85,15 @@ def draw_screen(self, *args: Any, **kwargs: Any) -> None:
 
     def show_help(self) -> None:
         help_view = HelpView(self)
+        dialog_height = min(50, help_view.height)
```
rht: 50 is probably too compact. It displays all the way to 'Scroll to bottom', whereas the current version (with '-') displays to 'Send a message'. It takes <1/3 of my vertical screen space on a 12.5" screen.
2019-02-12-123859_1522x767_scrot

matrixise: @rht I just updated, using the height of the screen and taking the half.

rht: Yeah looks better / less cramped to me.

rht: Incidentally it is also 50% of screen on webapp, though not sure if it continues to stretch to be 50% on larger screen,

zulipterminal/ui_tools/views.py
```diff
@@ -571,6 +571,7 @@ def __init__(self, controller: Any) -> None:
                   for binding in KEY_BINDINGS.values()]
         max_widths = [max(width) for width in zip(*widths)]
         self.width = sum(max_widths)
+        self.height = len(KEY_BINDINGS)
```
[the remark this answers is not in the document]
matrixise: sure, sorry, I have just pushed a new revision.

matrixise:
@rht Tell me if you prefer one commit (squash&merge) or many commits for this PR.

rht:
Usually, it is many commits within the theme of the PR, but each commits are 'correct' in that it doesn't have to be fixed by the next commit(s). i.e. in this case you should squash them.

matrixise:
fixed, just rebased the code and merge the 3 commits in only one.

zulipterminal/ui_tools/views.py
```diff
@@ -580,7 +580,7 @@ def __init__(self, controller: Any) -> None:
                     (max_widths[1], urwid.Text(", ".join(binding['keys'])))
                               ], dividechars=2),
                 tlcorner='', brcorner='', trcorner='', blcorner='',
-                rline=' ', lline=' ', bline='-', tline='',
```
neiljp: If we're removing this, I'm not sure what the point of the `LineBox` is now?

rht:
(Just want to additionally comment that without the dashes it looks smoother / less command-line-y (you see dashes in ASCII art).)

rht:
IMO, compact with alternating background color > compact (this PR) >  with dashes (current master). I think it is fine as a transition state?

zulipterminal/ui_tools/views.py
rht: As in, L577-585 are to be removed?

neiljp: I meant that the `LineBox` 'wrapper' around the `Columns` is rather less useful now; there is a left- and right- padding there, but I think that would be easier as a padding around the entire 'table' (rather than each entry), or just as spaces to the left and right of each entry.

neiljp:
TLDR: Given @amanagr 's comment, I'm going to merge now, but with the following concerns/caveats:
* we've definitely lost some action-to-key clarity with removing the lines
* half height + no lines shows maybe 3 fewer entries? So it's more compact, but less obvious?
(though, maybe a small 'window' helps make the absent lines clearer?)
* the linebox issue is not major & I think we can just remove the side spaces now too
* the referring issue has screenshots of 0.3.1, not master, so it's not a direct comparison there.

I have a provisional solution to most of these issues, so would encourage those who were interested in this change to review that upcoming PR.

This change is all the more relevant assuming we plan to integrate the readline keys into the online help, as that will add even more actions/keys to the list!

PR #299, opened 2019-02-13
Further README improvements

neiljp:
This incorporates some of the recently-raised points with some tidying & reorganising; briefly:
* Add `flush=True` (@sumanthvrao) to debugging tips, and add subsections
* Simplify/clarify the pipenv setup for development install (@punchagan ; not using pip3 yet, could do?)
* Move the 'edits aren't affecting zulip-term' section into debugging tips
* Move the remaining 'troubleshooting' section above the development section
* Make the 'Contributor guidelines' the start of the development section

The final result looks like this:
https://github.com/neiljp/zulip-terminal/tree/dev-docs-improvements

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,14 @@
 
 An interactive terminal interface for [Zulip](https://zulipchat.com).
 
+[Recent changes](CHANGELOG.md) | [Installation & Running](#Installation--running) | [Hot Keys](#hot-keys) | [Troubleshooting](#troubleshooting-common-issues) | [Development](#contributor-guidelines)
+
 [![Zulip chat](https://img.shields.io/badge/zulip-join_chat-brightgreen.svg)](https://chat.zulip.org/#narrow/stream/206-zulip-terminal)
 [![PyPI](https://img.shields.io/pypi/v/zulip-term.svg)](https://pypi.python.org/pypi/zulip-term)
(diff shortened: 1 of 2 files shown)


neiljp:
I just updated the PR to include
* use `pip3 -e .[dev]` in pipenv section (as discussed on #zulip-terminal with @punchagan)
* add a note in the troubleshooting section regarding crashes/tracebacks
* update changelog change to be a little broader

amanagr:
@neiljp Looks great in general. Since the Readme is getting larger I would suggest to include a table of contents at the top.

neiljp:
@amanagr Better with table (row) of changes commit? :) (removed the small recent-changes section too)

amanagr:
Yeah very innovative!

README.md
```diff
@@ -124,7 +122,59 @@ autohide=autohide
 
 Note: You can use `arrows`, `home`, `end`, `Page up` and `Page down` keys to move around in Zulip-Terminal.
 
-## Setting up a development environment
+## Troubleshooting: Common issues
```
amanagr: We should also add that users can use `Ctrl + L` to clear any screen distortions.

neiljp: @amanagr If this PR is good as it stands, I'm inclined to merge first?

Re `ctrl + l`, I think it would be useful, but isn't this a duplication of the hot-key help? Shouldn't we improve that text in preference to adding a separate section, or do you think both is warranted?

PR #301, opened 2019-02-15
Improve help menu: compact further, style lines, adapt height

neiljp:
This is a quick follow-up to the help menu observation and contribution by @matrixise in #298 and #297.

This PR:
* Removes side/top spaces
* Sets alternating background colors (using 'bar' style for now)
* Adjusts the menu height to be between (the minimum) of 3/4 window height (rows), and the number of help menu actions

The latter does increase the menu height slightly (it was set to 1/2 in #298), so any feedback through testing would be good.

Tagging @amanagr & @rht who also gave feedback on the PR by @matrixise.

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -91,9 +91,10 @@ class Controller:
                           title="Help Menu (up/down scrolls)"),
             self.view,
             align='center',
-            width=help_view.width+2,  # +2 from LineBox
             valign='middle',
-            height=rows//2
+            # +2 to both of the following, due to LineBox
+            width=help_view.width+2,
+            height=min(3*rows//4, help_view.number_of_actions)+2
         )
 
     def exit_help(self) -> None:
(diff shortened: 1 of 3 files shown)


zulipterminal/core.py
```diff
@@ -91,9 +91,10 @@ def show_help(self) -> None:
                           title="Help Menu (up/down scrolls)"),
             self.view,
             align='center',
-            width=help_view.width+2,  # +2 from LineBox
             valign='middle',
-            height=rows//2
+            # +2 to both of the following, due to LineBox
+            width=help_view.width+2,
+            height=min(3*rows//4, help_view.number_of_actions)+2
```
rht: No objection. I have no strong position on the exact fraction of the screen. Upon measuring the webapp's ratio more carefully, it's ~70%.

amanagr:
@neiljp looks great!

PR #305, opened 2019-02-17
model/append_message: Set proper flags for received messages.

amanagr:
Set the flags for the received messages from the `event` instead
of initializing them with empty list.

This marks the messages sent by user as read too.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -48,6 +48,7 @@ Event = TypedDict('Event', {
     'all': bool,
     # message:
     'message': Message,
+    'flags': List[str],
 }, total=False)  # Each Event will only have a subset of these
 
 OFFLINE_THRESHOLD_SECS = 140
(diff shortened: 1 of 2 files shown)


amanagr:
mypy tests are passing on travis but I am locally gettting `zulipterminal/core.py:271: error: Too many arguments for "deregister" of "Client"` @neiljp any ideas?

zulipterminal/model.py
```diff
@@ -485,7 +486,9 @@ def append_message(self, event: Event) -> None:
         Adds message to the end of the view.
         """
         response = event['message']
-        response['flags'] = []
+        # sometimes `flags` are missing in `event` so initialize
+        # an empty list of flags in that case.
+        response['flags'] = [] if not event.get('flags') else event['flags']
```
neiljp: I think this can be `event.get('flags', [])` ? That seems clearer?

amanagr: oh! I forgot we can do that.

amanagr:
@neiljp I have added proper test for this. Also, we were calling set_count every time so changed it to call only if 'read' not present.

neiljp:
@amanagr This looks good to me to merge now.

For the future I'm wondering if we should consider combining the `append_message` tests using parametrize or maybe fixtures?

amanagr:
Yeah!  there are more such test functions which can be combined including append_message.

PR #309, opened 2019-02-20
Model: rename 2 methods, use client API

neiljp:
This moves ZT to using the specific API calls available in the latest `zulip` package; this makes the code more concise.

We do need to depend upon a later zulip package, since there was a bug in an earlier version which meant that one of these changes didn't work.

I've also renamed two methods for clarity:
* `update_flags` -> `mark_message_ids_as_read`
* `update_presence` -> `_start_presence_updates`

The latter has additional documentation on what I think is the likely route to improve the way that presence updates occur.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -309,7 +309,7 @@ class TestMessageView:
         msg_view.read_message()
         assert msg_view.update_search_box_narrow.called
         assert msg_view.model.index['messages'][1]['flags'] == ['read']
-        self.model.update_flag.assert_called_once_with([1])
+        self.model.mark_message_ids_as_read.assert_called_once_with([1])
 
     def test_read_message_no_msgw(self, mocker, msg_view):
         # MSG_W is NONE CASE
@@ -317,7 +317,7 @@ class TestMessageView:
 
         msg_view.read_message()
 
-        self.model.update_flag.assert_not_called()
+        self.model.mark_message_ids_as_read.assert_not_called()
 
 
 class TestStreamsView:
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -229,9 +229,7 @@ class Model:
             request = dict(base_request, op='remove')
         else:
             request = dict(base_request, op='add')
-        response = self.client.call_endpoint(url='messages/flags',
-                                             method='POST',
-                                             request=request)
+        response = self.client.update_message_flags(request)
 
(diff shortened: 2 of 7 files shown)


tests/ui/test_ui_tools.py
```diff
@@ -309,15 +309,15 @@ def test_read_message(self, mocker, msg_box):
         msg_view.read_message()
         assert msg_view.update_search_box_narrow.called
```
sumanthvrao: @neiljp Since this PR contains changes along similar lines (being renaming functions for clarity),
Do you want to rename `test_init` in `Class TestMessageBox` to something more appropriate like `test_general_message` (Line number 848)?

neiljp: If you're meaning renaming `test_init`, I think that there is certainly work that can be done to simplify/clarify some of the test cases, though I think we can be more specific than `general_message`. I wouldn't say this belongs in this PR, but perhaps you meant this as a side question/comment, rather than a review?

sumanthvrao: Right! A quick `git grep` shows a lot of tests with the name `test_init`. Keeping this separate is what I meant. I should not have posted it as a review. Apologies for that.

zulipterminal/model.py
```diff
@@ -200,22 +205,19 @@ def react_to_message(self,
         # FIXME Only support thumbs_up for now
         assert reaction_to_toggle == 'thumbs_up'
 
-        endpoint = 'messages/{}/reactions'.format(message['id'])
         reaction_to_toggle_spec = dict(
             emoji_name='thumbs_up',
-            reaction_type='unicode_emoji',
-            emoji_code='1f44d')
+            emoji_code='1f44d',
+            reaction_type='unicode_emoji',  # FIXME Should be emoji_type?
```
amanagr: I think `reaction_type` is used for differentiating between 'unicode_emoji` and user create emojis.

amanagr:
@neiljp overall refactoring looks great to me!

zulipterminal/model.py
neiljp: I used `reaction_type` originally, just for the thumbs-up, and it works - but the python API method seems to indicate `emoji_type=unicode_emoji`. I just checked the backend and it has our approach, so I'll remove the comment and leave it as-is. We can always add an update if the API changes.

neiljp:
Waiting for tests, then will merge.

PR #310, opened 2019-02-23
Default theme: Match 256-color styles to existing, except deeper black

### LINKED ISSUE #283: Augment 'default' theme with 256-color version
This is because on urwid, sometimes the light version of the color is not properly rendered. ~See the warning at http://urwid.org/manual/displayattributes.html#bright-background-colors.~ the warning is only on background color, so not sure why some light colors are not rendered.

neiljp:
This should be fairly self-explanatory and works fine here.

This is split into 3 parts:
- Use a deeper/darker black in 256-color than otherwise, since it improves contrast
- Match colors in 256-color mode to previous ones
- Add whites (may not be necessary)

If in particular those who have experienced color mismatches could test this (@rht @rishig ?), this may resolve the issue in part of #225, and would as  whole resolve #283.

I was going to add a CHANGELOG entry, but was initially hesitant whether this is a bug-fix or feature; happy to know thoughts on that too, but if this does fix the color mismatches then probably more of a bug-fix?


amanagr:
I can't feel much difference in UI but changes in code are nice.

neiljp:
@amanagr If you saw the colors 'right' before, the only difference with this PR should be the darker background. This is based on the assumption that terminals showing 16-color codes in urwid properly (I'm assuming yours and mine included, eg. the green for msg_selected) migrate to 256 colors OK, whereas for those whose terminals weren't showing the 16-colors properly (urwid bug? terminal bug?) the 256-color specification works better but matches the colors to the 16-color shades.

rht:
2019-03-05-023232_1733x784_scrot
This black is what you guys usually see? I thought the slightly greenish background color is based on solarized (https://github.com/altercation/vim-colors-solarized). I don't see any hint of light green in msg_selected though.

rht:
By greenish background color, I meant, the entirety of it. On master it looks like the before of https://github.com/zulip/zulip-terminal/pull/199#issue-238951815.

neiljp:
I think we should resolve the color issues before making another release, hence adding to 0.3.2

neiljp:
I have amended this to use hex codes; they're definitely not quite as good a match as before, but it would be good to know if they look roughly right now.

rht:
Isn't the mapping between 256 and hex supposed to be exact? https://jonasjacek.github.io/colors/

neiljp:
@rht I'm not sure where my previous codes came from, maybe an old (mis)translation, but I agree with you, and have updated my local copy to these - I'll push an updated copy soon.

That said, I have ZT running side by side with old and new versions, and it's really *not* the same for me, though at least the colors are similar. Everything seems...more intense.

neiljp:
Left (old); Right (new)
image

neiljp:
Note: the silver bar colors should be darker; I missed updating that.

rht:
By using black (#000) as the reference point, the right one is definitely the absolute one. Maybe there are color0 - color16 modifications in `~/.Xdefaults` or `~/.Xresources`.

neiljp:
Pushed updates. Neither of those in my home folder.

neiljp:
This was resolved instead by #610, without the darker black.

Closing as a consequence - please speak up if you think this needs more work, or open another issue!

PR #312, opened 2019-02-24
Improve unread counts (refactor/simplify & fix few bugs)

neiljp:
This effort started as an attempt to simplify `set_count`, and in doing so the code looks a lot clearer.

Some of these commits could likely be combined, but since this code isn't well-tested in CI then I've left them separate for now to enhance clarity.

Two commits should address apparent bugs:
* Don't add counts to the total if it's muted
* Only reduce unread counts if the server reports success on marking messages as read

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -65,6 +65,10 @@ def asynch(func: Any) -> Any:
 
 
 def set_count(id_list: List[int], controller: Any, new_count: int) -> None:
+    # This method applies new_count for 'new message' (1) or 'read' (-1)
+    # (we could ensure this in a different way by a different type)
+    assert new_count == 1 or new_count == -1
+
     messages = controller.model.index['messages']
     for id in id_list:
         msg = messages[id]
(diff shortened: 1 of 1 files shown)


zulipterminal/model.py
```diff
@@ -232,12 +232,18 @@ def toggle_message_star_status(self, message: Dict[str, Any]) -> None:
...
             'flag': 'read',
             'op': 'add',
         })
-        set_count(id_list, self.controller, -1)  # FIXME Update?
+        # We should only update the local cache of counts if the
+        # message is marked successfully this way; server events
+        # will trigger updates of the index for other cases.
+        if response['result'] == 'success':
+            # FIXME: Failure should check partial success?
+            # Maybe we should update each message flag individually?
```
amanagr: Updating each message flag individually may result in a large number of api requests to the server and hence the server may block the client from sending further requests.

neiljp: I wasn't necessarily suggesting that approach, but I'm happy to drop this commit for now until we agree on a solution.

How do you like the other changes? Good to merge?

amanagr: yeah other changes look nice to me.

neiljp: OK, have dropped this commit - will merge now.

PR #316, opened 2019-02-27
Add gitlint with custom rules

neiljp:
I came across `gitlint` via both the main Zulip repo (at some point) and subsequently; after exploring it, I think it could be a useful addition to our workflow, for those who want it (only, at this point).

This PR adds the package to the dev environment, sets a configuration file with some customizations, and also a file with some custom rules to make it clear what may be missing when a rule fails.

The custom rules are written assuming that we want each commit to
* start with one or more 'areas', each followed by `: ` (colon+space), in lower case (except eg. `WIP`)
* have a capitalized description after the areas, ending in a full-stop/period (`.`)

Gitlint is not set up to run in CI, and users must add the git hook manually (via the instructions), so currently this is just an optional tool. Also, one can just say that we're ignoring gitlint (not in CI, etc), so I'm not sure if we benefit from the `ignore-by-title` section.

I may squash these two commits - any thoughts on that? It is new code, but it did have some kind of history :)
(also we could remove some of the commented automatic code in `.gitlint`, but that could be useful? I've trimmed some already)


.gitlint
```diff
@@ -0,0 +1,54 @@
...
+
+[body-max-line-length]
+# ZT: This matches the Zulip preference for short body line lengths
+line-length=76
+
+[area-formatting]
+# This is a ZT custom rule; these are excluded from the lower-case area rule
+exclusions=WIP, DEBUG
+
+[ignore-by-title]
```
amanagr: I am getting this error when I run `gitlint install-hook`:
`Config Error: No such rule 'ignore-by-title'`

README.md
```diff
@@ -232,6 +232,7 @@ Once you have a development environment set up, you might find the following use
 | Build test coverage report | `pytest --cov-report html:cov_html --cov=./` | `pipenv run pytest --cov-report html:cov_html --cov=./` |
 | Check type annotations | `./tools/run-mypy` | `pipenv run ./tools/run-mypy` |
 
+If you plan to add git commits and submit pull-requests (PRs), then we highly suggest installing the `gitlint` commit-message hook by running `gitlint install-hook` (or `pipenv run gitlint install-hook` with pipenv setups). While the content still depends upon your writing skills, this ensures a more consistent formatting between different commits by different authors.
```
amanagr: We should also add instructions to run it properly here.

.gitlint
neiljp: What version of gitlint do you have? I have 0.10.0

Are you using an IDE or something like that, which may add git hooks otherwise?

README.md
neiljp: Do you mean describing that it will check and maybe ask the yes/no/edit question?
Could you describe more what you mean by this?

amanagr: I mean that we should specify gitlint will check commit message after every `git commit XYZ` command. Also add some commands to check commit message lints for last n commits.

.gitlint
amanagr: Apparently there are two ways to install gitlint. via `apt` and `pip`. If user has already installed gitlint via `apt` then this error occurs.
Uninstalling apt version and then removing the `.git/hooks/commit-msg` file removes the error.

amanagr:
Since it's working fine now I like this tool :)

.gitlint
neiljp: Was this an older version of gitlint, or some other issue, do you know?

README.md
neiljp: I've pushed a new version that should address these points. Let me know if they cover what you expect.

.gitlint
amanagr: I am sure but it can be an old version of gitlint.

neiljp:
OK, pushing these in now - let's see how people like it (or not!)

PR #318, opened 2019-03-01
Clarify Organization URL (was Realm URL) with examples.

neiljp:
Following some confusion over pre-existing use of 'Zulip Realm URL', this moves to use the form used in the desktop app, namely indicating 'Organization URL' and giving some examples.

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -87,7 +87,18 @@ def fetch_zuliprc(zuliprc_path: str) -> None:
                       "\nPlease enter your credentials to login into your"
                       " Zulip organization."
                       "\n" +
-                      in_color('blue', "Zulip-Realm URL: "))
+                      "\nNOTE: The " + in_color('blue', "Zulip URL") +
+                      " is where you would go in a "
+                      "web browser to log in to Zulip." +
+                      "\nIt often looks like one of the following:" +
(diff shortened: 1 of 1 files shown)


zulipterminal/cli/run.py
```diff
@@ -87,7 +87,18 @@ def fetch_zuliprc(zuliprc_path: str) -> None:
...
                       " Zulip organization."
                       "\n" +
-                      in_color('blue', "Zulip-Realm URL: "))
+                      "\nNOTE: The " + in_color('blue', "Zulip URL") +
+                      " is where you would go in a "
+                      "web browser to log in to Zulip." +
+                      "\nIt often looks like one of the following:" +
+                      in_color('green', "\n   your-org.zulipchat.com") +
+                      " (Zulip-hosted servers)" +
+                      in_color('green',"\n   zulip.your-org.com") +
```
mregdos: missing whitespace after ',' here (and another below) would fail PEP8 check.

I like the clarification added. That part was slightly confusing to me when I launched zulip-terminal for the first time.

neiljp: @mregdos Thanks for the check - I forgot to run manually, though the CI picked it up :)

Glad you liked the clarification :+1:

neiljp:
CI issues should be fixed; if they pass OK then given the positive feedback I'll merge.

PR #319, opened 2019-03-02
snakeviz/requirements: Remove snakeviz from requirements.

amanagr:
Snakeviz install tornado v6.0 which requires python >=3.5. So, to
provize support for zt on python 3.4 snakeviz is being removed.


Pipfile
```diff
@@ -19,5 +19,4 @@ pytest-pep8 = "==1.0.6"
 pytest-mock = "==1.7.1"
 pytest-cov = "==2.5.1"
 pudb = "==2017.1.4"
-snakeviz = "==0.4.2"
```
neiljp: Would it be more effective to install this dependent on the python version, the same as for `typing`?

zulipterminal/cli/run.py
```diff
@@ -263,7 +263,8 @@ def main(options: Optional[List[str]]=None) -> None:
             prof.dump_stats("/tmp/profile.data")
             print("Profile data saved to /tmp/profile.data")
             print("You can visualize it using e.g."
-                  "`snakeviz /tmp/profile.data`")
+                  "`snakeviz /tmp/profile.data`"
+                  "NOTE: Use `pip install snakeviz` to install snakeviz")
```
neiljp: This *could* do a quick check to see if the command is present via running `snakeviz -h` and checking the result, if we wanted to smooth the user experience?

neiljp:
See #321 for an alternative approach.

PR #321, opened 2019-03-03
requirements: Explicitly install tornado ~5.1 for snakeviz on py 3.4.

neiljp:
The latest tornado is now 6.0, which doesn't support python 3.5.

Rather than remove snakeviz or only install in some python versions, this requires installation of the earlier tornado version, which is still available.


amanagr:
This looked like a better approach! Thanks Neil!

PR #323, opened 2019-03-07
Fix recipient email footer for self-messages

sumanthvrao:
Reproduce bug: 
* Narrow to PM with yourself.
* click HotKey 'r' or 'c' on any of the messages sent.
* A footer appears with no email address in `To:` field.

Fix: 
* We explicitly check if a message is self message. We create a special recipient list for this.

(This Bug fix was solved as a part of #278. But to avoid bug-fix amidst feature adds, it was moved to this separate PR)

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -140,6 +140,18 @@ class MessageBox(urwid.Pile):
         else:
             raise RuntimeError("Invalid message type")
 
+        if self.message['type'] == 'private':
+            self.recipients_names = ', '.join(list(
+                        recipient['full_name']
+                        for recipient in self.message['display_recipient']
+                        if recipient['email'] != self.model.user_email
+                    ))
+            self.recipients_emails = ', '.join(list(
+                        recipient['email']
+                        for recipient in self.message['display_recipient']
+                        if recipient['email'] != self.model.user_email
+                    ))
+
         super(MessageBox, self).__init__(self.main_view())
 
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -192,11 +203,14 @@ def stream_header(self) -> Any:
...
-            if recipient['email'] != self.model.user_email
-        ))
+        if self._is_private_message_to_self():
+            self.recipients = self.message['display_recipient'][0]['full_name']
+        else:
+            self.recipients = ', '.join(list(
+                recipient['full_name']
+                for recipient in self.message['display_recipient']
+                if recipient['email'] != self.model.user_email
+            ))
```
neiljp: As per my comment on the previous commit, we don't now use `self.recipients` :)

sumanthvrao: Are both the comments you left here referring to the same thing? Which is I need to change in `private_header` function as I am still referring to the old `self.recipients` variable there.
Re tests: Will add them :)

Apologies about this. Missed to spot this. :(

neiljp: Yes, same area - I think we clarified this on czo?

neiljp:
Merged with a very small change to combine lines.
https://github.com/zulip/zulip-terminal/commit/b7b43757cfe0159e5754a293ee30641c4e63d739

sumanthvrao:
@neiljp Thanks for the merge :).
This was particularly an irritating bug from my part as I used to make lots of self notes.

Issue #222, opened 2018-12-28
Implement autocomplete for @mention in message box.


amanagr:
This feature implementation could use the autocomplete feature provided by urwid_readline library in the first version.

amanagr:
Sure!

zulipbot:
Hello @ishanrai05, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

ishanrai05:
@zulipbot  abandon

neiljp:
@ishanrai05 Please let us know if there is anything else (an issue or something not listed) that'd you like to work on :)

omi10859:
Is it still open and can be claimed?

amanagr:
Yes @omi10859 you work on it.

omi10859:
@zulipbot claim

zulipbot:
Welcome to Zulip, @omi10859! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

omi10859:
@zulipbot claim

neiljp:
For me, one important change to include is that we should make this case-independent in the same way the user-list is sorted, eg. if I searched for 'Eric', it should not find them not just using '@e' but also with '@E'. It already does this if you just do '@<tab>' and cycle through, but I expect other people will want this based on using some first letter(s) rather than cycling through the whole/active user list.

[...]

PR #324, opened 2019-03-10
Themes: Add preliminary "dark solarized" theme.

neiljp:
After @rht mentioned solarized, I thought I would give a try at a dark solarized theme.

This is not tidy right now, but uses simplified (short hex) colors. I'm not overly familiar with this side of things, but it looks OK :) I'm leaving this as WIP/draft to get some feedback

I've taken the same approach as in #310 with a color dict and referencing it later.
I started with some 'translations', which are the comments above the theme itself. However, I've since adjusted some colors further.

Notable changes at this point: (that I recall; may list more)
* "Blue" is used for highlighting in each panel, which I think really helps with consistency
* active message is marked bold

Note that the solarized colors don't match the urwid colors, but we're explicitly setting a 256-color mode in urwid, and the non-256 colors aren't working for everyone right now.

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -28,6 +28,79 @@ required_styles = {
     'starred',
 }
 
+SOLAR_BASE = dict(
+    # Base03='#002b36',
+    # Base02='#073642',
+    # Base01='#586e75',
+    # Base00='#657b83',
+    # Base0='#839496',
+    # Base1='#93a1a1',
+    # Base2='#eee8d5',
+    # Base3='#fdf6e3',
+    # Yellow='#b58900',
+    # Orange='#cb4b16',
+    # Red='#dc322f',
+    # Magenta='#d33682',
+    # Violet='#6c71c4',
+    # Blue='#268bd2',
+    # Cyan='#2aa198',
+    # Green='#859900',
+
+    # Base03='#023',
+    # Base02='#034',
+    # Base01='#567',
+    # Base00='#678',
+    # Base0='#899',
+    # Base1='#9aa',
+    # Base2='#eed',
+    # Base3='#ffe',
+    # Yellow='#b80',
+    # Orange='#c41',
+    # Red='#d32',
+    # Magenta='#d38',
+    # Violet='#67c',
+    # Blue='#28d',
+    # Cyan='#2a9',
+    # Green='#890',
+# base03='#1c1c1c',
+# base02='#262626',
+# base01='#585858',
+# base00='#626262',
+# base0='#808080',
+# base1='#8a8a8a',
+# base2='#e4e4e4',
+# base3='#ffffd7',
+# yellow='#af8700',
+# orange='#d75f00',
+# red='#d70000',
+# magenta='#af005f',
+# violet=	'#5f5faf',
+# blue='#0087ff',
+# cyan='#00afaf',
+# green='#5f8700',
+
+Base03='#111',
+Base02='#222',
+Base01='#555',
+Base00='#666',
+Base0='#888',
+Base1='#888',
+Base2='#eee',
+Base3='#ffd',
+Yellow='#a80',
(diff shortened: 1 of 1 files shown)


zulipterminal/config/themes.py
```diff
@@ -28,6 +28,79 @@
...
+# yellow='#af8700',
+# orange='#d75f00',
+# red='#d70000',
+# magenta='#af005f',
+# violet=	'#5f5faf',
+# blue='#0087ff',
+# cyan='#00afaf',
+# green='#5f8700',
+
+Base03='#111',
```
rht: I tried `h234` and `h235`. The color is different from the `#111` and `#222` approximation.

rht: The only way to get `#1c1c1c` accurately is via `h234`. That is, assuming that h17-h255 is not tampered by the user in `.Xdefaults`. I also bumped the pending urwid 24-bit merge. The 24-bit version of solarized has chalkboard-like dark green bg color.

rht: `#262626`

rht: Testing full solarized base03: `#002b36`, base02: `#073642`.

neiljp: @rht We could use 'h' values, but according to the experience on #310 I think we should stick to hex for now, and we'll need a 256-color at least for the moment, and can add 24-bit later, but will still need the 256-color approximation?

If you agree with that:
* how is the use of colors?
* if you've seen solarized being used, are the colors good choices?
* do you see the expected named colors (as opposed to the grey in kitty (?) for selected, for example)

rht: Yet https://github.com/altercation/vim-colors-solarized/blob/master/colors/solarized.vim uses the `h###` version for the 256 mode. I think it is unlikely for h colors higher than 16 to be altered. Maybe put a warn message that says "if you alter your h1-h256, some of the colors in the theme will look different."

> how is the use of colors?

The background color is probably too black.

> do you see the expected named colors (as opposed to the grey in kitty (?) for selected, for example)

Selected messages are blue.

neiljp: @rht 
a) I think we should maintain equivalents in the code, which provide the best matches we can, for both cases, in case we want to provide a `h` mode but also a `#` mode for cases where terminal colors are varied? So we can list both in comments and default to one for the moment, perhaps.
b) There are two background colors? We could use the other (`Base02`) ?
c) Right; this is perhaps an orthogonal change, but with a lack of response on #310 I considered this an extra aspect to experiment with; do you like it?

rht: > I think we should maintain equivalents in the code, which provide the best matches we can, for both cases, in case we want to provide a `h mode` but also a `# mode` for cases where terminal colors are varied?

The `# mode` approximation is likely specific to Urwid (and might be stable across platforms). Though this means having to maintain 4 versions per theme (24-bit, `h mode`, `# mode`, 16 colors). Usually it is automatically selected to do 24-bit, fallback to h mode, then fallback to 16 colors. With 2 versions for 256-color, it's probably too many choices / too complicated for the users to have to check if the h mode is consistent with a screenshot then fallback to # mode.


> b) There are two background colors? We could use the other (Base02) ?

Base02 seems to be the main one. Base03 for stripes, e.g.
```
exe "hi! pandocTableZebraLight"              .s:fg_ptable.s:bg_base03.s:fmt_none
exe "hi! pandocTableZebraDark"               .s:fg_ptable.s:bg_base02.s:fmt_none
```

> c) Right; this is perhaps an orthogonal change, but with a lack of response on #310 I considered this an extra aspect to experiment with; do you like it?

#310 with # mode works on my terminal with green on msg_selected. But again, this is because previously the h mode used the number 1-16 for some of the colors.

neiljp:
Closing in favor of #1175, though color discussion may still be relevant.

PR #327, opened 2019-03-17
Model: Consolidate register calls & improve error handling

neiljp:
This is a follow-up on some earlier refactoring I worked on, inspired by a bug-report by @omi10859 on the #zulip-terminal stream recently. This is not directly targeted at this bug, but brings all the event-registration code into one method, which was previously only used for re-registration, with common error-handling too. This also adds retries for re-connecting, since previously if this failed we just raised an Exception (leaving the thread boundary!)

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -608,16 +608,20 @@ class Model:
                 self.msg_list.log[msg_pos] = new_msg_w
                 self.controller.update_screen()
 
-    def _register_desired_events(self) -> None:
+    def _register_desired_events(self) -> bool:
         try:
             response = self.client.register(event_types=Model.event_types,
                                             client_gravatar=True,
                                             apply_markdown=True)
-        except zulip.ZulipError as e:
-            raise ServerConnectionFailure(e)
-        self.max_message_id = response['max_message_id']
-        self.queue_id = response['queue_id']
-        self.last_event_id = response['last_event_id']
+        except zulip.ZulipError:
+            return False
+
+        if response['result'] == 'success':
+            self.max_message_id = response['max_message_id']
+            self.queue_id = response['queue_id']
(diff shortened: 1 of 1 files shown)


zulipterminal/model.py
```diff
@@ -296,32 +296,9 @@ def _update_initial_data(self) -> None:
...
-                )
-                self.initial_data.update(response)
-                self.max_message_id = response['max_message_id']
-                self.queue_id = response['queue_id']
-                self.last_event_id = response['last_event_id']
-            except zulip.ZulipError as e:
-                raise ServerConnectionFailure(e)
-            except Exception as e:
-                # There was some error in the request sent.
-                if response['result'] == 'error':
```
sumanthvrao: Are we considering this case where an error occurs in the request?

sumanthvrao:
I just had that one clarification above. Everything else looks great!

zulipterminal/model.py
neiljp: The code now explicitly checks for the inverse, `response['result'] == 'success'` instead; we don't get the explicit type of error now, ie. there is now no handling of general exceptions.

Regarding other exceptions:
 * We could catch other exceptions that we understand from the `register` call, but it should just be `ZulipError`? If there are others to handle, we'd want to distinguish them.
 * This code is now all called in extra threads (though was already before, in one case!), so we would (should!) have to pass *unhandled* general exceptions/errors back cleanly into the main thread for logging/shutdown as we do with exceptions in the main thread. That's a bit more complex than this PR was intended to be.

As far as I'm aware we've not had any exceptions logged (and reported) via the code that was removed except the bug I mentioned - so my feeling would be to explore these two cases separately.

sumanthvrao: All-right! This definitely clears things. Thanks for the detailed explanation.

Issue #171, opened 2018-11-27
Fix issues with message merging (hiding content headers) based on star status


neiljp:
When two messages have identical author and very similar date/time, messages are shown adjacently, without repeating either information. These messages appear 'merged', though will highlight separately when scrolled through.

Starring/unstarring a message should 'split' any merged messages, but currently does not do so reliably. The rationale for this is that the content header applies to the entire (apparent) message, or group of messages, and this includes the star marker.

In addition, if a message is separate, it should not just have a star/no-star status, ie. no repeated author/datetime but a `*` (or no star); rather, probably the datetime should be repeated.

The code handling this is entirely in the `MessageBox` class `main_view` function.

varadgunjal:
@zulipbot claim

neiljp:
@varadgunjal Are you working on this?

rohitsanjay:
I want to work on this but I don't know where to start.

rohitsanjay:
I want to work on this one but I need help in where to start.

neiljp:
You might find it useful to explore this by checking the current tests, and adding a new or amended test for the situation(s) described in the issue. if you need more clarification, please ask on #zulip-terminal or here.

shreyamalviya:
@zulipbot claim

zulipbot:
Hello @shreyamalviya, it looks like you've currently claimed 1 issue in this repository. We encourage new contributors to focus their efforts on at most 1 issue at a time, so please complete your work on your other claimed issues before trying to claim this issue again.

We look forward to your valuable contributions!

shreyamalviya:
@neiljp, I'd like to work on this but I'm waiting for a review on another PR, so I can't claim this. Could I please be assigned to this issue? Thanks!

shreyamalviya:
@zulipbot claim

neiljp:
@shreyamalviya Let me know if you need reassigning; was this the screenshot PR? Are you still working on that too?

shreyamalviya:
@neiljp, yes, I just have to change that one image as we discussed on czo (#286).

zulipbot:
Hello @shreyamalviya, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

[...]

PR #335, opened 2019-03-30
Shows/updates content header for message when starred

shreyamalviya:
Part of #171

Splits and shows content header for a message (having a similar timestamp as the previous message(s)) when starred.

For multiple messages in succession having similar timestamps, initially, often times, it would not show that a message is starred when the user presses `*`, although it updates on other Zulip platforms.

Now, however, in such cases, the message is shown as starred along with the timestamp of that message.

~~_What's left:_ When a message is starred, the messages following it do not split from the starred message, making it difficult for the user to figure out if it is one big message which is starred or if there are multiple messages and only one is starred.~~ Fixed.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -116,6 +116,8 @@ class Model:
             'typing': self.handle_typing_event,
             'update_message_flags': self.update_message_flag_status,
         }  # type: Dict[str, Callable[[Event], None]]
+        self.message_star_toggled = False
+        # True if message star status toggled
 
     def get_focus_in_current_narrow(self) -> Union[int, Set[None]]:
         """
@@ -586,6 +588,7 @@ class Model:
         if flag_to_change != 'starred':
             return
 
+        self.message_star_toggled = True
         indexed_message_ids = set(self.index['messages'])
         message_ids_to_mark = set(event['messages'])
 
@@ -618,6 +621,25 @@ class Model:
                 msg_pos = self.msg_list.log.index(msg_w)
                 self.msg_list.log[msg_pos] = new_msg_w
                 self.controller.update_screen()
+
+                if self.message_star_toggled:  # Updates next content_header
+                    try:
+                        next_msg_box =\
+                                self.msg_list.log[msg_pos+1].original_widget
+                        next_msg_w_list = create_msg_box_list(
+                                        self, [next_msg_box.message['id']],
+                                        last_message=next_msg_box.last_message)
+                        if not next_msg_w_list:
+                            return
+                        else:
+                            next_new_msg_w = next_msg_w_list[0]
+                            next_msg_pos = self.msg_list.log.index(
+                                                self.msg_list.log[msg_pos+1])
+                            self.msg_list.log[next_msg_pos] = next_new_msg_w
+                            self.controller.update_screen()
+                            self.message_star_toggled = False
+                    except IndexError:
+                        self.message_star_toggled = False
 
     def _register_desired_events(self, *, fetch_data: bool=False) -> bool:
         fetch_types = None if not fetch_data else [
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -423,23 +423,31 @@ class MessageBox(urwid.Pile):
                      .days)),
             'timestamp': (message['last']['time'] is not None and
                           message['this']['time'] != message['last']['time']),
-            'star_status': (message['this']['is_starred'] !=
-                            message['last']['is_starred']),
         }
         any_differences = any(different.values())
 
-        if any_differences:  # Construct content_header, if needed
-            TextType = Dict[str, Tuple[Optional[str], str]]
-            text = {key: (None, ' ')
-                    for key in ('author', 'star', 'time')}  # type: TextType
+        add_content_header = False  # Construct content_header if == True
+
+        TextType = Dict[str, Tuple[Optional[str], str]]
+        text = {key: (None, ' ')
+                for key in ('author', 'star', 'time')}  # type: TextType
+        if any_differences:
+            add_content_header = True
             if any(different[key] for key in ('recipients', 'author', '24h')):
                 text['author'] = ('name', message['this']['author'])
-            if message['this']['is_starred']:
-                text['star'] = ('starred', "*")
             if any(different[key]
                    for key in ('recipients', 'author', 'timestamp')):
                 text['time'] = ('time', message['this']['time'])
+        if message['this']['is_starred']:
+            add_content_header = True
+            text['star'] = ('starred', "*")
+            text['time'] = ('time', message['this']['time'])
+        elif message['last']['is_starred'] and\
+                not message['this']['is_starred']:
+            add_content_header = True
+            text['time'] = ('time', message['this']['time'])
 
+        if add_content_header:
             content_header = urwid.Columns([
                 ('weight', 10, urwid.Text(text['author'])),
                 (1, urwid.Text(text['star'], align='right')),
@@ -480,7 +488,7 @@ class MessageBox(urwid.Pile):
         # Build parts together and return
         parts = [
             (recipient_header, recipient_header is not None),
-            (content_header, any_differences),
+            (content_header, add_content_header),
             (content, True),
             (reactions, reactions != ''),
         ]
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1252,8 +1252,12 @@ class TestMessageBox:
         varied_message = dict(message, **to_vary_in_each_message)
         msg_box = MessageBox(varied_message, self.model, varied_message)
         view_components = msg_box.main_view()
-        assert len(view_components) == 1
-        assert isinstance(view_components[0], Padding)
+        if to_vary_in_each_message == {'flags': ['starred']}:
+            assert len(view_components) == 2
+            assert isinstance(view_components[1], Padding)
+        else:
+            assert len(view_components) == 1
+            assert isinstance(view_components[0], Padding)
 
     def test_main_view_generates_EDITED_label(self, mocker,
                                               messages_successful_response):


zulipterminal/ui_tools/boxes.py
```diff
@@ -373,23 +373,31 @@ def main_view(self) -> List[Any]:
                      .days)),
             'timestamp': (message['last']['time'] is not None and
                           message['this']['time'] != message['last']['time']),
-            'star_status': (message['this']['is_starred'] !=
-                            message['last']['is_starred']),
         }
         any_differences = any(different.values())
 
+        content_header = 0
+        flag = 0 # content_header needed if == 1
```
sumanthvrao: A more descriptive variable name for example `is_starred_content_header` or something which means the same would help in better readability here. 

(*Note: inline comments are generally followed by 2 spaces before #, else linting fails*)

sumanthvrao (commented):
@shreyamalviya Thanks for working this :)

To address the question you asked on `czo`, 
* When the start status is toggled on a message in View, the keypress calls `toggle_messages_star_status` in `model.py`. 
* This intern calls `update_message_flag_status` which calls `update_rendered_view` 
* This basically does exactly the same thing you mentioned, takes the log of messages and updates/re-displays the one you starred (This effect is then seen in the terminal/ front-end)

To give a gap between the newly starred and following messages, you should probably do the check here and display it accordingly. Let me know if you need clarification on any point :)

zulipterminal/ui_tools/boxes.py
shreyamalviya: Updated. 
Thanks!

neiljp (commented):
I saw that @sumanthvrao messaged you on #zulip-terminal, which I believe was regarding this latest update? I'm not sure if you want to respond or make any changes following that, before I also look in more detail.

That said, the changes here do seem larger than I would expect, and I'm not sure we can rely on the initial state and user updates in zulip-terminal being the only way that the data can change (ie. use `Model.message_star_toggled` that you introduced), as the status can change on the server via other clients. That is why we rely on `is_starred` in the messages themselves, unless there is a part I am missing that needs the new part you introduced.

Note you have no newline at the end of one file, which is why there is the small red symbol; I've not checked to see if the test fails are related to that or not.

shreyamalviya:
> That said, the changes here do seem larger than I would expect, and I'm not sure we can rely on the initial state and user updates in zulip-terminal being the only way that the data can change (ie. use Model.message_star_toggled that you introduced), as the status can change on the server via other clients. That is why we rely on is_starred in the messages themselves, unless there is a part I am missing that needs the new part you introduced.

@neiljp, every time the user presses `*`, the function `update_rendered_view` is called (after `toggle_message_star_status` and `update_message_flag_status`).

In order for there to be a split before the next message, `update_rendered_view` needs to be called for the next message as well. I could've just modified this function so that it would cause the message box of the current message and that of the next message to be updated but that seemed very redundant since `update_rendered_view` is being called at other places as well.

Hence, I introduced `Model.message_star_toggled` which basically checks if `toggle_message_star_status` was called. If true, only then it updates the message box of the next message as well.

> Note you have no newline at the end of one file, which is why there is the small red symbol; I've not checked to see if the test fails are related to that or not.

That, and some of the lines are too long. I'll fix that in the next commit.

neiljp (commented):
If you star-toggle a message *not* in ZT, but have ZT running, none of the code in model.py you added will be triggered. Think about the flow here.

I gave the PR a try, and it seems to be an improvement in general :+1:
When I have a block of 'merged' messages, it picks-out the starred (or not) messages with just a date :+1:  

That said, there were two issues that I noticed while testing manually:
* I get a traceback, at least when I toggle the star at the end
* I expected groups of starred messages to 're-merge', but they did not

The latter point is less of an issue of course, but my feeling was that if messages merged and there was no star, then merging when they were all starred made sense too?

shreyamalviya:
>If you star-toggle a message not in ZT, but have ZT running, none of the code in model.py you added will be triggered. Think about the flow here.

Right, I'll read up on that and fix it. Although I think I'll be requiring some help with this, so I'll ask on czo.

> That said, there were two issues that I noticed while testing manually:
> * I get a traceback, at least when I toggle the star at the end
> * I expected groups of starred messages to 're-merge', but they did not

Fixed the first issue. 
As for the second issue, I'm not sure about the idea... It'll be hard for the user to figure out whether it is one long message which is starred or multiple messages which are all starred.

shreyamalviya:
@neiljp, are there any changes required in this?

zulipterminal/model.py
```diff
@@ -576,7 +579,9 @@ def update_message_flag_status(self, event: Event) -> None:
 
     def update_rendered_view(self, msg_id: int) -> None:
         # Update new content in the rendered view
+        pos = 0  # Tracks position
         for msg_w in self.msg_list.log:
+            pos += 1
```
neiljp: Can we use `msg_pos + 1` in the second block rather than keeping track of another variable like this?

zulipterminal/ui_tools/boxes.py
```diff
@@ -373,23 +373,32 @@ def main_view(self) -> List[Any]:
...
-            'star_status': (message['this']['is_starred'] !=
-                            message['last']['is_starred']),
         }
         any_differences = any(different.values())
 
-        if any_differences:  # Construct content_header, if needed
-            TextType = Dict[str, Tuple[Optional[str], str]]
-            text = {key: (None, ' ')
-                    for key in ('author', 'star', 'time')}  # type: TextType
+        content_header = None
```
neiljp: This line doesn't seem to be necessary?

zulipterminal/ui_tools/boxes.py
```diff
@@ -373,23 +373,32 @@ def main_view(self) -> List[Any]:
...
+        is_content_header = 0  # Construct content_header if == 1
```
neiljp: Would this read clearer as `add_content_header`, and as a `bool`?

zulipterminal/ui_tools/boxes.py
```diff
@@ -430,7 +439,9 @@ def main_view(self) -> List[Any]:
         # Build parts together and return
         parts = [
             (recipient_header, recipient_header is not None),
-            (content_header, any_differences),
+            (content_header, any_differences or
```
neiljp: Can the condition be `is_content_header` (or `add_content_header`), or even just `content_header`?

neiljp (commented):
@shreyamalviya Sorry for the delay -  I was half way through a review and must have lost my session. Here's some more detailed feedback.

zulipterminal/model.py
```diff
@@ -590,6 +595,23 @@ def update_rendered_view(self, msg_id: int) -> None:
                 self.msg_list.log[msg_pos] = new_msg_w
                 self.controller.update_screen()
 
+                if self.message_star_toggled:  # Updates next content_header
```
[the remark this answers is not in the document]
shreyamalviya: Right... I'm surprised I missed that out...

> Would it be equivalent to add an option to update_rendered_view to also update any next message?

I did that initially but I thought it was unnecessary to update the next message every time `update_rendered_view` was called, even when it wasn't needed, so I introduced `message_star_toggled`.

I could reset `message_star_toggled` to `False` or just update the next message every time.
What seems better to you?

zulipterminal/model.py
```diff
@@ -590,6 +595,23 @@ def update_rendered_view(self, msg_id: int) -> None:
...
+                                        last_message=next_msg_box.last_message)
+                        if not next_msg_w_list:
+                            return
+                        else:
+                            next_new_msg_w = next_msg_w_list[0]
+                        next_msg_pos = self.msg_list.log.index(
+                                            self.msg_list.log[pos])
+                        self.msg_list.log[next_msg_pos] = next_new_msg_w
+                        self.controller.update_screen()
+                    except:
```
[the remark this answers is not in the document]
shreyamalviya: Yes, to make sure it doesn't crash in case the last message is starred.

Should I add that it's an IndexError? What else do you suggest to avoid this?

zulipterminal/model.py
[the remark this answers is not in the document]
shreyamalviya: > I could reset message_star_toggled to False or just update the next message every time.

If I do the latter, though, it fails the tests.

```
self = <tests.model.test_model.TestModel object at 0x7fc48eb73f60>, mocker = <pytest_mock.MockFixture object at 0x7fc48eab76a0>
model = <zulipterminal.model.Model object at 0x7fc48ea79ac8>, response = {'message_id': 1, 'rendered_content': '<p>Foo is Boo</p>'}
index = {'edited_messages': {1}, 'messages': {1: {'content': '<p>Foo is Boo</p>', 'id': 1}, 2: {'content': 'Boo is not Foo', 'id': 2}}}

    @pytest.mark.parametrize('response, index', [
        ({
            'message_id': 1,
            'rendered_content': '<p>Foo is Boo</p>'
        }, {
            'messages': {
                1: {
                    'id': 1,
                    'content': 'Boo is Foo',
                },
                2: {
                    'id': 2,
                    'content': "Boo is not Foo"
                }
            },
            'edited_messages': set()
        })
    ])
    def test_update_message(self, mocker, model, response, index):
        model.index = index
        model.msg_list = mocker.Mock()
        mock_msg = mocker.Mock()
        another_msg = mocker.Mock()
        model.msg_list.log = [mock_msg, another_msg]
        mock_msg.original_widget.message = index['messages'][1]
        another_msg.original_widget.message = index['messages'][2]
        mocker.patch('zulipterminal.model.create_msg_box_list',
                     return_value=[mock_msg])
        model.update_message(response)
        assert model.index['messages'][1]['content'] == \
            response['rendered_content']
        assert model.msg_list.log[0] == mock_msg
        # Ensure the index shows the message has been updated (edited)
        assert model.index['edited_messages'] == {1}
>       self.controller.update_screen.assert_called_once_with()
E       AssertionError: Expected 'update_screen' to be called once. Called 4 times.
E       
E       pytest introspection follows:

tests/model/test_model.py:623: AssertionError
```

shreyamalviya:
@neiljp, does this require any changes?

Issue #330, opened 2019-03-24
Update user list after application is started


neiljp:
The list of users in the right panel should be updated as the app finds out about it, as in the webapp.

Given the natural visibility of this panel in the no-autohide view (and an expectation it is the 'current' list of users) this seems like a high priority change.

sumanthvrao:
@zulipbot claim

sumanthvrao:
![](https://github.trello.services/images/mini-trello-icon.png) [Update users automatically](https://trello.com/c/CUyHQpSw/3-update-users-automatically)

PR #336, opened 2019-03-31
Use response from server to update user list after application has started.

sumanthvrao:
Version 1a included only notifying the server of user presence. Version 1b now includes using the response obtained by the server to refresh the users list once the application has started.

Fixes: #330

**TODO**: Focus jumps to left column view at the interval of every minute. This is more likely to be because of `self.body = urwid.Columns(body, focus_column=0)` statement in `build_window_frame` function. Any suggestions on how to handle the jump?

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -189,6 +189,16 @@ class Model:
             current_ids = self.index['all_starred']
         return current_ids.copy()
 
+    def _notify_server_of_presence(self) -> None:
+        response = self.client.update_presence(
+                request={
+                    # TODO: Determinal `status` from terminal tab focus.
+                    'status': 'active' if self.new_user_input else 'idle',
+                    'new_user_input': self.new_user_input,
+                }
+            )
+        self.new_user_input = False
+
     @asynch
     def _start_presence_updates(self) -> None:
         """
@@ -198,14 +208,7 @@ class Model:
         # FIXME: Version 2: call endpoint with ping_only=True only when
         #        needed, and rely on presence events to update
         while True:
-            response = self.client.update_presence(
-                request={
-                    # TODO: Determinal `status` from terminal tab focus.
-                    'status': 'active' if self.new_user_input else 'idle',
-                    'new_user_input': self.new_user_input,
-                }
-            )
-            self.new_user_input = False
+            self._notify_server_of_presence()
             time.sleep(60)
 
     @asynch
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -189,7 +189,7 @@ class Model:
             current_ids = self.index['all_starred']
         return current_ids.copy()
 
-    def _notify_server_of_presence(self) -> None:
+    def _notify_server_of_presence(self) -> Dict[str, Any]:
         response = self.client.update_presence(
                 request={
                     # TODO: Determinal `status` from terminal tab focus.
@@ -198,17 +198,24 @@ class Model:
                 }
             )
         self.new_user_input = False
+        return response
 
     @asynch
     def _start_presence_updates(self) -> None:
         """
-        Notify server of user's presence each minute (version 1a)
+        Call `_notify_server_of_presence` every minute (version 1a).
+        Use 'response' to update user list (version 1b).
         """
-        # FIXME: version 1b: Also use 'response' to update user list
         # FIXME: Version 2: call endpoint with ping_only=True only when
         #        needed, and rely on presence events to update
         while True:
-            self._notify_server_of_presence()
+            response = self._notify_server_of_presence()
+            if response['result'] == 'success':
+                self.initial_data['presences'] = response['presences']
+                self.users = self.get_all_users()
+                if hasattr(self.controller, 'view'):
+                    self.controller.view.users_view.update_user_list(
+                        user_list=self.users)
             time.sleep(60)
 
     @asynch
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -411,17 +411,27 @@ class RightColumnView(urwid.Frame):
                                               header=search_box)
 
     @asynch
-    def update_user_list(self, search_box: Any, new_text: str) -> None:
-        if not self.view.controller.editor_mode:
+    def update_user_list(self, search_box: Any=None,
+                         new_text: str="",
+                         user_list: Any=None) -> None:
+
+        assert ((user_list is None and search_box is not None) or
+                (user_list is not None and search_box is None and
+                 new_text == ""))
+
+        if not self.view.controller.editor_mode and not user_list:
             return
         # wait for any previously started search to finish to avoid
         # displaying wrong user list.
         self.search_lock.acquire()
+        if user_list:
+            self.view.users = user_list
         users = self.view.users.copy()
         users_display = users.copy()
-        for user in users:
-            if not match_user(user, new_text):
-                users_display.remove(user)
+        if new_text:
+            for user in users:
+                if not match_user(user, new_text):
+                    users_display.remove(user)
         self.body = self.users_view(users_display)
         self.set_body(self.body)
         self.view.controller.update_screen()
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -406,6 +406,7 @@ class RightColumnView(urwid.Frame):
             trcorner=u'─', blcorner=u'─', rline=u'',
             bline=u'─', brcorner=u'─'
             )
+        self.allow_update_user_list = True
         self.search_lock = threading.Lock()
         super(RightColumnView, self).__init__(self.users_view(),
                                               header=search_box)
@@ -420,6 +421,8 @@ class RightColumnView(urwid.Frame):
                  new_text == ""))
 
         if not self.view.controller.editor_mode and not user_list:
+            return
+        if not self.allow_update_user_list and new_text == "":
             return
         # wait for any previously started search to finish to avoid
         # displaying wrong user list.
@@ -470,9 +473,11 @@ class RightColumnView(urwid.Frame):
 
     def keypress(self, size: Tuple[int, int], key: str) -> str:
         if is_command_key('SEARCH_PEOPLE', key):
+            self.allow_update_user_list = False
             self.set_focus('header')
             return key
         elif is_command_key('GO_BACK', key):
+            self.allow_update_user_list = True
             self.user_search.set_edit_text("Search People")
             self.body = UsersView(
                 urwid.SimpleFocusListWalker(self.users_btn_list))
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -694,6 +694,16 @@ class TestRightColumnView:
         right_col_view.users_view.assert_called_with(assert_list)
         set_body.assert_called_once_with(right_col_view.body)
 
+    def test_update_user_presence(self, right_col_view, mocker,
+                                  user_list):
+        set_body = mocker.patch(VIEWS + ".urwid.Frame.set_body")
+
+        right_col_view.update_user_list(user_list=user_list)
+
+        right_col_view.search_lock.acquire.assert_called_once_with()
+        right_col_view.users_view.assert_called_with(user_list)
+        set_body.assert_called_once_with(right_col_view.body)
+
     @pytest.mark.parametrize('users, users_btn_len, editor_mode, status', [
         (None, 1, False, 'active'),
         ([{


zulipterminal/model.py
```diff
@@ -179,24 +179,40 @@ def get_message_ids_in_current_narrow(self) -> Set[int]:
...
         """
-        # FIXME: version 1b: Also use 'response' to update user list
-        # FIXME: Version 2: call endpoint with ping_only=True only when
-        #        needed, and rely on presence events to update
-        while True:
-            response = self.client.update_presence(
+        self._notify_server_of_presence()
+        self._start_presence_updates()
+
+    def _notify_server_of_presence(self) -> None:
```
neiljp: I like that you've extracted this method, but perhaps extract it first with a refactor commit, then add features?

zulipterminal/model.py
```diff
@@ -179,24 +179,40 @@ def get_message_ids_in_current_narrow(self) -> Set[int]:
             current_ids = self.index['all_starred']
         return current_ids.copy()
 
-    @asynch
-    def _start_presence_updates(self) -> None:
+    def _start_presence_initially(self) -> None:
```
neiljp: Do you have a reason for including this method? It's only called once and the 2 calls can just go inline in the `__init__` if needed?

zulipterminal/model.py
```diff
@@ -179,24 +179,40 @@ def get_message_ids_in_current_narrow(self) -> Set[int]:
...
+        response = self.client.update_presence(
                 request={
                     # TODO: Determinal `status` from terminal tab focus.
                     'status': 'active' if self.new_user_input else 'idle',
                     'new_user_input': self.new_user_input,
                 }
             )
-            self.new_user_input = False
+        self.new_user_input = False
+        self.response = response
```
neiljp: Is `self.response` something you just added? Can it just be a return value?

zulipterminal/ui.py
```diff
@@ -79,6 +79,11 @@ def footer_view(self) -> Any:
         text_header = self.get_random_help()
         return urwid.AttrWrap(urwid.Text(text_header), 'footer')
 
+    def show_updated_right_column(self, model: Any) -> Any:
+        self.users = model.users
+        self.right_panel = self.right_column_view()
+        super(View, self).__init__(self.build_window_frame())
```
neiljp: Is this necessary? What are you trying to do?

zulipterminal/model.py
sumanthvrao: Will do so :+1:

zulipterminal/model.py
sumanthvrao: Yes we can add it in `__init__` since it's asynchronous.

zulipterminal/model.py
sumanthvrao: @neiljp I made it an attribute of the object, as it would avoid passing a huge response as a function parameter. What do you suggest here?

zulipterminal/ui.py
sumanthvrao: I am trying to call only the `build_window_frame()` method here. As both the left and middle windows are already built. However only the right panel - which is the user's list has to be refreshed (built again), but now with the new user's list.

zulipterminal/model.py
```diff
@@ -188,15 +199,8 @@ def _start_presence_updates(self) -> None:
...
-            response = self.client.update_presence(
-                request={
-                    # TODO: Determinal `status` from terminal tab focus.
-                    'status': 'active' if self.new_user_input else 'idle',
-                    'new_user_input': self.new_user_input,
-                }
-            )
-            self.new_user_input = False
             time.sleep(60)
+            self._notify_server_of_presence()
```
neiljp: If these two lines are reordered, do you need to call `_notify...` from `__init__`?

zulipterminal/model.py
neiljp: I'm not sure precisely what you meant by asynchronous here, but it seems to be a good refactor :+1:

zulipterminal/model.py
neiljp: The data could be large, but the returned data is just a new name for the same data, as is typical in python.

To put the question back to you: why do you think it is better to set it as an attribute?

zulipterminal/model.py
```diff
@@ -189,18 +192,21 @@ def _notify_server_of_presence(self) -> None:
                 }
             )
         self.new_user_input = False
+        self.response = response
 
     @asynch
     def _start_presence_updates(self) -> None:
         """
-        Notify server of user's presence each minute (version 1a)
+        Use 'response' to update user list (version 1b).
```
neiljp: The 'every minute' can stay here :)

zulipterminal/ui.py
neiljp: I was trying to understand why you were using `super`, really.

That said, can you just use `update_screen()` instead of splitting the function and rebuilding it each time? That is, do you really need the commit that splits the method and makes `build_window_frame`?

zulipterminal/model.py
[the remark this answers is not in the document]
sumanthvrao: To elaborate,
during initial setup, we call `_notify...` once to inform server of user's presence, we would then call `_start_presence.....` (sleep) and continue to setup, views, UI, etc for startup. Once 60 seconds have passed, we only update server of new details and update the same immediately in our UI.

The 2 lines were re-ordered as we would else display user status in UI which was 60 seconds ago. I choose to treat the 2 calls (1st - initial setup and 2nd - every other time) to be different as this involved not repeating `_notify...` code in 2 places.

zulipterminal/model.py
```diff
@@ -181,6 +181,9 @@ def get_message_ids_in_current_narrow(self) -> Set[int]:
         return current_ids.copy()
 
     def _notify_server_of_presence(self) -> None:
+        """
+        Notify server of user's presence every minute.
```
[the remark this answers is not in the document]
sumanthvrao: True :+1: . will remove this.

zulipterminal/model.py
sumanthvrao: I was thinking of his from a C perspective (that it would be a pass-by-value here). Oops . It's a pass-by-reference.

zulipterminal/model.py
sumanthvrao: Done :+1:

zulipterminal/ui.py
sumanthvrao: I did this in the current iteration. Did away with the first commit, using `update_screen()` to reflect the changes in the Right User column. I also replaced the `super` call with a normal one. However the changes don't seem to work. I am a bit puzzled here. We are building the UI every 60 seconds here?

sumanthvrao:
@neiljp I did away with the first commit. Using `update_screen()`. Also instead of `super` I am doing a simple call to `main_window()`. This, however, does not seem to work. That is, the right-column does not get updated. It is pretty baffling. I do not seem to understand the issue here. What do you suggest?

sumanthvrao:
> Have you checked that the functions as they are can just work for an update vs the initial setup?

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: To rephrase: if you switch the order of the `sleep` and `_notify_server_of_presence` here (about lines 202/203), and have `_start_presence_updates` in `__init__`, without the `_notify_server_of_presence`, I think the net result should be almost identical, right?

zulipterminal/model.py
```diff
@@ -180,7 +180,7 @@ def get_message_ids_in_current_narrow(self) -> Set[int]:
             current_ids = self.index['all_starred']
         return current_ids.copy()
 
-    def _notify_server_of_presence(self) -> None:
+    def _notify_server_of_presence(self) -> Any:
```
neiljp: I think you should be able to tighten up this response type a little.

zulipterminal/ui.py
```diff
@@ -79,6 +79,12 @@ def footer_view(self) -> Any:
         text_header = self.get_random_help()
         return urwid.AttrWrap(urwid.Text(text_header), 'footer')
 
+    def show_updated_right_column(self, model: Any) -> Any:
+        self.users = model.users
+        self.right_panel = self.right_column_view()
+        self.main_window()
+        self.controller.update_screen()
```
neiljp: We discussed the issues you were having here on chat.zulip.org; I suspect the easier solution to integrate with the existing interface is probably to avoid adding this method and instead use `RightColumnView.update_user_list`, possibly adjusted slightly.

zulipterminal/model.py
```diff
@@ -189,18 +189,21 @@ def _notify_server_of_presence(self) -> None:
                 }
             )
         self.new_user_input = False
+        return response
 
     @asynch
     def _start_presence_updates(self) -> None:
         """
-        Notify server of user's presence each minute (version 1a)
+        Use 'response' to update user list every minute (version 1b).
```
neiljp: This docstring change seems to reflect the added functionality, but you've lost some information; do you understand what?

zulipterminal/ui.py
amanagr: Yes, this is the proper way to update the list. We don't want want re-initialising the view every minute might have other problems like what happens if the user is searching for users and we re-initialise the view?

zulipterminal/model.py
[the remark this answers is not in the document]
sumanthvrao: Oops. misunderstood you previously. Yup. They are identical. Made the changes.

zulipterminal/model.py
sumanthvrao: That I should add `Notify server of user's presence each minute` as a doc-string to the `_notify_server_of_presence` function ? :)

zulipterminal/model.py
sumanthvrao: Done.

zulipterminal/ui.py
sumanthvrao: Thanks for the tip @neiljp and @amanagr! I have tweaked `update_user_list` to display the updated user list. Let me know how the PR looks.

zulipterminal/ui_tools/views.py
```diff
@@ -411,17 +411,28 @@ def __init__(self, width: int, view: Any) -> None:
                                               header=search_box)
 
     @asynch
-    def update_user_list(self, search_box: Any, new_text: str) -> None:
-        if not self.view.controller.editor_mode:
+    def update_user_list(self, search_box: Any=None,  # type: ignore
+                         new_text: str=None,
+                         user_list: Any=None) -> None:
```
sumanthvrao: I wanted to tighten this up. But it is leading to a cascade of mypy - 'unable to infer type' in further places.

zulipterminal/ui_tools/views.py
```diff
@@ -406,22 +406,34 @@ def __init__(self, width: int, view: Any) -> None:
...
+
+        assert ((user_list is None and search_box is not None and
+                 new_text is not None) or
+                (user_list is not None and search_box is None and
+                 new_text is None))
+
+        if not user_list and not self.view.controller.editor_mode:
             return
```
shreyamalviya: Why is this `not self.view.controller.editor_mode`? From what I understand, this basically means that if the user is *not* searching, it'll return? I might be getting confused so please correct me if I'm wrong.

sumanthvrao: Ah. Thanks for bringing this up @shreyamalviya ! This will definitely help pen down my though process behind the addition :+1: 
The `update_user_function` in itself gets called in 2 ways :+1: 
* 1: When the `urwid.connect` on self.user_search gets triggered  
* 2: Due to this new feature of update_user_list. 

When it gets triggered from part one, then we must return at `self.view.controller.edit_mode` as the user is not searching. (same reason you mentioned). However, If we call from the second position, we need to go ahead and update the user_list (rest of function). Therefore we check the extra condition here.

sumanthvrao:
[Update users automatically](https://trello.com/c/CUyHQpSw/3-update-users-automatically)

zulipterminal/ui_tools/views.py
```diff
@@ -411,17 +411,30 @@ def __init__(self, width: int, view: Any) -> None:
...
+
+        if not self.view.controller.editor_mode and not user_list:
             return
         # wait for any previously started search to finish to avoid
         # displaying wrong user list.
         self.search_lock.acquire()
+        if user_list:
+            self.view.users = user_list
         users = self.view.users.copy()
         users_display = users.copy()
```
neiljp: I think this is ok to stay for a v1 of this new feature, but I think it'd be good to see these copies avoided for the `user_list` case, and simplified further using a comprehension otherwise.

zulipterminal/model.py
```diff
@@ -204,8 +204,9 @@ def _start_presence_updates(self) -> None:
             self.initial_data['presences'] = response['presences']
             self.users = self.get_all_users()
             if hasattr(self.controller, 'view'):
-                self.controller.view.users_view.update_user_list(
-                    user_list=self.users)
+                if self.controller.view.users_view.allow_update_user_list:
+                    self.controller.view.users_view.update_user_list(
+                        user_list=self.users)
```
neiljp: It seems a little strange having this logic in the model; wouldn't it be simpler in the view, where the variable is?

zulipterminal/ui_tools/views.py
```diff
@@ -411,17 +411,30 @@ def __init__(self, width: int, view: Any) -> None:
...
+    def update_user_list(self, search_box: Any=None,
+                         new_text: str="",
+                         user_list: Any=None) -> None:
+
+        assert ((user_list is None and search_box is not None and
+                 new_text != "") or
+                (user_list is not None and search_box is None and
+                 new_text == "") or
+                (user_list is None and search_box is not None and
+                 new_text == ""))
```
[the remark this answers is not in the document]
sumanthvrao: Done :+1:

zulipterminal/ui_tools/views.py
sumanthvrao: Agreed. Will keep this in mind :+1:

zulipterminal/model.py
sumanthvrao: Done :+1: 
I added a 
```if not self.allow_update_user_list and new_text == "":
            return
```
Inside `update_user_list`, which basically checks if user began search and if the function is not being called from `urwid.connect_signal` (no new alphabet entered to do search) - i.e the function must have been called from the thread to update user list (60s) and hence we can return safely.

zulipterminal/model.py
```diff
@@ -198,17 +198,23 @@ def _notify_server_of_presence(self) -> None:
...
+        Call `_notify_server_of_presence` every minute (version 1a).
+        Use 'response' to update user list (version 1b).
         """
-        # FIXME: version 1b: Also use 'response' to update user list
         # FIXME: Version 2: call endpoint with ping_only=True only when
         #        needed, and rely on presence events to update
         while True:
-            self._notify_server_of_presence()
+            response = self._notify_server_of_presence()
+            self.initial_data['presences'] = response['presences']
```
amanagr: @sumanthvrao response object is not guaranteed to have `presences`. Also, ` self._notify_server_of_presence()` should be in try catch statement.

sumanthvrao: Ah. All-right. Makes sence. Will do the additions :+1:

sumanthvrao: This may also be the issue @neiljp was facing (KeyError?)

amanagr: Oops, `run.py` will catch error in `self._notify_server_of_presence()` , so no need to put in try-catch statement. ;)

amanagr: we should always check for `if response['result'] == 'success':`

sumanthvrao: Ah. Got it. Yup, am adding the response check for success.

neiljp: @amanagr Good review catch in the checking for success here - I've not reproduced the issue I had previously with this PR, in this updated version :+1:  I'm not convinced that the try/catch would help here, as `_start_presence_updates` is `@asynch`, so the exceptions are in different threads, and that doesn't just resolve itself automatically (as possibly previously discussed elsewhere).

neiljp:
@sumanthvrao Merged with a few textual changes only - great to have this merged! :tada:

PR #337, opened 2019-04-02
Undoing thumbs-up reaction now always works.

### LINKED ISSUE #334: Can't undo thumbs-up reaction in new ZT session
When a person adds a thumbs-up reaction to a message, if they exit the terminal session and start a new one, the reaction can't be undone through ZT.

Also, a pre-existing reaction (previously added from any other Zulip platform) can't be toggled unless the same session of ZT has been running since the reaction was added.

konraddeka:
This fixes #334 : previously reaction couldn't have been undone, unless it was added in current session of zulip-terminal.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -210,10 +210,14 @@ class Model:
             emoji_code='1f44d',
             reaction_type='unicode_emoji',
             message_id=str(message['id']))
-        existing_reactions = [reaction['emoji_code']
-                              for reaction in message['reactions']
-                              if ('user_id' in reaction['user'] and
-                                  reaction['user']['user_id'] == self.user_id)]
+        existing_reactions = [
+            reaction['emoji_code']
+            for reaction in message['reactions']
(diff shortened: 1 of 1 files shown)


amanagr:
@konraddeka thanks for working on this issue. This works as expected.
Please modify the commit message to something like: 
```
react_to_message: Fix :+1: reaction toggle.

__Properly describe here what problem it fixes__
Fixes #334 
```

konraddeka:
@amanagr @neiljp thank you for input, I changed the commit message as suggested.

amanagr:
LGTM, @neiljp ?

neiljp:
Right - this fixes things fine for now :+1:  

We should watch out if there's a fix in the server (zulip/zulip#12049) at some point as this will make this simpler again :)

neiljp:
@konraddeka Thanks! You might want to check out the follow-up commit I just added, where I just made a small change to extend the tests to ensure we don't break this :)

PR #339, opened 2019-04-04
UI/Model: Add feature to edit sent messages.

amanagr:


amanagr:
@neiljp thanks for the review. Can you reproduce the error ? I am not getting any errors.

I don't think this issue is related to 'topic' vs 'subject'.

neiljp:
@amanagr I just reproduced with this PR, rebased against master.

1. Go to a test message I wrote recently in #**test here**
2a. Edit the topic (in topic bar) -> traceback
2b. Edit the topic (in 'edit message') -> traceback

Editing the content is fine, but that's not the enhancement (in that regard) in this PR.

neiljp:
Are you breaking this up into other PRs, now? I saw #344 but wasn't sure if it was  ready or still being updated.

amanagr:
It's ready. :) I will try and reproduce the error.

neiljp:
@amanagr We may have miscommunicated, since from my perspective the separate elements were between sending and receiving. From that perspective the first step is the `update_message` commit, which has the error I observed.

The other behavior (sending from ZT) depends on that commit; I'm not sure if you changed behavior in that respect?

amanagr:
I think we can work in this PR itself. Also, @neiljp I am not seeing any tracebacks, both editing topics and content works smoothly. Can u try and debug?

neiljp:
@amanagr If you're happy with the rewrite & tests, I plan to merge a version with the my 'fix code' commit squashed into yours as per the top 3 commits at this branch: https://github.com/neiljp/zulip-terminal/commits/update-message-with-topic

The rewritten tests emphasize the updated message only, but examine different combinations of changed subject and content. If you test it without the updated 'fix code' commit I added, you'll see the tests fail, in a similar way I experienced while testing manually. Note that your commit now has the code change (to be squashed with the fix), and mine has test updates - I removed the test changes from your commit.

If this fix/tests look good I'll push them and we can rebase and examine this in terms of editing from ZT?

amanagr:
Yeah looks good! You can go ahead and rebase. Sorry for the late response. Thanks for the improvements.

neiljp:
@amanagr I just cherry-picked those commits :+1:

amanagr:
@punchagan @neiljp I have updated the PR with resolved conflicts. This seems ready for editing sent messages and topics. Future improvements will include:
* Editing topic of last n messages in a topic. ("change_later")
* Editing topic of all messages in a topic. ("change all")
Let me know your thoughts on this.

tests/ui/test_ui.py
```diff
@@ -58,6 +59,14 @@ def test_set_footer_text_default(self, view, mocker):
         view._w.footer.set_text.assert_called_once_with(['some help text'])
         view.controller.update_screen.assert_called_once_with()
 
+    def test_set_footer_text_with_duration(self, view, mocker):
+        mocker.patch('zulipterminal.ui.View.get_random_help',
+                     return_value=['some help text'])
+        view.set_footer_text("custom text", 0.1)
+        view._w.footer.set_text.assert_called_with(['some help text'])
```
neiljp: Minor: maybe extract the common text as a test function parameter?

tests/ui/test_ui_tools.py
```diff
@@ -1217,6 +1218,105 @@ def test_main_view_generates_EDITED_label(self, mocker,
...
+        assert write_box.is_edit_mode is False
+
+        # MEET ALL THE CONDITIONS
+        msg_box.message['timestamp'] = time.time()
+        msg_box.model.user_id = message['sender_id']
+        msg_box.keypress(size, key)
+
+        write_box.msg_write_box.set_edit_text.assert_called_once_with("MSG")
+        assert write_box.is_edit_mode is True
+
```
neiljp: Whenever I see these sections within one test, I wonder whether it could be 'parametrize'd. I think this would make it clearer. With `id`s, it could be even clearer, particularly if a test subsequently fails.

tests/ui/test_ui_tools.py
```diff
@@ -1217,6 +1218,105 @@ def test_main_view_generates_EDITED_label(self, mocker,
...
+        # MSG SENT BY OTHER USER
+        msg_box.model.user_id = 2  # != sender_id
+        msg_box.keypress(size, key)
+
+        write_box.msg_write_box.set_edit_text.assert_not_called()
+        assert write_box.is_edit_mode is False
+
+        # MEET ALL THE CONDITIONS
+        msg_box.message['timestamp'] = time.time()
```
neiljp: While I see that this can work, it would seem less fragile to just mock it to return the same value?

zulipterminal/ui_tools/boxes.py
```diff
@@ -529,6 +547,37 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+        elif is_command_key('EDIT_MESSAGE', key):
+            if self.message['sender_id'] != self.model.user_id:
+                self.model.controller.view.set_footer_text(
+                        " You can't edit messages sent by other users.", 3)
+                return key
+            # Time since the message was sent by user
+            time_since_msg_sent = time() - self.message['timestamp']
+            # Time limit set by realm admin
+            edit_time_limit = self.model.initial_data[
+                    'realm_message_content_edit_limit_seconds']
```
neiljp: I would drop the two time-related statements down and break the `elif` below; this keeps these variables closer to where they're used.

Also, if the variables are clear enough, or there are hopefully-clear (!) error messages as here, I would drop the comments - I would prefer a blank line to these comments, as long as the code is sufficiently clear :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -18,6 +18,8 @@ def __init__(self, view: Any) -> None:
         super(WriteBox, self).__init__(self.main_view(True))
         self.model = view.model
         self.view = view
+        self.is_edit_mode = False  # type: bool
```
neiljp: Maybe add an `in` to the variable name? *Maybe* remove the `is`?
(though see below)

zulipterminal/ui_tools/boxes.py
```diff
@@ -18,6 +18,8 @@ def __init__(self, view: Any) -> None:
...
+        self.msg_id = 0  # Defined when a message is edited
```
neiljp: It might be clearer in the text (and remove the need for this comment) if you named the variable differently?

At risk of combining two related states/meanings (though ensuring they're consistent), would it make sense to combine these two variables as an `Optional[int]` ?

zulipterminal/ui_tools/boxes.py
```diff
@@ -82,19 +84,35 @@ def stream_box_view(self, button: Any=None, caption: str='',
...
+                    )
+                else:
+                    success = self.model.send_private_message(
+                        recipients=self.to_write_box.edit_text,
+                        content=self.msg_write_box.edit_text
+                    )
             if success:
                 self.msg_write_box.edit_text = ''
+                if self.is_edit_mode:
+                    self.is_edit_mode = False
```
neiljp: You're not resetting `msg_id` here? (though see other comment re combining them)

tests/ui/test_ui.py
amanagr: I think it's fine since we are not testing it here.

amanagr:
@neiljp I have updated this PR with the requested changes.

zulipterminal/ui_tools/boxes.py
```diff
@@ -82,19 +83,35 @@ def stream_box_view(self, button: Any=None, caption: str='',
...
         if is_command_key('SEND_MESSAGE', key):
-            if not self.to_write_box:
-                success = self.model.send_stream_message(
-                    stream=self.stream_write_box.edit_text,
-                    topic=self.title_write_box.edit_text,
-                    content=self.msg_write_box.edit_text
-                )
+            if self.msg_edit_id:
+                if not self.to_write_box:
+                    success = self.model.update_stream_message(
```
sumanthvrao: While this definitely works, I thought it a little odd to check whether the 'topics' box has been set (or None) to determine whether it is a stream / PM. Maybe we could check the message type like always? Just a suggestion (v2?)

tests/ui/test_ui.py
neiljp: `'some help text'` ?

amanagr: `test_set_footer_text_default` is test for that, right?

zulipterminal/ui.py
```diff
@@ -67,13 +69,19 @@ def get_random_help(self) -> List[Any]:
...
-    def set_footer_text(self, text_list: Optional[List[Any]]=None) -> None:
+    @asynch
+    def set_footer_text(self, text_list: Optional[List[Any]]=None,
+                        duration: int=0) -> None:
         if text_list is None:
             text = self.get_random_help()
         else:
-            text = text_list
+            # Append additional space custom text for clear appearance
+            text = [" "] + [text_list]  # type: ignore
```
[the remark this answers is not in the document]
amanagr: Thanks included!

zulipterminal/ui_tools/boxes.py
amanagr: We will have to provide writebox with the whole msg, so not sure about this.

tests/ui/test_ui.py
neiljp: I agree that the other test uses that text; I think the issue is which combination of parameters are used? This test sets a duration with a random help test? This doesn't actually correspond to how you're then using the code, which is with specific text and a duration?

amanagr:
@neiljp I have resolved all conflicts in this PR.

zulipterminal/ui_tools/boxes.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, view: Any) -> None:
         super(WriteBox, self).__init__(self.main_view(True))
         self.model = view.model
         self.view = view
+        self.msg_edit_id = False  # type: Optional[int]
```
neiljp: Should this be `None`? This seems like a mypy bug, assuming no type coercion or 'Falsey' type-checking? Are we using the latest mypy?

tests/ui/test_ui_tools.py
```diff
@@ -1235,6 +1235,44 @@ def test_main_view_generates_EDITED_label(self, mocker,
...
+        varied_message = dict(message_fixture, **to_vary_in_each_message)
+        key = 'e'
+        size = (20,)
+        msg_box = MessageBox(varied_message, self.model, message_fixture)
+        msg_box.model.user_id = 1
+        msg_box.model.initial_data = dict()
+        msg_box.model.initial_data[
+            'realm_allow_message_editing'] = editing_allowed
+        msg_box.model.initial_data[
+            'realm_message_content_edit_limit_seconds'] = 60
```
neiljp: Why not combine the previous lines to initialize `initial-data` in one statement?

tests/ui/test_ui_tools.py
```diff
@@ -1235,6 +1235,44 @@ def test_main_view_generates_EDITED_label(self, mocker,
...
+        msg_box.model.client.get_raw_message.return_value = {
+            'raw_content': "Edit this message"
+        }
+        write_box = msg_box.model.controller.view.write_box
+        write_box.msg_edit_id = False
```
neiljp: `None`?

zulipterminal/ui_tools/boxes.py
```diff
@@ -575,6 +592,34 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+            if time_since_msg_sent >= edit_time_limit:
+                self.model.controller.view.set_footer_text(
+                        " Time Limit for editing the message has"
+                        " been exceeded.", 3)
+                return key
+            self.keypress(size, 'enter')
+            msg_id = self.message['id']
+            msg = self.model.client.get_raw_message(msg_id)['raw_content']
+            write_box = self.model.controller.view.write_box
+            write_box.msg_edit_id = msg_id
```
neiljp: Is `write_box` the same as `self` here?

amanagr: No - MessageBox

zulipterminal/ui_tools/boxes.py
amanagr: Would it be better with `0` instead?

neiljp: I think that `None` makes it obvious that it is invalid, whereas `0` could be a valid id.

zulipterminal/ui_tools/boxes.py
neiljp: I think this is reasonable to keep for now, but perhaps we can look into this later - this isn't as obvious as it could be, certainly.

amanagr:
The best way to disable the boxes would be to add a method to `ReadlineEdit` to disable keypress( I will open a PR there for this and then we can add it here.)
Also, the msg cannot go anywhere else since we are not sending stream and to fields to the server.

tests/ui/test_ui_tools.py
```diff
@@ -1235,6 +1235,43 @@ def test_main_view_generates_EDITED_label(self, mocker,
             assert label[0].text == '(EDITED)'
             assert label[1][1] == 11
 
+    @pytest.mark.parametrize('to_vary_in_each_message, editing_allowed,\
+                             pass_test', [
```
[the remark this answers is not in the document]
amanagr: Done!

zulipterminal/ui_tools/boxes.py
amanagr: Cool! Done.

neiljp:
@amanagr Merged :)

Issue #326, opened 2019-03-11
Last 30 messages being muted causes crash when starting up ZT


sumanthvrao:
Steps to repro:
* Mark Stream #Denmark to be muted (Web-App) in User 1 (Eg: Iago).
* In User 2 (Eg: Hamlet) on send 30 odd (around 40 to be on safe side) messages to stream #Denmark. (Through ZT or Web-app)
* Open user Iago from ZT causing it to crash.

Complete Traceback: 
```
ERROR:root:Can't set focus, ListBox is empty
Traceback (most recent call last):
  File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/cli/run.py", line 238, in main
    autohide_setting).main()
  File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/core.py", line 38, in __init__
    self.view = View(self)
  File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/ui.py", line 39, in __init__
    super(View, self).__init__(self.main_window())
  File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/ui.py", line 84, in main_window
    self.center_panel = self.message_view()
  File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/ui.py", line 46, in message_view
    self.search_box)
  File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/ui_tools/views.py", line 284, in __init__
    msg_list = MessageView(model)
  File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/ui_tools/views.py", line 34, in __init__
    self.set_focus(self.focus_msg)
  File "/home/sumanthvrao/.local/share/virtualenvs/zulip-terminal-sRIAC_7x/lib/python3.6/site-packages/urwid/listbox.py", line 578, in set_focus
    raise IndexError("Can't set focus, ListBox is empty")
IndexError: Can't set focus, ListBox is empty
```

sumanthvrao:
**Reason why there is a crash**: In the default all-messages view, ZT loads last 30 messages, but doesn't display muted messages. All of them being muted results in no messages being in focus.

**Preferred approach**: If muted messages are found then an equal number of messages are downloaded. (*This specific algorithm may not work if all the streams are muted.*)

sumanthvrao:
@zulipbot add "bug" "help wanted"

timabbott:
[...]

PR #341, opened 2019-04-06
[WIP] views/utils: Fetch last 30 non-muted messages.

sumanthvrao:
Muted messages are not displayed in default All_message view. This causes less then 30 messages to be displayed in case there were any previous muted messages. We now keep on fetching until there are 30 non-muted messages, which can be displayed in default view.

(This is a step in the direction of solving #326)

### DIFF
--- a/zulipterminal/ui_tools/utils.py
+++ b/zulipterminal/ui_tools/utils.py
@@ -45,6 +45,9 @@ def create_msg_box_list(model: Any, messages: Union[None, Iterable[Any]]=None,
         last_msg = msg
     if focus_msg is not None:
         model.set_focus_in_current_narrow(focus_msg)
+    model.muted_msg_count = muted_msgs
+    if message_list:
+        model.last_msg_id = message_list[0]['id']
     return w_list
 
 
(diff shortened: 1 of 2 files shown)


sumanthvrao:
@neiljp Let me know your thoughts on this. Whether this is the right direction to proceed in.

zulipterminal/ui_tools/utils.py
```diff
@@ -45,6 +45,9 @@ def create_msg_box_list(model: Any, messages: Union[None, Iterable[Any]]=None,
         last_msg = msg
     if focus_msg is not None:
         model.set_focus_in_current_narrow(focus_msg)
+    model.muted_msg_count = muted_msgs
+    if message_list:
+        model.last_msg_id = message_list[0]['id']
```
neiljp: I'm not sure if you wanted a review of this approach, but setting these attributes is only applicable to each fragment of message list created using this function, which seems strange.

sumanthvrao:
[Last 30 messages crash](https://trello.com/c/tjT8F5uo/4-last-30-messages-crash)

sumanthvrao:
This is being developed in another PR #446 with a better approach. I'm closing this due to the same reason.

PR #342, opened 2019-04-07
boxes: Reduce message padding after introduction of (EDITED).

neiljp:
This makes a little more space for the message text, but would appreciate what people think compared to master after the merge of #328.


neiljp:
Just updated with a 5-padding on the right, which seems a little clearer than 2, perhaps due to being able to scan for the time.

neiljp:
After various discussion on #zulip-terminal on chat.zulip.org in 'Width-scaling', I've merged this - hopefully that will lead to a broader audience for any feedback, which of course is welcome in that stream or here :)

PR #345, opened 2019-04-13
BUGFIX: Fix sending to multiple recipients when narrowed.

neiljp:
In Model.append_message we now split the recipient email string from the
narrow into a list which is mapped to frozenset of user-id, instead of
previously assuming we could look up the recipient string as a single
recipient email.

CHANGELOG.md updated.


sumanthvrao:
Just tested this manually! Nice bug-fix @neiljp :)

neiljp:
I was going to add some tests, but I'm thinking we could leave this for a general cleanup of the append tests?

PR #346, opened 2019-04-13
model: Prevent user from reacting to their own message.

sumanthvrao:
In the web-app users cannot react to their own message;
This is done by not providing the option in the UI;
We do a similar approach here by doing a noop when
the hotkey is used to react on a user's own message.

Tests amended.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -240,6 +240,7 @@ class TestModel:
             for er in existing_reactions
         ]
         message = dict(
+            sender_id=5,  # any
             id=msg_id,
             reactions=full_existing_reactions)
         reaction_spec = dict(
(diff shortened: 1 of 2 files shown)


sumanthvrao:
@neiljp Something I noticed while solving a related issue. Wrote up a quick PR fix for this. Let me know if this is all-right.

mregdos:
> In the web-app users cannot react to their own message; This is done by not providing the option in the UI;

I am able to react to my own messages in web-app. For example: 
[here](https://chat.zulip.org/#narrow/stream/7-test-here/topic/Test.20self-reactions/near/732707)
. The option to do so is contained within message's menu ("add emoji reaction" which is above "copy link to conversation"). Therefore I am not comfortable with removing such feature from zulip-terminal, since it would introduce inconsistency.

sumanthvrao:
Nice find @mregdos! Apologies for just assuming this. Closed the issue due to the basis being wrong :)

Issue #147, opened 2018-07-30
Show hint for pressing SHIFT for selecting with the mouse.


aero31aero:
Currently, trying to select anything with a mouse in zulip-terminal immediately deselects the text in xterm. The solution is to provide a hint for the following detail:

> Most terminal UI frameworks implement pressing SHIFT and then dragging to select text as you would expect (passing that mouse input to your terminal emulator).

I believe we do not have any use of selecting text in zt so far, so it could be possible to attach an event listener and then show a small message hinting them to use SHIFT and select.

Conversation that inspired the issue: https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/xterm.20copy.2Fpaste.3F

neiljp:
Closed via #349.

PR #349, opened 2019-04-16
Display a hint when the user tries to select text with mouse

vonknio:
Implements #147 

The hint is displayed in the footer where random hints are displayed. The key suggested is `Fn + Alt` on OS X and `Shift` elsewhere. `Fn + Alt` [worked for many people](https://github.com/mitmproxy/mitmproxy/issues/597#issuecomment-163819671) in another project using Urwid. `Shift` works for me on Linux but not on OS X.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
 from time import ctime
 from datetime import datetime
+from sys import platform
 from typing import Any, Dict, List, Tuple, Union, Optional
 
 import emoji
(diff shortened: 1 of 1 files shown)


vonknio:
@neiljp 
> @vonknio Very minor, but perhaps the more specific `self.displaying_selection_hint`?

Makes sense. I've changed the variable name

rht:
> Also, do we have another tester for OSX? (I'm guessing @rht ?)

I don't have access to any physical Darwin devices and there is no AWS instance for it either.

neiljp:
@rht Thanks for your response; I'm not sure who else has OSX access?

amanagr:
Tested on my friend's mac, works fine! Thanks @vonknio for this.

aero31aero:
Yes, this looks neat and fits what I was looking for in a feature like this.

neiljp:
I appear to not have received notifications about this, but given the feedback this looks great to merge :)
@vonknio Thanks for this! :+1:

PR #353, opened 2019-04-25
core: Fix action of 'z' on self-messages.

sumanthvrao:
**Steps to repro the bug**: 
* Send a test message to yourself (PM)
* Go to All Private messages narrow.
* Use z on any message with self.
The narrow does not change to message with self instead switched to All messages.

**Commit message**: 
When 'z' hotkey was used on in the All private messages
narrow, on self messages, it did not switch narrow
to PM with oneself. We now handle this logic explicitly.

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -174,6 +174,8 @@ class Controller:
             emails = [recipient['email']
                       for recipient in button.message['display_recipient']
                       if recipient['email'] != self.model.client.email]
+            if not emails and len(button.message['display_recipient']) == 1:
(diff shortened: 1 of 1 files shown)


sumanthvrao:
@neiljp yes, I also didn't find tests for code surrounding keypress `z`.

shreyamalviya:
Tested. Looks good! :+1:

neiljp:
@shreyamalviya Thanks for the test report; given the simplicity, I'm merging this now so we can fix the bug.

@sumanthvrao Thanks for identifying this and providing the fix!

Issue #135, opened 2018-07-20
Show message info on pressing `i`.


amanagr:
Display the focused message's info on pressing <kbd>i</kbd>. Message info should comprise of:

- [x] Sender details
- [x] Message Edit History
- [x] Details of emoji - who reacted
...

skapil:
@zulipbot claim

zulipbot:
Welcome to Zulip, @skapil! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

skapil:
@zulipbot claim

Pradhvan:
@amanagr I would like to take this issue up

Pradhvan:
@zulipbot claim

varadgunjal:
@neiljp I started working on this without realizing I needed to stake claim to this. If it's still open can I take you up on your offer of help?

Pradhvan:
@neiljp hey, yes I am still working on this. Sorry had been a bit busy with job search and festival coming up so could not give it much time, now everything is done so I could work on it again :)  @varadgunjal if you don't mind you can pick up another issue :)

neiljp:
@Pradhvan @varadgunjal 
I thought I'd replied to this already, but maybe you could collaborate, or discuss on here or chat.zulip.org (on #zulip-terminal) how you plan to go about this?

varadgunjal:
@neiljp As mentioned above, since I was requested to tackle another issue, I stopped working on this and am looking for other opportunities to contribute to the codebase.

Pradhvan:
@varadgunjal if you haven't picked up any other issue, I would be happy to collaborate on the current issue :smile:

varadgunjal:
[...]

PR #347, opened 2019-04-16
Improve recipient bar layout and design

sumanthvrao:
This involves redoing the top bar, to better represent the current narrow and remove redundant information.
*All message view* - 

all_msg

*Stream view* -
stream

Currently it supports for :
* All messages,
* All Private messages,
* Streams
* Stream and topic
* PM's
* Group PM's
* Starred messages

Fixes: #340 
**Based on discussion in** - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/top.20recipient.20bar/near/733211

PR #354, opened 2019-04-26
Displays message info on pressing `i`, edit history on pressing `e`

shreyamalviya:
As of now, the following details are shown when a user presses `i` on a message:
* date/time (with seconds)
* sender
* sender's email ID
* recipient (`stream -> message` in case of a stream message, `recipient name` in case of a private message, "`Self`" in case of a message to oneself or a private message sent to them by someone else)
* reactions (of the form - `name of reaction: name of user who reacted` sorted alphabetically, "`---None---`" if no reactions)
* edit history (pressing `e` shows another window with the edit history)

Issues:
* When displaying reactions, the 'thumbs-up' reaction is shown as `+1` sometimes, and as `thumbs_up` other times. I can't seem to figure out the reason behind this behavior.

Issue #135

### DIFF
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -148,6 +148,11 @@ KEY_BINDINGS = OrderedDict([
         'keys': {'i'},
         'help_text': 'View message information',
     }),
+    ('EDIT_HISTORY', {
+        'keys': {'e'},
+        'help_text': 'View edit history from message information box',
+        'excluded_from_random_tips': True,
+    }),
     ('QUIT', {
         'keys': {'ctrl c'},
         'help_text': 'Quit',
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -14,7 +14,7 @@ from zulipterminal.helper import asynch
 from zulipterminal.model import Model, GetMessagesArgs, ServerConnectionFailure
 from zulipterminal.ui import View, Screen
 from zulipterminal.ui_tools.utils import create_msg_box_list
-from zulipterminal.ui_tools.views import HelpView, MsgInfoView
+from zulipterminal.ui_tools.views import HelpView, MsgInfoView, EditHistoryView
 from zulipterminal.config.themes import ThemeSpec
 from zulipterminal.ui_tools.views import PopUpConfirmationView
 
@@ -114,6 +114,10 @@ class Controller:
         msg_info_view = MsgInfoView(self, msg)
         self.show_pop_up(msg_info_view,
                          "Message Information (up/down scrolls)")
+
+    def show_edit_history(self, msg: Any) -> None:
+        edit_history_view = EditHistoryView(self, msg)
+        self.show_pop_up(edit_history_view, "Edit History (up/down scrolls)")
 
     def search_messages(self, text: str) -> None:
         # Search for a text in messages
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -484,6 +484,8 @@ class MessageBox(urwid.Pile):
             edited_label_size = 0
             left_padding = 8
 
+        self.message['content_to_show'] = content
+
         content = urwid.Padding(
             urwid.Columns([
                 (edited_label_size,
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -767,12 +767,12 @@ class MsgInfoView(urwid.ListBox):
                             "\n"
                             for reaction in msg['reactions']])
             reactions[-1] = reactions[-1].rstrip("\n")
-
         msg_info = OrderedDict([
             ('Date & Time', time.ctime(msg['timestamp'])[:-5]),
             ('Sender', msg['sender_full_name']),
             ('Sender\'s Email ID', msg['sender_email']),
             ('Reactions', reactions if msg['reactions'] else '---None---'),
+            ('Edit History', 'Press \'e\' to view'),
             ])
 
         widths = [(len(field)+7,
@@ -798,4 +798,150 @@ class MsgInfoView(urwid.ListBox):
     def keypress(self, size: Tuple[int, int], key: str) -> str:
         if is_command_key('GO_BACK', key) or is_command_key('MSG_INFO', key):
             self.controller.exit_popup()
+        elif is_command_key('EDIT_HISTORY', key):
+            self.controller.show_edit_history(self.msg)
         return super(MsgInfoView, self).keypress(size, key)
+
+
+class EditHistoryView(urwid.ListBox):
+
+    def __init__(self, controller: Any, msg: Any):
+        self.controller = controller
+        self.msg = msg
+
+        self.width = 70
+        self.height = 0  # Calculated depending on contents to show
+        self.to_show = []  # type: List[urwid.Pile]
+
+        if 'edit_history' in self.msg:
+            # Current version
+            timestamp = time.ctime(msg['last_edit_timestamp'])[:-5]
+            topic = [msg['subject']]
+            # tracks topic for versions with no topic change
+            orig_topic = topic
+            if 'prev_subject' in msg['edit_history'][0]:
+                topic.append(" (edited)")
+            content = [msg['content_to_show']]
+            # tracks content for versions with no content change
+            orig_content = content
+            self.version_contents = []  # type: List[Any]
+
+            self.version_contents.extend([
+                urwid.Columns([
+                    urwid.AttrWrap(
+                        urwid.Text(topic), 'edit_topic'),
+                    (20, urwid.AttrWrap(
+                        urwid.Text(timestamp, align='right'),
+                        'edit_time'))
+                        ]),
+                urwid.AttrWrap(
+                    urwid.Text("(Current version)"), 'edit_history'),
+                urwid.Text([content, "\n"])
+                ])
+            if not msg['edit_history'][0]['user_id'] == msg['sender_id']:
+                # if edit_author not sender, find author by user id
+                edit_author = ''
+                for user in self.controller.view.users:
+                    if user['user_id'] == msg['edit_history'][0]['user_id']:
+                        edit_author = user['full_name']
+                self.version_contents.extend([
+                    urwid.AttrWrap(
+                        urwid.Text(["edited by ", edit_author, "\n"],
+                                   align='right'), 'edit_history')])
+
+            self.to_show.append(urwid.Pile(self.version_contents))
+
+            self.height += self.to_show[-1].rows((self.width,))
+
+            # Edited versions (all except current & original)
+            for version in range(0, len(msg['edit_history'])-1):
+                # each version contains previous version's content and subject
+                # timestamp of future version matches with
+                # current version's contents
+                self.version_contents = []
+                timestamp = time.ctime(msg['edit_history'][version+1]
+                                          ['timestamp'])[:-5]
+                if 'prev_subject' in msg['edit_history'][version]:
+                    topic = [msg['edit_history'][version]['prev_subject']]
+                    orig_topic = topic
+                    if 'prev_subject' in msg['edit_history'][version+1]:
+                        if msg['edit_history'][version+1]['prev_subject'] !=\
+                           msg['edit_history'][version]['prev_subject']:
+                            topic.append(" (edited)")
+                else:
+                    topic = orig_topic
+                    if 'prev_subject' in msg['edit_history'][version+1]:
+                        topic.append(" (edited)")
+                if 'prev_content' in msg['edit_history'][version]:
+                    content = [msg['edit_history'][version]['prev_content']]
+                    orig_content = content
+                else:
+                    content = orig_content
+                edit_author_id = msg['edit_history'][version+1]['user_id']
+                self.edit_history_content(topic,
+                                          timestamp,
+                                          content,
+                                          edit_author_id)
+                self.to_show.append(urwid.Pile(self.version_contents))
+                self.height += self.to_show[-1].rows((self.width,))
+
+            # Original version
+            timestamp = time.ctime(msg['timestamp'])[:-5]
+            if 'prev_subject' in msg['edit_history'][-1]:
+                topic = [msg['edit_history'][-1]['prev_subject']]
+            else:
+                topic = orig_topic
+            if 'prev_content' in msg['edit_history'][-1]:
+                content = [msg['edit_history'][-1]['prev_content']]
+            else:
+                content = orig_content
+            self.edit_history_content(topic,
+                                      timestamp,
+                                      content,
+                                      self.msg['sender_id'])
+            self.to_show.append(urwid.Pile(self.version_contents))
+            self.height += self.to_show[-1].rows((self.width,))
+
+        # Unedited
+        else:
+            self.to_show = urwid.Pile([
+                (urwid.AttrWrap(
+                    urwid.Text("Unedited"), 'edit_history'))
+            ])
+            self.to_show[-1].rows((self.width,))
+
+        self.log = urwid.SimpleFocusListWalker(self.to_show)
+
+        super(EditHistoryView, self).__init__(self.log)
+
+    def edit_history_content(self,
+                             topic: List[str],
+                             time: str,
+                             content: List[Any],
+                             edit_author_id: int) -> None:
+        self.version_contents.extend([
+                            urwid.Columns([
+                                (urwid.AttrWrap(
+                                    urwid.Text(topic), 'edit_topic')),
+                                (20, urwid.AttrWrap(
+                                    urwid.Text(time, align='right'),
+                                    'edit_time'))]),
+                            (urwid.Text([content, "\n"]))])
+        if not edit_author_id == self.msg['sender_id']:
+            # if edit_author not sender, find author by user id
+            for user in self.controller.view.users:
+                if user['user_id'] == edit_author_id:
+                    edit_author = user['full_name']
+            self.version_contents.extend([
+                            urwid.AttrWrap(urwid.Text(["edited by ",
+                                                       edit_author, "\n"],
+                                                      align='right'),
+                                           'edit_history')])
+
+    def keypress(self, size: Tuple[int, int], key: str) -> str:
+        if is_command_key('MSG_INFO', key):
+            self.controller.exit_popup()
+        elif is_command_key('GO_BACK', key) or\
+                is_command_key('EDIT_HISTORY', key):
+            self.controller.show_msg_info(self.msg)
+        return super(EditHistoryView, self).keypress(size, key)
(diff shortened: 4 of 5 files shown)


sumanthvrao:
@shreyamalviya Looks good :+1: . Here's my feedback. Hope its helpful :)
* I see there is a `Merge commit`. using `git pull --rebase upstream master` instead of the normal pull would avoid this merge commit. You can maybe drop this commit and rebase now.
* You may want to split your PR into further smaller commits, so that the review process is easy.  Each commit can be like a logical unit, for example, in a commit a feature (or part of code) may be added and a test to check the features correctness. This method was advised to me during my early contribution phase and has stuck on ever since :).

Based on some manual testing :
* I use `gruvbox` as my default theme, the white on white-background for some parts of the Info box seems little difficult to read.
* The sender name, date& time - seem a little redundant, don't you think? provided they are already visible in the normal layout.
* I see that you have calculated the width of the Message Info box based on how much content there is to show, but to reduce the complexity, we could have a fixed width? What do you think?
* For messages with no reactions, the reactions part, we could avoid listing that row in info box, instead of the `---None---`. What do you think about this?

shreyamalviya:
@sumanthvrao, thanks a lot for that! 

> * I use `gruvbox` as my default theme, the white on white-background for some parts of the Info box seems little difficult to read.

I noticed the same for the help menu too. We should probably fix that as well.

> * The sender name, date& time - seem a little redundant, don't you think? provided they are already visible in the normal layout.

I thought the same initially but I felt it seemed a little incomplete without those...?

> * I see that you have calculated the width of the Message Info box based on how much content there is to show, but to reduce the complexity, we could have a fixed width? What do you think?

We could do that but in cases where the content to be shown is really short, the info box will unnecessarily take up space, and in cases where the content is longer than expected, it'll look quite chaotic. Again, we could probably do that when we're showing the help menu since the content is static there.

> * For messages with no reactions, the reactions part, we could avoid listing that row in info box, instead of the `---None---`. What do you think about this?

As a user, I'd personally prefer that I be told there aren't any reactions for the message, rather than the whole field just disappearing. But then that's me... What would you as a user prefer to see?

sumanthvrao:
@shreyamalviya To clarify re: fixed width : I meant the Message Info box width could be fixed and the height could be variable (for the same reason you pointed out).
Re last point: Yeah. Let's wait for other feedback's to regarding this.

shreyamalviya:
@sumanthvrao, I was actually referring to the msg info box's width as well when I said the above.

neiljp:
@shreyamalviya Another approach to consider in the popup, inspired by my feedback re the code being similar to in `MessageBox`, would be to have something like a list of `MessageBox`s for the Edit history, in which case a lot of the code is already there - though that would replace some of your current implementation.

zulipterminal/ui_tools/views.py
```diff
@@ -587,5 +587,38 @@ def __init__(self, controller: Any) -> None:
...
+                urwid.Columns([
+                    urwid.Text(msg_info_keys[msg_info_values.index(field)]),
+                    (max_widths[1], urwid.Text(field))
+                ], dividechars=2),
+                None if index % 2 else 'bar')
+             for index, field in enumerate(msg_info_values)])
+
+        self.number_of_actions = len(self.log)
+
+        super(MsgInfoView, self).__init__(self.log)
```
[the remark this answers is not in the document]
sumanthvrao: As a datapoint -> an example of a refactor commit would be https://github.com/zulip/zulip-terminal/commit/2a69ec5fb9509314fa0d968a3ec98c3c4d89463e. You could add the 'Refactor' in the commit message and even move the commit up, before you introduce any commits with features.

zulipterminal/model.py
```diff
@@ -656,3 +656,31 @@ def poll_for_events(self) -> None:
                 for event_type, do_action_for in self.event_actions.items():
                     if event_type == event['type']:
                         do_action_for(event)
+
+    def get_msg_info(self, msg: Dict[str, Any]) -> Dict[str, Any]:
+        msg_info = {
+            'DATE & TIME': time.ctime(msg['timestamp'])[:-5],
+            'SENDER': msg['sender_full_name'],
+            'SENDER\'S EMAIL ID': msg['sender_email'],
+            'RECIPIENT': msg['display_recipient'] + " -> " + msg['subject'] if
```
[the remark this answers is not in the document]
shreyamalviya: > This kind of information is already calculated for the MessageBox; could you reuse some of that functionality to simplify this?

It's all mostly stored in variables within functions (such as `stream_header()` which calculates `Recipient` in the msg info box) which can't be accessed outside the function right now (as far as I know). Is there some simple way to do this?

zulipterminal/config/keys.py
```diff
@@ -136,6 +136,11 @@
         'keys': {'i'},
         'help_text': 'View message information',
     }),
+    ('EDIT_HISTORY', {
+        'keys': {'e'},
+        'help_text': 'View edit history',
+        'excluded_from_random_tips': True,
+    }),
```
neiljp: We may have to change this key later, if we do plan to use this key for editing messages.

zulipterminal/model.py
```diff
@@ -656,31 +656,3 @@ def poll_for_events(self) -> None:
...
-                        else reaction['emoji_name'] +
-                        ": " +
-                        reaction['user']['full_name'] +
-                        "\n"
-                        for reaction in msg['reactions']])
-            if msg['reactions']
-            else '---None---',
-            # 'EDIT HISTORY':
-            }
-        return msg_info
```
neiljp: It seems a bit strange to introduce this in the first commit and then remove it in this commit?

Also, if you can make this commit just about adding the edit history, that would probably be sufficient for one commit.

zulipterminal/core.py
```diff
@@ -266,6 +266,20 @@ def _finalize_show(self, w_list: List[Any]) -> None:
...
+        msg_info = MsgInfoView(self, msg)
+        self.loop.widget = urwid.Overlay(
+            urwid.LineBox(msg_info,
+                          title="Message information"),
+            self.view,
+            align='center',
+            valign='middle',
+            # +2 to both of the following, due to LineBox
+            width=msg_info.width+2,
+            height=min(3*rows//4, msg_info.number_of_actions)+2
```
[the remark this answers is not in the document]
neiljp: I'll wait for the next iteration to see the longer window, as we discussed on #zulip-terminal :)

zulipterminal/core.py
```diff
@@ -83,20 +83,33 @@ def update_screen(self) -> None:
...
         )
 
+    def show_help(self) -> None:
+        help_view = HelpView(self)
+        self.show_pop_up(help_view, "Help Menu (up/down scrolls)")
+
+    def show_msg_info(self, msg: Any) -> None:
+        msg_info_view = MsgInfoView(self, msg)
+        self.show_pop_up(msg_info_view,
+                         "Message Information (up/down scrolls)")
```
neiljp: This refactor commit is good to see :+1: As you may have seen in Tim's comment in #learning, it can be clearer if you adapt the code first (like this), then add the feature in a follow-up.

zulipterminal/core.py
```diff
@@ -83,20 +83,33 @@ def update_screen(self) -> None:
...
+
+    def show_edit_history(self, msg: Any) -> None:
+        edit_history_view = EditHistoryView(self, msg)
+        self.show_pop_up(edit_history_view, "Edit History (up/down scrolls)")
+
```
neiljp: This class doesn't exist at this stage in the sequence of commits, so it seems a bit strange to introduce it in this commit - maybe with the edit-history commit next/later?

zulipterminal/ui_tools/views.py
```diff
@@ -587,5 +588,108 @@ def __init__(self, controller: Any) -> None:
...
+
+        if msg['type'] == 'stream':
+            recipient = msg['display_recipient'] + " -> " + msg['subject']
+        elif not msg['display_recipient'][0]['full_name'] ==\
+                msg['sender_full_name']:
+            recipient = msg['display_recipient'][0]['full_name']
+        else:
+            recipient = 'Self'
+
+        msg_info = {
```
neiljp: Using a dict here doesn't ensure that the results will always be in this order, at least for early enough versions of python3. For 3.6+ we should be good, but for now an `OrderedDict` may be preferable - we use this elsewhere for this reason.

zulipterminal/ui_tools/views.py
```diff
@@ -587,5 +588,108 @@ def __init__(self, controller: Any) -> None:
...
+            else '---None---',
+            'Edit History': 'Press \'e\' to view',
+            }
+        msg_info_keys = list(msg_info.keys())
+        msg_info_values = list(msg_info.values())
+        widths = [(len((msg_info_keys[msg_info_values.index(field)]))+7,
+                  max(len(reaction_users) for reaction_users in field)
+                  if isinstance(field, list)
+                  else len(field)+2)
+                  for field in msg_info_values]
```
neiljp: Can you use `msg_info.items()` to simplify this?

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: Let's wait until we're closer to the final version; we may be able to refactor and use that code, but I'm sure we can all see the idea right now in any case :)

zulipterminal/ui_tools/views.py
```diff
@@ -589,3 +590,107 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+            'Reactions': sorted(
+                    [reaction['emoji_name'] +
+                        ": " +
+                        reaction['user']['full_name'] +
+                        "\n"
+                        for reaction in msg['reactions']])
+            if msg['reactions']
+            else '---None---',
+            'Edit History': 'Press \'e\' to view',
+            })
```
neiljp: Unfortunately passing a dict to an `OrderedDict` doesn't preserve the order, as the dict you pass in is not guaranteed to be ordered.

zulipterminal/ui_tools/views.py
```diff
@@ -589,3 +590,107 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+        self.controller = controller
+        self.msg = msg
+
+        if msg['type'] == 'stream':
+            recipient = msg['display_recipient'] + " -> " + msg['subject']
+        elif not msg['display_recipient'][0]['full_name'] ==\
+                msg['sender_full_name']:
+            recipient = msg['display_recipient'][0]['full_name']
+        else:
+            recipient = 'Self'
```
neiljp: I'm not sure about this private-message logic:
* while perhaps obvious, we've generally included 'you' (or 'self' here) within the list of recipients we display
* this doesn't seem to work for multi-recipient PMs

zulipterminal/ui_tools/views.py
```diff
@@ -589,3 +590,107 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+                          else 6, 27)
+
+        self.log = urwid.SimpleListWalker(
+            [urwid.AttrWrap(
+                urwid.Columns([
+                    urwid.Text(field[0]),
+                    (max_widths[1], urwid.Text(field[1]))
+                ], dividechars=2),
+                None if index % 2 else 'bar')
+             for index, field in enumerate(msg_info.items())])
```
neiljp: Maybe something like the following would simplify this further? (or other variable names, of course)
`for index, (title, data) in enumerate(msg_info.items()`

zulipterminal/ui_tools/views.py
```diff
@@ -589,3 +590,107 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+            'Date & Time': time.ctime(msg['timestamp'])[:-5],
+            'Sender': msg['sender_full_name'],
+            'Sender\'s Email ID': msg['sender_email'],
+            'Recipient': recipient,
+            'Reactions': sorted(
+                    [reaction['emoji_name'] +
+                        ": " +
+                        reaction['user']['full_name'] +
+                        "\n"
+                        for reaction in msg['reactions']])
```
neiljp: This leads to a trailing newline in the reactions section each time.

zulipterminal/ui_tools/views.py
```diff
@@ -589,3 +590,107 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+            if msg['reactions']
+            else '---None---',
+            'Edit History': 'Press \'e\' to view',
+            })
+
+        widths = [(len(field[0])+7,
+                  max(len(reaction_users) for reaction_users in field[1])
+                  if isinstance(field[1], list)
+                  else len(field[1])+2)
+                  for field in msg_info.items()]
```
neiljp: I'd again suggest using a tuple in this for loop rather than indexing `field`, ie. using the `for key, value in some_dict.items()` style.

amanagr:
@shreyamalviya are still working on this?

shreyamalviya:
@amanagr, yes! Just got done with my college exams/assessments. I'll be getting back on to this in a day or two.

zulipterminal/core.py
```diff
@@ -83,21 +83,25 @@ def update_screen(self) -> None:
...
+                          title=display_text),
             self.view,
             align='center',
             valign='middle',
             # +2 to both of the following, due to LineBox
-            width=help_view.width+2,
-            height=min(3*rows//4, help_view.number_of_actions)+2
+            width=to_show.width+2,
+            height=to_show.height+2 if hasattr(to_show, 'height')
+                  else min(3*rows//4, to_show.number_of_actions)+2
```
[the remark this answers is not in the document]
amanagr: This appears to we working fine for tall windows, was this for a previous version?

zulipterminal/ui_tools/views.py
amanagr: I think we should remove the `Recipient` part from message info, since we are very clearly showing it in normal view.

zulipterminal/core.py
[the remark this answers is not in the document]
shreyamalviya: @neiljp, I don't understand what you mean?

zulipterminal/ui_tools/views.py
shreyamalviya: I think removing it makes sense after #347. @neiljp, what do you say?

zulipterminal/core.py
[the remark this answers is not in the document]
neiljp: Well, can each class that wants to be used as a popup just set a height attribute?

zulipterminal/ui_tools/views.py
```diff
@@ -604,3 +605,100 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+        return super(MsgInfoView, self).keypress(size, key)
+
+
+class EditHistoryView(urwid.ListBox):
+    def __init__(self, controller: Any, msg: Any):
+        self.controller = controller
+        self.msg = msg
+
+        self.width = 70
+        self.height = 10
```
neiljp: Could this be adjusted based on the content available?

zulipterminal/ui_tools/views.py
```diff
@@ -604,3 +605,100 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+
+        if 'edit_history' in self.msg:
+            self.log = urwid.SimpleListWalker(
+                    [urwid.Pile([
+                        (urwid.AttrWrap(
+                            urwid.Text(
+                                time.ctime(
+                                    version['timestamp'])[:-5]), 'bar')),
+                        (urwid.Text(
+                                "\n*---Subject Edited---*\n" +
```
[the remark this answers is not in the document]
shreyamalviya: Agreed. Working on that right now.

I also noticed that if a message is edited, its timestamp shown in the normal view is still the time that the original message was sent. Is that on purpose or should it be the time that the message was last edited?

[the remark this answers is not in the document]
neiljp: I suspect the original time is fixed, as otherwise it would be out of order when edited?

amanagr:
Thanks @shreyamalviya for your efforts. I understand it can be a great learning curve to pull all this together in one.  This is what I feel can be improved overall:
* I feel comments are majorly missing in classes. 
* Your commit messages are too short, try to include more information about the changes you have made. 
* Try using the message_fixture in conftest.py instead of creating a new message fixture for the class. You can modify the message inside the test function to make it adapt to the test. `test_main_view_compact_output` is a good example in `test_ui_tools.py`.

amanagr:
@neiljp I think it can be useful to have the raw_message content in-case of rendering issues. `lxml` can be used to do it easily.

shreyamalviya:
@amanagr, just made the changes.

neiljp:
@shreyamalviya I've merged 2½ of these commits, namely the first two and the tests in the last commit which pertain to that first View. The remainder of the commits in this PR after rebasing should now be related to showing edit history.

I had to resolve some conflicts after merging the muting-stream commits from @sumanthvrao, and also removed some commented-out code, switched to the 'height' approach I mentioned previously and added a README entry for the 'i' key. I may have made a few other adjustments, but it should still look solidly like your code - but if you have any concerns please let me know. 

You might find it helpful for commit summaries to use the gitlint tool that we have set up; it's of course not required, but it helps with suggesting formatting improvements.

neiljp:
@shreyamalviya Just a note here to note that I tested this and it seems to be slightly broken right now; there are no connections to load the edit window. I've got a minimal fix, so let me know what the current state of your local code is and we can work towards a solution.

zulipterminal/ui_tools/views.py
```diff
@@ -798,4 +798,176 @@ def __init__(self, controller: Any, msg: Any) -> None:
...
+        super(EditHistoryView, self).__init__(self.log)
+
+    def edit_history_content(self,
+                             topic: List['str'],
+                             time: str,
+                             content: List[Any],
+                             edit_author_id: int) -> None:
+        # display who edited if edit_author not sender
+        if edit_author_id == self.msg['sender_id']:
+            self.to_show.append(
```
neiljp: Rather than appending to `self.to_show` from within the method, it may be clearer if this method returned the `Pile` to append?

zulipterminal/ui_tools/views.py
```diff
@@ -798,4 +798,176 @@ def __init__(self, controller: Any, msg: Any) -> None:
...
+                            (20, urwid.AttrWrap(
+                                urwid.Text(timestamp, align='right'),
+                                'edit_time'))
+                            ]),
+                        urwid.AttrWrap(
+                            urwid.Text("(Current version)"), 'edit_history'),
+                        urwid.Text([content]),
+                        urwid.AttrWrap(
+                            urwid.Text(["edited by ", edit_author, "\n"],
+                                       align='right'), 'edit_history'),
```
neiljp: Could you simplify the above two `Pile`s using the same `edit_message_history`, at least if you adjust slightly how that method works?

zulipterminal/ui_tools/boxes.py
```diff
@@ -484,6 +484,8 @@ def main_view(self) -> List[Any]:
             edited_label_size = 0
             left_padding = 8
 
+        self.message['content_to_show'] = content
```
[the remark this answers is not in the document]
shreyamalviya: Yes. Where should it be according to you?

neiljp:
@shreyamalviya Thanks for your work on this! After the original message information popup added some time ago, I just merged #663 which is reworked version of the edit history :tada:

With that merge, we should have covered the material from this PR, so I'm closing this now.

PR #356, opened 2019-05-12
requirements: Explicitly install pygments 2.3.1 for pudb on py 3.4.

sumanthvrao:
The recent release of pygments 2.4.0 does not have support for python 3.4. It throws the following error in CI :
```
An error occurred while installing pygments==2.4.0! Will try again.
No matching distribution found for pygments==2.4.0
(from -r /tmp/pipenv-fafr2rmh-requirements/pipenv-if7u64fy-requirement.txt (line 1))
```
Setting the requirements of pygments explicitly to 2.3.1 is a workaround in this case.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -18,6 +18,7 @@ pytest = "==3.4.2"
 pytest-pep8 = "==1.0.6"
 pytest-mock = "==1.7.1"
 pytest-cov = "==2.5.1"
+pygments= "==2.3.1"  # For pudb, python 3.4
 pudb = "==2017.1.4"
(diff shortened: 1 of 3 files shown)


neiljp:
@sumanthvrao Thanks for identifying this issue and providing a workaround! I added extra detail into the commit text and reformatted the lines slightly (missing comma & space) but this is now merged :+1: 

I also filed this upstream as inducer/pudb#340.

sumanthvrao:
Thanks for the merge @neiljp :)

PR #361, opened 2019-05-24
run: Add UploadCommand class to automate release process.

amanagr:
Moved this code form https://github.com/zulip/zulint/blob/master/setup.py.

I will just leave this pull request open for now and test merge it when making the next release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,8 +47,15 @@ dev_helper_deps = [
 class UploadCommand(Command):
     """Support setup.py upload."""
 
-    description = 'Build and publish the package.'
-    user_options = []
+    description = "build and upload release to test/main server."
+    "Read Making a new release wiki of zt on github "
+    "for proper guidelines on making a new release."
+    user_options = [
+        # The format is (long option, short option, description).
+        ('final-release=', None,
+         'set it to "True" to push the release to the main PyPI server.'
+         'This also adds git tags and pushes them upstream.'),
+    ]
 
     @staticmethod
     def status(s):
@@ -56,7 +63,8 @@ class UploadCommand(Command):
         print('\033[1m{0}\033[0m'.format(s))
 
     def initialize_options(self):
-        pass
+        # Make release to test server by default.
+        self.final_release = None
 
     def finalize_options(self):
         pass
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -44,6 +45,55 @@ def long_description():
...
+            here = os.path.abspath(os.path.dirname(__file__))
+            rmtree(os.path.join(here, 'dist'))
+        except OSError:
+            pass
+
+        self.status('Building Source and Wheel (universal) distribution…')
+        os.system('{0} setup.py sdist bdist_wheel --universal'.format(
+                                                            sys.executable))
+
+        if self.final_release == 'True':
```
amanagr: I haven't tested if this works.

neiljp (commented):
I think automating this will really help with the release process :+1:

Initial thoughts:
* This generates the packages fine, and I saw the package on test.pypi.org that you pointed to, though haven't tested the upload myself (as mentioned on czo, it won't install due to dependency issues).
* A short piece of documentation might be good, for existing/new committers to refer to.
* How does this chunk differ from the zulint code? You've modified it for the test upload, right? Could you have one commit for the zulint code, and others for the subsequent changes to make this clearer?
* This depends on twine and wheel? Not sure if you want a separate `release_deps` in `setup.py`?
* Running this tagged this branch on my local copy as 0.3.1; do we want it to always tag?
(We likely don't want to tag a release on a test upload?)
* This does work, but a followup might be good with pathlib?
* Do we want `--universal`? We don't support python2, and while `python_version` clarifies this, it might be simpler to avoid this? Maybe even `--python-tag=py34.py35.py36.py37`? That isn't automatic from the setup.py tags unfortunately, from what I can tell.
* Any changes could be backported to zulint or elsewhere that has similar code in zulip, if relevant?

amanagr:
I have made most of the suggested changes.
I will document release process in https://github.com/zulip/zulip-terminal/wiki/Making-a-new-release-%F0%9F%8E%89.
I don't think we need to add twine and wheel as release deps in setup, I will include it in the docs since they are quite common known deps.
I am not sure if pathlib really make an impact here.
Yeah, we can move the code to zulint or other places. Thanks @neiljp for great recommendations!

timabbott:
As a random sidenote, I generally recommend committing developer documentation like a "release checklist" in `docs/` folder rather than using the GitHub wikis; that way content in it shows up in `git grep` and is more likely to be updated as changes happen.

neiljp:
@amanagr After discussion over the Summer, I'm wondering if this would be simpler to have as a separate script so that we don't have to include it in `setup.py`? For example, the `python-zulip-api` repo has a `release-packages` script - which admittedly is perhaps more complex than we need.

amanagr:
or leave it? I now think that we should manually verify each step of the release process. While this method is great, I think I will fear using it 😨 !

amanagr:
Closing since this likely needs a complete setup change to be useful now.

PR #362, opened 2019-05-24
model: Fix append message event for streams.

sumanthvrao:
Ensure that a new stream message gets appended to the current stream (or topic in current stream) only
if the recipient and current narrow matches. Tests have been added.


neiljp:
@sumanthvrao Thanks for this, this was a good find :+1:  I made a few small changes to the ids, a variable name and the summary, then merged as 42aed7973e46ea6c2f4697ac197928645dd224c4 since this is an important bugfix to have out there, I think!

The tests could be improved further, I think - for example, you added the stream(_to_stream) case, but there is not a test for the other failure you observed (same topic, different stream), which could perhaps be added.

sumanthvrao:
Thanks for the quick merge @neiljp!
The ids seem more descriptive now :+1: 
I will get working on a test for the other failure observed.

Issue #360, opened 2019-05-23
Drop support for py3.4 & migrate to py3.5+ code and any new features


sumanthvrao:
Py3.4 has reached End Of Life, upstream, making it reasonable for us to drop support for it.
This occurs as a result of third party packages (example: pygments, requests) having already stopped support for py3.4 in their latest release (on PyPI).  
There are various stages at which we could implement this transition : 
* [x] Removing it from CI
* [x] Changing the version requirements/suggestions
* [ ] Removing some simplifications that were also done due to python 3.4, and thinking around possible improvements using 3.5.

rht:
This issue can be closed.

neiljp:
@rht I think this was previously left open for the remaining element, namely removing python3.4 backwards-compatible entries. From memory, and checking the changelog, this is maybe mainly multiple-unpacking (https://www.python.org/dev/peps/pep-0448/), which I remember having to avoid when we had 3.4 included previously.

Do you know any tools that help with transitions like this?

rht:
> Do you know any tools that help with transitions like this?

I doubt there is a tool that caters to a specific PEP change.

neiljp:
@rht I was meaning for migrating between python versions, like for 2-to-3, but minor versions.

PR #363, opened 2019-05-25
requirements: Remove support for python 3.4.

neiljp:
Python 3.4 is now end-of-line, and dependencies (those below & requests)
are now expected to support 3.5+. This simplifies things as below, but
also may enable removal of old compatibility code and new features only
available in python 3.5+.

Removes:
* testing 3.4 in Travis CI
* labelled python version support in setup.py for 3.4
* pinning of pygments & tornado package versions in setup.py/Pipfile/requirements.txt
* need to explicitly install typing package in setup.py/Pipfile/requirements.txt

Also updates reference to 3.5 in README.md, and adds to CHANGELOG.md.

This is a first stage towards #360.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ cache:
     - $HOME/.cache/pipenv
 matrix:
   include:
-    - python: 3.4
-      name: "Python 3.4 pytest (default linux)"
     - python: 3.5
       name: "Python 3.5 pytest (default linux)"
     - python: 3.6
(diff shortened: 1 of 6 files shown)


sumanthvrao:
Looks great @neiljp!

sumanthvrao:
I don't mind getting this merged as-is, just so our tests on CI are passing. 
[ Another change to keep in mind (not so important) would be to update the [metadata badge](https://img.shields.io/pypi/pyversions/zulip-term.svg) , now that we are moving towards py3.5+ only.]

neiljp:
@sumanthvrao I believe the pypi support badges are automatic from pypi, so will change once we make a release.

PR #364, opened 2019-05-25
test_model: Add tests for append message event for streams.

sumanthvrao:
This commit adds tests regarding another issue addressed by #362,
in which,  messages with same topic names appear in (different)
streams which is not a recipient of the message.


sumanthvrao:
@neiljp This should wrap-up #326. Sorry for leaving this test out in the first place.

neiljp:
@sumanthvrao Thanks for the test update! I manually tested the new tests by reverting the previous bugfix commit - and both failed :) :+1: Merged with a minor reword.

sumanthvrao:
Good to know of the failure (in this case ) :) 
Thanks @neiljp!

PR #365, opened 2019-05-26
model: Fix append message event for PMs.

sumanthvrao:
This PR is a fix for the bug discussed here - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/group.20PMs.20in.20user.20narrow.2E/near/745808.


neiljp:
@sumanthvrao Thanks for this extra fix to `append_message` :+1: I remember broadening this code to support group PMs, but missed this fix at the time. Also, the `Model.recipients` refactor was on my to-do list from a while back, so it was good to see this being useful as well as a cleanup.

sumanthvrao:
Thanks for the merge @neiljp! The refactor has made the flow more readable :)
This PR has been merged as - 1de1697c6860c763d45442eeabc848d4bd8f582d and 2 before

PR #369, opened 2019-05-31
Switch to no-autohide view by default

amanagr:


sumanthvrao:
I think this change was long overdue ;). Looks good!

amanagr:
I will be making further onboarding improvements in different PRs, so this is ready for review.

README.md
```diff
@@ -65,7 +65,7 @@ site=https://realm.zulipchat.com
 [zterm]
 # Theme can also be set to 'gruvbox' ('blue' & 'light' are older themes needing work)
 theme=default
-# Autohide can also be set to 'no_autohide', to always show the left and right panels
+# Autohide can also be set to 'autohide', to hide the left and right panels by default
```
neiljp: I think this needs to be slightly clearer; perhaps "... to only show the left and right panels when required" ?
(or 'necessary', or 'focused'?)

amanagr: How about - If you want to view only messages, you can set `autohide` as the default setting. Then, left and right panels will only be visible when focused.

neiljp: @amanagr How about the following?
```
# Autohide defaults to 'no_autohide', but can be set to 'autohide' to hide the left & right panels except when focused
```

I think it would be good to get this merged ASAP for new devs or people running from master.

amanagr: Cool, done!

sumanthvrao: The comment text great! LGTM :)

PR #370, opened 2019-06-01
ui: Send request to change column focus form the `View`.

amanagr:
urwid doesn't move cursor to the newly focused column if View doesn't
change it focus to the focused column.

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -171,9 +171,11 @@ class TestView:
         view.keypress(size, 'down')
         super_view.assert_called_once_with(size, 'down')
 
-    def test_keypress_w(self, view, mocker):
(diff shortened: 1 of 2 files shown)


amanagr:
@neiljp I don't think there will be any changes to test cases since they `assert view.body.focus_col == 2/0` is already included in the test cases for the respective key presses. As we discussed (here)[https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/.23T370] this change is still needed since this is a bug for no_autohide view.

neiljp:
@amanagr As discussed elsewhere, your original fix is now merged, with more complete tests in a followup commit by myself that are more complete/reliable. Thanks for confirming what the problem was here, as it wasn't obvious at first! :+1:

PR #373, opened 2019-06-04
State support for PyPy, and test on pypy3.5

neiljp:
This is a low priority, I expect, but I noticed that pypy supports python3.5 (or 3.6) and also that urwid claims support for it too.

The challenge here is that the `Pipfile` has standard and dev sections, but we need an intermediate one for mypy, at least if we want to run tests with pypy - but not mypy, since `typed-ast` doesn't appear to have a build for pypy.

At this WIP stage I have commented out mypy in the Pipfile, so it would have to be manually installed by developers who use `pipenv`, like it is in Travis with this commit.

This appears to run the tests fine, but of course doesn't test actually running it.


neiljp:
I've tested running this with a variant of the Docker support in #464; 0.4.0 seems to work fine with pypy 3.5 and 3.6 on Debian stretch.

amanagr:
@neiljp great! It should significantly reduce Travis running time. LGTM!
(It would be better if we could find a better alternative than removing mypy from Pipfile.)

neiljp:
@amanagr I'm not sure that this will reduce Travis running time - it's *another* task to run. The main aspect is to claim/confirm/test/support an additional platform.

neiljp:
Merging this for now; we can adapt this if we receive feedback, or perhaps if the testing becomes too onerous.

PR #374, opened 2019-06-04
Some miscellaneous fixes/refactors.

sumanthvrao:
This PR adds a fixture for stream_dict;
It can potentially simplify common code as illustrated with the follow-up commit (
bb039d3)

The tests data [here](https://github.com/zulip/zulip-terminal/pull/315/files#diff-484462fced51d1a06b1d93b4a44dd535R903-R946) can be an extension to this.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -49,12 +49,9 @@ class WriteBox(urwid.Pile):
         ]
         self.focus_position = 1
 
-    def stream_box_view(self, button: Any=None, caption: str='',
-                        title: str='') -> None:
+    def stream_box_view(self, caption: str='', title: str='') -> None:
         self.set_editor_mode()
         self.to_write_box = None
-        if caption == '' and button is not None:
-            caption = button.caption
         self.msg_write_box = ReadlineEdit(multiline=True)
         self.stream_write_box = ReadlineEdit(
             caption=u"Stream:  ",
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -564,6 +564,8 @@ class MessageBox(urwid.Pile):
             pad_count += 1
 
     def selectable(self) -> bool:
+        # Returning True, indicates that this widget
+        # is designed to take focus.
         return True
 
     def mouse_event(self, size: Tuple[int, int], event: Any, button: Any,


zulipterminal/ui_tools/boxes.py
```diff
@@ -486,9 +486,6 @@ def main_view(self) -> List[Any]:
         ]
         return [part for part, condition in parts if condition]
 
-    def selectable(self) -> bool:
-        return True
-
```
neiljp: I'm pretty sure this is not 'dead', but rather necessary? It could benefit from a comment; I believe it allows the user to select a message with a click, or otherwise 'receive focus' in the list - right @amanagr ?

sumanthvrao: Woops! I think you are right here. http://urwid.org/reference/widget.html#urwid.Widget.selectable has information what we want. Perhaps a comment above it as you mentioned would help. 
Sorry for the mistake.

sumanthvrao:
> That said, the balance for test fixtures I see here is that really the fixtures are desperate to be determined algorithmically, balanced against all the duplication everywhere. Also, in cases like `stream_dict`, small cases can be sufficient, rather than needing to eg. refer to a full example. That said, there are quite a few tests with duplication, or _almost_ duplication, and it can get _very_ verbose!

The number of tests which duplicated the `stream_dict` code was 'large' enough that I though of factoring them out. As you said, we may want to reduce the fixture to just include the factored out values (stream id : 205 and 5) instead of going with the full fixture. Is that okay for now?

zulipterminal/ui_tools/boxes.py
neiljp: I think it would be reasonable to amend this commit to add a comment.

neiljp:
Regarding the later commits, one approach might be to examine what each test would benefit from - one stream_dict? the test running through various alternatives of stream_dict? one big stream_dict? It may not matter here, but it may be that having various fixtures which provide variants of these could be useful. We can discuss if you like.

sumanthvrao:
Sure. I like the idea of having different types of fixtures for `stream_dict` here.

zulipterminal/ui_tools/boxes.py
amanagr: Yeah, a comment would be nice there!

sumanthvrao: Done. Updated this with the commit :+1:

sumanthvrao:
I amended the last 2 commits here. So one introduces the new fixture, while the other replaces some common code with the new fixture.

sumanthvrao:
With the addition of #387 (further improvements along the way), Which was what I was going for with the last 2 comments here. I will update this to only the first commit. Will add further changes/refactors as and when I find them.

sumanthvrao:
The last commit here, would be a follow-up for #423.

zulipterminal/ui_tools/boxes.py
```diff
@@ -49,12 +49,9 @@ def private_box_view(self, button: Any=None, email: str='') -> None:
         ]
         self.focus_position = 1
 
-    def stream_box_view(self, button: Any=None, caption: str='',
-                        title: str='') -> None:
+    def stream_box_view(self, caption: str='', title: str='') -> None:
         self.set_editor_mode()
         self.to_write_box = None
-        if caption == '' and button is not None:
-            caption = button.caption
```
neiljp: The writebox still uses 'caption' at this time.

neiljp:
@sumanthvrao I just merged these small changes - small but trimming the code means less to read/learn :)

PR #375, opened 2019-06-04
Improve user experience of reading messages.

amanagr:
This changes following behaviors:
* Don't read the first unread messages when zt starts.
* Don't read all the message in `self.log` when narrowing.

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -254,15 +254,15 @@ class Controller:
 
     def _finalize_show(self, w_list: List[Any]) -> None:
         focus_position = self.model.get_focus_in_current_narrow()
-
         if focus_position == set():
             focus_position = len(w_list) - 1
         assert not isinstance(focus_position, set)
         self.model.msg_view.clear()
-        self.model.msg_view.extend(w_list)
-        self.editor_mode = False
         if focus_position >= 0 and focus_position < len(w_list):
-            self.model.msg_list.set_focus(focus_position)
+            self.model.msg_view.extend(w_list, focus_position)
+        else:
+            self.model.msg_view.extend(w_list)
+        self.editor_mode = False
 
     def deregister_client(self) -> None:
         queue_id = self.model.queue_id
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -19,6 +19,44 @@ from zulipterminal.ui_tools.utils import create_msg_box_list
 from zulipterminal.ui_tools.boxes import UserSearchBox, StreamSearchBox
 
 
+class ModListWalker(urwid.SimpleFocusListWalker):
+
+    def set_focus(self, position: int) -> None:
+        # When setting focus via set_focus method.
+        self.focus = position
+        self._modified()
+        if hasattr(self, 'read_message'):
+            self.read_message()
+
+    def _set_focus(self, index: int) -> None:
+        # This method is called when directly setting focus via
+        # self.focus = focus_position
+        if not self:
+            self._focus = 0
+            return
+        if index < 0 or index >= len(self):
+            raise IndexError('focus index is out of range: %s' % (index,))
+        if index != int(index):
+            raise IndexError('invalid focus index: %s' % (index,))
+        index = int(index)
+        if index != self._focus:
+            self._focus_changed(index)
+        self._focus = index
+        if hasattr(self, 'read_message'):
+            self.read_message()
+
+    def extend(self, items: List[Any],
+               focus_position: Optional[int]=None) -> int:
+        if focus_position is None:
+            focus = self._adjust_focus_on_contents_modified(
+                slice(len(self), len(self)), items)
+        else:
+            focus = focus_position
+        rval = super(urwid.MonitoredFocusList, self).extend(items)
+        self._set_focus(focus)
+        return rval
+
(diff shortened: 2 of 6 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -179,6 +179,15 @@ def update_search_box_narrow(self, message_view: Any) -> None:
 
     def read_message(self) -> None:
         # Message currently in focus
+        if hasattr(self.model.controller, "view"):
+            self.view = self.model.controller.view
+        else:
+            return
```
neiljp: Can this just use a local variable for view? I'm not sure why you're setting it on the object rather than just a local variable in the method.

Since `self.view` is not used later at all, it could also be written with `getattr` with a default, or with an exception block?

amanagr: I might have done it out of habit. Thanks for pointing out.

zulipterminal/ui_tools/views.py
```diff
@@ -177,12 +177,21 @@ def update_search_box_narrow(self, message_view: Any) -> None:
             )
         self.model.controller.update_screen()
 
-    def read_message(self) -> None:
+    def read_message(self, index: int=-1) -> None:
```
neiljp: This `index` seems unused?

sumanthvrao: Correct me if I am wrong, but this seems to be a default parameter passed with the callback which is being caught with 'index' variable?

amanagr: Yes Sumanth it is. @neiljp `index` will have to remain unused.

amanagr:
@neiljp `G` & `J` keys are working fine for me as well. I have added tests for the changes I made to the functions in SimpleFocusListWalker.

zulipterminal/ui_tools/views.py
neiljp: To clarify, this is an expected parameter for a callback method provided to `set_focus_changed_callback`?
(this seems not well documented?)

zulipterminal/ui_tools/views.py
```diff
@@ -187,10 +187,12 @@ def read_message(self, index: int=-1) -> None:
...
         self.update_search_box_narrow(msg_w.original_widget)
+        # If this the last message in the view and focus is set on this message
+        # then read the message.
+        last_message_focused = (curr_pos == len(self.log) - 1)
         # Only allow reading a message when middle column is
-        # in focus. Changing the contents of self.log also changes
-        # focus automatically.
-        if view.body.focus_col != 1:
+        # in focus.
+        if view.body.focus_col != 1 and not last_message_focused:
```
neiljp: If I understand what you're trying to do here (and by de Morgan's), would this be clearer as:
```python
not(focus_col == 1 or last_message_focused)
```
If so, are you trying to enable the last message to be read even if the (column) focus is not on the message column?
In what circumstances does this happen?

zulipterminal/core.py
```diff
@@ -254,15 +254,15 @@ def show_all_starred(self, button: Any) -> None:
...
         assert not isinstance(focus_position, set)
         self.model.msg_view.clear()
-        self.model.msg_view.extend(w_list)
-        self.editor_mode = False
         if focus_position >= 0 and focus_position < len(w_list):
-            self.model.msg_list.set_focus(focus_position)
+            self.model.msg_view.extend(w_list, focus_position)
+        else:
+            self.model.msg_view.extend(w_list)
+        self.editor_mode = False
```
neiljp: You've moved the code around a little here; I thought this commit would be focused on a substitution of the different ListWalker, so is this a separate change?

zulipterminal/ui_tools/views.py
amanagr: yes.

zulipterminal/ui_tools/views.py
sumanthvrao: I too had the same doubt here regarding when the second commit here would do?

sumanthvrao:
Just to confirm, one  change this Pr brings is that if I have around 20 unread messages in a stream and `Enter` it, The focus now jumps to the first unread message (and reads it), the unread count would be reduced by 1.

The earlier approach we had was, we would mark all messages currenly visible as read and decrease the unread count accordingly. The second commit of this PR title sort off hints at this, so I was confused if the old behaviour was still expected here?

zulipterminal/ui_tools/views.py
amanagr: It happens when the user has only one unread message and open's zt. 
This will be the behavior without this change:
Screenshot (1)
Screenshot (2)

amanagr:
Yes, the old behavior was not intentional. We don't want to read too many messages at once.

zulipterminal/ui_tools/views.py
sumanthvrao: Ah thanks for the clarification @amanagr. it would help to have that line in commit description, if possible.

neiljp:
@amanagr Thanks for your work on this; I've merged the very slightly amended branch I showed you :tada:

I think this is a great improvement and changes the "expected from the webapp but not in this way" behavior to a simpler one, which is admittedly slightly different from the webapp but acts in a much more controlled way :) I'm interested to hear feedback from users!

PR #378, opened 2019-06-05
themes/buttons: Set background color as default provided by terminal.

### LINKED ISSUE #377: Support terminal background-transparency
This was raised by @aero31aero on chat.zulip.org on #zulip-terminal.

The solution may involve setting the background to `''`, or at least this worked for them at the time.

neiljp:
This should also enable support for transparent-background terminals, so
fixes #377.

For me this currently enables a default deeper black background; I don't have a transparent-background terminal, but perhaps @aero31aero could confirm?

This achieves the same result as the first commit of #310, but likely doesn't solve any other color issues.

The only issue I can see is if the default background might be set as white, in which case we should make this a configurable setting, eg. only enable via a `--transparent` flag or a config option.


sumanthvrao:
Can confirm this. ZT background changes to the same transparency, when theme is changed to 'default'.

aero31aero:
Can confirm this PR works perfectly for tmux/zsh with guake as the emulator. This is really pleasing to look at.

amanagr:
LGTM.

neiljp:
@sumanthvrao @aero31aero @amanagr Thanks for the reviews!

PR #379, opened 2019-06-06
Introduce new `message` fixture.

amanagr:
@neiljp this is just an example usage, we can use it in other places too.


tests/conftest.py
```diff
@@ -83,110 +83,130 @@ def logged_on_user():
...
-@pytest.fixture
+
+@pytest.fixture(params=[
+    stream_msg_template, pm_template, group_pm_template
+])
+def message(request):
+    """
+    Acts as a parametrize fixture for stream msg, pms and group_pms.
+    """
+    return deepcopy(request.param)
```
neiljp: My suggestion would be to actually leave this first commit as a refactoring and move adding the `message` fixture to the separate (next) commit, where it can be used. So you then have 'refactor' and 'add fixture and use it' commits.

tests/ui/test_ui_tools.py
```diff
@@ -1206,38 +1206,6 @@ def test_main_view_content_header_without_header(self, mocker, message,
...
-                'full_name': 'Iago'
-            }],
-            'flags': [],
-            'content': '<p>what are you planning to do this week</p>',
-            'reactions': [],
-            'sender_full_name': 'Alice',
-            'timestamp': 1532103879,
-        }
-    ])
     @pytest.mark.parametrize('to_vary_in_each_message', [
```
neiljp: This looks rather strange, since message was a parameter and is now a fixture, but no code is added here.

Not just for this reason, but I'm wondering if a slightly more verbose name for `message` might be good, though these things are difficult to name well.

amanagr:
I haven't noticed any speed change and since the fixtures are quite small in size I don't think it should. @sumanthvrao we don't have to explicitly work on it, just try to use this in whichever area you are working or create new ones.

tests/conftest.py
amanagr: Updated

tests/ui/test_ui_tools.py
amanagr: replaced with `message_fixture`.

sumanthvrao:
I just went through this PR and find the fixture helpful. This will definitely  help us make some verbose paramterize parameter simpler (as you have done in one of the commits). I think we can go ahead and merge this as the changes are pretty useful.

@amanagr I got your point here. Maybe as and when I encounter some tests which can benefit from a similar refactor you have done here, I will end-up submitting a PR for them.

amanagr:
commit 711c2af4cd3fccf3d80d4f6480ec501f812598e0 (HEAD -> message_fixture, origin/message_fixture)
Author: Aman Agrawal <f2016561@pilani.bits-pilani.ac.in>
Date:   Thu Jun 6 21:20:02 2019 +0530

    tests: Use `message_fixture` to parametrize messasges.

    The `message_fixture` can be used to test over all types of messages.

commit 8cce69c7b1f2633e7ce0dc73fb9f7c2b9bcf1ca2
Author: Aman Agrawal <f2016561@pilani.bits-pilani.ac.in>
Date:   Sat Jun 8 09:18:11 2019 +0530

    conftest: Create message fixture for all types of messages.

    This fixture can be used in any function to test all the types
    of messages -- stream messages/pms/group pms.

commit d873ca12d724f893282e07d96cbc0f4627ea549d
Author: Aman Agrawal <f2016561@pilani.bits-pilani.ac.in>
Date:   Thu Jun 6 23:45:26 2019 +0530

    conftest: Refactor `empty_index` to use template messages.

commit ce5a88607ec5b0ca8a3c63a3a5ca274bb63722ec
Author: Aman Agrawal <f2016561@pilani.bits-pilani.ac.in>
Date:   Thu Jun 6 19:50:25 2019 +0530

    conftest: Refactor `message_successful_response`.

    Extract the 3 types of msgs out of message_successful_response.
    Deepcopy is used to avoid modification of the original fixture when the
    test function is parametrized


This is the real commit order. Github is showing them differently.

@sumanthvrao cool!

neiljp:
@amanagr Merged with a few final adjustments as agreed as 82afcc605bb0d7169160eef2a6b9e2da0bd123bc and the previous 3 commits.

neiljp:
I look forward to this opening new test improvements and refactoring :+1:

PR #381, opened 2019-06-09
model: Report connection error at startup

vonknio:
When I launched zulip-term with no internet connection, I got a general crash message even though there's a special message defined for `ServerConnectionFailure` (`run:239-245`). The exception was thrown from `model/_update_initial_data`, where I think it was meant to be captured and reported as `ServerConnectionFailure`, and that's the exact change I've made.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -322,13 +322,21 @@ class Model:
                                                 anchor=None),
                 'register': executor.submit(self._register_desired_events,
                                             fetch_data=True),
-            }  # Dict[str, Future[Any]]
+            }  # Dict[str, Future[bool]]
 
             # Wait for threads to complete
             wait(futures.values())  # type: ignore
 
-        results = {name: future.result()  # type: ignore
-                   for name, future in futures.items()}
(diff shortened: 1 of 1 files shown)


vonknio:
@neiljp Not sure I understood you well, are you suggesting retuning a boolean instead of raising `ServerConnectionFailure` in `_update_initial_data`?

> return value (and so type) could be more specific based on the values returned by the threaded methods. I expect other typing improvements are then possible too.

Requests seem to return booleans, so I've adjusted the types correspondingly.

neiljp:
@vonknio The changes you made were exactly as I hoped would be possible; I removed the last `# type: ignore` and made the commit text a little more detailed, with the result merged as ba0e4f994a878ea9edfaf9454b24b6ce8bb52844

For the record, the main issue I recall noting was returning `None` in the nested function; I think this is fine, but feel free to chat with me about this further if you were confused by my message.

neiljp:
Thanks @vonkio! Not only is this tidier with the improved type hints, but of course the bug fix is great and behaves more as we expected :tada:

PR #382, opened 2019-06-09
Display 'topics' in a stream

sumanthvrao:
### **List of aspects this PR covers**:
* View topics in a stream by toggling `t` on a stream button.
* Scroll through topic list with `up/down` or `j/k`.
* Clicking `Enter` on any topic button, brings you to the topic narrow.
* Navigating between Open topic view and messages can be done with `left` and `right`. This restores the focus.
* Scroll through the topics with mouse.
* Unread_counts are shown for topic buttons.
* When a new message arrives, we update the UI to show the re-arranged order.
* Search for a topic in topic_list with `q` (Same as stream hostkey).
* Muted topics are marked so. Their messages are not added to the message view.
* Topic names are extended to the corners to provide as much space as possible.
* Toggle `t` on open topic View, closes it and brings you back to the streambutton you clicked on.

### Follow-ups
* Handle topic edits.
* Event based approach for muting topics.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -406,7 +406,7 @@ class TestStreamsView:
         mocker.patch(VIEWS + ".urwid.connect_signal")
         mocker.patch(VIEWS + ".threading.Lock")
         self.view = mocker.Mock()
-        self.search_box = mocker.patch(VIEWS + ".PanelSearchBox")
+        self.stream_search_box = mocker.patch(VIEWS + ".PanelSearchBox")
         stream_btn = mocker.Mock()
         stream_btn.stream_name = "FOO"
         self.streams_btn_list = [stream_btn]
@@ -416,8 +416,9 @@ class TestStreamsView:
         assert stream_view.view == self.view
         assert stream_view.log == []
         assert stream_view.streams_btn_list == self.streams_btn_list
-        assert stream_view.search_box
-        self.search_box.assert_called_once_with(stream_view, 'SEARCH_STREAMS')
+        assert stream_view.stream_search_box
+        self.stream_search_box.assert_called_once_with(
+            stream_view, 'SEARCH_STREAMS')
 
     @pytest.mark.parametrize('new_text, expected_log', [
         ('f', ['FOO', 'foo', 'fan']),
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -406,7 +406,7 @@ class TestStreamsView:
         mocker.patch(VIEWS + ".urwid.connect_signal")
         mocker.patch(VIEWS + ".threading.Lock")
         self.view = mocker.Mock()
-        self.search_box = mocker.patch(VIEWS + ".StreamSearchBox")
+        self.search_box = mocker.patch(VIEWS + ".PanelSearchBox")
         stream_btn = mocker.Mock()
         stream_btn.stream_name = "FOO"
         self.streams_btn_list = [stream_btn]
@@ -417,7 +417,7 @@ class TestStreamsView:
         assert stream_view.log == []
         assert stream_view.streams_btn_list == self.streams_btn_list
         assert stream_view.search_box
-        self.search_box.assert_called_once_with(stream_view)
+        self.search_box.assert_called_once_with(stream_view, 'SEARCH_STREAMS')
 
     @pytest.mark.parametrize('new_text, expected_log', [
         ('f', ['FOO', 'foo', 'fan']),
@@ -789,7 +789,7 @@ class TestRightColumnView:
     @pytest.fixture(autouse=True)
     def mock_external_classes(self, mocker):
         self.view = mocker.Mock()
-        self.user_search = mocker.patch(VIEWS + ".UserSearchBox")
+        self.user_search = mocker.patch(VIEWS + ".PanelSearchBox")
         self.connect_signal = mocker.patch(VIEWS + ".urwid.connect_signal")
         self.line_box = mocker.patch(VIEWS + ".urwid.LineBox")
         self.thread = mocker.patch(VIEWS + ".threading")
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -474,6 +474,9 @@ class TestTopicsView:
         self.log = mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker",
                                 return_value=[])
         self.stream_button = stream_button
+        mocker.patch(VIEWS + ".urwid.connect_signal")
+        mocker.patch(VIEWS + ".threading.Lock")
+        self.topic_search_box = mocker.patch(VIEWS + ".PanelSearchBox")
         self.view = mocker.Mock()
         self.view.controller = mocker.Mock()
         topic_btn = mocker.Mock()
@@ -485,6 +488,26 @@ class TestTopicsView:
         assert topic_view.log == []  # topic_view patches this
         assert topic_view.stream_button == self.stream_button
         assert topic_view.view == self.view
+        assert topic_view.topic_search_box
+        self.topic_search_box.assert_called_once_with(
+            topic_view, 'SEARCH_TOPICS')
+
+    @pytest.mark.parametrize('new_text, expected_log', [
+        ('f', ['FOO', 'foo', 'fan']),
+        ('foo', ['FOO', 'foo']),
+        ('FOO', ['FOO', 'foo']),
+    ])
+    def test_update_topics(self, mocker, topic_view, new_text, expected_log):
+        self.view.controller.editor_mode = True
+        new_text = new_text
+        search_box = "SEARCH_BOX"
+        topic_view.topics_btn_list = [
+            mocker.Mock(topic_name=topic_name) for topic_name in [
+                'FOO', 'foo', 'fan', 'boo', 'BOO']]
+        topic_view.update_topics(search_box, new_text)
+        assert [topic.topic_name for topic in topic_view.log
+                ] == expected_log
+        self.view.controller.update_screen.assert_called_once_with()
 
     @pytest.mark.parametrize('topic_name, topic_initial_log,\
                               topic_final_log', [
@@ -517,6 +540,7 @@ class TestTopicsView:
     def test_keypress_EXIT_TOGGLE_TOPIC(self, mocker, topic_view):
         key = "t"
         size = (200, 20)
+        mocker.patch(VIEWS + '.urwid.Frame.keypress')
         topic_view.view.left_panel = mocker.Mock()
         topic_view.view.left_panel.contents = [mocker.Mock(), mocker.Mock()]
         topic_view.keypress(size, key)
@@ -526,6 +550,7 @@ class TestTopicsView:
     def test_keypress_GO_RIGHT(self, mocker, topic_view):
         key = "l"
         size = (200, 20)
+        mocker.patch(VIEWS + '.urwid.Frame.keypress')
         topic_view.view.body.focus_col = None
 
         topic_view.keypress(size, key)
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -482,6 +482,8 @@ class TestTopicsView:
         topic_btn = mocker.Mock()
         topic_btn.caption = "BOO"
         self.topics_btn_list = [topic_btn]
+        self.header_list = mocker.patch(VIEWS + ".urwid.Pile")
+        self.divider = mocker.patch(VIEWS + ".urwid.Divider")
         return TopicsView(self.topics_btn_list, self.view, self.stream_button)
 
     def test_init(self, mocker, topic_view):
@@ -491,6 +493,9 @@ class TestTopicsView:
         assert topic_view.topic_search_box
         self.topic_search_box.assert_called_once_with(
             topic_view, 'SEARCH_TOPICS')
+        self.header_list.assert_called_once_with([topic_view.stream_button,
+                                                  self.divider('─'),
+                                                  topic_view.topic_search_box])
 
     @pytest.mark.parametrize('new_text, expected_log', [
         ('f', ['FOO', 'foo', 'fan']),
@@ -561,7 +566,7 @@ class TestTopicsView:
         size = (200, 20)
         mocker.patch(VIEWS + '.TopicsView.set_focus')
         topic_view.keypress(size, key)
-        topic_view.set_focus.assert_called_once_with('header')
+        topic_view.header_list.set_focus.assert_called_once_with(2)
 
     def test_keypress_esc(self, mocker, topic_view):
         key = "esc"
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -552,11 +552,24 @@ class TestTopicsView:
         size = (200, 20)
         mocker.patch(VIEWS + '.urwid.Frame.keypress')
         topic_view.view.body.focus_col = None
-
         topic_view.keypress(size, key)
-
         assert topic_view.view.body.focus_col == 1
         topic_view.view.show_left_panel.assert_called_once_with(visible=False)
+
+    def test_keypress_SEARCH_TOPICS(self, mocker, topic_view):
+        key = 'q'
+        size = (200, 20)
+        mocker.patch(VIEWS + '.TopicsView.set_focus')
+        topic_view.keypress(size, key)
+        topic_view.set_focus.assert_called_once_with('header')
+
+    def test_keypress_esc(self, mocker, topic_view):
+        key = "esc"
+        size = (200, 20)
+        mocker.patch(VIEWS + '.TopicsView.set_focus')
+        topic_view.keypress(size, key)
+        topic_view.set_focus.assert_called_once_with("body")
+        assert topic_view.log == self.topics_btn_list
 
 
 class TestUsersView:
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -183,9 +183,9 @@ class View(urwid.WidgetWrap):
             self.left_panel.keypress(size, 'q')
             self.show_right_panel(visible=False)
             self.show_left_panel(visible=True)
-            self.stream_w.search_box.set_edit_text("")
+            self.stream_w.stream_search_box.set_edit_text("")
             self.controller.editor_mode = True
-            self.controller.editor = self.stream_w.search_box
+            self.controller.editor = self.stream_w.stream_search_box
             return key
         elif is_command_key('HELP', key):
             # Show help menu
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -177,15 +177,20 @@ class View(urwid.WidgetWrap):
             self.controller.editor_mode = True
             self.controller.editor = self.user_search
             return key
-        elif is_command_key('SEARCH_STREAMS', key):
+        elif (is_command_key('SEARCH_STREAMS', key) or
+                is_command_key('SEARCH_TOPICS', key)):
             # jump stream search
             self.body.focus_position = 0
             self.left_panel.keypress(size, 'q')
             self.show_right_panel(visible=False)
             self.show_left_panel(visible=True)
-            self.stream_w.stream_search_box.set_edit_text("")
+            if self.left_panel.is_in_topic_view:
+                search_box = self.topic_w.topic_search_box
+            else:
+                search_box = self.stream_w.stream_search_box
+            search_box.set_edit_text("")
             self.controller.editor_mode = True
-            self.controller.editor = self.stream_w.stream_search_box
+            self.controller.editor = search_box
             return key
         elif is_command_key('HELP', key):
             # Show help menu
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -767,54 +767,28 @@ class SearchBox(urwid.Pile):
         return key
 
 
-class UserSearchBox(urwid.Edit):
+class PanelSearchBox(urwid.Edit):
     """
-    Search Box to search users in real-time.
+    Search Box to search streams, topics or users in real-time.
     """
-
-    search_text = ("Search [" +
-                   ", ".join(keys_for_command("SEARCH_PEOPLE")) +
-                   "]: ")
-
-    def __init__(self, user_view: Any) -> None:
-        self.user_view = user_view
-        super(UserSearchBox, self).__init__(edit_text=self.search_text)
+    def __init__(self, panel_view: Any,
+                 search_command: str='SEARCH_STREAMS') -> None:
+        self.panel_view = panel_view
+        self.search_command = search_command
+        self.search_text = ("Search [" +
+                            ", ".join(keys_for_command(search_command)) +
+                            "]: ")
+        super(PanelSearchBox, self).__init__(edit_text=self.search_text)
 
     def keypress(self, size: Tuple[int, int], key: str) -> str:
         if is_command_key('ENTER', key):
-            self.user_view.view.controller.editor_mode = False
-            self.user_view.set_focus("body")
-        if is_command_key('GO_BACK', key):
-            self.user_view.view.controller.editor_mode = False
+            self.panel_view.view.controller.editor_mode = False
+            self.panel_view.set_focus("body")
+            if hasattr(self.panel_view, 'log') and len(self.panel_view.log):
+                self.panel_view.body.set_focus(0)
+        elif is_command_key('GO_BACK', key):
+            self.panel_view.view.controller.editor_mode = False
             self.set_edit_text(self.search_text)
-            self.user_view.set_focus("body")
-            self.user_view.keypress(size, 'esc')
-
-        return super(UserSearchBox, self).keypress(size, key)
-
-
-class StreamSearchBox(urwid.Edit):
-    """
-    Search Box to search streams in real-time.urwid
-    """
-
-    search_text = ("Search [" +
-                   ", ".join(keys_for_command("SEARCH_STREAMS")) +
-                   "]: ")
-
-    def __init__(self, stream_view: Any) -> None:
-        self.stream_view = stream_view
-        super(StreamSearchBox, self).__init__(edit_text=self.search_text)
-
-    def keypress(self, size: Tuple[int, int], key: str) -> str:
-        if is_command_key('ENTER', key) and len(self.stream_view.log):
-            self.stream_view.view.controller.editor_mode = False
-            self.stream_view.set_focus("body")
-            self.stream_view.body.set_focus(0)
-        if is_command_key('GO_BACK', key):
-            self.stream_view.view.controller.editor_mode = False
-            self.set_edit_text(self.search_text)
-            self.stream_view.set_focus("body")
-            self.stream_view.keypress(size, 'esc')
-
-        return super(StreamSearchBox, self).keypress(size, key)
+            self.panel_view.set_focus("body")
+            self.panel_view.keypress(size, 'esc')
+        return super(PanelSearchBox, self).keypress(size, key)
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -255,10 +255,11 @@ class StreamsView(urwid.Frame):
         self.log = urwid.SimpleFocusListWalker(streams_btn_list)
         self.streams_btn_list = streams_btn_list
         list_box = urwid.ListBox(self.log)
-        self.search_box = PanelSearchBox(self, 'SEARCH_STREAMS')
-        urwid.connect_signal(self.search_box, 'change', self.update_streams)
+        self.stream_search_box = PanelSearchBox(self, 'SEARCH_STREAMS')
+        urwid.connect_signal(self.stream_search_box,
+                             'change', self.update_streams)
         super(StreamsView, self).__init__(list_box, header=urwid.LineBox(
-            self.search_box, tlcorner=u'─', tline=u'', lline=u'',
+            self.stream_search_box, tlcorner=u'─', tline=u'', lline=u'',
             trcorner=u'─', blcorner=u'─', rline=u'',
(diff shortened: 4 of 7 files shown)


zulipterminal/model.py
```diff
@@ -311,6 +313,19 @@ def get_messages(self, *,
...
+        """
+        Fetch all topics of a stream (Version 1)
+        """
+        # FIXME: Version 2: Fetch only last 'n' topics
+        #        for each stream.
+        for stream_id in self.stream_dict.keys():
+            response = self.client.get_stream_topics(stream_id)
+            if response['result'] == 'success':
+                self.index['topics'][stream_id] = [topic['name'] for \
+                      topic in response['topics']]
```
neiljp: The startup time for this is definitely too lengthy. With a carefully chosen method you should be able to run this in parallel with the other initial data-fetch methods in `_update_initial_data` (whose name could now be more appropriate). You might even split the list of stream_ids to work on them concurrently, as long as you can fetch the data back together safely.

shreyamalviya:
Just a suggestion: instead of having `*` before each topic, it could show the number of unreads.
Something like:
```
(99) Topic 1
(7) Topic 2
(0) Topic 3
(0) Topic 4
```

Also, I think it would look better if the topics were shown next to the streams list, and not over it?

amanagr:
@sumanthvrao great work! Few thoughts:
* Change the color of the stream button for which topics are displayed (i.e. highlight the stream)
* Add support for `j` and `k` hotkeys.

sumanthvrao:
@neiljp @amanagr @shreyamalviya Here is a new version of the topics PR. I thank you guys for the quick feedback based on the last draft; Here are some changes/improvements in the current version:

* I have removed the overlay widget, and now you can switch to the main message column and return back to the topic column smoothly;
* The startup time has improved from 15 seconds on loading `czo` to 5 seconds for me. This is due to a threading model with 4 workers here. I have commented out the part where the code waits for the 4 threads to complete (Just to gain 2 seconds);
* Using `t` toggles the view now.
* The `*` has been removed at the start of the topic.
* 'z' , 'S' and 'n' work without any hassle.
* The stream whose topic is currently open is highlighted in 'yellow/orange' color.
* `j` and `k` keys work in Topic View

I have updated the original PR description with the next step of 'TODO' stuff. Tests have not yet been fixed/added hence this is still in the 'Draft mode'.

sumanthvrao:
Just updated this with another iteration, Here are the changes from previous versions:
* It now re-orders topics based on 'most recent' messages.
* Shows muted topics by marking them with 'M'.
* Handles updating topic-wise unread count when topic view is toggled.

Would love to hear feedback about this. I also noticed a small issue with the 're-ordering' topics logic and will update that soon.

sumanthvrao:
*update*:
* New commit added which clarifies the logic for updating 'topics' in index and toggled topic view.

@neiljp Do you have some time to give a quick feedback regarding this?

shreyamalviya:
This feels much better, now!

Just one thing that I think may be troublesome for some is - if I open the topics list for some stream and go back to navigating the streams with the topics still open, in order for me to go back to navigating the topics, I have to go all the way down to the end of the streams' list, and then into the topics' list.

Perhaps `tab` could enable the user to switch between streams and topics?

sumanthvrao:
Thanks for feedback @shreyamalviya; Thanks for pointing this out. This was definitely something I wanted to bring up on #zulip-term.

zulipterminal/model.py
```diff
@@ -316,6 +318,22 @@ def get_topics_in_stream(self, stream_list: List[int]) -> None:
...
+        Distribute stream ids across threads in order to fetch
+        topics concurrently.
+        """
+        executor = ThreadPoolExecutor(max_workers=workers)
+        list_of_streams = list(self.stream_dict.keys())
+        thread_objects = []
+        for i in range(workers):
+            split_list = list_of_streams[i::workers]
+            thread_objects.append(executor.submit(
+                self.get_topics_in_stream, split_list))
```
neiljp: Would a list comprehension be clearer?

zulipterminal/model.py
```diff
@@ -316,6 +318,22 @@ def get_topics_in_stream(self, stream_list: List[int]) -> None:
...
+        # FIXME : Should we wait here for threads to
+        # complete?
+        # wait(thread_objects)
```
neiljp: In principle we may not need this: the code seems to work fine as it is; I've not reviewed enough yet to see whether this is since the UI spends time loading while this is happening (ie. we're lucky?) or that it really is unnecessary. I suppose it depends how quickly the UI might grab a topics list, relative to how fast the data is loading, and if that's an issue. Not waiting allows the code to carry on, which makes it appear faster.

If we use a contextmanager form of the ThreadPoolExecutor, then waiting happens automatically when it tidies up, but I think it would be OK to include an explicit wait if we want one. While we could add it later, this will also ensure that if we move to longer-lasting centralized threads for this kind of thing, then we still wait *here* (if intended).

zulipterminal/model.py
```diff
@@ -305,6 +305,17 @@ def get_messages(self, *,
             return True
         return False
 
+    def get_topics_in_stream(self, stream_list: List[int]) -> None:
+        """
+        Fetch all topics of a stream (Version 1)
```
neiljp: `...all topics with specified stream_id's and indexes their names` ?

zulipterminal/model.py
```diff
@@ -305,6 +305,17 @@ def get_messages(self, *,
             return True
         return False
 
+    def get_topics_in_stream(self, stream_list: List[int]) -> None:
```
neiljp: I think this could just be an `Iterable[int]` ?

zulipterminal/ui_tools/buttons.py
```diff
@@ -19,10 +19,12 @@ def __init__(self, controller: Any, caption: str,
                  show_function: Callable[..., Any], width: int,
                  prefix_character: Union[str, Tuple[Any, str]]='\N{BULLET}',
                  text_color: Optional[str]=None,
-                 count: int=0) -> None:
+                 count: int=0, stream: str='', topic: str='') -> None:
         self.caption = caption
         self.prefix_character = prefix_character
         self.count = count
+        self.title = topic
+        self.stream = stream
```
neiljp: It seems like these two could just go into the `TopicButton` class?

neiljp:
Another testing note: when writing a new message in a new topic, I expect it to appear in the topics list if it's the appropriate one, and possibly to append it to the topic - it does this, but doesn't scroll to it, a bit like when a new message arrives, we do scroll? This is possibly something we should break into a separate issue and discuss further, including how it should apply to messages.

tests/model/test_model.py
```diff
@@ -224,6 +224,21 @@ def test_get_message_ids_in_current_narrow(self, mocker, model,
         model.index = index
         assert current_ids == model.get_message_ids_in_current_narrow()
 
+    @pytest.mark.parametrize("response, stream_list, expected_index", [
+        ({'result': 'success', 'topics': [{'name': 'Foo'}, {'name': 'Boo'}]},
+         [23], {23: ['Foo', 'Boo']}),
+        ({'result': 'success', 'topics': []},
+         [23], {23: []})
```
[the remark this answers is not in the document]
sumanthvrao: Added a failure case. Let's handle it regularly for this. I assume you are talking about to 'pytest.xfail'?

zulipterminal/model.py
sumanthvrao: Will do so. But I am curious to know why you are asking me to do the change to `Iterable` from `List`?

zulipterminal/model.py
```diff
@@ -316,6 +318,22 @@ def get_topics_in_stream(self, stream_list: List[int]) -> None:
                 self.index['topics'][stream_id] = [topic['name'] for
                                                    topic in response['topics']]
 
+    def wrapper_get_topics(self, workers: int) -> None:
```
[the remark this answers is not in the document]
sumanthvrao: Done :+1:

zulipterminal/model.py
```diff
@@ -316,6 +318,22 @@ def get_topics_in_stream(self, stream_list: List[int]) -> None:
...
+        """
+        Distribute stream ids across threads in order to fetch
+        topics concurrently.
+        """
+        executor = ThreadPoolExecutor(max_workers=workers)
```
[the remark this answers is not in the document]
sumanthvrao: Agreed!

zulipterminal/model.py
```diff
@@ -316,6 +318,22 @@ def get_topics_in_stream(self, stream_list: List[int]) -> None:
...
+        list_of_streams = list(self.stream_dict.keys())
+        thread_objects = []
+        for i in range(workers):
+            split_list = list_of_streams[i::workers]
```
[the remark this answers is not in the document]
sumanthvrao: Yeah! I find this pythonic :)

zulipterminal/model.py
sumanthvrao: Yeah! I don't see why not.

zulipterminal/model.py
sumanthvrao: From what I observed, The topic fetch is pretty quick. However if we try to press `t` on a stream button as soon as the UI is loaded (1-2 seconds) no topics can be seen in the TopicView. In any case, I am now pushing a slight different version which handles thread safety also, albeit, its slightly slower (~6 seconds). This was partly the reason I didnt go with the context manager form initislly to shave around (2 seconds) of startup time.

zulipterminal/ui_tools/buttons.py
sumanthvrao: Ah, yes. I forgot to update this. Have changed it in the new version.

sumanthvrao:
Thanks for taking a look at this @neiljp! 
> There are other things we could include in the UI, such as making it clearer which stream we're looking at topics from, maybe including color, and there still seems to be an extra indent in the topics (on both sides?)

The current UI design completely replaces the stream view with the topic view. I had previous kept a color mapping indicating the stream whose topic view is toggled, but with this new design I am not sure how we will do that. Will remove the extra indent. Should not be too much of a problem.

> My only concern with the parallel fetch is that everything is modifying the index in parallel; that may be fine, but to guarantee that, we could pass the data back using the futures and merge it together.

Yeah! I haven't done this yet, will look into it. Maybe we can use locks for updating the index (Though I am not sure how it will impact speed).

> Another testing note: when writing a new message in a new topic, I expect it to appear in the topics list if it's the appropriate one, and possibly to append it to the topic

I noticed this quite recently when trying out the PR on `czo`. I did not spot this issue when using the dev server as the number of topics are quite small there. I remember reading some parts of the code which does this modification for Message List; will have a look at it. 

Merging the first few commits seems like a good idea. Will reduce this PR from exploding exponentially :)

zulipterminal/model.py
neiljp: From what I remember of the method, it was just looking at the elements inside of it; it could have been a tuple (so still a sequence), but the order itself is also not important, which probably makes iterable the more general form.

If the method ends up requiring a list, then we can put it as such (for the sake of mypy & docs), but until it does, this frees up the passing-in code to use whatever it likes.

tests/model/test_model.py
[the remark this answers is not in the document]
neiljp: Just that what happens if this code fails? Is there code to handle it? Are there tests to examine that?

We may want to handle this more generally, both with regard to code & tests, but for now we can get this working on a case by case basis.

neiljp:
Another note quickly while I remember, though I've not checked for this: are topic edits handled?

sumanthvrao:
No, topic edits are yet to be added (Amended the PR Description, thanks!)

sumanthvrao:
@neiljp topics now extend to the far left and right ends of the topic view; It crops a little to the right only when there are unread counts to display/ or 'M' to be shown, otherwise it extends to the full length (with `..`)

The traceback which you reported in `update_topic_index` has also been taken care of :+1:

zulipterminal/model.py
```diff
@@ -327,6 +329,29 @@ def exception_safe_result(self, future: 'Future[bool]') -> bool:
...
             return False
 
+    def fetch_all_topics(self, workers: int) -> None:
+        """
+        Distribute stream ids across threads in order to fetch
+        topics concurrently.
+        """
+        with ThreadPoolExecutor(max_workers=workers) as executor:
+            list_of_streams = list(self.stream_dict.keys())
+            thread_objects = {}  # type: Dict[int, Future[bool]]
```
neiljp: This statement isn't necessary any more, but you can move the type hint to the end of the next statement instead.

zulipterminal/model.py
neiljp: I see you've updated this; let me know if you want to discuss this further.

zulipterminal/model.py
```diff
@@ -305,6 +306,21 @@ def get_messages(self, *,
             return True
         return False
 
+    def get_topics_in_stream(self, stream_list: Iterable[int]) -> bool:
```
neiljp: The commit that adds this method states it fetches topics at startup, but that commit doesn't fetch them all at startup - that's in another commit.

zulipterminal/model.py
neiljp: Startup takes ~8s for me (timing manually) with the current PR.
This compares to ~4-6s with master.

Note that the switch to the contextmanager form of the `ThreadPool` (as in the updated form now) means that it waits as part of the tidying up process; I tested switching that back and it seems to take 1-2s less time. The `wait` *seems* to make not much difference to this?

zulipterminal/model.py
```diff
@@ -327,6 +329,29 @@ def exception_safe_result(self, future: 'Future[bool]') -> bool:
...
+        # ~ 4 seconds as has to wait for threads to complete.
+            # wait(thread_objects.values())
+        results = {
+            str(name): self.exception_safe_result(thread_object)
+            for name, thread_object in thread_objects.items()
+        }  # type: Dict[str, bool]
+        if not all(results.values()):
+            failures = ['Thread'+str(name) for name,
+                        result in results.items() if not result]
+            raise ServerConnectionFailure(", ".join(failures))
```
neiljp: I'm not sure we really mind which thread failed here?
(adding this as a single comment, since for some reason GH generated multiple reviews already)

zulipterminal/model.py
sumanthvrao: Agreed. I'll amend the commit title to something like 
`Fetch all topics with specified streams_ids'

zulipterminal/model.py
sumanthvrao: Oops, I had missed this updated. I think my question was justified. Thanks for the explanation :)

zulipterminal/model.py
sumanthvrao: All right. Wanted to have a bit more discussion regarding this.

zulipterminal/model.py
sumanthvrao: Done :+1:. Note that for space constraint, I had to move this a bit.

zulipterminal/model.py
sumanthvrao: Noted. Yes, I can repro the same timings as you. I read similarly, and was mostly sure the wait wasn't required (hence commented it). Will remove it in the upcoming iteration.

zulipterminal/model.py
sumanthvrao: Updated this based on the discussion. :+1:

zulipterminal/model.py
sumanthvrao: Updated :+1:

sumanthvrao:
@neiljp @amanagr This is ready for a review. I am excited to see what you guys think of this.

neiljp:
@sumanthvrao I just merged the initial data commits; I did make the two small changes where I mentioned, and adjusted a commit title, but it should be fairly straightforward to rebase.

sumanthvrao:
@neiljp Thanks for merging the first few commits. I have rebased the PR. :+1:

zulipterminal/ui_tools/views.py
```diff
@@ -305,6 +305,41 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+                self.keypress(size, 'up')
+                return True
+            elif button == 5:
+                self.keypress(size, 'down')
+                return True
+        return super(TopicsView, self).mouse_event(size, event, button, col,
+                                                   row, focus)
+
+    def keypress(self, size: Tuple[int, int], key: str) -> str:
+        if is_command_key('TOGGLE_TOPIC', key):
```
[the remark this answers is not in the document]
sumanthvrao: Thanks @amanagr . I remember support for `ESC` back when it was an overlay widget, but may have removed it in an iteration. Will add it :+1:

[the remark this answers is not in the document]
sumanthvrao: Added `ESC` button as a separate commit for now.

tests/ui/test_ui_tools.py
```diff
@@ -1640,3 +1645,33 @@ def test_text_content(self, mocker,
...
+class TestTopicButton:
+    @pytest.mark.parametrize('width, count, stream_id, title, stream_name', [
+        (8, 2, 86, 'topic1', 'Django'),
+        (9, 1, 14, 'topic2', 'GSoC'),
+        (25, 1000, 205, 'topic3', 'PTEST'),
+    ])
+    def test_init_calls_top_button(self, mocker, width, count, title,
+                                   stream_id, stream_name):
+        controller = mocker.Mock()
+        controller.model.stream_dict = {
```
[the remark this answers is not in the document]
sumanthvrao: @neiljp Not for this. I test for the `__init__` part here which works better with different TopicButton creations (More cases) rather than having a single `topic_button` fixture which calls init once. 

For more info, the 10th commit - `models: Mark 'muted' topics as when displaying topic view.` does include a topic_button fixture.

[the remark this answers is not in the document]
sumanthvrao: Re-reading this, It seems you are asking if there is a 'stream_dict' fixture. *Oops*

Not yet. It is in the BuddyPR PR?

zulipterminal/ui_tools/boxes.py
```diff
@@ -620,6 +620,7 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
                 else:
                     self.model.controller.narrow_to_user(self)
             elif self.message['type'] == 'stream':
+                self.stream_name = self.caption
```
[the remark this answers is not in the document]
sumanthvrao: Yeah! makes sense to move it to `__init__`. :+1:

zulipterminal/ui_tools/buttons.py
```diff
@@ -181,12 +181,20 @@ def _narrow_with_compose(self, button: Any) -> None:
         self._view.write_box.private_box_view(self)
 
 
-class TopicButton(urwid.Button):
-    def __init__(self, stream_id: int, topic: str, model: Any) -> None:
-        self.caption = model.stream_dict[stream_id]['name']  # stream name
+class TopicButton(TopButton):
```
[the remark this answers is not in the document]
sumanthvrao: Done :+1:

zulipterminal/ui_tools/views.py
```diff
@@ -415,8 +415,8 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
             if stream_topic is None:
                 return key
             stream, topic = stream_topic
-            self.controller.narrow_to_topic(TopicButton(stream, topic,
-                                                        self.model))
+            self.controller.narrow_to_topic(TopicButton(topic, stream,
```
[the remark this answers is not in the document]
sumanthvrao: No, particular reason. I can switch it back to how it was in both places.

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
sumanthvrao: Done :+1:

zulipterminal/ui_tools/buttons.py
```diff
@@ -18,11 +18,12 @@ class TopButton(urwid.Button):
     def __init__(self, controller: Any, caption: str,
                  show_function: Callable[..., Any], width: int,
                  prefix_character: Union[str, Tuple[Any, str]]='\N{BULLET}',
-                 text_color: Optional[str]=None,
+                 text_color: Optional[str]=None, is_topic: bool=False,
```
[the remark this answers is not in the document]
sumanthvrao: @neiljp This maybe slightly tricky. Generalizing this to the other buttons I feel is not so easy and after doing a quick mock-up the other don't benefit much from space increse. I pushed a newer version which uses a simpler logic rather than passing a flag from child -> parent, which may solve your worry regarding that.

I had tried before to use `prefix_character` for account spacing. The problem with that is the on hover/focus of the button, there are to large gaps -> after vs current
after
current

tests/ui/test_ui_tools.py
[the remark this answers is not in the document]
neiljp: Yes, apologies. I think it may possibly be in the test-refactoring PR #387 which will fix #385.

neiljp:
@sumanthvrao I merged 5 more commits to get a minimal useful topics UI (adjusting after #423 and not showing the right panel with autohide on, and a few commit text changes), jumping over the width-adjustment commit for now. I'm not sure which ones are dependent on others, but this may make the remaining 9 even easier to deal with :)

sumanthvrao:
@neiljp Rebased and updates this. The major changes from previous iteration is
* I renamed some variables in `update_topics_list`.
* Added tests for topics_search (which is quite similar to stream search) + test for 2 more keypresses related to this
* Added a new commit which redesigns the UI for topicsView with the stream_button on top.

zulipterminal/model.py
```diff
@@ -637,6 +640,21 @@ def append_message(self, event: Event) -> None:
                 set_count([response['id']], self.controller, 1)
             self.controller.update_screen()
 
+    def update_topic_index(self, response: Dict[str, Any]) -> None:
+        """
+        Update topic order in index based on incoming message.
+        """
+        msg_topic = response['subject']
+        topic_index_log = self.index['topics'][response['stream_id']]
```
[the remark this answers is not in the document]
sumanthvrao: Yeah. Sounds simpler.

zulipterminal/model.py
```diff
@@ -637,6 +640,21 @@ def append_message(self, event: Event) -> None:
...
+        for topic_iterator in range(len(topic_index_log)):
```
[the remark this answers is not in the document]
sumanthvrao: Yup! It will

zulipterminal/model.py
```diff
@@ -597,6 +597,18 @@ def append_message(self, event: Event) -> None:
...
+        # If the topic view is toggled for incoming message's
+        # recipient stream, then we re-arrange topic buttons
+        # with most frequent at the top.
+        if (hasattr(self.controller, 'view') and
+            self.controller.view.left_panel.is_in_topic_view and
+            response['type'] == 'stream' and
+            response['stream_id'] == self.controller.view.
+                topic_w.stream_button.stream_id):
+
+            self.controller.view.topic_w.update_topics_list(response)
```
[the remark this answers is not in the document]
sumanthvrao: Done :+1:

zulipterminal/ui_tools/views.py
```diff
@@ -314,6 +314,24 @@ def __init__(self, topics_btn_list: List[Any], view: Any,
...
+                self.log.insert(0, self.log.pop(topic_iterator))
+                return
+        # No previous topics with same topic names are found
+        # hence we create a new topic button for it.
+        new_topic_button = TopicButton(msg_topic,
+                                       response['stream_id'],
+                                       self.view.controller,
+                                       self.view.LEFT_WIDTH,
+                                       0)
+        self.log.insert(0, new_topic_button)
```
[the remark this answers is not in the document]
sumanthvrao: Yeah. seems so. I am only passing in the `stream_id` and `topic_name` now.

tests/ui/test_ui_tools.py
```diff
@@ -1834,10 +1835,40 @@ def test_text_content(self, mocker, count, title, expected_text):
...
+        controller = mocker.Mock()
+        controller.model.muted_topics = muted_topics
+        controller.model.stream_dict = {
+            205: {'name': stream_name}
+        }
+        params = dict(controller=controller,
+                      width=40,
+                      count=0)
+        topic_button = TopicButton(stream_id=205,
+                                   topic=title, **params)
```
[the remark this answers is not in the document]
sumanthvrao: I removed the unused fixture for now. For this test we don't really need it as we assert for `button.__init__`

tests/model/test_model.py
```diff
@@ -562,14 +562,15 @@ def test__stream_info_from_subscriptions(self, initial_data, streams,
...
+        mocker.patch('zulipterminal.model.Model.update_topic_index')
         index_msg = mocker.patch('zulipterminal.model.index_messages',
                                  return_value={})
         model.msg_list = mocker.Mock()
         create_msg_box_list = mocker.patch('zulipterminal.model.'
                                            'create_msg_box_list',
                                            return_value=["msg_w"])
         model.msg_list.log = []
-        event = {'message': {'id': 0}}
+        event = {'message': {'id': 0, 'type': 'private'}}
```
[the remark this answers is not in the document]
sumanthvrao: @neiljp I made use of `message_fixture` here, which does solve the problem. It also is not specific to one type as the fixture iterates through stream, private and groupPM. I have added this in the same commit. Let me know if I should split this commit.

sumanthvrao:
@neiljp Updated this as per your feedback.

zulipterminal/ui_tools/views.py
```diff
@@ -315,12 +315,15 @@ def __init__(self, topics_btn_list: List[Any], view: Any,
         self.stream_button = stream_button
         self.list_box = urwid.ListBox(self.log)
         self.topic_search_box = PanelSearchBox(self, 'SEARCH_TOPICS')
+        self.header_list = urwid.Pile([self.stream_button,
+                                       urwid.Divider('-'),
```
neiljp: I'm not sure if we need these dashed lines here? Solid lines might be cleaner?

zulipterminal/ui.py
```diff
@@ -177,15 +177,21 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
-            self.controller.editor_mode = True
-            self.controller.editor = self.stream_w.stream_search_box
+            if self.left_panel.is_in_topic_view:
+                self.topic_w.topic_search_box.set_edit_text("")
+                self.controller.editor_mode = True
+                self.controller.editor = self.topic_w.topic_search_box
+            else:
+                self.stream_w.stream_search_box.set_edit_text("")
+                self.controller.editor_mode = True
+                self.controller.editor = self.stream_w.stream_search_box
```
neiljp: The only difference here is which box, so maybe just set a variable in the conditional and do the common actions after that?

zulipterminal/ui_tools/views.py
```diff
@@ -311,9 +311,34 @@ def __init__(self, topics_btn_list: List[Any], view: Any,
...
+        if not self.view.controller.editor_mode:
+            return
+        # wait for any previously started search to finish to avoid
+        # displaying wrong topics list.
+        with self.search_lock:
+            topics_to_display = self.topics_btn_list.copy()
+            for topic_button in self.topics_btn_list:
+                if not topic_button.topic_name.lower().startswith(
+                        new_text.lower()):
+                    topics_to_display.remove(topic_button)
```
neiljp: You've moved to the locking version here; what do you feel about the list comprehension approach?

On a related note, this only matches the start of topics; we could (later?) support anywhere in the word, or similar?
Eg. "Install error" wouldn't be matched with "error".

zulipterminal/ui_tools/boxes.py
```diff
@@ -767,54 +767,29 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
     """
-
-    search_text = ("Search [" +
-                   ", ".join(keys_for_command("SEARCH_PEOPLE")) +
-                   "]: ")
-
-    def __init__(self, user_view: Any) -> None:
-        self.user_view = user_view
-        super(UserSearchBox, self).__init__(edit_text=self.search_text)
+    def __init__(self, panel_view: Any, search_command: str='') -> None:
```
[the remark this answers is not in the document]
sumanthvrao: Right. I can make this default to be 'SEARCH_STREAM`

zulipterminal/ui_tools/boxes.py
```diff
@@ -767,54 +767,29 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
-        if is_command_key('ENTER', key) and len(self.stream_view.log):
-            self.stream_view.view.controller.editor_mode = False
-            self.stream_view.set_focus("body")
-            self.stream_view.body.set_focus(0)
-        if is_command_key('GO_BACK', key):
-            self.stream_view.view.controller.editor_mode = False
+        if (is_command_key('ENTER', key) and
+                ((self.search_command == 'SEARCH_STREAMS' and
+                  len(self.panel_view.log)) or
+                 self.search_command == 'SEARCH_PEOPLE')):
```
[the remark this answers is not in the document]
sumanthvrao: We do need to check if the `log` is empty so that we don't accidentally set focus on an empty widget and cause it to crash. Further, this occurs only in case of SEARCH_STREAMS and later SEARCH_TOPCIS.

I have made it look a bit better with an approach like this :
```
if is_command_key('ENTER', key):
            self.panel_view.view.controller.editor_mode = False
            self.panel_view.set_focus("body")
            if hasattr(self.panel_view, 'log') and len(self.panel_view.log):
                self.panel_view.body.set_focus(0)
```

zulipterminal/ui_tools/views.py
sumanthvrao: Seems fair to support the List comprehensions :+1: 

Agreed. Something along the lines of `match_user` in helper.py can be used for that.

zulipterminal/ui.py
sumanthvrao: Sounds fair :+1:

zulipterminal/ui_tools/views.py
sumanthvrao: Sure. Solid lines look better also.

neiljp:
@sumanthvrao This should now be fully merged with the adjustments discussed :tada:

PR #388, opened 2019-06-15
ui: Restrict max width of center column for easier reading.

aero31aero:
Adds a config option layout=space-between to make the center
column fix width and add space between the three columns on
wider screens. On smaller screens, the original behavior is
preserved.

This mimics the webapp's idea of having the line-length be
small so that the text is easy to read even on larger screens.

## Screenshots:

1. Small Width:
image

2. Medium Width:
image

3. Large Width:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ If you see further text, like the following, `zulip-term` should be loading!
 Loading with:
    theme 'default' specified with no config.
    autohide setting 'autohide' specified with no config.
+   layout setting 'fill' specified with no config.
 Welcome to Zulip.
 ```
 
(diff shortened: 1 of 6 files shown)


aero31aero:
@neiljp should I rebase this or this feature isn't needed now?

neiljp:
@aero31aero I would suggest seeing what people think; I'm not sure how much work is necessary to rebase? We could discuss on #zulip-terminal (did we already?), and/or open an issue? I expect something like this could definitely be useful, though I don't normally have a window that large personally, but I'm not sure where people would prefer the 'scaling' to take place: between the panels and the message list, or on the outside of the panels? Or maybe even filling extra space with the 'topics' view all the time, rather than toggling the view...?

PR #389, opened 2019-06-15
bugfix: Avoid calls to 'update_screen' before setting 'update_pipe'.

sumanthvrao:
A race condition exists between calling `update_user_list` from `_start_presence_updates` (asynch) which in turn calls 'Controller.update_screen' before setting 'update_pipe' property during initialisation of
Controller.main(). This causes the following traceback during some cases.
```
Loading -Exception in thread Thread-5:
  File "/home/sumanthvrao/.pyenv/versions/3.6.5/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/sumanthvrao/.pyenv/versions/3.6.5/lib/python3.6/threading.py", line 864, in run
  File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/ui_tools/views.py", line 440, in update_user_list
  File "/home/sumanthvrao/Git_Hub/zulip-terminal/zulipterminal/core.py", line 81, in update_screen
    os.write(self.update_pipe, b'1')
AttributeError: 'Controller' object has no attribute 'update_pipe'
```

We now check for the attribute explicitly before calling 'update_user_list'.

Fixes: #391

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -78,7 +78,8 @@ class Controller:
 
     def update_screen(self) -> None:
         # Write something to update pipe to trigger draw_screen
-        os.write(self.update_pipe, b'1')
+        if hasattr(self, 'update_pipe'):
+            os.write(self.update_pipe, b'1')
 
     def draw_screen(self, *args: Any, **kwargs: Any) -> None:
         self.loop.draw_screen()


sumanthvrao:
Have updated this with the modification as per the discussion here - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Early.20update_screen.20.23T389/near/754375

neiljp:
Fixed via 6f744a4782c831808781fa67957e35ff7ec24ba5 with a little rewording.

PR #390, opened 2019-06-15
model: Use event callback mapping when registering desired events

neiljp:
Now only one data-structure is necessary for mapping event names (to
associated callback methods), rather than relying upon a separate list
which could become unsynchronized with the mapping.

The switch to an OrderedDict simplifies testing.

Discussed at https://chat.zulip.org/#narrow/stream/49-development-help/topic/event.20names.2Ftypes/near/753673


amanagr:
Seems like a good way to check if we are handling all the events we are registering to. LGTM.

neiljp:
@amanagr @sumanthvrao Thanks for the feedback! :+1:

PR #392, opened 2019-06-18
WIP: Tutorial.

amanagr:
This PR brings following changes:
* Shows a Tutorial at startup.
* Allows user to change settings before loading UI.

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -12,7 +12,7 @@ from urwid import set_encoding
 from zulipterminal.core import Controller
 from zulipterminal.model import ServerConnectionFailure
 from zulipterminal.config.themes import (
-    THEMES, all_themes, complete_and_incomplete_themes
+    all_themes, complete_and_incomplete_themes
 )
 from zulipterminal.version import ZT_VERSION
 
@@ -200,7 +200,8 @@ def main(options: Optional[List[str]]=None) -> None:
         else:
             theme_to_use = zterm['theme']
         available_themes = all_themes()
-        if theme_to_use[0] not in available_themes:
+        theme = theme_to_use[0]
+        if theme not in available_themes:
             print("Invalid theme '{}' was specified {}."
                   .format(*theme_to_use))
             print("The following themes are available:")
@@ -239,7 +240,8 @@ def main(options: Optional[List[str]]=None) -> None:
                 sys.exit(1)
             boolean_settings[setting] = (zterm[setting][0] == valid_values[0])
         Controller(zuliprc_path,
-                   THEMES[theme_to_use[0]],
+                   theme,
+                   zuliprc_path,
                    **boolean_settings).main()
     except ServerConnectionFailure as e:
         print(in_color('red',
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -211,17 +211,6 @@ def main(options: Optional[List[str]]=None) -> None:
                   "using -t/--theme options on command line.")
             sys.exit(1)
 
-        print("Loading with:")
-        print("   theme '{}' specified {}.".format(*theme_to_use))
-        complete, incomplete = complete_and_incomplete_themes()
-        if theme_to_use[0] in incomplete:
-            print(in_color('yellow',
-                           "   WARNING: Incomplete theme; "
-                           "results may vary!\n"
-                           "      (you could try: {})".
-                           format(", ".join(complete))))
-        print("   autohide setting '{}' specified {}."
-              .format(*zterm['autohide']))
         # For binary settings
         # Specify setting in order True, False
         valid_settings = {
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -157,6 +157,7 @@ def parse_zuliprc(zuliprc_str: str) -> Dict[str, Any]:
         'theme': ('default', NO_CONFIG),
         'autohide': ('no_autohide', NO_CONFIG),
         'notify': ('disabled', NO_CONFIG),
+        'tutorial': ('view', NO_CONFIG),
     }
 
     if 'zterm' in zuliprc:
@@ -216,6 +217,7 @@ def main(options: Optional[List[str]]=None) -> None:
         valid_settings = {
             'autohide': ['autohide', 'no_autohide'],
             'notify': ['enabled', 'disabled'],
+            'tutorial': ['view', 'skip'],
         }
         boolean_settings = dict()  # type: Dict[str, bool]
         for setting, valid_values in valid_settings.items():
--- a/zulipterminal/config/tutorial.py
+++ b/zulipterminal/config/tutorial.py
@@ -1,0 +1,18 @@
+# Tutorial to display at start
+TUTORIAL = [
+    ('active', 'Welcome to Zulip!\n'),
+    '\nPlease go through this totorial to understand the \n'
+    'basic workflow of zulip-terminal.',
+    ('unread', '\nMoving around\n'),
+    '- You can use arrow keys or `j`/`k`/`l`/`m` keys to move around and'
+    '`esc` to go back or move out of narrow.',
+    ('unread', '\nFor Reading\n'),
+    '- Use the left sidebar (or the `n` key) to review messages '
+    'thread-by-thread.\n',
+    '- Use the End key to mark all messages in the current view as read.',
+    ('unread', '\nFor Writing\n'),
+    '- To reply to a message or use `r` or `Enter`.\n',
+    '- You can type your message in the bottom big box and send it using'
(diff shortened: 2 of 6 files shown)


amanagr:
I think this is mostly ready feature wise, only things remaining are:
* Adding tests
* Improve Tutorial
* Restructure commits.

I would appreciate any feedback on this! (code wise too)

zulipterminal/config/tutorial.py
```diff
@@ -0,0 +1,24 @@
...
+A few basic workflows are useful when getting oriented.
+
+Moving around
+
+- You can use arrow keys or `j`/`k`/`l`/`m` keys to move around and
+ `esc` to go back or move out of narrow.
+
+For reading
+
+- Use the left sidebar (or the `n` key) to review messages thread-by-thread.
```
shreyamalviya: Maybe "review unread messages topic-by-topic", instead? Also, we can navigate through unread streams using the left sidebar, but unread topics using `n`, so maybe separate those two?

zulipterminal/config/tutorial.py
```diff
@@ -0,0 +1,24 @@
...
+- Use the down arrow key (`↓`/`j`) to move the cursor to the next message.
+- Use the End key to mark all messages in the current view as read.
```
shreyamalviya: `end` instead of End?

zulipterminal/config/tutorial.py
```diff
@@ -0,0 +1,24 @@
...
+
+For writing
+
+- To reply to a message, click on it, or use `r` or `Enter`.
+- You can type your message in the bottom big box and send it using `ctrl d` or `alt enter`.
+
+\n\n
+"""
```
shreyamalviya: I think we should mention at the end that pressing `?` would open up the help menu for more.

sumanthvrao:
@amanagr I had a go with this today. It's great overall! :+1:  I had few topics to address/request. Futher discussion can take place on these of course :
* In this PR, adding `h` to skip tutorial hence-forth is a nice touch. But suppose, I want to again enable it on startup then the only way I can do that is by changing the zulilprc `tutorial` flag manually. So, perphaps, like we do with the `theme` setting, we can add a flag (`--enable-tutorial = True`) so that the arg-parser can use it to override the setting in zuliprc 
* The only thing I felt odd about the startuppage was the way text was aligned. Maybe we can stick to left/right indentation rather than centre. Not sure how this will look though.

I skimmed through the code in each commit but haven't yet done an indepth code-wise review.

amanagr:
@sumanthvrao I thinking of doing a followup PR after this where we can toggle any setting in the `LoadingView`. User can just press `a` to switch from `autohide` to `no_autohide` and so on...

amanagr:
@neiljp this is already getting hard to manage, can we merge some initial commits?

amanagr:
@neiljp @sumanthvrao I would love to have this PR reviewed before proceeding further with writing tests. This brings a huge change to how zt loads up and I would welcome suggestions on if we would do it better in a different way.

zulipterminal/config/tutorial.py
```diff
@@ -0,0 +1,18 @@
+# Tutorial to display at start
+TUTORIAL = [
```
neiljp: To echo @shreyamalviya's point previously, after this text (independently of showing the tutorial text), I think it would be good to show a '"Remember you can use '?' to explore the available shortcut/hot keys" message or similar. That would act as a 'look <here> for more keys than we just introduced' as well as a 'remember you can get help <here>' if the tutorial is not enabled.

zulipterminal/config/tutorial.py
```diff
@@ -0,0 +1,18 @@
...
+    ('active', 'Welcome to Zulip!\n'),
+    '\nPlease go through this totorial to understand the \n'
+    'basic workflow of zulip-terminal.',
+    ('unread', '\nMoving around\n'),
+    '- You can use arrow keys or `j`/`k`/`l`/`m` keys to move around and'
+    '`esc` to go back or move out of narrow.',
+    ('unread', '\nFor Reading\n'),
+    '- Use the left sidebar (or the `n` key) to review messages '
+    'thread-by-thread.\n',
+    '- Use the End key to mark all messages in the current view as read.',
```
neiljp: As Shreya said, this key needs a highlight/tag.

zulipterminal/cli/run.py
```diff
@@ -200,7 +200,8 @@ def main(options: Optional[List[str]]=None) -> None:
         else:
             theme_to_use = zterm['theme']
         available_themes = all_themes()
-        if theme_to_use[0] not in available_themes:
+        theme = theme_to_use[0]
```
neiljp: Based on other renaming here, `theme_name`?

This first commit feels like a combination of 'preparing for other commits' elements, the most obvious of which is working with just the theme names in run.py, and only accessing the theme itself in core.py. The other element is setting the zuliprc_path in core.py, which perhaps should go with another commit. Then you can describe the intent of this first commit in terms of just that theme_name vs theme change.

zulipterminal/cli/run.py
```diff
@@ -239,7 +240,8 @@ def main(options: Optional[List[str]]=None) -> None:
                 sys.exit(1)
             boolean_settings[setting] = (zterm[setting][0] == valid_values[0])
         Controller(zuliprc_path,
-                   THEMES[theme_to_use[0]],
+                   theme,
+                   zuliprc_path,
```
neiljp: You're passing this parameter twice now?

zulipterminal/core.py
```diff
@@ -39,10 +39,7 @@ def __init__(self, config_file: str, theme_name: str,
         self.client = zulip.Client(config_file=config_file,
                                    client='ZulipTerminal/{} {}'.
                                           format(ZT_VERSION, platform()))
-        self.model = Model(self)
-        self.view = View(self)
-        # Start polling for events after view is rendered.
-        self.model.poll_for_events()
+        self.init_model_view()
```
neiljp: Currently `self.view` should exist before `initialize_loop` (which you just introduced), but this could cause self.view to not be set at this point.

I've not reviewed all the commits, but do we really need this to be asynch?

neiljp (commented):
@amanagr I see this is getting big now, so thanks for the extra ping to review this :+1: 

I've left some feedback inline, some actually regarding the implementation generally and also on some early commits, but mostly in terms of high-level thoughts as to how it behaves.

I think the concept and functionality is really great at this point, though I have some concerns about the implementation I've read so far. You seem to add code in one commit and then delete it in another; it would be clearer if you did that as a 'move' in just one commit, so we can see what is being lost vs what is being added.

zulipterminal/core.py
amanagr: We don't need `self.view` to initialize_loop :) `initialize_loop` loads `LoadingView` first and then when we have `self.view` ready, we display that. Also, In the latter commits I have split the function into two parts, first I load model, then initialize loop and then load the view.

zulipterminal/ui_tools/views.py
```diff
@@ -842,3 +844,53 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+            if key == "enter":
+                # display the view
+                self.controller.init_view()
+                self.controller.loop.widget = self.controller.view
+                self.controller.loop.screen.register_palette(
+                    self.controller.theme)
+                self.controller.update_screen()
+            if key == 'h' and self.controller.wait_after_loading:
+                # Add tutorial=skip to the generated zuliprc
+                self.controller.update_zuliprc('tutorial', 'skip')
```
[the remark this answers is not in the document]
amanagr: This is a great feature, which I think I will add as a v2 or possibly when we have less commits to keep track of in this PR. For now, users will have to remove `tutorial=skip` from the zuliprc to see this menu back. (We can add this info to the README.)

zulipterminal/core.py
```diff
@@ -337,7 +316,7 @@ def spinning_cursor() -> Any:
                 for cursor in '|/-\\':
                     yield cursor
 
-        def loading_text(spinner: List[str]) -> List[Any]:
```
sumanthvrao: `List[str]` is still the type here right?

sumanthvrao (commented):
@amanagr I gave this a review. I had a go with this for most of today and it works wonderfully. 

I like how we can easily toggle settings interactively with the help of the landing page this PR takes us to. I left a few comments clarifying some small points and one which has traceback from an error I received. Most of the commits from 2nd till 7th don't work on their own. They appear to crash `zulip-term` when I run in the middle of a rebase. The left alignment of text looks better than the previous centre alignment :+1:.

As Neil mentioned in one of his comments, the way we display text in the landing page (tutorial page) can be discussed further. Maybe we can show it in the same style as the help view. ;)

amanagr:
@sumanthvrao thanks for your thoughts. I will look into them.

PR #393, opened 2019-06-18
bugfix: Ensure 'self.update' property is set after messages are fetched.

sumanthvrao:
On startup, the event handler for tracking incoming messages would
be registered with 'append message' as callback. However, the 'self.update'
variable would not be set to true yet. This is due to 'get_messages' not
setting the parameter within the conditional when 'query_range'
and length of response are equal.

I have added a failing test for the earlier seen bug here.

Steps to reproduce this:

* Logged in (as Hamlet) from ZT. I have around 75 unread messages across 5 streams.
* Remained in the default All messages view and do not enter any stream/PM.
* Log in from web-app as another user (aaron) and sent a message to a stream (devel). The unread messages don't change in ZT.

However, If after I login and select a stream (after narrowing), and then repeat this process, the unread message does increase.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -371,6 +371,13 @@ class TestModel:
         model.get_messages(num_after=0, num_before=0, anchor=0)
         assert model.update is False
 
+        # TEST `query_range` == no_of messages received
+        # len(messages_successful_response) = 3 so comparing this with
+        # num_after + num_before + 1 = 3, should set update to True.
+        model.update = False
+        model.get_messages(num_after=1, num_before=1, anchor=0)
+        assert model.update is True
+
     # FIXME This only tests the case where the get_messages is in __init__
     def test_fail_get_messages(self, mocker, error_response,
                                initial_data, num_before=30, num_after=10):
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -43,7 +43,7 @@ class TestModel:
         assert model.msg_view is None
         assert model.msg_list is None
         assert model.narrow == []
-        assert model.update is False
+        assert model.update_allowed is False
         assert model.stream_id == -1
         assert model.stream_dict == {}
         assert model.recipients == frozenset()
@@ -332,7 +332,7 @@ class TestModel:
         anchor = messages_successful_response['anchor']
         if anchor < 10000000000000000:
             assert model.index['pointer'][str(model.narrow)] == anchor
-        assert model.update is True
+        assert model.update_allowed is True
 
     def test_get_message_false_first_anchor(
             self, mocker, messages_successful_response, index_all_messages,
@@ -367,16 +367,16 @@ class TestModel:
         assert model.index['pointer'][str(model.narrow)] == 0
 
         # TEST `query_range` < no of messages received
-        model.update = False  # RESET model.update value
+        model.update_allowed = False  # RESET model.update_allowed value
         model.get_messages(num_after=0, num_before=0, anchor=0)
-        assert model.update is False
+        assert model.update_allowed is False
 
         # TEST `query_range` == no_of messages received
         # len(messages_successful_response) = 3 so comparing this with
-        # num_after + num_before + 1 = 3, should set update to True.
-        model.update = False
+        # num_after + num_before + 1 = 3, should set update_allowed to True.
+        model.update_allowed = False
         model.get_messages(num_after=1, num_before=1, anchor=0)
-        assert model.update is True
+        assert model.update_allowed is True
 
     # FIXME This only tests the case where the get_messages is in __init__
     def test_fail_get_messages(self, mocker, error_response,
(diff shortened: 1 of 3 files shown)


zulipterminal/model.py
```diff
@@ -300,7 +300,7 @@ def get_messages(self, *,
             if first_anchor and response['anchor'] != 10000000000000000:
                 self.index['pointer'][str(self.narrow)] = response['anchor']
             query_range = num_after + num_before + 1
-            if len(response['messages']) < (query_range):
+            if len(response['messages']) <= (query_range):
```
punchagan: You could also get rid of the unnecessary parens around `query_range` while you are at it.

tests/model/test_model.py
```diff
@@ -371,6 +371,13 @@ def test_get_message_false_first_anchor(
         model.get_messages(num_after=0, num_before=0, anchor=0)
         assert model.update is False
 
+        # TEST `query_range` == no_of messages received
+        # len(messages_successful_response) = 3 so comparing this with
+        # num_after + num_before + 1 = 3, should set update to True.
+        model.update = False
```
punchagan: I wonder if `model.is_upto_date` or `model.is_updated` is a better name, than `model.update`. But, I guess that can come as a separate change if we'd like to change it.

zulipterminal/model.py
sumanthvrao: Sure. Thanks for pointing it out! :)

tests/model/test_model.py
sumanthvrao: Definitely! I will add the change as a seperate for now :)

sumanthvrao: @amanagr What do you think of renaming `self.update` here?

amanagr:
@sumanthvrao this is great! For renaming I think `self.update_allowed` maybe better.

sumanthvrao:
@punchagan, @amanagr Thanks for your quick feedback regarding this! I have changed the property to `update_allowed` for a better description.

zulipterminal/model.py
```diff
@@ -300,7 +300,7 @@ def get_messages(self, *,
             if first_anchor and response['anchor'] != 10000000000000000:
                 self.index['pointer'][str(self.narrow)] = response['anchor']
             query_range = num_after + num_before + 1
-            if len(response['messages']) < (query_range):
+            if len(response['messages']) <= query_range:
```
neiljp: This seems like a reasonable fix.

zulipterminal/model.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, controller: Any) -> None:
         self.msg_view = None  # type: Any
         self.msg_list = None  # type: Any
         self.narrow = []  # type: List[Any]
-        self.update = False
+        self.update_allowed = False
```
neiljp: Reading the code, this variable is only used inside `Model`, and only between startup and the first `get_messages`?
Therefore, if we just renamed this, I would suggest something more along the lines of `__have_fetched_initial_messages`?

However, if my logic is correct: in core.py, `Model` must finish initializing (including `_update_initial_data(...)`, whose threads, including the initial call to `get_messages`, must join and finish), before the `Model.poll_for_events()` starts.

if that is correct, we could set a flag at the end of `Model.__init__` and assert on it at the start of `poll_for_events()` to ensure this doesn't change, or simply remove it entirely?

Perhaps @amanagr can shed more light on this; was this intended for another purpose?

punchagan: Also, while searching for the usages of `model.update` I found `core.py` has is setting `self.update` on the `Controller`, and this seems like a vestige of some old code, unless I'm missing something.

neiljp: @punchagan Agreed about that; I think I saw that a while back, but was focused on other refactoring at the time, and wanted to understand the data flow better before looking into it.

Is this flag perhaps intended as a way of ensuring the index isn't updated at the same time by `get_messages` and `append_message`, throughout the time ZT is running? (just checking the flag doesn't resolve this situation, of course, it just skips the append, which just "drops" the message, and it isn't added to the UI?)

amanagr: The purpose of `model.update` is to ensure that we only append message once all the messages have been fetched for a narrow and we used to reset update to false every time the narrow changed in controller. This seems to have been refactored to something else entirely. In `core.py` it should be `model.update` instead. @sumanthvrao can you fix this which you are at it?

amanagr:
Perhaps, this fix is not the solution to the original problem of unread count not being updated. The proper solution would be to just update the unread count and not append the new message in the index.

zulipterminal/model.py
amanagr: Now that it's all coming back to me, with this fix, we will be appending messages that are not immediately next to the last message.

amanagr:
@sumanthvrao I would be great to have you work on this, since you recently worked on unread counts and have a fresh pair of eyes for the whole update logic. Checklist items:
- [ ] Check new message are appended only after all the message in that narrow are visisble.
- [ ] Update unread count of the new message even if cannot index it at this moment.

sumanthvrao:
Discussion moved to https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/unread.20counts.20'unchanged'.20on.20startup/near/756152

sumanthvrao:
Thanks for the feedback and reviews.

Since this fix is not the apt one here, I am closing this PR. An issue has been filed for developments towards this (#395) and another PR has been opened which tacked the issue this addressed (#398 )

PR #394, opened 2019-06-18
Adds test to 'quoted' message PR.

sumanthvrao:
Tests added for e7421af798ada9a6729d5956d0cdeec9865ff9ae.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1274,6 +1274,51 @@ class TestMessageBox:
             assert write_box.msg_edit_id is None
             write_box.msg_write_box.set_edit_text.assert_not_called()
 
+    @pytest.mark.parametrize('raw_html, expected_content', [
+        ("""<blockquote>
+                <p>A</p>
+            </blockquote>
+            <p>B</p>""", "░ A\n\nB"),
+        ("""<blockquote>
+                <blockquote>
+                    <p>A</p>
+                </blockquote>
+                <p>B</p>
+            </blockquote>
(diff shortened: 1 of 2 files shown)


neiljp:
@sumanthvrao As per my PM to you, I've merged the first commit with slightly different language, so this PR can refocus on tests :)

sumanthvrao:
@neiljp This has been simplified and 'expected content' is more readable now. The parameters could be placed in a single line (without the need for ''' ), however that's now how the 'raw' html incoming from the server looks like. I have tried to preserve as much of the structure here as possible.

One thing to note here is that, newlines and spaces in the 'raw_html' affect the rendered text.

sumanthvrao:
@neiljp to clarify, do you mean in the tests or in the code (`main_view` of `boxes.py`)?

neiljp:
@sumanthvrao I meant in the code, so the test would be to just test the new method, and you wouldn't then need a lot of setup in the test, since it would be in the new method.

sumanthvrao:
@neiljp All right. I pushed a new commit here which extracts the function based on the point you mentioned above.

sumanthvrao:
@neiljp updated this as per our diccussion on `czo`.

amanagr:
@sumanthvrao It's will be good to have the current behavior as tests. LGTM.

neiljp:
@sumanthvrao Just merged this! :+1: Looking forward to developing these tests further and seeing if we can iron out the few bugs left with quoting :)

sumanthvrao:
great! Glad to have this merged :+1:

PR #397, opened 2019-06-20
keys: Add a key to view a message in the browser.

punchagan:


amanagr:
Works well on WSL too with the below addition in `~/.bashrc`. I will add them to the docs after this gets merged. Might we worth adding tests for this too. Thanks @punchagan.
```
export DISPLAY=:0
export BROWSER=/mnt/c/Program\ Files\ \(x86\)/Google/Chrome/Application/chrome.exe
```

sumanthvrao:
@punchagan This does open the respective message (in narrow) :+1:. I tested on both Dev server and `czo`.

However, It does give be the following traceback on the terminal:
```
[22746 ERROR: browser_process_sub_thread.cc(221)] Waited 3 ms for network service.
Opening in existing browser session.]
```
It currently taked about 2-3 seconds from pressing 'v' to it opening a new tab in browser.

punchagan:
> However, It does give be the following traceback on the terminal:
> 
> ```
> [22746 ERROR: browser_process_sub_thread.cc(221)] Waited 3 ms for network service.
> Opening in existing browser session.]
> ```
> 
> It currently takes about 2-3 seconds from pressing 'v' to it opening a new tab in browser.

What's your default browser?  Does executing `webbrowser.open` in a stand alone script behave similarly?

punchagan:
> It currently takes about 2-3 seconds from pressing 'v' to it opening a new tab in browser.

Does adding the args `new=1` or `new=2` to the open call, change anything? 

https://docs.python.org/3.7/library/webbrowser.html#webbrowser.open

sumanthvrao:
@punchagan A small addition to README for the new key-mapping (`v`) may help here.

sumanthvrao:
Thanks for this! Can confirm this is working great without any log messages or other tracebacks appearing on the terminal for both firefox and chrome as default browser.

amanagr:
@punchagan I think It would be great to add "How to set your default browser?" in the docs for this. or additing it to `zuliprc` and let zt take care of it, since then that would be a zt specific settings and should be preferred?

punchagan:
I don't think setting the default browser is something that `zulip-term` needs to worry about. The `webbrowser` module attempts to figure out the default browser, and I think that's sufficient. We could add a line documenting the `BROWSER` environment variable, but I'm not sure we need much more than that.

punchagan:
I've added a test, though I'm not sure how much value/confidence the test is adding. :)

neiljp:
@punchagan Thanks for initiating this work - the original motivation is now resolved through the merged #991 so I'm closing this :tada:

Issue #395, opened 2019-06-20
Improve co-ordination between `append_message` and `get_message`;


sumanthvrao:
This issue is filed to keep track of developments regarding this:

As of now, we do not append_message unless all the message in the narrow have been fetched. To elaborate, Suppose a user has 20 unread messages, then on startup, ZT fetches 30 old messages and 10 unread messages. Now, if a new message comes in we do not append it to the log (If we did, we would lose out on the other 10 unread messages); However, we do increase the unread count here, without storing it in the index.

The way we currently do this is by maitaining a flag `self.update` which is checked before 'appending_messages' and is set to True when `The messages fetched < the query range` (indicating all messages in that narrow have been fixed).

Potential issues here are:
* On startup, when the messages fetched (41) == query_range(41), when new messages come in, neither unread_count (or append_message) work.
* How can we prevent the index from being updated twice?
* Issues with the index, we lose track of previous messages/incoming messages (Possibly #372 )

PR #398, opened 2019-06-20
Fix handling un-read count for un-indexed messages.

sumanthvrao:
This PR is a step in the direction of fixing #395.

**TODO**:
Add tests

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -88,9 +88,16 @@ def set_count(id_list: List[int], controller: Any, new_count: int) -> None:
 
     messages = controller.model.index['messages']
     unread_counts = controller.model.unread_counts  # type: UnreadCounts
+    unindexed_messages = controller.model.unindexed_messages
 
     for id in id_list:
-        msg = messages[id]
+        # We check to see if the message we are altering count to is
+        # indexed or not, and fetch it from the respective collection.
+        if id in messages:
+            msg = messages[id]
+        else:
+            msg = unindexed_messages[id]
+            messages = unindexed_messages
 
         if msg['type'] == 'stream':
             key = (messages[id]['stream_id'], msg['subject'])
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -103,6 +103,8 @@ class Model:
 
         self.initial_data = {}  # type: Dict[str, Any]
 
+        self.unindexed_messages = {}  # type: Dict[int, Any]
+
         # Register to the queue before initializing further so that we don't
         # lose any updates while messages are being fetched.
         self._update_initial_data()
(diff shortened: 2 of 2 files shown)


sumanthvrao:
@amanagr provided you pointed out the in-correct fix in #393, what do you think of this approach

amanagr:
Thanks @sumanthvrao. I just checked the response we receive from the server and it includes `found_newest` in it. This I found out is a new parameter which server gives us(https://chat.zulip.org/api/get-messages). So, while czo gives us this new parameter, older versions of the webapp will not, so we should primarily use `found_newest` to determine if we have fetched all the messages in the narrow, and then use the logic you came up with to determine if there are more messages to be fetched.

sumanthvrao:
Great. Will check it our :+1: !

sumanthvrao:
Updated this :+1: 

@neiljp The first 3 commits here are quite independent, there were also some of the changes we were discussing in #393 . This includes, a small indentation change (commit 1), renaming `update` to `found_newest`, which is also the variable the API returns (@punchagan what do you think of this name?) and fixing it as a property of model (commit 2) , and making use of the `found_newest` to update the flag (Thanks for the information @amanagr ) (commit 3)

The later commits (Last 2) are sort of techniques I came up with, which works for now (regardless of them being the best solution). I would love to have your feedback on them.

zulipterminal/ui_tools/views.py
```diff
@@ -87,6 +87,7 @@ def load_new_messages(self, anchor: int) -> None:
         else:
             last_message = None
 
+        self.model.decrease_unindexed_message_count(new_ids)
```
neiljp: This is necessary for `load_old_messages`, too, I suspect?

zulipterminal/model.py
```diff
@@ -257,6 +258,17 @@ def mark_message_ids_as_read(self, id_list: List[int]) -> None:
         })
         set_count(id_list, self.controller, -1)  # FIXME Update?
 
+    def increase_unindexed_message_count(self,
+                                         message: Dict[str, Any]) -> None:
+        unindexed_messages[message['id']] = message
+        set_count([message['id']], self.controller, 1)
```
neiljp: Does this need to check for whether the message is already marked as read?

zulipterminal/helper.py
```diff
@@ -39,6 +39,7 @@
     all_starred=set(),
 )
 
+unindexed_messages = dict()  # type: Dict[int, Any]
```
neiljp: I understand why you might want to avoid having this in the Index itself, but I'm unconvinced it should be a global variable.

For now, placing this in the model might be simpler?

neiljp (commented):
@sumanthvrao While I've left some comments, I think this may be simpler if we associated the 'decrease' (ie. transfer from unindexed to indexed) with just being called whenever we successfully `get_messages`?

That would make it easier to keep the state in the `Model` and maybe even remove the need for at least one of the extra methods - and maybe rename the other?

Issue #141, opened 2018-07-24
Implement Push Notifications.


amanagr:
Show Push Notifications for user-mentions, wildcard-mentions and if stream notifications are enabled. Currently zt doesn't receive any flags since the web client treats zt as a bot and doesn't attach any flags to messages. This will be fixed in the release of zulip(i.e, v1.9.0). So we can implement this after this. 

Also, we should check  if the server isn't sending us any flags for the messages, then probably the server hasn't upgraded their zulip version and we don't display notifications for it either.

Currently there are two libraries we can use to implement this:
on linux: https://pypi.org/project/notify2/
on mac: https://github.com/setem/pync

zulipbot:
Hello @amanagr, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

zulipbot:
Hello @amanagr, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

neiljp:
@amanagr To clarify, did you mean message notifications like we now have implemented via #399 and subsequent work - meaning we can close this issue - or something more like mobile notifications or something else?

amanagr:
Yeah, with #399, this is fixed now.

PR #399, opened 2019-06-20
Notify user of pms/mentions/wildcard_mentions

amanagr:
TODO:
- [x] Test on linux
- [x] Test on macos
- [ ] Add instructions and common errors/fixes to readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ site=https://realm.zulipchat.com
 theme=default
 # Autohide defaults to 'no_autohide', but can be set to 'autohide' to hide the left & right panels except when focused.
 autohide=autohide
+# Notify defaults to 'disabled', but can be set to 'enabled' to display notifications.
+notify=enabled
 ```
 
 ### Notifications
--- a/README.md
+++ b/README.md
@@ -224,6 +224,25 @@ $ pipenv run pip3 install -e .[dev]
 $ pip3 install -e .[dev]
 ```
 
+### Notifications
+
+#### Linux
+
+The following command installs `notify-send` on Debian based systems, similar
+commands can be found for other linux systems as well.
+```
+sudo apt-get install libnotify-bin
+```
+
+#### WSL
+
+Run powershell as **admin** and run these commands to install dependencies for showing notifications:
+
+```
+set-executionpolicy remotesigned
+Install-Module -Name BurntToast
+```
+
 ### Development tasks
 
 Once you have a development environment set up, you might find the following useful, depending upon your type of environment:
--- a/README.md
+++ b/README.md
@@ -69,6 +69,25 @@ theme=default
 autohide=autohide
 ```
 
+### Notifications
+
+#### Linux
+
+The following command installs `notify-send` on Debian based systems, similar
+commands can be found for other linux systems as well.
+```
+sudo apt-get install libnotify-bin
+```
+
+#### WSL
+
+Run powershell as **admin** and run these commands to install dependencies for showing notifications:
+
+```
+set-executionpolicy remotesigned
+Install-Module -Name BurntToast
+```
+
 ## Hot Keys
 | Command                                               | Key Combination                               |
 | ----------------------------------------------------- | --------------------------------------------- |
@@ -222,25 +241,6 @@ $ pipenv run pip3 install -e .[dev]
 2. Install zulip-term, with the development requirements
 ```
 $ pip3 install -e .[dev]
-```
-
-### Notifications
-
-#### Linux
-
-The following command installs `notify-send` on Debian based systems, similar
-commands can be found for other linux systems as well.
-```
-sudo apt-get install libnotify-bin
-```
-
-#### WSL
-
-Run powershell as **admin** and run these commands to install dependencies for showing notifications:
-
-```
-set-executionpolicy remotesigned
-Install-Module -Name BurntToast
 ```
 
 ### Development tasks
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ import codecs
 from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
 from zulipterminal.version import ZT_VERSION
+from zulipterminal.helper import in_color, LINUX, WSL, MACOS
 
 
 class PyTest(TestCommand):
@@ -90,3 +91,22 @@ setup(
         'mypy_extensions>=0.4',
     ],
 )
+
+
+print(in_color('green', '\nZulip Terminal was successfully installed!'))
+if MACOS:
+    sys.exit(0)
+print('\nTo ensure notifications are displayed, you may wish to run the '
+      'following:\n')
+if WSL:
+    print(in_color('blue', 'Run powershell as "admin" and run these commands'
+                   ' to install dependencies for showing notifications:'))
+    print(in_color('red', '\nset-executionpolicy remotesigned\nInstall-Module'
+                          ' -Name BurntToast\n'))
+elif LINUX:
+    print(in_color('blue', 'The following command installs `notify-send` on'
+                   ' Debian based systems, similar commands can be found for'
+                   ' other linux systems as well.'))
+    print(in_color('red', '\nsudo apt-get install libnotify-bin\n'))
+print('You can refer to https://github.com/zulip/zulip-terminal#notifications'
+      ' for more info on showing notifications.')
--- a/setup.py
+++ b/setup.py
@@ -110,3 +110,4 @@ elif LINUX:
     print(in_color('red', '\nsudo apt-get install libnotify-bin\n'))
 print('You can refer to https://github.com/zulip/zulip-terminal#notifications'
       ' for more info on showing notifications.')
+print("Also, make sure `notify=enabled` is set in your zuliprc file!")
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -735,6 +735,19 @@ class TestModel:
         else:
             notify.assert_not_called
 
+    @pytest.mark.parametrize('notify_enabled, is_notify_called', [
+        (True, True),
+        (False, False),
+    ])
+    def test_notify_users_enabled(self, mocker, model, message_fixture,
+                                  notify_enabled, is_notify_called):
+        message_fixture.update({'sender_id': 2, 'flags': ['mentioned']})
+        model.controller.notify_enabled = notify_enabled
+        model.user_id = 1
+        notify = mocker.patch('zulipterminal.model.notify')
+        model.notify_user(message_fixture)
+        assert notify.called == is_notify_called
+
     @pytest.mark.parametrize('response, update_call_count, new_index', [
         ({  # Only subject of 1 message is updated.
             'message_id': 1,
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -568,12 +568,14 @@ class TestModel:
         create_msg_box_list = mocker.patch('zulipterminal.model.'
                                            'create_msg_box_list',
                                            return_value=["msg_w"])
+        model.notify_user = mocker.Mock()
         model.msg_list.log = []
         event = {'message': message_fixture}
 
         model.append_message(event)
 
         assert len(model.msg_list.log) == 1  # Added "msg_w" element
+        model.notify_user.assert_called_once_with(event['message'])
         (create_msg_box_list.
          assert_called_once_with(model, [message_fixture['id']],
                                  last_message=None))
@@ -588,12 +590,14 @@ class TestModel:
         create_msg_box_list = mocker.patch('zulipterminal.model.'
                                            'create_msg_box_list',
                                            return_value=["msg_w"])
+        model.notify_user = mocker.Mock()
         model.msg_list.log = [mocker.Mock()]
         event = {'message': message_fixture}
 
         model.append_message(event)
 
         assert len(model.msg_list.log) == 2  # Added "msg_w" element
+        model.notify_user.assert_called_once_with(event['message'])
         # NOTE: So we expect the first element *was* the last_message parameter
         expected_last_msg = model.msg_list.log[0].original_widget.message
         (create_msg_box_list.
@@ -609,6 +613,7 @@ class TestModel:
         create_msg_box_list = mocker.patch('zulipterminal.model.'
                                            'create_msg_box_list',
                                            return_value=["msg_w"])
+        model.notify_user = mocker.Mock()
         model.msg_list.log = [mocker.Mock()]
         set_count = mocker.patch('zulipterminal.model.set_count')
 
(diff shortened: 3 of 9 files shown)


amanagr:
@neiljp I feel this is ready from the wsl perspective. It requires someone to test it on Linux and Mac OS, then add steps to get this working, common problems and their solution to README.

sumanthvrao:
@amanagr This works great and was much needed!

I tested this on a ubuntu 16.04. After installing `notify2`, It was throwing an `dbus` error 
```
configure: error: Package requirements (dbus-1 >= 1.8) were not met). 
```
However running `sudo apt-get install libdbus-1-dev libdbus-glib-1-dev` did the job (Source: https://github.com/autokey/autokey/issues/58), So its probably worht mentioning in the documentation. 

I tested it manually for PM and `@mention` notification. Right now, there is no sound associated with the notification pop-up I recieve, however that's probably setting I need to configure (Allowing terminal notifications with a default sound).

amanagr:
Thanks @sumanthvrao for the valuable feedback. I will include them in the documentation.

neiljp:
@amanagr I assume this is intended to fix #141? I only manually tested this.

I installed notify2 as per setup.py, and then when that caused a traceback (we should add an error message?), I added libdbus-1-dev (via apt) to allow dbus-python==1.2.8 to install. However, that didn't give notifications when I sent a PM from my bot to myself, which did happen if I used the webapp (which indicated I didn't need a special notification server). I installed the glib package @sumanthvrao mentioned, and ensured it rebuilt the dbus-python wheel. **That still didn't work, until I tried to change to a narrow, from the initial one, where notifications were then coming through :)** I think it still worked when I changed back to the previous dbus-python wheel, so it's possible we don't need the second dev package for it?

While I was trying to get the original version to work, I saw mention of notify-send, which it turns out is in libnotify-bin (via apt) in Debian. That produces the notifications via the command-line. I've tried it, but not explored further in terms of embedding it in ZT. That could be a simpler or alternative implementation, since we'd just call out to a script - it's just one system package to install, and no extra wheel etc?

amanagr:
Yeah, looks like notify-send could be a good option.

sumanthvrao:
@neiljp The issue you talked about above is more indicative of the issue we were aiming to solve in #393. Which got merged as a part of 9456865890e90b5db13ab0c72d4e48adbf75bcde. Could you give this a try again as I believe the issue shouldn't exist anymore?

neiljp:
@sumanthvrao This still doesn't seem to trigger for me unless I enter a narrow, ie. notify_user does not get called, so that issue appears still to be present.

zulipterminal/helper.py
```diff
@@ -347,3 +354,20 @@ def match_user(user: Any, text: str) -> bool:
...
+    FNULL = open(os.devnull, 'w')
+    command = ""
+    if WSL:
+        command = ('powershell.exe "New-BurntToastNotification'
+                   ' -Text \'{}\', \'{}\'"'.format(title, text))
+    elif MACOS:
+        command = ("osascript -e 'display notification \"{}\" with title"
+                   " \"{}\"'".format(text, title))
+    elif LINUX:
+        command = 'notify-send {} {}'.format(title, text)
```
neiljp: This worked for me when both had `"{}"` instead, to ensure the command considers them as two arguments rather than (likely) many more.

zulipterminal/helper.py
```diff
@@ -347,3 +354,20 @@ def match_user(user: Any, text: str) -> bool:
...
+    if command:
+        res = subprocess.run(shlex.split(command),
+                             stdout=FNULL, stderr=subprocess.STDOUT)
```
neiljp: `stdout=subprocess.DEVNULL` ?

zulipterminal/helper.py
```diff
@@ -347,3 +354,20 @@ def match_user(user: Any, text: str) -> bool:
...
+    if WSL:
+        command = ('powershell.exe "New-BurntToastNotification'
+                   ' -Text \'{}\', \'{}\'"'.format(title, text))
+    elif MACOS:
+        command = ("osascript -e 'display notification \"{}\" with title"
+                   " \"{}\"'".format(text, title))
+    elif LINUX:
+        command = 'notify-send {} {}'.format(title, text)
+    if command:
+        res = subprocess.run(shlex.split(command),
```
neiljp: Is the shlex actually required, if we construct the command differently?

zulipterminal/helper.py
amanagr: Thanks!

zulipterminal/helper.py
amanagr: oh! I didn't knew.

zulipterminal/helper.py
amanagr: I think it will be better than manually splitting text and feeding it. We will have to take care of a lot of corner cases (probably use some regex to split nested quotes)

amanagr:
Webapp uses message's id as tags on notifications to ensure they are unique per message. I searched for notify-send and couldn't find a similar feature in its docs. I am not sure if we could really implement it on terminal.

amanagr:
The issue was bcoz of `self.found_newest`, It was working for me as I had all the messages read in the `all messages` narrow while you didn't.

zulipterminal/helper.py
neiljp: You're referring to the quotes in the commands you've written, or substituted text?

amanagr: Substituted text.

README.md
```diff
@@ -220,6 +220,25 @@ $ pipenv run pip3 install -e .[dev]
...
 ```
 
+### Notificaitons 
+
+#### Linux
+
+Please run the following command to install dependencies for showing notifications:
+
+```
+sudo apt-get install notify-osd pulseaudio-utils libnotify-bin
```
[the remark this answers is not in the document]
amanagr: So for now I will just mention this is the way to install for debain based systems, and users can find similar solutions for other OS online.

setup.py
```diff
@@ -90,3 +91,16 @@ def long_description():
...
+    print(in_color('blue', 'Now, run powershell as "admin" and run these '
+                           'commands to install dependencies'
+                           ' for showing notifications:'))
+    print(in_color('red', '\nset-executionpolicy remotesigned\nInstall-Module'
+                          ' -Name BurntToast\n'))
+elif LINUX:
+    print(in_color('blue', 'Now, run the following command to install'
+                           ' dependencies for showing notifications:'))
+    print(in_color('red', '\nsudo apt-get install notify-osd pulseaudio-utils'
+                          ' libnotify-bin\n'))
```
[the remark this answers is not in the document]
amanagr: We count just print a link to the notification section?

zulipterminal/helper.py
```diff
@@ -410,3 +418,21 @@ def canonicalize_color(color: str) -> str:
         return color.lower()
     else:
         raise ValueError('Unknown format for color "{}"'.format(color))
+
+
+@asynch
+def notify(title: str, text: str) -> None:
```
neiljp: As per another comment, this might be clearer as `html_text`; this also ensures that the `text` variable is clearly separate.

tests/model/test_model.py
```diff
@@ -671,10 +678,31 @@ def test_append_message(self, mocker, user_profile, response,
...
+            'not_mentioned'])
+    def test_notify_users_calling_msg_type(self, mocker, model,
+                                           message_fixture, vary_each_msg,
+                                           types_when_notify_called):
+        message_fixture.update(vary_each_msg)
+        model.user_id = 1
+        notify = mocker.patch('zulipterminal.model.notify')
+        model.notify_user(message_fixture)
+        if message_fixture['type'] in types_when_notify_called:
+            assert notify.called
```
neiljp: I know we're limiting this test for now, so unless you want to include how notify is called, perhaps add a TODO here?

setup.py
```diff
@@ -90,3 +91,22 @@ def long_description():
...
+print('\nTo ensure notifications are displayed, you may wish to run the '
+      'following:\n')
+if WSL:
+    print(in_color('blue', 'Run powershell as "admin" and run these commands'
+                   ' to install dependencies for showing notifications:'))
+    print(in_color('red', '\nset-executionpolicy remotesigned\nInstall-Module'
+                          ' -Name BurntToast\n'))
+elif LINUX:
+    print(in_color('blue', 'The following command installs `notify-send` on'
+                   ' debain based systems, similar commands can be founds for'
```
neiljp: Same typos ;) Having the text in only one place is good.

Could you add a note to the issue for generating the README partially from keys.py, as this seems related?

zulipterminal/cli/run.py
```diff
@@ -219,9 +208,26 @@ def main(options: Optional[List[str]]=None) -> None:
...
         print("   autohide setting '{}' specified {}."
               .format(*zterm['autohide']))
+        # For binary settings
+        # Specify setting in order True, False
+        valid_settings = {
+            'autohide': ['autohide', 'no_autohide'],
+        }
+        settings = dict()
+        for setting in valid_settings:
+            values = valid_settings[setting]
```
neiljp: `for setting, valid_values in valid_settings.items()`?

zulipterminal/cli/run.py
```diff
@@ -219,9 +208,26 @@ def main(options: Optional[List[str]]=None) -> None:
                            format(", ".join(complete))))
         print("   autohide setting '{}' specified {}."
               .format(*zterm['autohide']))
+        # For binary settings
+        # Specify setting in order True, False
+        valid_settings = {
+            'autohide': ['autohide', 'no_autohide'],
+        }
+        settings = dict()
```
neiljp: Perhaps type-annotate this, which will indicate the expected types with less need for a comment?

Along similar lines, you could rename this to `boolean_settings`?

zulipterminal/model.py
```diff
@@ -594,6 +595,20 @@ def handle_typing_event(self, event: Event) -> None:
                 else:
                     raise RuntimeError("Unknown typing event operation")
 
+    def notify_user(self, message: Dict['str', Any]) -> None:
+        recipient = ''
```
neiljp: minor: move below the conditional?

zulipterminal/model.py
```diff
@@ -594,6 +595,20 @@ def handle_typing_event(self, event: Event) -> None:
                 else:
                     raise RuntimeError("Unknown typing event operation")
 
+    def notify_user(self, message: Dict['str', Any]) -> None:
```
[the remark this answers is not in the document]
amanagr: Weird that mypy didn't catch this. :thinking:

amanagr:
@neiljp updated with the requested changes!

README.md
```diff
@@ -223,6 +223,25 @@ $ pipenv run pip3 install -e .[dev]
 $ pip3 install -e .[dev]
 ```
 
+### Notificaitons
+
+#### Linux
+
+The following command installs `notify-send` on debain based systems, similar
```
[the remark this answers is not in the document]
neiljp: @amanagr You capitalized this, but didn't change the spelling?

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: Types can be defined this way, since it is allowed for forward references.

zulipterminal/helper.py
```diff
@@ -421,3 +429,21 @@ def canonicalize_color(color: str) -> str:
         return color.lower()
     else:
         raise ValueError('Unknown format for color "{}"'.format(color))
+
+
+@asynch
+def notify(title: str, html_text: str) -> None:
+    document = lxml.html.document_fromstring(html_text)
+    html_text = document.text_content()
```
neiljp: One reason to change the function parameter name, was that it would separate the input format (html_text) from the `text_content()` result, ie. just 'text'. That is the case, right?

setup.py
neiljp: Still 'Debain'.

zulipterminal/cli/run.py
```diff
@@ -219,9 +208,25 @@ def main(options: Optional[List[str]]=None) -> None:
...
+        # Specify setting in order True, False
+        valid_settings = {
+            'autohide': ['autohide', 'no_autohide'],
+        }
+        boolean_settings = dict()  # type: Dict[str, bool]
+        for setting, valid_values in valid_settings.items():
+            if zterm[setting][0] not in valid_values:
+                print("Invalid {} setting '{}' was specified {}."
+                      .format(setting, *zterm[setting]))
+                print("The following options are available:")
```
neiljp: Should we change 'options' to 'values'? Here or in the next commit?

setup.py
```diff
@@ -90,3 +91,22 @@ def long_description():
...
+                   ' to install dependencies for showing notifications:'))
+    print(in_color('red', '\nset-executionpolicy remotesigned\nInstall-Module'
+                          ' -Name BurntToast\n'))
+elif LINUX:
+    print(in_color('blue', 'The following command installs `notify-send` on'
+                   ' Debain based systems, similar commands can be found for'
+                   ' other linux systems as well.'))
+    print(in_color('red', '\nsudo apt-get install libnotify-bin\n'))
+print('You can refer to https://github.com/zulip/zulip-terminal#notifications'
+      ' for more info on showing notifications.')
```
neiljp: Did you try installing the package and testing this?

`python setup.py install` does show these messages, but `pip install .` does not.

zulipterminal/helper.py
amanagr: Oh! Now, I understand what you meant by the renames. 😬

zulipterminal/cli/run.py
amanagr: I think options is a more clear wording for it? ( + more user-friendly word)

setup.py
amanagr: Looks very tricky to get it working for pip install . I only checked for `python setup.py install ` prior to this. I tried a few things but nothing is working. Any suggestions?

zulipterminal/cli/run.py
neiljp: This isn't too important, I should have prefaced my comment with 'minor:' :)

setup.py
neiljp: I did briefly research this before, but those solutions also don't appear to work with setup.py. I think this comes down to using wheels. You do get the output if you use '-vvv', but we can't expect that.

Can we integrate this element into the onboarding PR instead?

amanagr: Yeah, that sounds like a better idea. I will include it in the onboarding PR.

neiljp:
@amanagr I just merged the commits we agreed upon, without the optional extra tuple change I indicated we could do. Good work with the testing :) I'm looking forward to using this! :tada:

I do think it would be useful to (optionally) show notifications for selected streams via the streams properties, but that is fine in a follow-up post-discussion :)

amanagr:
Thanks!,  I think it will be a while before we have stream settings.

Issue #273, opened 2019-01-21
Crashed if stream color in short syntax


k-k-mygento:
Hi there.

zulip-term crashed if stream color in short syntax. I.e. `#f00` instead of `#ff0000`. Why it's short? I set it before using web-version of zulip to red via color-picker

```
zulip-term
Loading with:
   theme 'gruvbox' specified in zuliprc file.
   autohide setting 'autohide' specified in zuliprc file.
Welcome to Zulip.
Loading |['food-stalker', 19, '#123456', True] <- debug string print(properties) from ui_tools/buttons.py:71
['general', 4, '#f00', False]
Traceback (most recent call last):
Loading /  File "/tmp/zt/lib/python3.5/site-packages/zulipterminal/cli/run.py", line 169, in main
    autohide_setting).main()
  File "/tmp/zt/lib/python3.5/site-packages/zulipterminal/core.py", line 38, in __init__
    self.view = View(self)
  File "/tmp/zt/lib/python3.5/site-packages/zulipterminal/ui.py", line 35, in __init__
    super(View, self).__init__(self.main_window())
  File "/tmp/zt/lib/python3.5/site-packages/zulipterminal/ui.py", line 89, in main_window
    self.left_column = self.left_column_view()
  File "/tmp/zt/lib/python3.5/site-packages/zulipterminal/ui.py", line 38, in left_column_view
    self.left_col_w = LeftColumnView(self)
  File "/tmp/zt/lib/python3.5/site-packages/zulipterminal/ui_tools/views.py", line 463, in __init__
    self.streams_view(),
  File "/tmp/zt/lib/python3.5/site-packages/zulipterminal/ui_tools/views.py", line 492, in streams_view
    ) for stream in self.view.pinned_streams]
  File "/tmp/zt/lib/python3.5/site-packages/zulipterminal/ui_tools/views.py", line 492, in <listcomp>
    ) for stream in self.view.pinned_streams]
  File "/tmp/zt/lib/python3.5/site-packages/zulipterminal/ui_tools/buttons.py", line 72, in __init__
    self.color = color[:2] + color[3] + color[5]
IndexError: string index out of range
Zulip Terminal has crashed!

```

So when I set channel(stream) color to something like `#123456` via web-version, term works fine.

amanagr:
@k-k-mygento thanks for reporting this issue. ZT assumes streams colours to be defined in #XXXXXX style. It should be an easy fix.

vrongmeal:
[...]

PR #400, opened 2019-06-20
Handle thread colors of the form '#xxx'

alex-ozdemir:
Our machinery for rendering thread colors in
   1. the thread list in the left column and
   2. thread titles in the center column

assumed that colors would be strings of the form '#xxxxxx'.

However, the web interface allows users to specifiy colors of the form
'#xxx', the rendering of which would cause the terminal client to crash.

Now the terminal client knows how to handle both forms, and will produce
a targetted error message in the face of future unknown formats.

Credits to #273 for originally discovering the problem, and @sadjad for
bringing it to my attention and helping me solve it.

### DIFF
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -100,19 +100,10 @@ class StreamButton(TopButton):
                  count: int=0) -> None:
         # FIXME Is having self.stream_id the best way to do this?
         # (self.stream_id is used elsewhere)
-        caption, self.stream_id, orig_color, is_private = properties
+        caption, self.stream_id, color, is_private = properties
         self.model = controller.model
         self.count = count
         self.view = view
-
-        # Simplify the color from the original version & add to palette
-        # TODO Should this occur elsewhere and more intelligently?
-        if len(orig_color) == 7:  # modern zulip server format
-            color = ''.join(orig_color[i] for i in (0, 1, 3, 5))
-        elif len(orig_color) == 4:  # possible with zulip servers <=1.9.0 ?
-            color = orig_color
-        else:
-            raise RuntimeError("Unknown color format: '{}'".format(orig_color))
 
         for entry in view.palette:
             if entry[0] is None:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -319,7 +319,7 @@ def initial_data(logged_on_user):
             'name': 'Django',
             'push_notifications': False,
             'email_address': '',
-            'color': '#94c849',
+            'color': '#94c849',     # Color in '#xxxxxx' format
             'in_home_view': True
         }, {
             'audible_notifications': False,
@@ -333,7 +333,7 @@ def initial_data(logged_on_user):
             'name': 'GSoC',
             'push_notifications': False,
             'email_address': '',
-            'color': '#c2c2c2',
+            'color': '#ccc',        # Color in '#xxx' format
             'in_home_view': True
         }, {
             # This is a private stream;
@@ -640,9 +640,9 @@ def streams():
     `initial_data` fixture.
     """
     return [
-        ['Django', 86, '#94c849', False],
-        ['GSoC', 14, '#c2c2c2', False],
-        ['Secret stream', 99, '#c3c3c3', True],
+        ['Django', 86, '#9c4', False],
+        ['GSoC', 14, '#ccc', False],
+        ['Secret stream', 99, '#ccc', True],
     ]
 
 
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -2,6 +2,7 @@ import pytest
 
 from zulipterminal.helper import (
     index_messages,
+    canonicalize_color
 )
 from typing import Any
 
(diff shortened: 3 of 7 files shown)


alex-ozdemir:
Hmm. The tests seem to be choking on the mocking machinery. I'm unfamiliar with how it works. If anyone has time to take a look, I'd appreciate. You can run what I believe to be a minimal failing example with `pytest  -k test_narrow_to_topic`

neiljp:
@alex-ozdemir Thanks for looking into this. I originally worked on and merged a fix for this, but the server was modified such that this should not be an issue in current zulip server versions; we hoped the fix worked, but did not have a confirmation of this, I think. Have you reproduced this original (closed) issue (#273), and if so with which zulip server version? I assume you're running with the latest zulip-terminal from git?

I can see that the code you have looks like it should resolve such an issue, and if it works for you then I think we should consider merging it. However, at some point we should also consider combining the original fix and this one, to avoid multiple solutions. If you wish to pursue that you are welcome to do so.

alex-ozdemir:
Hey @neiljp. Thanks for looking at this. Yeah, I was able to reproduce the crash described in #273. Our server version is `2.0.0-1352-g46c31910e`. Perhaps this is one of the affected server versions?

I see your changes now, and I agree that we should do our two fixes in the same way. I'll put up another commit within 3 days which unifies our approaches.

neiljp:
@alex-ozdemir I look forward to your next iteration - if you'd like to drop by #**zulip-terminal** on chat.zulip.org then you can chat about this with us; my understanding is that this bug shouldn't be present on that server, so you should be able to use `zulip-term` if you want!

Re the server version, we should support this old color version in any case, so it is good to have your server version to test against for now!

zulipterminal/ui_tools/buttons.py
```diff
@@ -109,14 +110,7 @@ def __init__(self, properties: List[Any],
         caption, self.stream_id, orig_color, is_private = properties
 
         # Simplify the color from the original version & add to palette
-        # TODO Should this occur elsewhere and more intelligently?
-        if len(orig_color) == 7:  # modern zulip server format
-            color = ''.join(orig_color[i] for i in (0, 1, 3, 5))
-        elif len(orig_color) == 4:  # possible with zulip servers <=1.9.0 ?
```
neiljp: We should probably retain these comments, at least as a pointer as to why this is necessary, or which versions may be relevant.

zulipterminal/ui_tools/color.py
```diff
@@ -0,0 +1,13 @@
+from re import match, ASCII
```
amanagr: Thanks @alex-ozdemir for working on this! I don't think we need to create a new file here. Just moving it to helper.py should be fine.

zulipterminal/ui_tools/color.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def canonicalize_color(color: str) -> str:
```
amanagr: @alex-ozdemir I would also suggest adding test cases for it. You can look at the existing tests for getting an overview of how it should look like.

zulipterminal/ui_tools/buttons.py
alex-ozdemir: Agreed. I'll move them to `canonicalized_color`

alex-ozdemir:
Agreed -- doing canonicalization while creating the model is better.

alex-ozdemir:
Hey there,

Thanks to @amanagr  and @neiljp for your advice regarding the best way to implement this fix! Summarizing the current state of the PR:
* color canonicalization is implemented as a function in `helper.py`.
* canonicalization is done at model load-time.
* canonicalization has unit tests (with parameters taken from the prior render-time integration tests)
* the earlier render-time integration tests of canonicalization have been removed, since canonicalization now occurs at model load-time.
* the model load tests now include values which test canonicalization, to serve as integration tests for canonicalization.

One aspect I'm a bit concerned about is the new integration tests. Right now, the way that the model load tests test canonicalization is not immediately obvious. It would be nice to have a test set that tests color canonicalization at model load-time in a more targeted way. I'd appreciate any advice on how to write such tests.

neiljp:
@alex-ozdemir Thanks for working on this further - this will be a good bugfix :+1:

Could you perhaps squash some of the commits, at least partially (possibly parts of commits, too), to make clear logical commits? The commits you have tell the story of how you reached the result, but for example within this PR you have the canonicalization function in maybe three different files in different commits - whereas it just needs adding in one place/commit, when we add it from this PR :) If you're adding a bugfix for a previous commit, within the same PR, then after a certain stage of development we find it generally makes sense to correct it and have it in just one commit.

Also, I'm not sure how many commits you'll end up with, but it would be useful to have a commit title style closer to those you'll see in our commit history.

This will just make it easier to understand what has changed and for what reasons, and so make it easier to review :)

alex-ozdemir:
Of course. I've rebuilt two commits: 
1. one for creating/testing the core function, and
2. another for integrating it into the model and updating integration tests accordingly 

Such that each commit represents a change to the codebase and the tests that verify it.

neiljp:
@alex-ozdemir Thanks for your work on this :+1: I just rebased, solved the conflicts compared to some recent merges, and got this merged to master. I also moved to a standard docstring format and adjusted the commit text slightly.

The second commit is at 5801358b379d6ae86a6305b20e44f94081b11092, and the other just before.

We've merged some other improvements since you worked on this, so please let us know if you have any  other feedback or problems! As I said before, you're also welcome to join us on chat.zulip.org in #**zulip-terminal**.

alex-ozdemir:
Thanks for getting this across the finish line, @neiljp. I appreciate it, and am glad that my colors will work now :smile:! :rainbow:

PR #403, opened 2019-06-24
utils: Avoid displaying messages from unsubscribed streams.

sumanthvrao:
This is a bugfix to the issue reported at 
* https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Dealing.20with.20unsubscribed.20streams/near/754229
* https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/how.20to/near/736274

In this PR, we tackle this issue, by not displaying messages which belong to 'unsubscribed' streams. Further, we could also first build the stream_dict with data from all streams and then update it with more information from subscribed streams. This has some challenges (Eg:  there is no 'color' mentioned for unsubscribed streams as the API call for get_all_streams returns minimal information). This was partly the reason a simpler fix like this PR was introduced.

### DIFF
--- a/tests/ui/test_utils.py
+++ b/tests/ui/test_utils.py
@@ -86,12 +86,14 @@ def test_is_muted(mocker, msg, narrow, muted_streams, muted_topics, muted):
     assert return_value is muted
 
 
-@pytest.mark.parametrize('narrow, messages, focus_msg_id, muted, len_w_list', [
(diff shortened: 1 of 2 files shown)


amanagr:
Thanks @sumanthvrao this was a much needed fix.

neiljp:
@sumanthvrao Thanks! Just merged; another bug fixed :tada:

The next step would probably be to add stream add/remove events re subscriptions.

PR #404, opened 2019-06-24
show_help: Use double-lines around help to improve contrast.

neiljp:
These line characters don't take up extra space, but make the menu stand out better.


neiljp:
A version of the first commit was merged after discussion on #zulip-terminal.

PR #405, opened 2019-06-26
helpview: Introduce scroll-bar for scrolling through HelpView.

sumanthvrao:
This is more of a *food for thought* PR (hence the 'draft') in hopes that it may possibly be an option we may want to consider one day. This introduces the feature of having a scroll-bar in ListBox's whose purpose would be to 

* Aid navigation (UX wise).
* Give a visual sense of 'how much more' there is to scroll. 
* Possibly avoid the Up/Down in the title.

As it stands the feature isn't implemented yet in `urwid` and has been adapted from https://github.com/rndusr/stig/blob/master/stig/tui/scroll.py. 

A screenshot of this implemented on the HelpView, looks as follows:

scroll_bar

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -85,15 +85,11 @@ class Controller:
         self.loop.draw_screen()
 
     def show_help(self) -> None:
-        double_lines = dict(tlcorner='╔', tline='═', trcorner='╗',
-                            rline='║', lline='║',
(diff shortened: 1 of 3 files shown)


neiljp:
An alternative to this may be something like this https://github.com/AFoeee/additional_urwid_widgets/wiki/IndicativeListBox, though that does use space at the top and bottom.

PR #408, opened 2019-06-28
ZT redesign UI layout for stream and user search boxes

sumanthvrao:
This tweaks the UI layout my making th following adjustments
* Avoids top border line in user and stream search box
* Rewords the search message in MessageSearch Box to extend to `Search messages: `
* Extend the top-left and top-right borders of central message view to make them look more vertical
redesign_UI_layout

### DIFF
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -45,8 +45,8 @@ class View(urwid.WidgetWrap):
     def message_view(self) -> Any:
         self.middle_column = MiddleColumnView(self, self.model, self.write_box,
                                               self.search_box)
-        return urwid.LineBox(self.middle_column, bline="",
-                             title=u'Messages', tline=u'─')
+        return urwid.LineBox(self.middle_column, bline="", title=u'Messages',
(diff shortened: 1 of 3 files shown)


neiljp:
As discussed on the stream, other than seeing more consensus on whether this seems like a slightly cleaner look overall, I think we should also think about the colon (`:`) in the message pane - possibly remove it?

sumanthvrao:
Discussion continued https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Redesign.20Search.20bar.20.26.20msg.20box/near/761496

neiljp:
@sumanthvrao Thanks for getting this idea implemented :+1: As per the discussions on czo I've merged the line changes, but not the textual commit.

I'll close this, since I think we should discuss how to unify the search box designs, and possibly track that via an issue. That could involve updating how we handle the message search, to make the most of space.

neiljp:
#413 is a follow-up to this, which adds some thicker lines to emphasize each column title.

Issue #402, opened 2019-06-21
Improve quoted text rendering (mixing '>' and ```quote ?)


neiljp:
The following quoted text renders incorrectly (see 'quote zt FAIL' topic on #**test here** on chat.zulip.org).

@sumanthvrao Perhaps a followup PR for you at some point?

The text is below this line, but due to markdown-within-markdown limits, it doesn't appear to work that well.

------

```quote
```quote
>foo

bar

>foo2

bar2
```

baz
```

end

**Comment by sumanthvrao** [comment]
@zulipbot claim

**Comment by zulipbot** [comment]
Hello @sumanthvrao, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

**Comment by neiljp** [comment]
There is another issue with this, visible here:
https://chat.zulip.org/#narrow/stream/127-integrations/topic/Semaphore.20webhook.20broken/near/849794
(ZT only adds an indent to the last line of the quote, though does style it all appropriately)

**Comment by zee-bit** [comment]
> ZT only adds an indent to the last line of the quote

This is not necessarily true. It checks for the number of `child-tags` inside `blockquote`, and if its greater than 1, it skips the first line and indents the rest. Probably, because of this line in `indent_quoted_message()` :
```python
child_iterator = child_list[1:]
```
Example: https://chat.zulip.org/#narrow/stream/7-test-here/topic/quote.20zt.20FAIL/near/1127598

PR #409, opened 2019-06-28
test_ui_tools: Add failing test for nested blockquote.

sumanthvrao:
Address the failing bug with a test which is marked to fail(#402)

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1362,13 +1362,34 @@ class TestMessageBox:
             <p>C</p>""", "░ A\n░ B\n\nC"),
         ("""<blockquote>
                 <p><a href='https://chat.zulip.org/'</a>czo</p>
-            </blockquote>""", "░ [czo](https://chat.zulip.org/)\n")
+            </blockquote>""", "░ [czo](https://chat.zulip.org/)\n"),
+        pytest.param("""<blockquote>
+                            <blockquote>
+                                <p>A<br>
+                                B</p>
+                            </blockquote>
+                        </blockquote>
+            """, "░ ░ A\n░ ░ B",
+                     marks=pytest.mark.xfail(reason="rendered_bug")),
+        pytest.param("""<blockquote>
+                            <blockquote>
+                                <p>A</p>
+                            </blockquote>
+                            <p>B</p>
+                            <blockquote>
+                                <p>C</p>
+                            </blockquote>
+                        </blockquote>
+        """, "░ ░ A\n░ B\n░ ░ C",
+                     marks=pytest.mark.xfail(reason="rendered_bug")),
     ], ids=[
         "quoted level 1",
         "quoted level 2",
         "quoted level 3",
         "multi-line quoting",
-        "quoting with links"
+        "quoting with links",
+        "multi-line level 2",
+        "quoted level 2-1-2",
     ])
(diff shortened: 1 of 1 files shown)


neiljp:
@sumanthvrao Does this actually contain the source which triggers #402 yet? It looks like what we may have discussed privately, but not the extra case there?

sumanthvrao:
@neiljp Not really. But, fundamentally both this and the one you mentioned in the #test-here boil down to a common problem (bug) in the code wherein a `child` in the `child_list` does not get padded properly, if it itself is a blockquote.

tests/ui/test_ui_tools.py
```diff
@@ -1302,13 +1302,21 @@ def test_keypress_edit_message(self, mocker, message_fixture,
...
+                        </blockquote>
+            """, "░ ░ A\n░ ░ B", marks=pytest.mark.xfail(reason="nested bq"))
     ], ids=[
         "quoted level 1",
         "quoted level 2",
         "quoted level 3",
         "multi-line quoting",
-        "quoting with links"
+        "quoting with links",
+        "nested blockquote"
```
neiljp: Is this not more like "quoted multiline level 2"?  (though we could likely skip the 'quoted' in the ids)

tests/ui/test_ui_tools.py
```diff
@@ -1302,13 +1302,21 @@ def test_keypress_edit_message(self, mocker, message_fixture,
...
                 <p><a href='https://chat.zulip.org/'</a>czo</p>
-            </blockquote>""", "░ [czo](https://chat.zulip.org/)\n")
+            </blockquote>""", "░ [czo](https://chat.zulip.org/)\n"),
+        pytest.param("""<blockquote>
+                            <blockquote>
+                                <p>A<br>
+                                B</p>
+                            </blockquote>
+                        </blockquote>
+            """, "░ ░ A\n░ ░ B", marks=pytest.mark.xfail(reason="nested bq"))
```
neiljp: I would give the reason as more like "rendering bug", rather than the id?

tests/ui/test_ui_tools.py
```diff
@@ -1302,13 +1302,21 @@ def test_keypress_edit_message(self, mocker, message_fixture,
...
     ], ids=[
         "quoted level 1",
         "quoted level 2",
         "quoted level 3",
         "multi-line quoting",
-        "quoting with links"
```
neiljp: Trailing commas at the end of the last item means you don't get these diffs in future :)

tests/ui/test_ui_tools.py
sumanthvrao: Agreed. That is a better description

tests/ui/test_ui_tools.py
sumanthvrao: Done :+1:

tests/ui/test_ui_tools.py
sumanthvrao: Thanks for sharing this :+1: (I had never though of this before)

sumanthvrao:
@neiljp updated this with the correct failing tests. Will work on solving these as a follow-up.

neiljp:
@sumanthvrao Thanks for following-through with this tests PR after the quotes indent feature :+1: this will be a good pointer to solving #402 and ensuring we don't have a regression on the status we currently have :)

PR #413, opened 2019-06-30
ui/views: Set title lines in columns in bold

neiljp:
This is something of a follow-up to the recent UI changes in #T408 to keep the columns separated, where it minimized the lines and made the UI more columnar, but now perhaps became a bit less clear which elements were 'titles'.

This small commit emphasizes the Streams, Messages and Users titles, by using a different line around it.


sumanthvrao:
The thicker lines add an awesome touch and are more visually appealing! Great idea @neiljp :)

amanagr:
LGTM!

PR #414, opened 2019-07-02
writebox: Set msg_edit_id to None if user exits from the compose box.

amanagr:
```quote
Looks like there is a bug in the editing message feature we have. To repro :

send a message in a stream
Click e and add some text
Press 'ESC' before you send the message.
Press 'r' on the last message, which you wanted to edit
And type another message and send it
Observe that this message replaces (the original message) and acts as if you were editing the message.
```
Thanks @neiljp and @sumanthvrao for informing. This should fix it.


sumanthvrao:
This fixes the bug for me @amanagr! Looks good :+1:

neiljp:
@amanagr Thanks for the prompt bugfix! While this strictly doesn't lead to dataloss, it's definitely a bug!

Merged with adjusted text, as we discussed in PMs.

PR #418, opened 2019-07-05
Properly handle topic updates.

amanagr:


zulipterminal/model.py
```diff
@@ -714,6 +714,7 @@ def update_rendered_view(self, msg_id: int) -> None:
                 msg_pos = self.msg_list.log.index(msg_w)
                 self.msg_list.log[msg_pos] = new_msg_w
                 self.controller.update_screen()
+                return
```
neiljp: This first commit is mainly tidying/optimizing, I assume?

If we're adding this line, can we make use of this tidying commit to also merged the `else` clause above into this code?

This commit title seems a little unclear (update_message?)

tests/model/test_model.py
```diff
@@ -656,33 +656,35 @@ def test_append_message(self, mocker, user_profile, response,
...
+        }, 2, {
             'subject': 'new subject',
             'content': '<p>new content</p>',
         }),
-        ({}, {
+        ({}, 0, {
             'subject': 'old subject',
             'content': 'old content',
         }),
     ])
```
neiljp: Two of these tests fail after this second commit; perhaps add some ids while you're doing this, and it will make it clearer?

tests/model/test_model.py
```diff
@@ -656,71 +656,110 @@ def test_append_message(self, mocker, user_profile, response,
...
-            expected_message['content']
-        assert model.index['messages'][msg_id]['subject'] == \
-            expected_message['subject']
-        assert model.msg_list.log[0] == mock_msg
-        # Ensure the index shows the message has been updated (edited)
-        assert model.index['edited_messages'] == {msg_id}
+        assert model.index == new_index
         assert model.update_rendered_view.call_count == update_call_count
 
-        # TEST FOR FALSE CASES
```
neiljp: This should be in a separate test, rather than removed completely? eg. `test_update_message_where_message_not_in_index` ?

tests/model/test_model.py
```diff
@@ -656,71 +656,110 @@ def test_append_message(self, mocker, user_profile, response,
...
         assert model.msg_list.log == log
 
-    @pytest.mark.parametrize('response, update_call_count, expected_message', [
+    @pytest.mark.parametrize('response, update_call_count, new_index', [
         ({
             'subject': 'new subject',
             'message_ids': [1],
         }, 1, {
-            'content': 'old content',
+            'messages': {
```
neiljp: For the changes in this third commit, are you broadening the test to show how surrounding messages are affected?

amanagr: yes

tests/model/test_model.py
amanagr: The last commit checks for this case.

tests/model/test_model.py
amanagr: Done!

zulipterminal/model.py
```diff
@@ -708,6 +708,14 @@ def update_rendered_view(self, msg_id: int) -> None:
         for msg_w in self.msg_list.log:
             msg_box = msg_w.original_widget
             if msg_box.message['id'] == msg_id:
+                # Remove the message if it no longer belongs in the current
+                # narrow.
+                if len(self.narrow) == 2 and\
+                        msg_box.message['subject'] != self.narrow[1][1]:
+                    self.msg_list.log.remove(msg_w)
+                    self.controller.update_screen()
+                    return
```
sumanthvrao: should we `return` or `continue` here?

neiljp:
@amanagr I've not been able to test/review this fully again yet, but editing from within ZT does seem to update the headers really well now :tada: (albeit in a little jumpy fashion, perhaps?)

amanagr:
Yeah, it's a little jumpy, but I don't think we can do anything about it?

neiljp:
@amanagr Thanks for this :+1: Another bug fixed, and other behavior and test coverage is improved too :)

I've merged this mostly as-is, with a few extra test ids, and adjusting commit text slightly.

PR #419, opened 2019-07-05
Search box improvements

neiljp:
This is an attempt to work upon the remaining concept in #408, namely to improve the consistency between the search boxes.

Since each area (streams, messages, users) appears clearer with the 'bold headers' of #413, it seems more feasible to reduce text duplication in the UI and simplify the search text to the minimal form used in the message search box. This PR first extracts the common locations for that text, then:
* unifies the documentation/help to the title of the right panel (Users vs People)
* simplifies the text to just 'Search:' for the Streams and Users areas
* Adds shortcut key 'labels' to each search box, as suggested by @rishig (?)

The last change could be diverted to another PR with similar changes, but it's search-box related and easier with shorter search-box text, so feedback on that would be useful even if it's not merged as part of this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ autohide=autohide
 | Narrow to all starred messages                        | <kbd>f</kbd>                                  |
 | Next Unread Topic                                     | <kbd>n</kbd>                                  |
 | Next Unread private message                           | <kbd>p</kbd>                                  |
-| Search People                                         | <kbd>w</kbd>                                  |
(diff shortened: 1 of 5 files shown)


sumanthvrao:
@neiljp The changes look good!. I like the `[hotket]` for starting searches, as that's somethings I forget easily :+1: The clean-up commits also look good to go.

neiljp:
Thanks for the feedback everyone - merged :)

PR #421, opened 2019-07-07
Refactor: Improve set_count() behaviour.

sumanthvrao:
This PR makes the set_count behaviour more consistent. 

The commits are broken as:
* Rename a variable in set_count() function for better understanding
* Extracting and distinguishing (read vs unread) two functions for setting unread_counts in model. This commit also contains how we can now use the two functions in set_count() for setting stream, pm unreads.
* Improving the reading behaviours for muted messages (Where by we set there unread_counts as well). This commits also sets unread_counts of unread[topics] which we had not done previously.

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -69,6 +69,25 @@ def asynch(func: Any) -> Any:
     return wrapper
 
 
+def update_unread_count(key: Union[int, Tuple[int, str]],
+                        unreads: Dict[Any, int]) -> int:
+    if key in unreads:
+        unreads[key] += 1
+    else:
+        unreads[key] = 1
+    return unreads[key]
+
+
+def update_read_count(key: Union[int, Tuple[int, str]],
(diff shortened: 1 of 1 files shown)


sumanthvrao:
Rebased this ;)

neiljp (commented):
@sumanthvrao I think we discussed this indirectly, but if we're merging the model & button update loops then I think it's clearer if they're separate statements for each update, in each branch. Also, the second commit both refactors and changes behavior?

Perhaps try that `update_count_by` approach first?

neiljp (commented):
@sumanthvrao I'm not entirely convinced on the use of the extra functions (hence my earlier comment on keeping separate statements), but I think in any case it may be clearer if you refactor to merge the two loops together first, perhaps using a similar approach to that we have already (with the `if key in unreads` block?), then add missing functionality.

sumanthvrao:
Sure. Sounds reasonable :+1:

PR #423, opened 2019-07-09
Refactor extra captions to streams and topics

neiljp:
In reviewing early commits of #382 I was examining the use of 'caption' on the buttons/boxes.
This PR is an attempt to make this clearer, and lead in to the (now) early commit of #382.

I would particularly be interested in @amanagr reviewing this, to give insight into whether this is moving away from some original design or not - though it seems cleaner to me :)

I think it may be worth examining the objects/types we pass to the narrow methods; the duck-typing works fine, but at least renaming of the 'button' parameters to the narrow methods, would definitely help developers reading the code understand that boxes can be passed too (for example)?

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -53,7 +53,8 @@ class TestController:
         controller.model.muted_topics = []
         controller.narrow_to_stream(stream_button)
         assert controller.model.stream_id == stream_button.stream_id
-        assert controller.model.narrow == [['stream', stream_button.caption]]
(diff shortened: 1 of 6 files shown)


amanagr:
@neiljp this definitely make things more understandable. LGTM!

sumanthvrao:
@neiljp Looks good to me as well.

neiljp:
@amanagr @sumanthvrao Thanks for looking at this, and glad you agree :)

PR #424, opened 2019-07-09
Opens message links popup menu

shreyamalviya:
As of now, this does the following:
- Opens menu with links on pressing `v`
- Narrows to a stream (but does nothing if not subscribed to stream)
- Narrows to PM with a user (has an issue for users with whom no previous private conversation has taken place)

**_Final functionality includes:_**
_Pressing enter on one of the buttons in the links popup menu does one of the following, depending on the link:_
_* narrow to stream_
_* narrow to user_
_* open image in system's default viewer_
_* open link in system's default browser_

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -6,11 +6,16 @@ from re import match, ASCII
 from threading import Thread
 from typing import (
     Any, Dict, List, Set, Tuple, Optional, DefaultDict, FrozenSet, Union,
-    Iterable, Callable
+    Iterator, Iterable, Callable
 )
 from mypy_extensions import TypedDict
+from contextlib import contextmanager
 
 import os
+import sys
+import requests
+import subprocess
+import tempfile
 
 Message = Dict[str, Any]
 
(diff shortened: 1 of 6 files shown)


zulipterminal/helper.py
```diff
@@ -393,22 +394,51 @@ def powerset(iterable: Iterable[Any],
...
+        # Tries to download media and open in default viewer
+        # If can't, opens in browser
+        img_name = url.split("/")[-1]
+        img_dir_path = os.path.join(tempfile.gettempdir(), "ZulipTerminal_media")
+        img_path = os.path.join(img_dir_path, img_name)
+        try:
+            os.mkdir(img_dir_path)
+        except FileExistsError:
+            pass
+        with requests.get(url, stream=True) as r:
```
amanagr: How about:
```python
        with requests.get(url,
                          auth=requests.auth.HTTPBasicAuth(client.email,
                                                          client.api_key),
                          stream=True) as r:
```

where client is controller.client.

shreyamalviya: That works. Thanks a lot, @amanagr!

neiljp:
@shreyamalviya Thanks for your initial work on this :+1: I'm going to close this now since we have a menu structure in place and PRs open for the remaining features (open in web-browser #698, open downloadable media #740).

If there were other features which you believe were in this PR and are not integrated, in those PRs, we'd appreciate if you were able to file an issue or mention it briefly on #**zulip-terminal** - thanks!

PR #426, opened 2019-07-10
refactor: Improve naming of stream/topic/user variables in set_count.

sumanthvrao:
This is a follow-up for #8792c64 which makes the variable
names in 'set_count' more descriptive.

* Rename streams -> stream_buttons_log, with iterator stream_button.
* Rename users -> user_buttons_log, with iterator user_button.
* Rename topics -> topic_buttons_log, with iterator topic_button.

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -100,12 +100,12 @@ def set_count(id_list: List[int], controller: Any, new_count: int) -> None:
     while not hasattr(controller, 'view'):
         time.sleep(0.1)
 
-    streams = controller.view.stream_w.log
+    stream_buttons_log = controller.view.stream_w.log
     is_open_topic_view = controller.view.left_panel.is_in_topic_view
     if is_open_topic_view:
-        topics = controller.view.topic_w.log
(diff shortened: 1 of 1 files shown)


sumanthvrao:
@neiljp Does this follow-up look good?

sumanthvrao:
@amanagr What are your thoughts on renaming these variables in `set_count` function?

amanagr:
LGTM!

neiljp:
@sumanthvrao I just merged this with a slight commit text change - this is an essentially independent commit which stands alone in improving the descriptive nature, and I 'refactored' the text slightly :)

neiljp:
@sumanthvrao As a follow-up, I wonder about changing the `update_count` method to be `update_count_by` with a +/- value rather than an absolute value? That would simplify the code quite substantially? Do we need absolute values at any point? Could we hide the `.count` attribute?

sumanthvrao:
@neiljp That sounds like a good simplification. No, we do not need to absolute values at any point. We could this further in the stream.

PR #429, opened 2019-07-10
bugfix: Change fake propertly `caption` of divider to `stream_name`.

amanagr:
This stops stream_search from raising error if pinned streams are
present.

@neiljp FYI.


neiljp:
@amanagr Thanks for following up on this; I was focused on ensuring the transition worked in general in order to get #382 elements merged and missed ensuring this was updated in the sweep.

This is now merged with a small typo fix and a few comment/commit-text changes.

PR #431, opened 2019-07-11
bugfix: Make stream search case-insensitive.

amanagr:


neiljp:
@amanagr This was something that was bugging me - cool to have it :+1: I just merged with a few textual adjustments to the commit text.

As a minor extension we could consider searching 'inside' names? Another comparison may be to the search in the user list, where we use `casefold`?

neiljp:
@amanagr The other follow-up might be to make that cleaner with a list comprehension, rather than copying and removing certain elements?

Issue #425, opened 2019-07-10
Add button to indicate mentions (@theloggedinuser)


neiljp:
We have most of the webapp buttons available in the UI; it seems logical to add the mentions button too - we even have spare space for it, as it stands :)

sumanthvrao:
@zulipbot claim

zulipbot:
Hello @sumanthvrao, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

zulipbot:
Hello @sumanthvrao, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

neiljp:
This is complete, since the remainder of #433 (written over the Summer by @sumanthvrao) has now been reviewed and merged.

PR #433, opened 2019-07-16
Add button to view @ mentions.

sumanthvrao:
This PR adds support to view `@` mentioned messages. Any feedback regarding this is welcomed.

Solves #425

### DIFF
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -106,6 +106,18 @@ class PMButton(TopButton):
                          width=width)
 
 
+class MentionedButton(TopButton):
+    def __init__(self, controller: Any, width: int, count: int=0) -> None:
+        button_text = ("Mentions         [" +
+                       keys_for_command("ALL_MENTIONS").pop() +
+                       "]")
+        super().__init__(controller, button_text,
+                         controller.show_all_mentions,
+                         width=width,
+                         count=count,
+                         prefix_character='')
+
+
 class StarredButton(TopButton):
     def __init__(self, controller: Any, width: int) -> None:
         button_text = ("Starred messages [" +
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -11,8 +11,8 @@ from zulipterminal.config.keys import (
 from zulipterminal.helper import Message, asynch, match_user
 from zulipterminal.ui_tools.boxes import PanelSearchBox
 from zulipterminal.ui_tools.buttons import (
-    HomeButton, PMButton, StarredButton, StreamButton, TopicButton,
-    UnreadPMButton, UserButton,
+    HomeButton, MentionedButton, PMButton, StarredButton, StreamButton,
+    TopicButton, UnreadPMButton, UserButton,
 )
 from zulipterminal.ui_tools.utils import create_msg_box_list
 from zulipterminal.urwid_types import urwid_Size
@@ -661,6 +661,11 @@ class LeftColumnView(urwid.Pile):
                                        count=count,
                                        width=self.width)
 
+        self.view.mentioned_button = MentionedButton(
+            self.controller,
+            width=self.width,
+            count=self.model.unread_counts['all_mentions'])
+
         # Starred messages are by definition read already
         self.view.starred_button = StarredButton(self.controller,
                                                  width=self.width)
@@ -668,6 +673,7 @@ class LeftColumnView(urwid.Pile):
         menu_btn_list = [
             self.view.home_button,
             self.view.pm_button,
+            self.view.mentioned_button,
             self.view.starred_button,
         ]
         w = urwid.ListBox(urwid.SimpleFocusListWalker(menu_btn_list))
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -194,13 +194,15 @@ def test_powerset(iterable, map_func, expected_powerset):
         'all_msg': 9,
         'streams': {99: 1},
         'unread_topics': {(99, 'Some private unread topic'): 1},
+        'all_mentions': 0,
     }),
     ([1000], [['Secret stream', 'Some private unread topic']], {
         'all_msg': 11,
         'streams': {1000: 3},
         'unread_topics': {(1000, 'Some general unread topic'): 3},
+        'all_mentions': 0,
     }),
-    ([1], [], {})
+    ([1], [], {'all_mentions': 0})
 ], ids=['mute_private_stream_mute_general_stream_topic',
         'mute_general_stream_mute_private_stream_topic',
         'no_mute_some_other_stream_muted']
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -746,11 +746,15 @@ class TestModel:
           'display_recipient': [{'id': 5827}, {'id': 3212}]},
          [['pm_with', 'notification-bot@zulip.com']],
          frozenset({5827, 5}), []),
+        ({'type': 'stream', 'id': 1, 'stream_id': 1, 'subject': 'c',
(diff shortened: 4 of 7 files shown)


amanagr:
One thing I noticed is that if there are two message of the same narrow (say #zt>some topic), and if I narrow using  one of them via `S/z` from the `@mention` narrow to `some topic` and come back to `@mention` narrow and select the other msg of the same narrow (`some topic`) and press `z`, it's focus is still set on the previous message. This might come as a separate PR btw.

sumanthvrao:
@amanagr Not sure why, But I'm not able to reproduce this. How are you coming back to the `@mention` narrow? (By selecting it from the left sidebar?). 

My steps:
* Went to `@mention` narrow and clicked on `#checkin>sumanth` which has multiple mentioned messages. I clicked on the top and used `z`, which zoomed into Checkin stream
* I again went to `@mention` narrow choosing from side bar and did the same on the other message. Which took me to the correct focus.

Am I doing anything wrong?

amanagr:
I am also unable to reproduce it now :|

sumanthvrao:
@neiljp FYI. Updated this with tests. Its now ready for a review.

sumanthvrao:
@neiljp Resolved conflicts and rebased this. Do you have time to give it a review?

sumanthvrao:
This PR has be split up as :

* The first 3 commits adds the necessary support functions involving the new 'mentioned messages' narrow; These involve - create narrow, get id's in narrow and show all messages in the narrow. Tests have been added in the respective commits as well.
* The next 3 commits are UI related. They help register the new hotkey (`#`) , connect it via keypress and change the search bar text as well.
* The 7th and 8th commit - create the UI button for the 'mentioned messages' and connect it via the UI.
* The last commit classifies the unread messages so that the UI button can be marked with them.

I have separated the commits which involve the scenario when a mentioned message is received after ZT has started. I have kept these in a separate branch and will work on these shortly.

@neiljp @amanagr

zulipterminal/config/keys.py
```diff
@@ -140,6 +140,11 @@
         'help_text': 'Narrow to all starred messages',
         'key_category': 'navigation',
     }),
+    ('ALL_MENTIONS', {
+        'keys': {'#'},
+        'help_text': 'Narrow to all mentioned messages',
```
neiljp: Sync this with the README, of course :)

zulipterminal/ui.py
```diff
@@ -164,6 +164,9 @@ def keypress(self, size: Tuple[int, int], key: str) -> Optional[str]:
         elif is_command_key('ALL_STARRED', key):
             self.model.controller.show_all_starred(self)
             self.body.focus_col = 1
+        elif is_command_key('ALL_MENTIONS', key):
+            self.model.controller.show_all_mentions(self)
+            self.body.focus_col = 1
```
neiljp: Perhaps a little minor, but perhaps squash this and the README/keys commit, to avoid a potential state where the docs/help don't match the UI (even if the visual button doesn't exist, the key does work!)

zulipterminal/ui_tools/buttons.py
```diff
@@ -118,6 +118,18 @@ def __init__(self, controller: Any, width: int) -> None:
...
+class MentionedButton(TopButton):
+    def __init__(self, controller: Any, width: int, count: int=0) -> None:
+        button_text = ("Mentions [" +
+                       keys_for_command("ALL_MENTIONS").pop() +
+                       "]")
+        super().__init__(controller, button_text,
+                         controller.show_all_mentions,
+                         width=width,
+                         count=count,
+                         prefix_character='')
```
neiljp: There's no test for this, but this seems reasonable for now given the simplicity :)

zulipterminal/ui_tools/views.py
```diff
@@ -665,10 +665,16 @@ def menu_view(self) -> Any:
...
+        self.view.mentioned_button = MentionedButton(
+            self.controller,
+            width=self.width,
+            count=self.model.unread_counts['all_mentions'])
+
         menu_btn_list = [
             self.view.home_button,
             self.view.pm_button,
             self.view.starred_button,
+            self.view.mentioned_button,
```
neiljp: Perhaps we should switch the order here, to match the webapp, at least for now?
(the order of the `MentionedButton` code relative to the `StarredButton` too, if so)

zulipterminal/ui.py
sumanthvrao: Good catch!

zulipterminal/ui_tools/views.py
sumanthvrao: Sure. I changed the order of the buttons and code in respective places.

sumanthvrao:
@neiljp Thanks for the review. I have updated the PR with the changes requested.

README.md
```diff
@@ -116,6 +116,7 @@ Install-Module -Name BurntToast
 | Go to the last message                                | <kbd>G</kbd> / <kbd>end</kbd>                 |
 | Narrow to all private messages                        | <kbd>P</kbd>                                  |
 | Narrow to all starred messages                        | <kbd>f</kbd>                                  |
+| Narrow to all mentioned mentions                      | <kbd>#</kbd>                                  |
```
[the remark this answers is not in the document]
neiljp: I think we lost the `Narrow to` prefix in the latest version; did you mean to?

zulipterminal/core.py
```diff
@@ -233,6 +233,12 @@ def show_all_starred(self, button: Any) -> None:
                         anchor=None,
                         starred=True)
 
+    def show_all_mentions(self, button: Any) -> None:
+        # NOTE: Should we ensure we maintain anchor focus here?
```
neiljp: It *seems* to maintain focus so far without? However as with 'starred', perhaps add a note to indicate that, as we may benefit from it later?

zulipterminal/ui_tools/buttons.py
```diff
@@ -106,6 +106,18 @@ def __init__(self, controller: Any, width: int, count: int=0) -> None:
                          width=width)
 
 
+class MentionedButton(TopButton):
+    def __init__(self, controller: Any, width: int, count: int=0) -> None:
+        button_text = ("Mentions [" +
```
neiljp: Maybe we should adjust the spacing to look a little better, hard-coded for now?

zulipterminal/core.py
sumanthvrao: Added 👍

zulipterminal/ui_tools/buttons.py
sumanthvrao: Anything particular in mind?

sumanthvrao:
@neiljp What do you want the README/help text to say. I think `Messages in which you're mentioned` seems good.

The new commit message was because I squashed 2 commits from the previous version. I have updated the commit heading to be the first one of the squashed  commits.

I didn't exactly get what changes you wanted wrt button spacing. 

Besides this, if there are no more changes, let's get it merged ASAP :)

neiljp:
@sumanthvrao The first 5 are merged; now onto more UI :) :tada: 

I think the other work is updating while ZT is running, and possibly the all-mentions aspect?

sumanthvrao:
Thanks @neiljp !

sumanthvrao:
Thanks for the review @kaustubh-nair! Glad to see there are no errors. :)

neiljp:
@sumanthvrao I just merged the rest of these commits manually, after slightly adjusting the last one to fit with the recent changes to `set_count` - working well so far, thanks! :tada:

Some commits are quite small, but we put in a lot of related effort before this - though some extra tests may be warranted?

preetmishra:
@sumanthvrao Awesome work! This looks great. :+1:

sumanthvrao:
@neiljp Thanks for taking time to review this and I'm glag to finally get it merged. 

Thanks @preetmishra :)

PR #439, opened 2019-07-21
regression: views: Fix PopUp width for confirmation box.

sumanthvrao:
With 3426854d11857f46173fc9f9f9c3a98bb58b9119 we adjusted the Left panel width but the
PopUpConfimation box width was fixed width. With this
we now set the popup width based on the Left Panel width
set in UI.

Tests amended.


neiljp:
@sumanthvrao This returns the appearance to how it was, so I've merged it now :+1: I'm not sure we clarified that the overlap was guaranteed to match the left panel, but I agree this is a logical change to fix it to be this way for now. I've marked this bugfix rather than regression due to this (and the commit doesn't *cause* a regression) and improved the consistent use of the widget name.

PR #440, opened 2019-07-21
buttons: Improve logic for marking button with 'M'.

sumanthvrao:
This is a different approach towards marking streams as muted 'M'; We don't need to have the `-1` approach now and have a separate method in `TopButton` to do this. Further, we handle the logic of converting unread_count -> text in their respective functions.

Wanted feedback on this since before adjusting the tests to support the same.

### DIFF
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -43,9 +43,7 @@ class TopButton(urwid.Button):
 
     def update_count(self, count: int) -> None:
         self.count = count
-        if count < 0:
-            count_text = 'M'
-        elif count == 0:
+        if count == 0:
             count_text = ''
         else:
             count_text = str(count)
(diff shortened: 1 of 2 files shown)


sumanthvrao:
@neiljp @amanagr Thoughts on this?

zulipterminal/ui_tools/buttons.py
```diff
@@ -37,22 +37,25 @@ def __init__(self, controller: Any, caption: str,
...
             count_text = ''
         else:
             count_text = str(count)
+        self._w = self.widget(count_text)
+
+    def mark_button_muted(self) -> None:
+        # Ensure self.count has the correct unread counts
+        # by not updating it.
+        count_text = 'M'
+        self._w = self.widget(count_text)
```
neiljp: I wonder if it's worth introducing this method, since we already have `mark_muted` in the `StreamButton` (and removed it from `TopButton` as other buttons don't generally mute)? Perhaps override `update_count` instead, and call 'widget' where appropriate? (I'd be open to making that method more descriptive)

sumanthvrao: Sure. I realized that a similar approach you suggested would be better for this.

neiljp:
@sumanthvrao Merged; glad to have this refactor :+1:

Issue #224, opened 2018-12-28
Implement autocomplete for streams in the message box.


amanagr:
This issue is paired with #222 and should be straighforward to implement once #222 is implemented.

PR #343, opened 2019-04-09
Autocomplete for @mention 

omi10859:
issue: https://github.com/zulip/zulip-terminal/issues/222

PR #441, opened 2019-07-23
Autocomplete @mentions and #stream with keypress.

sumanthvrao:
This PR builds in the direction of #343 (credits goes to @omi10859) and adds support for @mentions and #stream names. 

* It uses `ctrl + f` keypress (same as windows CML hotkey) for autocompleting user names and stream names. 
* The version upgrade of readlines has been placed as first commit.
* To ensure consistency a small helper function `match_stream` is added. Further, this handles the case-insenstive comparison.
* Last two commits add support for autocompleting #streams and @mentions. 

Fixes: #224, #222


sumanthvrao:
@neiljp @amanagr Tests are yet to be added. But feedback is welcome.

amanagr:
I get this error when I press `r` on a message:
```quote
  File "/home/aman/zulip-terminal/zulipterminal/ui_tools/boxes.py", line 633, in keypress
    title=self.message['subject']
  File "/home/aman/zulip-terminal/zulipterminal/ui_tools/boxes.py", line 62, in stream_box_view
    key=next(iter(keys_for_command('AUTOCOMPLETE'))))
TypeError: enable_autocomplete() got an unexpected keyword argument 'key'
```

sumanthvrao:
@amanagr Do you have the latest version of `urwid_readlines` (0.9.0)?

amanagr:
oh! So, that's the issue!

sumanthvrao:
@amanagr Yup. The earlier version of `urwid_readlines` used to only take the callback as the parameter. :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -78,6 +85,22 @@ def stream_box_view(self, caption: str='', title: str='') -> None:
...
 
+    def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
+        if text.startswith('#'):
+            return self.autocomplete_streams(text, state)
+        else:
+            return text
+
+    def autocomplete_streams(self, text: str, state: int) -> Optional[str]:
+        streams_list = self.view.pinned_streams + self.view.unpinned_streams
+        stream_typeahead = ['#**%s**' % stream[0]
```
amanagr: We should use `.format` here.

zulipterminal/ui_tools/boxes.py
```diff
@@ -101,6 +103,16 @@ def autocomplete_streams(self, text: str, state: int) -> Optional[str]:
         except IndexError:
             return None
 
+    def autocomplete_mentions(self, text: str, state: int) -> Optional[str]:
```
amanagr: can we add support `@"aman agrawal"` then we press `Ctrl f` to autocomplete it as `@**Aman Agrawal** `?

amanagr:
Do we have something to go back to the previous autocomplete?

sumanthvrao:
> Do we have something to go back to the previous autocomplete?

To clarify, what do you mean by this exactly?

zulipterminal/ui_tools/boxes.py
sumanthvrao: Sounds great!

zulipterminal/ui_tools/boxes.py
sumanthvrao: Sure. Will add it :+1:

amanagr:
Like in a list of autocomplete -> ['a', 'b', 'c'], if we are at `b`, we should have a key for going to `a`.

sumanthvrao:
I hadn't thought of this. Sounds like a reasonable thing to do. I don't think `urwid_readlines` supports this at the moment, but It should be possible to implement from our end?

amanagr:
While you are at it, @neiljp and I have been thinking of moving urwid_readline inside zt; basically subclassing it for now and using it, since that would be easier for us to manage in the future & it's a lot quicker. I think you should just subclasss it and start developing. @neiljp thoughts?

zulipterminal/ui_tools/boxes.py
```diff
@@ -78,6 +87,34 @@ def stream_box_view(self, caption: str='', title: str='') -> None:
...
+        if text.startswith('@'):
+            return self.autocomplete_mentions(text, state)
+        elif text.startswith('#'):
+            return self.autocomplete_streams(text, state)
+        else:
+            return text
+
+    def autocomplete_mentions(self, text: str, state: int) -> Optional[str]:
+        users_list = self.view.users
+        user_typeahead = ['@**%s**' % user['full_name']
```
aero31aero: We also need to have an additional check here. If you have two users with the same name, the syntax we use is: `@**Name|userid**`. I wonder how we could differentiate them on the terminal though (maybe showing email?).

sumanthvrao: Great! Thanks @aero31aero .

sumanthvrao:
@neiljp 
Here's a changelog from the previous iteration. This serves as a v1 of typeaheads and I will make further follow-ups addressing more points which Aman and Rohit have mentioned.

* Moved the 'match_stream' commit into the commit which adds support for autocomplete #streanames 
* I have managed to split the README and keep more of the original commit which omi wrote.
* Silent mentions have been taken care of and tested.
* Tests have been added for all the 3 types - user mentions, user silent mentions and stream mentions.

zulipterminal/config/keys.py
```diff
@@ -198,7 +198,7 @@
     }),
     ('AUTOCOMPLETE', {
         'keys': {'ctrl f'},
-        'help_text': 'Autocomplete @mentions',
+        'help_text': 'Autocomplete @mentions and #stream_names',
```
neiljp: You could include this file again in the commit title in this later commit, or at least mention the keys/README change in the summary.

tests/ui_tools/test_boxes.py
```diff
@@ -60,3 +60,27 @@ def test_autocomplete_mentions(self, write_box, users_fixture,
...
+        ('#St', 0, '#**Stream 1**'),
+        ('#St', 1, '#**Stream 2**'),
+        ('#Stream 1', 0, '#**Stream 1**'),
+        ('No match', 0, None)
+    ])
+    def test_autocomplete_streams(self, write_box, streams_fixture,
+                                  text, state, required_typeahead):
+        write_box.view.pinned_streams = [[stream['name']]
+                                         for stream in streams_fixture]
+        write_box.view.unpinned_streams = []
```
neiljp: The unpinned streams being empty feels a little suspicious. I know it would affect the search order, but perhaps just split them up into the two types for now, eg. Stream 1/2 into unpinned, and that would mean the order would be the same? Or even put Stream 1/2 into pinned, to show how they come first?

@amanagr This makes me wonder if we should encode some streams as pinned by default in the streams fixture(s)?

zulipterminal/ui_tools/boxes.py
```diff
@@ -87,19 +87,30 @@ def stream_box_view(self, caption: str='', title: str='') -> None:
...
             return text
 
+    def autocomplete_mentions(self, text: str, state: int,
+                              index: int) -> Optional[str]:
+        # Handles both mentions and silent mentions.
+        users_list = self.view.users
+        if index == 1:
+            prefix_string = '@'
+        else:
+            prefix_string = '@_'
```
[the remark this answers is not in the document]
sumanthvrao: Agreed :+1:

zulipterminal/helper.py
```diff
@@ -385,6 +385,17 @@ def match_user(user: Any, text: str) -> bool:
     return False
 
 
+def match_stream(stream: Any, text: str) -> bool:
```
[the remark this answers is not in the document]
sumanthvrao: Moved this code to a new commmit. May help simplify 444 as well.

tests/ui_tools/test_boxes.py
sumanthvrao: Done. :+1:

sumanthvrao:
@neiljp Pushed a new iteration which address the above review feedback you gave. Thanks for it :+1:

neiljp:
@sumanthvrao Thanks for integrating @omi10859's work into this new PR :tada: I've merged the code as-is but excluded the last commit, which I think we can discuss ideas about separately.

The commit titles were generally fine already, but I adjusted the descriptive text a little, amended a typo and renamed the 'wrapper' title to represent that this commit primarily adds silent mention support. I've left the match_streams commit separate, though I think due to the way the stream name is accessed we would need to adapt it to reuse it with stream search.

amanagr:
```quote
@amanagr I'm skeptical about the speech mark version; do we also add single quotes? underscores/hyphens to separate words? The webapp doesn't accept this either.
```
I think this would be unnecessary once we have all the possible suggestions displayed in footer.

PR #444, opened 2019-07-29
Refactor locking & search filters

neiljp:
This PR comprises two changes to improve search code, that I first saw some time ago:
* Simplifying locking using contextmanagers
* Tidying up the searching code, which used a lot of copying, using list comprehensions

The first needs a little change to the tests; the latter looks cleaner to me, but also seems a lot faster, at least on chat.zulip.org - possibly due to the number of users.

The improvement in search speed exposes a small bug - which I don't believe is caused by these changes.

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -270,15 +270,14 @@ class StreamsView(urwid.Frame):
             return
         # wait for any previously started search to finish to avoid
         # displaying wrong stream list.
-        self.search_lock.acquire()
-        streams_display = self.streams_btn_list.copy()
-        for stream in self.streams_btn_list:
-            if not stream.stream_name.lower().startswith(new_text.lower()):
-                streams_display.remove(stream)
-        self.log.clear()
-        self.log.extend(streams_display)
-        self.view.controller.update_screen()
-        self.search_lock.release()
+        with self.search_lock:
+            streams_display = self.streams_btn_list.copy()
+            for stream in self.streams_btn_list:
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -271,10 +271,11 @@ def update_streams(self, search_box: Any, new_text: str) -> None:
         # wait for any previously started search to finish to avoid
         # displaying wrong stream list.
         with self.search_lock:
-            streams_display = self.streams_btn_list.copy()
-            for stream in self.streams_btn_list:
-                if not stream.stream_name.lower().startswith(new_text.lower()):
-                    streams_display.remove(stream)
+            streams_display = [
+                stream
+                for stream in self.streams_btn_list.copy()
```
sumanthvrao: As you mentioned the `match_stream` helper function could be used here. In fact, the `match_stream` logic may even benefit stream search.

zulipterminal/ui_tools/views.py
```diff
@@ -536,11 +537,12 @@ def update_user_list(self, search_box: Any=None,
...
-            users_display = users.copy()
             if new_text:
-                for user in users:
-                    if not match_user(user, new_text):
-                        users_display.remove(user)
+                users_display = [
+                    user for user in users if match_user(user, new_text)
+                ]
+            else:
+                users_display = users
```
sumanthvrao: Would `users_to_display` be a slightly better name for this?

neiljp: I was focusing on adjusting the logic in the first instance.

PR #446, opened 2019-07-31
Ensure all-messages narrow is not empty via is-private/unmuted-streams

### LINKED ISSUE #326: Last 30 messages being muted causes crash when starting up ZT

neiljp:
This should resolve #326; the approach is somewhat WIP but is functional and could be merged once tests are updated/added.

The approach taken is as follows:
* extend fetching messages to be from some narrow only (optionally)
* keep track of (initial) unmuted streams
* if there are no unmuted messages at startup, fetch more.

This is more deterministic than the original approach discussed and proposed in #341.

@sumanthvrao This was the alternative approach I thought of and probably discussed with you.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -114,12 +114,48 @@ class Model:
          self.pinned_streams, self.unpinned_streams) = stream_data
 
         self.muted_topics = self.initial_data['muted_topics']
+
+        # Fetch additional messages to ensure non-empty all-messages narrow
+        # (caused by muted streams)
+        messages_not_muted = [
+            message
+            for message in self.index['messages'].values()
+            if (message['type'] == 'stream' and
+                message['stream_id'] not in self.muted_streams and
+                ([message['display_recipient'], message['subject']]
+                 not in self.muted_topics))
+        ]
+        if len(messages_not_muted) == 0:
+            self.__fetch_additional_initial_messages(workers=1)
+
         self.unread_counts = classify_unread_counts(self)
 
         self.fetch_all_topics(workers=5)
 
         self.new_user_input = True
         self._start_presence_updates()
+
+    def __fetch_additional_initial_messages(self, *, workers: int) -> None:
+        assert workers > 0
+        unmuted_stream_narrows = [
+            [['stream', self.stream_dict[stream_id]['name']]]
+            for stream_id in self.initial_unmuted_streams
+        ]
+        with ThreadPoolExecutor(max_workers=workers) as executor:
+            futures = {
+                executor.submit(self.get_messages,
+                                num_after=10,
+                                num_before=30,
+                                anchor=None,
+                                narrow=narrow)
+                for narrow in unmuted_stream_narrows + [[['is', 'private']]]
+            }
+
+            # Wait for threads to complete
+            wait(futures)
+
+        if not all(self.exception_safe_result(future) for future in futures):
+            raise ServerConnectionFailure("fetching_additional_initial_messages")
 
     def get_focus_in_current_narrow(self) -> Union[int, Set[None]]:
         """
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -311,10 +311,13 @@ class Model:
 
     def get_messages(self, *,
                      num_after: int, num_before: int,
-                     anchor: Optional[int]) -> bool:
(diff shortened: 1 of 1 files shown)


zulipterminal/model.py
```diff
@@ -110,17 +110,53 @@ def __init__(self, controller: Any) -> None:
...
         self._start_presence_updates()
 
+    def __fetch_additional_initial_messages(self, *, workers: int) -> None:
+        assert workers > 0
+        unmuted_stream_narrows = [
+            [['stream', self.stream_dict[stream_id]['name']]]
+            for stream_id in self.initial_unmuted_streams
+        ]
+        with ThreadPoolExecutor(max_workers=workers) as executor:
+            futures = {
```
amanagr: @neiljp did you profile this? I think fetching messages for all the unmuted_stream_narrows will take a lot of time. (Assume some realm has like 100 unmuted streams)

neiljp:
I wrote the following, and I think it could be useful, but there *may* be an API-based solution which would make this easier, though Tim's comment on #326 originally pushed me away from that idea.

-------

@amanagr I am totally open to discussion on this, and I think it could be a case of working on various iterations, as some would 'fix' the crash, but wouldn't give a full solution.

My understanding is that the current situation is that we attempt to fetch 30 messages initially, but if all these messages are muted (in muted streams/topics) then we end up with no messages in the all-messages narrow and can crash. If we take the approach of wanting to show at least one message (assuming one is present!) then we need to do at least one additional initial fetch (the alternative being just showing an 'empty' narrow as per the almost ready #278 from @sumanthvrao - for some time!).

Fetching from the all messages narrow would give *recent* messages, but depending on where they are (maybe lots are muted?) then the same situation holds, and the client could keep fetching messages back in time, over and over again, and maybe just keep finding muted messages. This is what concerned me about #326.

The alternative is to be selective on a per-narrow basis; the obvious solution is PMs, which are guaranteed not to be muted, but while this should avoid a crash, doing this alone doesn't seem representative of showing all messages in time order - given that there could be recent un-muted stream messages not shown in that interleaved view if we just fetched PMs. This is why I added the fetching from un-muted streams.

These unmuted streams are only those subscribed, not all those on the server, as I understand it. While people could be subscribed to a lot, this implementation is currently only triggered if people get a lot of messages in muted streams. That is an obvious use-case of muting, but depends on the balance of messages people receive in muted vs non-muted streams.

One additional factor is that we could exclude unpinned streams from the secondary fetch, though I'm not a fan of that approach either, as while they're "less important", we still show content from them, unliked muted.

PR #447, opened 2019-08-01
gitlint: Allow certain filenames to be capitalized in commit title.

sumanthvrao:
This is as-per discussions here (https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/edit.20gitlint.20rules/near/775070)

* We relax the rule for capitalization for words such as (README, CHANGELOG, LICENSE) so that gitlint doesn't throw errors when these are used in their capital forms in the commit titles.
* We add the above words to the exclusion list and check if any captial word occuring in the title does not belong to this list. (In which case we raise an error)

### DIFF
--- a/tools/gitlint-extra-rules.py
+++ b/tools/gitlint-extra-rules.py
@@ -39,8 +39,17 @@ class AreaFormatting(CommitRule):
             exclusions_text = " (or {})".format(exclusions_text)
         error = ("Areas at start of title should be lower case{}, "
                  "followed by ': '".format(exclusions_text))
+
+        def deny_capital_text(text: str) -> bool:
+            if text in exclusions:
+                return False
+            if not text.islower():
+                return True
+            return False
+
         for area in title_components[:-1]:
-            if not (area.islower() or area in exclusions) or ' ' in area:
+            if (any(deny_capital_text(word) for word in area.split('/')) or
+                    ' ' in area):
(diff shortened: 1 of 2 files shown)


sumanthvrao:
@neiljp FYI. Fairly quick one to review, if its in the right path.

tools/gitlint-extra-rules.py
```diff
@@ -40,7 +40,8 @@ def validate(self, commit: Any) -> Optional[List[RuleViolation]]:
         error = ("Areas at start of title should be lower case{}, "
                  "followed by ': '".format(exclusions_text))
         for area in title_components[:-1]:
-            if not (area.islower() or area in exclusions) or ' ' in area:
+            if (any(word not in exclusions for word in area.split('/')
+                    if word.isupper()) or ' ' in area):
```
neiljp: This logic isn't quite right, as it lets through eg. "Foo/keys/Readme: Something."

sumanthvrao: @neiljp Updated

neiljp:
@sumanthvrao Thanks, this is useful! I just merged this with an extended commit title/summary, to represent that this doesn't just extend the exclusions list, but also checks on a per-word (within an area) basis - with is a separate bugfix/feature of note :tada:

PR #449, opened 2019-08-06
core: Improve consistency of narrowing behavior.

neiljp:
This introduces a _narrow_to method, which takes the button and a kwargs parameter which is used to specify the narrow to switch to. I wanted to move towards a change like this some time ago, but the methods were too different/complex at the time.

This reduces a *lot* of duplication (+16/-92), and the narrow code is so much cleaner!

This seems to make narrow-changing behave a little better, but could still be improved further. While I thought initially that this might fix 311, there still appear to be issues around this.

@amanagr Given the original differences between the implementation in some of these methods and that you wrote these originally, I would particularly appreciate any feedback regarding this change, as it is not simply a refactoring.

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -142,14 +142,17 @@ class Controller:
         self.loop.widget = PopUpConfirmationView(self, question,
                                                  mute_this_stream)
 
-    def narrow_to_stream(self, button: Any) -> None:
-        already_narrowed = self.model.set_narrow(stream=button.stream_name)
+    def _narrow_to(self, button: Any, **narrow: Any) -> None:
+        already_narrowed = self.model.set_narrow(**narrow)
         if already_narrowed:
             return
 
         self.model.found_newest = False
+
         # store the steam id in the model (required for get_message_ids...)
-        self.model.stream_id = button.stream_id
+        if hasattr(button, 'stream_id'):  # FIXME Include in set_narrow?
+            self.model.stream_id = button.stream_id
+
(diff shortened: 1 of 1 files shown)


zulipterminal/core.py
```diff
@@ -142,14 +142,17 @@ def stream_muting_confirmation_popup(self, button: Any) -> None:
...
+    def _narrow_to(self, button: Any, **narrow: Any) -> None:
+        already_narrowed = self.model.set_narrow(**narrow)
         if already_narrowed:
             return
 
         self.model.found_newest = False
+
         # store the steam id in the model (required for get_message_ids...)
-        self.model.stream_id = button.stream_id
+        if hasattr(button, 'stream_id'):  # FIXME Include in set_narrow?
```
amanagr: I think it's okay, since we will have to set it in two functions anyway.

neiljp: I left this note for now, as it honestly feels like if we set a narrow then we should be able to do this there, rather than handling it separately.

neiljp:
@amanagr @sumanthvrao I was very close to merging this, but would like to hold off slightly until we discuss further how we expect narrowing-while-focused (on a message) to behave, to ensure this change doesn't mask unintended behavior - ie. that later we end up needing to make separate narrowing methods.

I've introduced this in #z-t on czo in 'Keep context on narrow?'.

neiljp:
Also 'Focus on narrowing' topic in #**design**.

PR #454, opened 2019-08-08
requirements: Upgrade urwid-readline from 0.9 to 0.10.

sumanthvrao:
This upgrade resolves the issue filed at [rr-/urwid_readline#10](https://github.com/rr-/urwid_readline/issues/10) which makes autocomplete work when there are 2/more newlines. 

Thanks to folks upstream for the quick fix.


neiljp:
LGTM @sumanthvrao :+1: Great to see this fixed so quickly :)

Merged with a reference to it fixing a bug and reference to the upstream issue.

PR #455, opened 2019-08-09
requirements: Upgrade pytest to 5.0.1.

amanagr:
This adds many new features to pytest.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -13,7 +13,7 @@ lxml = "==4.2.3"
 mypy_extensions = ">=0.4"
 
 [dev-packages]
-pytest = "==3.4.2"
+pytest = "==5.0.1"
 pytest-pep8 = "==1.0.6"
 pytest-mock = "==1.7.1"
 pytest-cov = "==2.5.1"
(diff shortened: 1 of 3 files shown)


amanagr:
The warnings seem very strange. We will possibly will have to filter them.

neiljp:
I agree we may have to, but there may be alternative options to set (for now) or it could be that things are actually being flagged accurately. The last update of `pytest-pep8` seems some time ago now, which was what made me wonder if we should use zulint or pycodestyle directly.

amanagr:
Agreed, I will remove pytest-pep8 then and add pycodestyle with zulint. I will filter the other warnings.

sumanthvrao:
@amanagr I don't know much about zulint. Would this bring any difference in style compared to how we have stuff currently?

amanagr:
We can customize it to use whatever style we want. :)

sumanthvrao:
Sounds cool! :)

neiljp:
@amanagr See #485. Once that is merged, or before, we can open an issue regarding improving linting, perhaps.

neiljp:
This is superseded by the now-merged #485, as discussed above, so closing now.

PR #460, opened 2019-08-15
views: Avoid wrapping users_btn_list in SimpleFocusListWalker twice.

sumanthvrao:
Class `UsersView` wraps up incoming `users_btn_list` in a
SimpleFocusListWalker object. It is not necessary for the
incoming data to be wrapped in the ListWalker as well.

Tests amended to support changes.

Tested:
Previous behaviour of
* User search
* Updating user list at 60-second intervals 

are intact.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -871,11 +871,8 @@ class TestRightColumnView:
         }]
         self.view.controller.editor_mode = editor_mode
         user_btn = mocker.patch(VIEWS + ".UserButton")
-        mocker.patch(VIEWS + ".UsersView")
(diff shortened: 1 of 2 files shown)


sumanthvrao:
@neiljp Testing proves there is no behaviour change. But I'd like to hear your thoughts on this.

neiljp:
@amanagr This looks reasonable to me - any thoughts on specific requirements for this to remain?

sumanthvrao:
@amanagr Any thoughts on this?

neiljp:
@sumanthvrao Merging this now, as it's small and seems good :+1: 

@amanagr Please review if you get a chance, retroactively :)

PR #462, opened 2019-08-16
core: Avoid writing newline to stdout after UI has loaded.

### LINKED ISSUE #456: Handle distorted UI on ZT startup - Race bug
Sometimes on ZT startup, the UI tends to be not fully loaded. These cause gaps (dark patches) in certain parts of the UI like users/stream lists, etc. 

A temporary fix for this can be to REDRAW the UI at the end of the startup. However, this is not the correct fix. We must identify why this is happening and fix the race.

pasted_image

sumanthvrao:
Possibly fixes #456.

@amanagr FYI :)

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -59,7 +59,6 @@ class Controller:
             time.sleep(0.1)
             sys.stdout.write('\b'*len(next_spinner))
 
-        sys.stdout.write('\n')
         self.capture_stdout()
 
     def capture_stdout(self, path: str='debug.log') -> None:


amanagr:
I think that line was there so that errors get printed nicely in a new line; but now, they seem to be printing nicely now even without that line. LGTM!

neiljp:
@sumanthvrao Good catch! I've a test account that this reliably triggered with, and this definitely seems like it works for that account, as well as being a good fix from a code point of view :+1: The output in case of an error doesn't seem affected either, I agree - thanks for that feedback @amanagr :+1:

I'm merging this now; let's see if we observe any more cases of #456.

Issue #194, opened 2018-12-15
Add/modify specific test(s) to expect a Divider may be in the stream list


neiljp:
After adding pinned-messages separated by a `urwid.Divider()`, a few bugs appeared; the Divider is masquerading as a StreamButton, but that should be taken into account, and ideally the tests should have failed when I added the Divider. Based on that, we need more/modified tests for this.

neiljp:
This would be a good issue if you either know tests/pytest, or want to learn it - we would prefer code to have tests in PRs if possible, so this issue starts from the other side and would likely involve *just* test additions/changes.

devkant:
@zulipbot claim

zulipbot:
Welcome to Zulip, @devkant! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

devkant:
@neiljp I am learning pytest. Do I need to recreate the traceback ?

neiljp:
@devkant I would suggest reading existing tests which examine/use the stream list.

It was some time ago, but I believe my intent with this issue was to explore why adding the `Divider` initially just appeared to work, but later bugs showed up since of course it was not actually a `StreamButton` and did not have all the same attributes. So, this could be about ensuring that the `Divider` (as well as other elements) have the expected properties, ie. at least for `Divider` the faked `stream_name` and `stream_id` attributes in `view.py`, in tests which expect to use those attributes on the elements.

zulipbot:
Hello @devkant, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

[...]

PR #463, opened 2019-08-16
Add test for divider attributes

devkant:
Fixes: #194

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1001,6 +1001,42 @@ class TestLeftColumnView:
             for topic, count in zip(topic_list, unread_count_list)
         ])
 
+    @pytest.mark.parametrize('pinned', powerset([1, 2, 99, 1000]))
+    def test_stream_buttons(self, mocker, streams, pinned, width=40):
+        self.view.unpinned_streams = [s for s in streams if s[1] not in pinned]
+        self.view.pinned_streams = [s for s in streams if s[1] in pinned]
+
+        class button:
+            pass
+
+        def altbutton(stream, *args, **kwargs):
+            obj = button()
+            obj.stream_name = stream[0]
+            obj.stream_id = stream[1]
+            return obj
+
+        def altdivider(char):
+            obj = button()
+            return obj
+
+        def altlinebox(view, *args, **kwargs):
+            return view
+
+        def altstreamsview(btnlist, *args, **kwargs):
+            return btnlist
+
+        stream_button = mocker.patch(VIEWS + '.StreamButton',
+                                     side_effect=altbutton)
+        stream_view = mocker.patch(VIEWS + '.StreamsView',
+                                   side_effect=altstreamsview)
+        divider = mocker.patch(VIEWS + '.urwid.Divider',
+                               side_effect=altdivider)
+        line_box = mocker.patch(VIEWS + '.urwid.LineBox')
+        mocker.patch(STREAMBUTTON + ".mark_muted")
+        left_col_view = LeftColumnView(width, self.view)
+        for i in left_col_view.view.stream_w:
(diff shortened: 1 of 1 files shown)


sumanthvrao:
@devkant Thanks for the PR! One quick thing I would like to point out is that we don't make use of `git merge` workflow. We use a rebase based work-flow (This avoids the merge commits like the one occurring here). More details can be found here - https://zulip.readthedocs.io/en/latest/git/pull-requests.html#id1. You could `squash` the first two commits as well.

devkant:
@neiljp  It would be great if could review it.

devkant:
@neiljp I have used side_effect and objects of a temporary class button to create a list of buttons. It fails the test if we remove the attributes from divider. Please review it.

tests/ui/test_ui_tools.py
```diff
@@ -1001,6 +1001,42 @@ def test_topics_view(self, mocker, stream_button, width=40):
...
+
+        def altdivider(char):
+            obj = button()
+            return obj
+
+        def altlinebox(view, *args, **kwargs):
+            return view
+
+        def altstreamsview(btnlist, *args, **kwargs):
+            return btnlist
```
neiljp: You shouldn't need these classes/functions, as that's either what the mock does for you, or you actually want to test the code itself.

tests/ui/test_ui_tools.py
```diff
@@ -1001,6 +1001,42 @@ def test_topics_view(self, mocker, stream_button, width=40):
...
+
+        stream_button = mocker.patch(VIEWS + '.StreamButton',
+                                     side_effect=altbutton)
+        stream_view = mocker.patch(VIEWS + '.StreamsView',
+                                   side_effect=altstreamsview)
+        divider = mocker.patch(VIEWS + '.urwid.Divider',
+                               side_effect=altdivider)
```
neiljp: The above 3 mocks shouldn't be necessary, as we're wanting to test the actual code, not a mock. Also, for future reference, you don't test/use the mocks later, so you don't need to assign them to variables.

tests/ui/test_ui_tools.py
```diff
@@ -1001,6 +1001,42 @@ def test_topics_view(self, mocker, stream_button, width=40):
...
+        line_box = mocker.patch(VIEWS + '.urwid.LineBox')
+        mocker.patch(STREAMBUTTON + ".mark_muted")
+        left_col_view = LeftColumnView(width, self.view)
+        for i in left_col_view.view.stream_w:
+            assert hasattr(i, 'stream_name') and hasattr(i, 'stream_id')
```
neiljp: While pytest is pretty clever at identifying which element of an expression is failing, it's probably clearer if you have these on separate lines.

tests/ui/test_ui_tools.py
```diff
@@ -1001,6 +1001,42 @@ def test_topics_view(self, mocker, stream_button, width=40):
...
+        stream_button = mocker.patch(VIEWS + '.StreamButton',
+                                     side_effect=altbutton)
+        stream_view = mocker.patch(VIEWS + '.StreamsView',
+                                   side_effect=altstreamsview)
+        divider = mocker.patch(VIEWS + '.urwid.Divider',
+                               side_effect=altdivider)
+        line_box = mocker.patch(VIEWS + '.urwid.LineBox')
+        mocker.patch(STREAMBUTTON + ".mark_muted")
+        left_col_view = LeftColumnView(width, self.view)
+        for i in left_col_view.view.stream_w:
```
neiljp: Have you looked at the `stream_w`? You should be able to `print` it; `pytest` should display it in the details at the end, and/or you can use the `-s` option to show the output during each individual 'dot' of the test. I'm not convinced it contains what you want.

neiljp:
@devkant Thanks for giving this a go :+1:

PR #464, opened 2019-08-17
docker: add dockerfiles and readme

littleli:


amanagr:
I don't have docker installed on my system to review it. @neiljp ?

sumanthvrao:
Having docker installed on my system, I gave this a try and it definitely works :+1: . However I have a development version zulip installed hence my `zuliprc` file was found at `~/zuliprc` rather than in `~/.zulip/zuliprc` (Which is the case in production environment)`

docker/README.md
```diff
@@ -0,0 +1,20 @@
...
+git clone --depth=1 git@github.com:zulip/zulip-terminal.git
+cd zulip-terminal
+
+# build docker image with prefered Dockerfile
+docker build -t zulip-terminal:latest -f docker/Dockerfile.alpine .
+
+# run zulip-terminal image
+# and setup your zulip for the first time
+# next time ~/.zulip/zuliprc file will be used
+docker run -it -v ~/.zulip:/.zulip zulip-terminal:latest --config-file /.zulip/zuliprc
```
amanagr: As per @sumanthvrao comments, I agree that we should use `~/zuliprc` here for config file location. Since that is the default location where zuliprc is generated on a normal run.

docker/Dockerfile.buster
```diff
@@ -0,0 +1,26 @@
+FROM python:3.7-buster AS builder
+
+RUN useradd --user-group --create-home zulip 
+# && apt-get update \
+# && apt-get install -y libffi-dev libxml2-dev libxslt1-dev
```
neiljp: Can these lines be removed, since we don't run them?

docker/Dockerfile.buster
```diff
@@ -0,0 +1,26 @@
...
+RUN set -ex; python3 -m venv zulip-terminal-venv \
+&& . zulip-terminal-venv/bin/activate \
+&& pip3 --disable-pip-version-check install zulip-term \
+&& rm -rf /home/zulip/.cache
+
+FROM python:3.7-slim-buster
+
+RUN useradd --user-group --create-home zulip \
+&& apt-get update \
+&& apt-get install -y libxml2 libxslt1.1 \
```
neiljp: It is maybe tricky to test this, since many systems depend on this libraries and so it seems inadvisable to remove them, but do the binary wheels of our dependencies include these files in any case?

docker/Dockerfile.buster
```diff
@@ -0,0 +1,26 @@
...
+RUN useradd --user-group --create-home zulip 
+# && apt-get update \
+# && apt-get install -y libffi-dev libxml2-dev libxslt1-dev 
+
+USER zulip
+WORKDIR /home/zulip
+
+RUN set -ex; python3 -m venv zulip-terminal-venv \
+&& . zulip-terminal-venv/bin/activate \
+&& pip3 --disable-pip-version-check install zulip-term \
```
neiljp: We should emphasize that this only installs the version from PyPI, *not* the current version from git master.

Was this your intent? You mention cloning from git, so that is confusing. 

In any case, it would be useful to perhaps have options for building an image from pypi and from git?

docker/README.md
neiljp: a) I'm not sure if it's worth including the `--config-file` option in the run command, if we have to specify the file-mapping using `-v`?
b) The mapping should probably be a little clearer, and I'm not sure if this actually works without adjustment?

Would something like this be sufficient and clearer?
```
docker run -it -v path_to_zuliprc_file:/.zulip zulip-terminal:latest
```
with the following in the dockerfile
```
CMD ["--config-file", "/.zulip"]

File: docker/Dockerfile.buster (hunk 2 above)
Comment (littleli) [reply docker/Dockerfile.buster]: You are right. During development I've found that base image already contains these:
```
libffi-dev is already the newest version (3.2.1-9).
libxml2-dev is already the newest version (2.9.4+dfsg1-7+b3).
libxslt1-dev is already the newest version (1.1.32-2).
```
I'll update that

File: docker/README.md (hunk 1 above)
Comment (littleli) [reply docker/README.md]: It depends. 

a) You can actually choose to run without specifying volume. Then CMD from the Dockerfile is applied and a user is obliged to specify server and credentials every time. I feel like this can be a reasonable default.

b) I'm personally not very happy that my whole `$HOME` directory is used as a volume. That's why I picked subdirectory `$HOME/.zulip` where config file `zuliprc` sits. Now that I think about it maybe even `/etc/zulip/zuliprc` could be a good place in Unix tradition.

File: docker/Dockerfile.buster (hunk 2 above)
Comment (littleli) [reply docker/Dockerfile.buster]: To answer your question. Testing really isn't difficult. It may be just tedious to remove cached images for the clear run. It's slightly better to use Alpine as a baseline, because it is barebone and you have to be more explicit with each build step.

The idea to use volume to keep the config file come afterwards.

File: docker/Dockerfile.buster (hunk 4 above)
Comment (littleli) [reply docker/Dockerfile.buster]: I used these lines from README. But to actually build in docker you need Dockerfile. Then it depends if you want to build from source or build from PyPI. I've to admit I know very little about Python package system.

File: docker/README.md (hunk 1 above)
Comment (neiljp) [reply docker/README.md]: a) I agree that asking for the file each time is a valid 'no harm' approach, but I think it is worth indicating the lack of persistence with that approach and then listing a second example with persistence? For users not very familiar with Docker or how ZT works, it might be useful to compare those two approaches.

b) I'm not sure which approach you're referring to, but my interpretation is that `-v path_to_zuliprc_file:/.zulip` only exposes one file as a volume?

File: docker/Dockerfile.buster (hunk 2 above)
Comment (neiljp) [reply docker/Dockerfile.buster]: I didn't initially explore Alpine as I didn't realize the build process was so separated from the build OS that this was possible - but that builds and runs for me with a few adjustments to the run command, as I noted elsewhere.

File: docker/Dockerfile.buster (hunk 3 above)
Comment (neiljp) [reply docker/Dockerfile.buster]: To follow-up, I don't believe we need lines 18 & 19?

Comment (littleli) [reply docker/Dockerfile.buster]: yes, I already removed them with the following commit

Comment (neiljp) [reply docker/Dockerfile.buster]: I was referring to the libraries, not the dev files.

File: docker/Dockerfile.buster (hunk 4 above)
Comment (neiljp) [reply docker/Dockerfile.buster]: It would be good to have this work more broadly in future, but for now just indicating it is the latest PyPI version seems important.

I assume the only reason we need the git clone currently is for the `dockerfile` we're using, so that could be worth clarifying too.

**Comment by neiljp** [comment]
@littleli This PR does appear to work as-is, so I'd be happy to merge soon if you do not intend to add to it :+1: I can foresee that we will likely want to refine this soon, if not here, to improve the user experience, including in terms of clarifying how to run (minimally vs saved-zuliprc), expected resource usage & running different ZT versions.

This has been useful already in a different way with regard to #373, which we could probably have tested with pyenv but works fine with docker too :)

File: docker/README.md [hunk 5]
```diff
@@ -0,0 +1,25 @@
...
+docker build -t zulip-terminal:latest -f docker/Dockerfile.alpine .
+
+# Run zulip-terminal image and setup your zulip for the first time
+# next time $HOME/.zulip/zuliprc file will be used
+mkdir $HOME/.zulip
+docker run -it -v $HOME/.zulip:/.zulip zulip-terminal:latest --config-file /.zulip/zuliprc
+```
+
+You can swap `docker/Dockerfile.alpine` for `docker/Dockerfile.buster` if you prefer
+running inside image bassed on Debian over Alpine.
```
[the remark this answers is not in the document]
littleli: Done

docker/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Now we have access to docker directory
+
+# To build docker image with prefered Dockerfile, here based on Alpine
+docker build -t zulip-terminal:latest -f docker/Dockerfile.alpine .
+
+# Run zulip-terminal image and setup your zulip for the first time
+# next time $HOME/.zulip/zuliprc file will be used
+mkdir $HOME/.zulip
+docker run -it -v $HOME/.zulip:/.zulip zulip-terminal:latest --config-file /.zulip/zuliprc
```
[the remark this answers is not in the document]
littleli: Done

docker/README.md
```diff
@@ -0,0 +1,25 @@
+# Running in Docker
+
+To run `zulip-terminal` with docker you have to do following steps:
```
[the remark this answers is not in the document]
littleli: Done

docker/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+```sh
+# Clone repository and change current directory to it
+git clone --depth=1 git@github.com:zulip/zulip-terminal.git
+cd zulip-terminal
```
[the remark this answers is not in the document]
littleli: Done

littleli:
commit squashed and should be able to merge (even rebase)

PR #465, opened 2019-08-18
bugfx: buttons: Fix marking 'M' in muted topics.

sumanthvrao:
With 0b8052b we moved the logic to 'mark button
as muted' from TopButton into respective button classes
(StreamButtons).

However, we forgot to update the same call for muted TopicButtons.
This patch fixes the same.

### DIFF
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -226,7 +226,7 @@ class TopicButton(TopButton):
             self.mark_muted()
 
     def mark_muted(self) -> None:
-        self.update_count(-1)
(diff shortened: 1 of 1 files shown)


amanagr:
LGTM!

neiljp:
@sumanthvrao Thanks for finding this :+1: Agreed, LGTM - maybe some thoughts on a test to ensure that we don't miss this in future, as a followup?

sumanthvrao:
@neiljp Agreed! I am working on a PR which adds support to mute topics from within ZT. I will add the test in the PR. Thanks for the reminder :+1:

PR #466, opened 2019-08-18
bugfix: model: Fix is_muted_topic helper using list instead of tuples.

sumanthvrao:
Each element of `self.muted_topics` is a two element list containing stream_name and topic name. This patch fixes the `is_muted_topic` helper which checks if a given topic of a stream is muted or not.

Type annotations added.

Tests amended.


sumanthvrao:
@neiljp Should be a quick fix :)

neiljp:
@sumanthvrao Good to get this in, and the type annotations should help with avoiding regressions :+1:

PR #467, opened 2019-08-20
[WIP] - Mute topics from within ZT

sumanthvrao:
This PR adds support to mute topics from within ZT on `M` keypress:
* The first commit here introduces API call to toggle muted status.
* The second adds a method in core which builds the text and calls the PopUpconfirmationView
with the model function passed in as callback.
* We add functionality in buttons to restore/decrease unread count during appropriate events like (un)muting.
* We register the handler for muted topic event which handeles and event-based approach as well. This intern calls the mark_(un)muted methods of each buttons.
* We introduce the 'TOGGLE_MUTE_TOPIC' hotkey and maps it to its own category.
* Last commit involve connecting the Key added  earlier on a topic button so that it will be able to perform action of muting/unmuting a topic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,11 @@ Install-Module -Name BurntToast
 | Toggle topics in a stream                             | <kbd>t</kbd>                                  |
 | Mute/unmute Streams                                   | <kbd>m</kbd>                                  |
 
+### Topic list actions
+| Command                                               | Key Combination                               |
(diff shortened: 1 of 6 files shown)


sumanthvrao:
Tests are yet to be added, but feedback is of-course welcome.
@neiljp @amanagr

PR #468, opened 2019-08-25
Displays '(you)' after current user in list with white color

### LINKED ISSUE #437: Display '(You)' beside logged-in-user in user list.
It would benefit to mark  users who has logged-in (yourself) with a slightly different style compared to the rest of the users active (marked in yellow). Currently, we distinguish this subtly as the user name is the top name in the list.

One method was to mark `(YOU)` beside the logged in person, which is what the Web-app uses to distinguish. White color would be preferable.

We handle the logic of storing and sorting the user list in `get_all_users` in `model.py`, Which is likely the function to which the text must be added. Further, coloring schemes can be found in `config/themes.py` to be used for this. 

In case of getting stuck, feel free to join us in #zulip-terminal on [chat.zulip.org](https://chat.zulip.org/#narrow/stream/206-zulip-terminal)

khizirsiddiqui:
Fixes #437 
This PR appends (you) after the current user name in sidebar and gives it a white color.

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -564,6 +564,10 @@ class RightColumnView(urwid.Frame):
                 continue
             unread_count = (self.view.model.unread_counts['unread_pms'].
                             get(user['user_id'], 0))
+            if self.view.model.user_id == user['user_id']:
+                postfix_txt_markup = ('light grey', u" (you)")
+            else:
+                postfix_txt_markup = ''
             users_btn_list.append(
                 UserButton(
                     user,
@@ -571,6 +575,7 @@ class RightColumnView(urwid.Frame):
                     view=self.view,
                     width=self.width,
                     color=user['status'],
+                    postfix_txt_markup=postfix_txt_markup,
                     count=unread_count
                 )
             )
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -567,7 +567,7 @@ class RightColumnView(urwid.Frame):
             if self.view.model.user_id == user['user_id']:
                 postfix_txt_markup = ('light grey', u" (you)")
             else:
-                postfix_txt_markup = ''
+                postfix_txt_markup = ('', '')
             users_btn_list.append(
                 UserButton(
                     user,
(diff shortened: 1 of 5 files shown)


khizirsiddiqui:
@sumanthvrao Thanks for reviewing it. I have updated the code and I guess it works as expected, the ```full_name``` in ```active``` type color and ```(you)``` in light grey color.

sumanthvrao:
@khizirsiddiqui Thanks for sticking with this. This is definitley an overall improvement compared to last time.
* The old commits from the previous iteration still seem to be present in the PR. You could make use of `git rebase -i ` to fixup this commit history and squash the respective commits. [This article](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) might be helpful. Zulip has some docs on GitHub as well which are helpful - https://zulip.readthedocs.io/en/latest/git/pull-requests.html

The idea is to not make the change in the first place rather than making a change and fixing it in a follow-up commit. Your welcome to ask for help in the community for this, in case you get stuck :+1: 

* The `(you)` beside the name appears in a sparate line for me and appears in green color. 
image
 
* @khizirsiddiqui Do you plan to add tests for this?

khizirsiddiqui:
image
My name comes in green for me with 'you' in grey.
Yes, I'll try fixing that new line thing first, also rebase the commits.

sumanthvrao:
@khizirsiddiqui I am using gruvbox theme. Guess you are using a different one here.

Re rebase commits and newline: Great!

zulipterminal/ui_tools/views.py
```diff
@@ -564,13 +564,18 @@ def users_view(self, users: Any=None) -> Any:
                 continue
             unread_count = (self.view.model.unread_counts['unread_pms'].
                             get(user['user_id'], 0))
+            if self.view.model.user_id == user['user_id']:
+                postfix_txt_markup = ('light grey', u" (you)")
```
neiljp: Generally we tend to use style names so that this works better across themes. Was there a reason you picked this color? I saw you did have a theme there at some point, so perhaps just leave that addition, or use an existing name if you can find one that isn't too misleading?

Note that previously your style would not have worked on a transparent background terminal, as it specified the background color to be black.

PR #472, opened 2019-09-12
Add organization name to notifications & title bar

neiljp:
This idea was mentioned recently, and seems pretty useful when you have multiple organizations (realms), so I'm putting this small PR out there to see how well it works cross-platform and inspire discussion in this direction.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -309,6 +309,7 @@ def initial_data(logged_on_user, users_fixture, streams_fixture):
         'full_name': logged_on_user['full_name'],
         'email': logged_on_user['email'],
         'user_id': logged_on_user['user_id'],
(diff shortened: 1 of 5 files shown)


sumanthvrao:
@neiljp This idea looks fairly straightforward to me definitely a good thing to have when one is looked into multiple zulip realms :+1:

neiljp:
The updated version has a less minimal way of adding the `server_name` (organization - we could change that, but we have `server_url` and it seems reasonable to have 'name'?), the notification update, as well as adding the server name to the title bar.

There are also more test updates accordingly.

neiljp:
Merging this now, due to the low risk, and that we have further confirmation (via #zulip-terminal chat) that this should work on OSX.

PR #474, opened 2019-10-04
requirements: Upgrade pytest from 3.4.2 to 3.10.1.

neiljp:
Other than upgrading to a later pytest version, this commit also
resolves an issue with Travis, avoiding a need to specify an upgraded
attr package explicitly since this pytest depends on a later version.

Note that this can cause two warnings to be output by pytest re `inspect` being used in a deprecated way inside `urwid`, but I'm not sure there's much we can do here for now.

@amanagr @sumanthvrao The latter issue is causing failures in PRs right now, hence the high priority.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -13,7 +13,7 @@ lxml = "==4.2.3"
 mypy_extensions = ">=0.4"
 
 [dev-packages]
-pytest = "==3.4.2"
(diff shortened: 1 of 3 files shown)


neiljp:
@sumanthvrao Thanks for the feedback. It seems important to merge so other PRs don't get affected by it.

Issue #412, opened 2019-06-30
Style in-line emoji :+1: to match styling of its 'reactions' counterpart.


sumanthvrao:
Emojis like :+1:  are displayed as `:+1: ` are displayed without any special styling when used in a message (part of message content). We could potentially display them similar to the reaction :+1: , which get's displayed as `:thumbs_up:` (Shift `+` hotkey used on message toggles a like). We could change the color to match the reactions (one can be ~purple and another can be ~purple, bold)

Conversation link : https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Inline.20emoji.20styling/near/761561

sumanthvrao:
@zulipbot add "good first issue"

pranav-kirsur:
@zulipbot claim

zulipbot:
Welcome to Zulip, @pranav-kirsur! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

zulipbot:
Hello @pranav-kirsur, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

blenderskool:
@zulipbot claim

zulipbot:
Welcome to Zulip, @blenderskool! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
[...]

PR #476, opened 2019-10-13
Add in-line emoji styling to match reaction styling

blenderskool:
Emojis in messages are styled to match the styling of reactions. Reactions are additionally made bold to better differentiate between them.

Before:
before

After:
after

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1294,11 +1294,11 @@ class TestMessageBox:
         ('<div class="message_inline_image">'
          '<a href="x"><img src="x"></a></div>', []),
         ('<div class="message_inline_ref">blah</div>', []),
-        ('<span class="emoji">:smile:</span>', [':smile:']),
+        ('<span class="emoji">:smile:</span>', [('emoji', ':smile:')]),
         ('<div class="inline-preview-twitter"',
             ['[TWITTER PREVIEW NOT RENDERED]']),
-        ('<img class="emoji" title="zulip"/>', [':zulip:']),
-        ('<img class="emoji" title="github"/>', [':github:']),
+        ('<img class="emoji" title="zulip"/>', [('emoji', ':zulip:')]),
+        ('<img class="emoji" title="github"/>', [('emoji', ':github:')]),
     ], ids=[
         'empty', 'p', 'user-mention', 'group-mention', 'code', 'codehilite',
         'strong', 'em', 'blockquote',
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -21,6 +21,7 @@ required_styles = {
     'bar',
     'help',
     'emoji',
+    'reaction',
     'span',
     'link',
     'blockquote',
@@ -44,6 +45,8 @@ LIGHTGREEN = 'h142'  # bright_green
 LIGHTRED = 'h167'  # bright_red
 LIGHTREDBOLD = '%s, bold' % LIGHTRED
 GRAY = 'h244'  # gray_244
+LIGHTMAGENTA = 'h132'  # neutral_purple
+LIGHTMAGENTABOLD = '%s, bold' % LIGHTMAGENTA
 
 THEMES = {
     'default': [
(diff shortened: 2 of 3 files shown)


blenderskool:
Fixes #412

sumanthvrao:
@blenderskool you can run `pytest && ./tools/run-mypy` to run the tests locally. Are you familiar with pytest?  You can look at existing tests and see how you can fix the errors and add new tests. 

The diff in your commits shows that the second commit only has unnecessary spaces. You could drop the commit (`git rebase` can help you here)

Feel free to reach out incase you have further doubts.

blenderskool:
@sumanthvrao Ok I'll make the necessary changes as you mentioned 🙂

neiljp:
@blenderskool Thanks for working on this ;)

Did you test the extended colo[u]rs you added to the gruvbox scheme? Where did they come from?

You might like to use `gitlint` as described in the README, and also see how we generally style commit text based on looking at `git log` (or using `tig`).

While #412 (based on the original discussion) suggests using an entirely different style for reactions - and we can continue to do this - I'm wondering if we should reserve a 'bold' styling of whatever the base reactions colo[u]r is for when a user has contributed to that reaction @sumanthvrao @amanagr Thoughts?

neiljp:
@blenderskool Oh, and please pop in on the zulip stream at chat.zulip.org if you want to discuss any aspect of development :)

blenderskool:
@neiljp @sumanthvrao I updated the tests to match the added `emoji` styling. Also fixed the gruvbox color that I had added to a `neutral_purple` color I found here https://github.com/morhetz/gruvbox/blob/master/colors/gruvbox.vim

neiljp:
@blenderskool Thanks for the update - I'm glad to be able to get back to this! This looks *almost* ready to go, but I've described some aspects of how you seem to have structured the commits below. I could change them myself, but hopefully you might find the following useful to work through, though please ask questions if you don't follow or agree with my reasoning :)

Overall your branch passes the tests, but individually the commits don't. It's often useful to have lots of small 'working' commits during development (which perhaps is what you have here), but prior to merging (or a near-final PR) it's useful to have commits which individually don't break anything and are cleaner. In other projects, commits can get squashed together when a branch is merged, which hides this. However we rebase on merge here, and cleaner commits can make it easier to review code.

This behavior (not passing tests from one commit to another) can also suggest that elements in each commit aren't grouped quite right between commits. Could you comment on the following:
* Is the last commit a fix to the style in the first one? If so, since this is new code you're adding, do those two commits belong squashed together?
* Similarly, is the second commit a fix to the first?

Depending on how those commits get squashed together (or not), you can determine where the tests changes should go, so as to not break the fact that the tests pass from one commit to another :)

blenderskool:
@neiljp Thanks for the comments. I agree with what you've mentioned about the commit structure. I'll be more thoughtful regarding this in the future.
How do I go about fixing this with the commits I made in this branch?
- The last commit is indeed a fix for gruvbox style in the first commit.
- The second commit adds custom icon support to the first commit where I didn't notice that it was being handled separately.

neiljp:
@blenderskool Depending upon how big the changes are, you might find an **interactive rebase** useful, which allows you to reorder and merge (squash, or fixup - when you're confident!) commits. Sometimes this can lead to conflicts - but you can always escape with a `git rebase --abort` if things get confusing :)

If the later commits are fixes to the first commit - not just developments on top of it, which is a subtle difference - then particularly if they're small then they could just be all squashed together into one commit, with the commit text updated.

There are many git resources out there, but you may find the Zulip git guide useful, specifically this section at the moment: https://zulip.readthedocs.io/en/latest/git/fixing-commits.html

blenderskool:
@neiljp I squashed and reordered some commits which made logical sense. Thanks for the advice.

neiljp:
@blenderskool Thanks for working on this - I hope this was useful for you too. The code looks a lot simpler to understand when looking at just these commits :+1: I checked manually and things still look to be working fine overall :)

I ran `tools/lint-all && pytest` on your branch and after checking out `HEAD~`. The branch passes, as per CI results, but the latter intermediate checkout doesn't. This is since the tests are added prior to the code that implements them. Adding them afterwards instead would be OK, but generally we include the tests in the same commit as the code that would otherwise break them.

Given the above, but that the code is essentially fine, I've squashed the two commits and merged them manually as 9a708277d67d7285066e64d09e49da784fa197c2. Other than making a minor change to the commit text for accuracy (match -> similar to), I also adjusted the background color of the blue & light themes for the added style.

PR #480, opened 2019-11-18
Upgrade zulip to 0.6.1, avoiding uncaught SSLError

neiljp:
The new zulip library version allows catching of `ZulipError` in an
additional case, removing some cases where a requests exception was
thrown directly into zulip-terminal.

This avoids an uncaught exception, in cases such as being connected to a
network but not routed, such as if not signed in to a wifi router.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -5,7 +5,7 @@ name = "pypi"
 
 [packages]
 urwid = "==2.0.1"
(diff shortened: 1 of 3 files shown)


neiljp:
I see this as low risk, so I'm merging this now.

PR #481, opened 2019-11-23
Report reasons for initial connection failure

neiljp:
This PR first adjusts the error handling in functions used by `ThreadPool`s in the model, allowing for a reason to be passed back rather than just a boolean. That in turn allows for reporting of why an error has occurred to the user; initially this is simply for the first connection attempt to the server, which can diagnose rather useful situations such as incorrect API keys.

Tests are amended for the refactor, and a test added for the example case of an incorrect API key.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -22,7 +22,8 @@ class TestModel:
 
     @pytest.fixture
     def model(self, mocker, initial_data, user_profile):
-        mocker.patch('zulipterminal.model.Model.get_messages')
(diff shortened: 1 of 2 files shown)


neiljp:
Self-reviewed this and noticed that the `ZulipError` case was incorrect, and added a test case for that.

neiljp:
@sumanthvrao Thanks for the review; this should help with reporting errors, and maybe we can fine-tune the tests later :)

PR #483, opened 2019-11-29
Improve link rendering

neiljp:
This fixes a few outstanding issues in link rendering:
* Don't render extra copies of images/refs provided by the server
* Render simplified links (where text & link match) in the link style
* Improve handling of relative links, by using `urlparse` and `urljoin` to determine them & prepend the server URL
* Ensure that if we get a relative link which matches the actual text, we also simplify that rendering (the server simplifies the URL; we want to ensure the full link+text is simplified)

Tests are updated & added.


neiljp:
Interpreting Tim's feedback on #**development help** as positive, and given the incremental changes here, merging this now.

Issue #470, opened 2019-09-07
Add 'year' data when a message more than a year old is displayed.


sumanthvrao:
For messages which date back more than a year old, ZT does not display the year tag in the message box. We could improve the time formating in many ways but adding the year attribute would be a good first step. 

There might be potential issues locating the year with the 'star' to avoid it appearing in different locations, so for a basic-version, we can move the star to the right of the date.

A complete solution for this would be to use a logic similar to the web-app in `statis/js/timerender.js`.

zulipbot:
Hello @neiljp, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

subhu12:
@zulipbot claim

zulipbot:
Welcome to Zulip, @subhu12! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

neiljp:
@subhu12 Please note that this feature already has a PR active (#484). I have removed the help-wanted tag for now.

PR #484, opened 2019-12-01
Indicate message year, when it is not the current year

neiljp:
This is a fix for #470, has tests, and works functionally for me. Feedback welcome.

I propose that the extension ideas in #470 are placed into other issues, for things such as any styling improvements and potentially dynamic date changes as with the webapp.


amanagr:
LGTM, I am not sure about the star change that @sumanthvrao suggested.

neiljp:
@sumanthvrao I agree about the star looking odd. The first commit does resolve the issue, but I just added another if you'd like to try that.

@amanagr See how you like the star repositioning.

Perhaps long-term we could place the star next to the message.

sumanthvrao:
@neiljp The star repositioning looks good. :+1:

neiljp:
Merging now, given the positive review and feedback.

PR #485, opened 2019-12-08
Upgrade pytest from 3.10.1 to 5.3.5 (latest).

neiljp:
This is a replacement/extension for @amanagr's idea in #455. pytest was partially upgraded in 14c9161 already. Since #455 was opened, 5.3.1 is now available (cf 5.0.1 in #455 ).

Upgrading shows two types of external warnings:
* `urwid` uses some 'inspect' calls which are deprecated (since last upgrade to 3.10.1)
* `pytest-pep8` doesn't set up some kind of 'hook' to register a marker

To avoid confusing users with these warnings about external code, this commit works around
the existing and new issue by adding a `filterwarnings` entry for urwid, and explicitly
mentioning the pep8 `mark`, both in `setup.cfg`.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -13,7 +13,7 @@ lxml = "==4.2.3"
 mypy_extensions = ">=0.4"
 
 [dev-packages]
-pytest = "==3.10.1"
+pytest = "==5.3.5"
 pytest-pep8 = "==1.0.6"
 pytest-mock = "==1.7.1"
 pytest-cov = "==2.5.1"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 urwid==2.1.0
 zulip==0.6.3
-pytest==3.10.1
+pytest==5.3.5
 pytest-cov==2.5.1
 pytest-pep8==1.0.6
 pytest-mock==1.7.1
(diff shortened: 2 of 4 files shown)


amanagr:
Thanks @neiljp  for this awesome find. I ran `pytest` after doing `pip3 install .` on this branch. I get another warning for `lxml` library:
```
tests/core/test_core.py::TestController::test_narrow_to_stream
  /Users/amanagrawal/.local/share/virtualenvs/zulip-terminal-NCZzs1Kf/lib/python3.7/site-packages/bs4/builder/_lxml.py:61: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    if isinstance(parser, collections.Callable):

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

I tried upgrading `lxml` to the latest version `4.4.2` but I still get the same warning. I thing we should filter it for now.

neiljp:
@amanagr Right, I agree, this needs checking in each version of Python that we support, since pytest reports different results in each case. I have only really tested locally, which is 3.6.x.

neiljp:
Updated:
* pytest upgraded in separate commit to 5.3.5 (cf 5.3.1), from 3.10.1
* Filters make tests green for all CI :+1:

neiljp:
An update with the latest CI confirms that after the `urwid` upgrade, we no longer need to exclude that from the warnings :+1:

I plan to merge this soon, since this improves the developer experience.

rht:
Looks fine to me.

PR #486, opened 2019-12-11
FIX: #437 Added (you) beside logged-in user

wolf9970:
I piggy-backed off of the work begun by @khizirsiddiqui. All credit given where due. 

Added "(you)" beside logged in user. Color is pulled from themes. Displays all on one line.


neiljp:
@wolf9970 Thanks for following up the earlier work on this :+1:

It seems you have extra copies of some files added here in one of the commits; is that the case?

In terms of building on the previous commit, could you initially retain the previous commit separately to identify their work, and then put your changes into a follow-up?

It's also a little confusing that there is another user in the list of commits - is that you also?

wolf9970:
@neiljp Thanks for the input. Something odd was happening with my git installation so I tried pushing my changes on github.com which ended up causing this mess. Would it be best to close this PR and start over?

neiljp:
@wolf9970 PRs are generally just connected to a branch, so you can push updates to the connected branch and the pull-request commits can be changed completely - so you should be able to keep this PR open.

That said, from the naming scheme here, you seem to have this PR connected to your `master` branch (at least on github); that can work, but you might find it easier to have a local branch with a different name for a specific thing you're working on, then push that to github (which would be associated with any PR).

Depending on the scale of the changes and how comfortable you are with git, various options are available. You should be able to push from a branch name of your choice to your `master` branch on github, at least temporarily, for example.

I would suggest squashing the changes that (mostly) 'cancel out' together, and to grab the original commit you can `git cherry-pick` it onto the current branch.

PR #490, opened 2019-12-30
run: Rephrase the login screen messages.

preetmishra:
Issue:
When trying to login to the Zulip terminal, the error message being displayed after entering the wrong email or password is "Username or Password Incorrect!" but the user is never asked for a username. 

Fix:
Rephrasing the error message in `run.py`

Screenshots:

Before:
Screenshot from 2019-12-29 21-52-09

After:
Screenshot from 2019-12-30 20-15-07

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -92,7 +92,7 @@ def fetch_zuliprc(zuliprc_path: str) -> None:
                       "web browser to log in to Zulip." +
                       "\nIt often looks like one of the following:" +
                       in_color('green', "\n   your-org.zulipchat.com") +
-                      " (Zulip-hosted servers)" +
+                      " (Zulip cloud)" +
                       in_color('green', "\n   zulip.your-org.com") +
                       " (self-hosted servers)" +
                       in_color('green', "\n   chat.zulip.org") +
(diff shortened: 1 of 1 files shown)


amanagr:
LGTM!

timabbott:
Would it make sense to also change "Zulip-hosted servers" to "Zulip Cloud"?

amanagr:
While "Zulip-hosted servers" is more intuitive, if "Zulip Cloud" is the proper term for it, then we should show it there.

preetmishra:
@amanagr I changed it to "Zulip Cloud" and squashed the two commits into one.

neiljp:
@preetmishra Squashing the two commits together simplifies things, but it's useful to keep the commit title simpler and put details in the text below - perhaps look at other git commits for examples. Here, it could be that the previous title was still reasonable, but the details could be added further down.

Generally if you find yourself using something like 'and' in the title, either the commit is too big, or it can be summarized differently with details below :)

preetmishra:
@neiljp Thank you for pointing it out. I will keep these things in mind from now. :)
I referred to earlier commits and I have made the title simpler and have added a description.

sumanthvrao:
Nice catch @preetmishra .

A small suggestion would be to rename it to `Email` rather than `E-mail` so that it is consistent with other usages of the same term in the error message.

The rest of the changes LGTM as well.

preetmishra:
@sumanthvrao Thank you for reverting. I agree that it should be consistent. I have changed it to `Email`.

neiljp:
@preetmishra Thanks for these adjustments - I'm merging these now :+1: While they're small, they will improve the experience for users, including yourself I hope :)

preetmishra:
Thanks! Looking forward to contributing more.

PR #495, opened 2020-01-27
Added stream info view for streams

### LINKED ISSUE #494: Feature: Enable view of stream descriptions
One possible way this might work is showing this information when `i` is pressed, while a stream name is selected, much like when on a message a pop-up shows information on the message.

The data should be available in the subscriptions in the model, which you can examine based on the fixtures in `tests/conftest.py`, which can also be used to write tests to develop this functionality.

SGeetansh:
Fixes #494

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -14,7 +14,7 @@ from zulipterminal.helper import asynch
 from zulipterminal.model import Model, GetMessagesArgs, ServerConnectionFailure
 from zulipterminal.ui import View, Screen
 from zulipterminal.ui_tools.utils import create_msg_box_list
-from zulipterminal.ui_tools.views import HelpView, MsgInfoView
+from zulipterminal.ui_tools.views import HelpView, MsgInfoView, StreamInfoView
 from zulipterminal.config.themes import ThemeSpec
 from zulipterminal.ui_tools.views import PopUpConfirmationView
 
@@ -114,6 +114,10 @@ class Controller:
         msg_info_view = MsgInfoView(self, msg)
         self.show_pop_up(msg_info_view,
                          "Message Information (up/down scrolls)")
+
+    def show_stream_info(self, color: str, name: str, desc: str) -> None:
+        show_stream_view = StreamInfoView(self, color, name, desc)
+        self.show_pop_up(show_stream_view, " # {} ".format(name))
 
     def search_messages(self, text: str) -> None:
         # Search for a text in messages
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -559,7 +559,8 @@ class Model:
     def _stream_info_from_subscriptions(
             subscriptions: List[Dict[str, Any]]
     ) -> Tuple[Dict[int, Any], Set[int], List[List[str]], List[List[str]]]:
-        stream_keys = ('name', 'stream_id', 'color', 'invite_only')
+        stream_keys = ('name', 'stream_id', 'color',
+                       'invite_only', 'description')
 
         # Canonicalize color formats, since zulip server versions may use
         # different formats
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -123,7 +123,8 @@ class StreamButton(TopButton):
                  count: int=0) -> None:
         # FIXME Is having self.stream_id the best way to do this?
         # (self.stream_id is used elsewhere)
-        self.stream_name, self.stream_id, color, is_private = properties
+        self.stream_name, self.stream_id, self.color, is_private, \
+            self.description = properties
         self.model = controller.model
         self.count = count
         self.view = view
@@ -133,13 +134,16 @@ class StreamButton(TopButton):
                 background = entry[5] if len(entry) > 4 else entry[2]
                 inverse_text = background if background else 'black'
                 break
-        view.palette.append((color, '', '', '', color+', bold', background))
-        view.palette.append(('s' + color, '', '', '', inverse_text, color))
+        view.palette.append((
+            self.color, '', '', '', self.color+', bold', background))
+        view.palette.append((
+            's' + self.color, '', '', '', inverse_text, self.color))
 
         super().__init__(controller,
                          caption=self.stream_name,
                          show_function=controller.narrow_to_stream,
-                         prefix_character=(color, 'P' if is_private else '#'),
+                         prefix_character=(
+                            self.color, 'P' if is_private else '#'),
                          width=width,
                          count=count)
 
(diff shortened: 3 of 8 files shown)


SGeetansh:
I noticed that this breaks some of the stream tests because I am fetching descriptions too. Should I just add a description string in other tests that aren't using description or should I do something else?

zulipterminal/model.py
```diff
@@ -559,7 +559,8 @@ def get_all_users(self) -> List[Dict[str, Any]]:
     def _stream_info_from_subscriptions(
             subscriptions: List[Dict[str, Any]]
     ) -> Tuple[Dict[int, Any], Set[int], List[List[str]], List[List[str]]]:
-        stream_keys = ('name', 'stream_id', 'color', 'invite_only')
+        stream_keys = ('name', 'stream_id', 'color',
+                       'invite_only', 'description')
```
neiljp: This approach is fine for now - it certainly follows the current logic. It does seem ready for refactoring, though.

zulipterminal/ui_tools/views.py
```diff
@@ -839,6 +839,19 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
             self.controller.exit_popup()
         return super(PopUpConfirmationView, self).keypress(size, key)
 
+class StreamInfoView(urwid.ListBox):
+    def __init__(self, controller: Any, color: Any, name: Any, desc: Any) -> None:
```
neiljp: Same point about type hints here.

zulipterminal/ui_tools/views.py
```diff
@@ -839,6 +839,19 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
...
+        self.controller = controller
+        self.width = max(len(desc), len(name))
+        self.height = 3
+        self.name = urwid.Text((color, name), align='center')
+        self.desc = urwid.Text(desc, align='center')
```
neiljp: For these objects, do we need all of these as attributes, or are they just names manipulated and passed elsewhere?

zulipterminal/ui_tools/views.py
```diff
@@ -839,6 +839,19 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
             self.controller.exit_popup()
         return super(PopUpConfirmationView, self).keypress(size, key)
 
+class StreamInfoView(urwid.ListBox):
+    def __init__(self, controller: Any, color: Any, name: Any, desc: Any) -> None:
+        self.controller = controller
+        self.width = max(len(desc), len(name))
+        self.height = 3
```
neiljp: This fixed height looks a little awkward in practice, with short stream descriptions.

zulipterminal/ui_tools/buttons.py
```diff
@@ -174,6 +174,9 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
                 )
         elif is_command_key('TOGGLE_MUTE_STREAM', key):
             self.controller.stream_muting_confirmation_popup(self)
+        elif is_command_key('MSG_INFO', key):
```
neiljp: You've used `MSG_INFO` here and elsewhere, which works fine since it's the same key.

However, it would be useful to have this separately documented in the help menu and in the README; adding it manually to the latter and a new entry to the key metadata would be one way to resolve this.

SGeetansh:
@neiljp Thanks a lot for the review. I'll do the work and get back to you.

zulipterminal/ui_tools/views.py
SGeetansh: I am not sure about what do with the height and width. Could you suggest something?

zulipterminal/core.py
```diff
@@ -115,9 +115,9 @@ def show_msg_info(self, msg: Any) -> None:
         self.show_pop_up(msg_info_view,
                          "Message Information (up/down scrolls)")
 
-    def show_stream_info(self, color: Any, name: Any, desc: Any) -> None:
+    def show_stream_info(self, color: str, name: str, desc: str) -> None:
         show_stream_view = StreamInfoView(self, color, name, desc)
-        self.show_pop_up(show_stream_view, "Stream Description")
+        self.show_pop_up(show_stream_view, "Stream - {}".format(name))
```
neiljp: I'm wondering if the title might be cleaner just using eg `# announce` or `P some private stream`, as they are shown in the stream list, even without the color (color doesn't work?)

SGeetansh: I tried adding the color to the title of the popup. But I don't think it is possible with the curernt definition of the pop_up() function. Please correct me if I am wrong. And guide me on how to do it.

SGeetansh: I've given the tiltle a minimal look though.

SGeetansh:
Sorry for the delay. I was busy with something totally unrelated.
I couldn't find the character limit for the stream description, hence the height and width are still defined in the same way. 
If I get the character limit of the description, I can use that to break the description into several lines of fixed length and then the height can adjust according to the number of lines with some padding of course.

zulipterminal/core.py
```diff
@@ -115,6 +115,10 @@ def show_msg_info(self, msg: Any) -> None:
         self.show_pop_up(msg_info_view,
                          "Message Information (up/down scrolls)")
 
+    def show_stream_info(self, color: str, name: str, desc: str) -> None:
+        show_stream_view = StreamInfoView(self, color, name, desc)
+        self.show_pop_up(show_stream_view, "#{}".format(name))
```
neiljp: We could do with a space here?

Without color, I think the 'P' form we use for private streams would be confusing, so this is fine for now. If we can figure out color then that could be changed later.

zulipterminal/ui_tools/views.py
```diff
@@ -840,6 +840,21 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
         return super(PopUpConfirmationView, self).keypress(size, key)
 
 
+class StreamInfoView(urwid.ListBox):
+    def __init__(self, controller: Any, color: str,
+                 name: str, desc: str)-> None:
+        self.controller = controller
+        self.width = max(len(desc), len("Stream - {}".format(name)))+2
+        self.height = 2
+        log = [urwid.Text(desc, align='center')]
```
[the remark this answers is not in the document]
SGeetansh: It's my first time working with urwid. I saw some other code samples and thought this was the convention.

neiljp:
@SGeetansh Thanks for this work :+1: I hope it was a good introduction to contributing!

As I discussed with you, I rebased and added a few small changes, in addition to writing a more representative commit text, which I merged separately here:
https://github.com/zulip/zulip-terminal/commit/2826ef9b396bd80f73c8d57421cc3f48a02f8057

PR #497, opened 2020-01-28
Add unread_msgs structure to index

### LINKED ISSUE #496: Refactor index to better support unread messages
Currently, very few unread messages are retrieved during initial load. Due to lack of data about unread messages, issues like #434 cannot be solved. Solution is to:

Retrieve around 50k unread messages from server. Add data structures to the index to store info about unread messages, and to prevent retrieval being costly. When marking message as read, remove the message from the newly added data structure
For more info, refer to https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/.5BCode.20help.5D.20how.20does.20index.20work.3F

kaustubh-nair:
Fixes #496

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -183,8 +183,9 @@ def test_classify_unread_counts(mocker, initial_data, stream_dict,
     model.initial_data = initial_data
     model.muted_topics = muted_topics
     model.muted_streams = muted_streams
-    assert classify_unread_counts(model) == dict(classified_unread_counts,
-                                                 **vary_in_unreads)
+    assert classify_unread_counts(model) == (dict(classified_unread_counts[0],
+                                             **vary_in_unreads),
+                                             classified_unread_counts[1])
 
 
 @pytest.mark.parametrize('color', [
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1145,13 +1145,15 @@ class TestModel:
 
     def test_update_star_status_no_index(self, mocker, model):
         model.index = dict(messages={})  # Not indexed
-        event = dict(messages=[1], flag='starred', all=False)
+        event = dict(messages=[1], flag='starred', all=False, operation='add')
         mocker.patch('zulipterminal.model.Model.update_rendered_view')
+        set_count = mocker.patch('zulipterminal.model.set_count')
 
         model.update_message_flag_status(event)
 
         assert model.index == dict(messages={})
         model.update_rendered_view.assert_not_called()
+        set_count.assert_not_called()
 
     def test_update_star_status_invalid_operation(self, mocker, model):
         model.index = dict(messages={1: {'flags': None}})  # Minimal
@@ -1163,9 +1165,11 @@ class TestModel:
             'all': False,
         }
         mocker.patch('zulipterminal.model.Model.update_rendered_view')
+        set_count = mocker.patch('zulipterminal.model.set_count')
         with pytest.raises(RuntimeError):
             model.update_message_flag_status(event)
         model.update_rendered_view.assert_not_called()
+        set_count.assert_not_called()
 
     @pytest.mark.parametrize('event_message_ids, indexed_ids', [
         ([1], [1]),
@@ -1199,6 +1203,7 @@ class TestModel:
             'all': False,
         }
         mocker.patch('zulipterminal.model.Model.update_rendered_view')
+        set_count = mocker.patch('zulipterminal.model.set_count')
 
         model.update_message_flag_status(event)
 
@@ -1206,11 +1211,71 @@ class TestModel:
         for changed_id in changed_ids:
             assert model.index['messages'][changed_id]['flags'] == flags_after
         (model.update_rendered_view.
-         has_calls([mocker.call(changed_id) for changed_id in changed_ids]))
+         assert_has_calls([mocker.call(changed_id)
+                          for changed_id in changed_ids]))
 
         for unchanged_id in (set(indexed_ids) - set(event_message_ids)):
             assert (model.index['messages'][unchanged_id]['flags'] ==
                     flags_before)
+
+        set_count.assert_not_called()
+
+    @pytest.mark.parametrize('event_message_ids, indexed_ids', [
+        ([1], [1]),
+        ([1, 2], [1]),
+        ([1, 2], [1, 2]),
+        ([1], [1, 2]),
+        ([], [1, 2]),
+        ([1, 2], []),
+    ])
+    @pytest.mark.parametrize('event_op, flags_before, flags_after', [
+        ('add', [], ['read']),
+        ('add', ['read'], ['read']),
+        ('add', ['starred'], ['starred', 'read']),
+        ('add', ['read', 'starred'], ['read', 'starred']),
+        ('remove', [], []),
+        ('remove', ['read'], ['read']),   # msg cannot be marked 'unread'
+        ('remove', ['starred'], ['starred']),
+        ('remove', ['starred', 'read'], ['starred', 'read']),
+        ('remove', ['read', 'starred'], ['read', 'starred']),
+    ])
+    def test_update_read_status(self, mocker, model, event_op,
+                                event_message_ids, indexed_ids,
+                                flags_before, flags_after):
+        model.index = dict(messages={msg_id: {'flags': flags_before}
+                                     for msg_id in indexed_ids})
+        event = {
+            'messages': event_message_ids,
+            'type': 'update_message_flags',
+            'flag': 'read',
+            'operation': event_op,
+            'all': False,
+        }
+
+        mocker.patch('zulipterminal.model.Model.update_rendered_view')
+        set_count = mocker.patch('zulipterminal.model.set_count')
+
+        model.update_message_flag_status(event)
+
+        changed_ids = set(indexed_ids) & set(event_message_ids)
+        for changed_id in changed_ids:
+            assert model.index['messages'][changed_id]['flags'] == flags_after
+
+            if event_op == 'add':
+                model.update_rendered_view.assert_has_calls(
+                        [mocker.call(changed_id)])
+            elif event_op == 'remove':
+                model.update_rendered_view.assert_not_called()
+
+        for unchanged_id in (set(indexed_ids) - set(event_message_ids)):
+            assert (model.index['messages'][unchanged_id]['flags'] ==
+                    flags_before)
+
+        if event_op == 'add':
+            set_count.assert_called_once_with(list(changed_ids),
+                                              self.controller, -1)
+        elif event_op == 'remove':
+            set_count.assert_not_called()
 
     @pytest.mark.parametrize('narrow, event, called', [
         # Not in PM Narrow
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -34,7 +34,7 @@ class TestModel:
         # NOTE: PATCH WHERE USED NOT WHERE DEFINED
         self.classify_unread_counts = mocker.patch(
             'zulipterminal.model.classify_unread_counts',
-            return_value=[])
+            return_value=([], set()))
         self.client.get_profile.return_value = user_profile
         model = Model(self.controller)
         return model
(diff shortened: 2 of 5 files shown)


kaustubh-nair:
@neiljp Initially, I was thinking of storing only the message_ids in the index, but I'm not sure how much of this extra information ( like `stream_id`, `user_ids_string`) is useful. Do you think all these attributes need to be stored? The structure of `unread_msgs` is here https://zulip.readthedocs.io/en/latest/subsystems/unread_messages.html

kaustubh-nair:
@zulipbot add "in progress"

neiljp:
@kaustubh-nair We know the extra information is there in the `initial_data`; I would work towards a version that works simply first, perhaps leaving comments referring to the additional data available for quick reference later.

At a glance the code looks like it should broadly do the right thing, though we might be able to move things around and simplify them once it is more settled. You can probably squash the second commit with the first?

kaustubh-nair:
> @kaustubh-nair You may want to investigate integrating this into `classify_unread_counts` and the `UnreadCounts` data, though currently this may work with the data being separate.

@neiljp Integrating it into `classify_unread_counts` is a good idea. We could modify `UnreadCounts` to store the `message_ids` instead of only counts. But does it make sense to let `unread_msg_ids` remain in index? Why is the `UnreadCounts` structure kept separate from the index?

tests/helper/test_helper.py
```diff
@@ -177,14 +177,17 @@ def test_powerset(iterable, map_func, expected_powerset):
...
     model.stream_dict = stream_dict
     model.initial_data = initial_data
     model.muted_topics = muted_topics
     model.muted_streams = muted_streams
-    assert classify_unread_counts(model) == dict(classified_unread_counts,
-                                                 **vary_in_unreads)
+    model.index = initial_index
+    assert classify_unread_counts(model) == dict(
+            classified_unread_counts, **vary_in_unreads)
+    assert model.index['unread_msg_ids'] == index_unread_message_ids
```
neiljp: If `classify_unread_counts` returns a tuple as I mention elsewhere, then the function can be tested without reference to the index, which is simpler.

zulipterminal/model.py
```diff
@@ -280,6 +280,7 @@ def toggle_message_star_status(self, message: Dict[str, Any]) -> None:
     def mark_message_ids_as_read(self, id_list: List[int]) -> None:
         if not id_list:
             return
+        self.index['unread_msg_ids'].difference_update(id_list)
```
neiljp: This change should probably be accompanied by another test addition.

zulipterminal/helper.py
```diff
@@ -365,14 +368,21 @@ def classify_unread_counts(model: Any) -> UnreadCounts:
     )
 
     for pm in unread_msg_counts['pms']:
-        count = len(pm['unread_message_ids'])
+        msg_ids = pm['unread_message_ids']
+        count = len(msg_ids)
+        msg = {'type': 'pm', 'sender_id': pm['sender_id']}
+        unread_msgs.update(zip(msg_ids, [msg] * count))
```
kaustubh-nair: When updating counts of unread messages, we will need information about which stream/pm the message_id belongs to. 
Since the current index does not contain all the messages, simply storing message_ids will not be enough. Hence I've modified the index datastructure

zulipterminal/helper.py
```diff
@@ -131,11 +129,12 @@ def set_count(id_list: List[int], controller: Any, new_count: int) -> None:
     all_msg = controller.view.home_button
     all_pm = controller.view.pm_button
     for id in id_list:
-        user_id = messages[id]['sender_id']
+        # FIXME
+        #user_id = messages[id]['sender_id']
```
kaustubh-nair: This needs to be addressed properly, since `sender_id` data is not available for streams. Should I do a lookup from `index['messages']` ?

zulipterminal/helper.py
```diff
@@ -152,9 +151,10 @@ def set_count(id_list: List[int], controller: Any, new_count: int) -> None:
                                                    new_count)
                         break
             # FIXME: Update unread_counts['unread_topics']?
-            if ([messages[id]['display_recipient'], msg_topic] in
-                    controller.model.muted_topics):
-                add_to_counts = False
+            # FIXME
+            #if ([messages[id]['display_recipient'], msg_topic] in
+                    #controller.model.muted_topics):
+                #add_to_counts = False
```
kaustubh-nair: What does display_recipient mean?

sumanthvrao: `display_recipient` has the intended recipient of the message. In case of messages to stream this will be the stream name. If to a PM it will be things like User ID, email, full_name, etc. 

Try grepping `display_recipient` in our code-base and read through the tests. You'll get examples of how display_recipients can take on different forms.

zulipterminal/model.py
```diff
@@ -287,7 +287,6 @@ def mark_message_ids_as_read(self, id_list: List[int]) -> None:
             'flag': 'read',
             'op': 'add',
         })
-        set_count(id_list, self.controller, -1)  # FIXME Update?
```
kaustubh-nair: This is moved to `update_message_flags`. Also what is this 'Update' FIXME?

tests/model/test_model.py
```diff
@@ -1221,6 +1221,51 @@ def test_update_star_status(self, mocker, model, event_op,
...
+        ([1, 2], [1, 2]),
+        ([1], [1, 2]),
+        ([], [1, 2]),
+        ([1, 2], []),
+    ])
+    @pytest.mark.parametrize('event_op, flags_before, flags_after', [
+        ('add', [], ['read']),
+        ('add', ['read'], ['read']),
+        ('add', ['starred'], ['starred', 'read']),
+        ('add', ['read', 'starred'], ['read', 'starred']),
```
neiljp: You've got similar parametrization here as for the starred case. I understand that the `'remove` event shouldn't occur, but perhaps we should add defensive code to ensure this, and test this?

The overlap with the star code/test does make me wonder if that test and others related to `update_message_flag_status` might need reviewing. Have you checked those?

tests/model/test_model.py
```diff
@@ -1221,6 +1221,51 @@ def test_update_star_status(self, mocker, model, event_op,
...
+        mocker.patch('zulipterminal.model.Model.update_rendered_view')
+        set_count = mocker.patch('zulipterminal.model.set_count')
+
+        model.update_message_flag_status(event)
+
+        changed_ids = set(indexed_ids) & set(event_message_ids)
+        for changed_id in changed_ids:
+            assert model.index['messages'][changed_id]['flags'] == flags_after
+        (model.update_rendered_view.
+         has_calls([mocker.call(changed_id) for changed_id in changed_ids]))
```
neiljp: This should be `assert_has_calls`, which is also a bug in the other test.

tests/model/test_model.py
```diff
@@ -1205,13 +1205,60 @@ def test_update_star_status(self, mocker, model, event_op,
         changed_ids = set(indexed_ids) & set(event_message_ids)
         for changed_id in changed_ids:
             assert model.index['messages'][changed_id]['flags'] == flags_after
-        (model.update_rendered_view.
-         has_calls([mocker.call(changed_id) for changed_id in changed_ids]))
+            assert model.update_rendered_view.has_calls(
+                    [mocker.call(changed_id)])
```
kaustubh-nair: I've pushed this line inside the loop, since it's basically the same. Also added the assert statement.

tests/model/test_model.py
kaustubh-nair: @neiljp I've fixed a missing assertion in the other test. Apart from that I think it looks fine

zulipterminal/model.py
kaustubh-nair: Moved `difference update` to `update_message_flag_status`

tests/model/test_model.py
kaustubh-nair: Fixed

kaustubh-nair:
@neiljp I've reorganized the commits and fixed the other issues. The unread counts now get updated for indexed messages. The next step would be to store all the unread data in the index and modify `set_counts` to rely entirely on `index['unread_msgs']` instead of `index['messages']` to calculate unread counts.

tests/model/test_model.py
```diff
@@ -1205,13 +1205,60 @@ def test_update_star_status(self, mocker, model, event_op,
...
 
         for unchanged_id in (set(indexed_ids) - set(event_message_ids)):
             assert (model.index['messages'][unchanged_id]['flags'] ==
                     flags_before)
```
neiljp: Given the change to the method to add the `set_count`, we should ensure it is not called in this test?

tests/model/test_model.py
```diff
@@ -1205,13 +1205,60 @@ def test_update_star_status(self, mocker, model, event_op,
...
+    def test_update_read_status(self, mocker, model, event_op,
+                                event_message_ids, indexed_ids,
+                                flags_before, flags_after):
+        model.index = dict(messages={msg_id: {'flags': flags_before}
+                                     for msg_id in indexed_ids})
+        event = {
+            'messages': event_message_ids,
+            'type': 'update_message_flags',
+            'flag': 'read',
+            'operation': 'add',
```
neiljp: This value is hardcoded in this test, different to the starred one, so the add/remove in the parameters are not used.

You've also got only a few of the `'remove'` cases in the parametrization; we should cover more of these (like in the starred test) since remove+read gives the exception case.

neiljp:
@kaustubh-nair I'm not sure that `has_calls` is part of the mock features? I'd suggest printing it and checking it's doing what you think.

kaustubh-nair:
@neiljp Some important points:

- I think it's a good idea to merge the first commit, I'll fix the tests and push the changes soon.
- Re to your question about using `unread_msg_ids`: I don't think we can _do_ anything with just the unread message ids. If we want to update the unread counts, we need to know which stream/topic it belongs to. This is available in the initial data, which we do not store anywhere. So we can't really retrieve it using the message id from anywhere. Though, if have any solution in mind please let me know
- I just realized that the `set_counts` method isn't being tested anywhere? Can you please confirm this? If this is true, I could write a test for it before modifying it to accommodate the new datastructure (Maybe in a different pull, to avoid clutter) Also, I guess you could feel safer about making changes to `set_counts` if it is well tested.

tests/model/test_model.py
kaustubh-nair: Not really, a similar fix that has been applied in `test_update_unread_status` needs to be applied here. Would it be more appropriate to fix this in a different commit/PR ?

tests/model/test_model.py
kaustubh-nair: Added

tests/model/test_model.py
```diff
@@ -1199,19 +1203,80 @@ def test_update_star_status(self, mocker, model, event_op,
...
 
         model.update_message_flag_status(event)
 
         changed_ids = set(indexed_ids) & set(event_message_ids)
         for changed_id in changed_ids:
             assert model.index['messages'][changed_id]['flags'] == flags_after
         (model.update_rendered_view.
-         has_calls([mocker.call(changed_id) for changed_id in changed_ids]))
+         assert_has_calls([mocker.call(changed_id)
+                          for changed_id in changed_ids]))
```
kaustubh-nair: This was not being asserted. It's not related to this commit, I can add a separate commit if think it would be more appropriate

kaustubh-nair:
@neiljp I've fixed the tests. Also, I've added a comment above explaining a few other things

neiljp:
@kaustubh-nair Thanks for getting this work going! :+1:  I've merged the first commit as-is, but have expanded quite a bit on the commit text, as this feels like an important change. The commit is 013f898cb4cfc21f80f05fc78d86b8b1e3c32f61 :tada: While this just covers those in the index, this feels like a great first step :)

If you'd like to work on the message-ids extension, then feel free to continue it in this PR, or elsewhere if you'd prefer to reduce the github noise.

neiljp:
Re the message_ids, we do have this data to begin with in `unread_msgs` from the server, so perhaps a dict mapping message id to a 'category' (stream, topic, pm, huddle, ...) could work?

kaustubh-nair:
v2 is being continued in https://github.com/zulip/zulip-terminal/pull/535

PR #498, opened 2020-02-01
Refactor & improve narrowing logic

neiljp:
This is a more straightforward version of #449, splitting things up to make the main refactoring and some of the implicit changes more obvious.

This version also provides a parameter to the extracted method to specify the `anchor` for the narrow, which clarifies when message focus is retained on changing narrow (if possible).

The last commit then extends the message focus retention to all-messages and all-PMs narrowing; this is a specific change in behavior, but with this refactored version it should be more obvious what is happening.

NOTE: I've not changed the behavior for starred messages, as it didn't seem to be an issue - but we should investigate this.

NOTE: I may extract the `stream_id` setting into the model, but I think this is clean enough to merge now.


neiljp:
Just rebased to fix conflict after merging isort PR.

neiljp:
I'm merging this now since it'd be good to get in to clean up the narrowing logic, and there was positive feedback on the one-commit version (#449).

PR #499, opened 2020-02-02
Be consistent with import statements by linting/formatting using isort

neiljp:
This looks like a reasonable linting and auto-formatting approach, which generally has less of a one-size-fits-all approach, leaves our imports looking broadly as they currently do - yet automatically standardizing them :+1:

I know this has been discussed on the server repo; we have slightly different formatting (historically, at least).

It will be interesting to see if the soon-to-be-released v5 of isort has some exciting new features.

@SGeetansh We discussed this, so would be interested to know what you think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,9 @@ cache:
 matrix:
   include:
     - python: 3.5
-      name: "CPython 3.5 pytest (default linux)"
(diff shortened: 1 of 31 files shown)


SGeetansh:
Sure, isort is good to start with. Why not start using pylint or any other linting tool to improve the linting of the whole codebase instead?

PR #500, opened 2020-02-03
requirements: Update Urwid from 2.0.1 to 2.1.0.

rht:
Urwid 2.1.0 has neat features such as the true color support.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ url = "https://pypi.python.org/simple"
 name = "pypi"
 
 [packages]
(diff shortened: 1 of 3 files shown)


neiljp:
@rht Thanks for noticing this :+1: I don't see any problem in merging this promptly, so will do so shortly.

This should also simplify the need to exclude a pytest DeprecationWarning for urwid. It also claims to add support for python 3.7/3.8, which we test against and claim support for, so it's useful to know that they're officially supported in `urwid` too now!

PR #503, opened 2020-02-12
Improve CI speed & effectiveness

neiljp:
* Use `pip3` in travis instead of `pipenv` - this alone reduces runtime significantly
* Split linting and testing dependencies from general 'dev' dependencies in `setup.py`, allowing unnecessary package installation to be avoided in travis
* Now that jobs run faster individually, it's practical to set up linting in separate jobs, making it easier for developers to see what CI aspect failed
* On each platform, first install the package, then install testing packages for running tests; this is a partial fix towards #181.

I merged #502 pretty quickly so that this would be easier to discuss/review. However, I think the increased speed from pip over pipenv makes this a pretty clear improvement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,24 +19,24 @@ matrix:
       # Note that running isort on the lowest version should ensure changes are compatible
       script: ./tools/run-isort-check
     - python: 3.5
-      name: "Pytest - CPython 3.5 (default linux)"
+      name: "Install & pytest - CPython 3.5 (default linux)"
     - python: 3.6
-      name: "Pytest - CPython 3.6 (default linux)"
+      name: "Install & pytest - CPython 3.6 (default linux)"
     - python: 3.7
-      name: "Pytest - CPython 3.7 (default linux)"
+      name: "Install & pytest - CPython 3.7 (default linux)"
     - python: 3.8
-      name: "Pytest - CPython 3.8 (default linux)"
+      name: "Install & pytest - CPython 3.8 (default linux)"
     - python: pypy3.5
-      name: "Pytest - PyPy 3.5 (default linux)"
+      name: "Install & pytest - PyPy 3.5 (default linux)"
     - os: osx
-      name: "Pytest - Python 3 (OS X; xcode10)"
+      name: "Install & pytest - Python 3 (OS X; xcode10)"
       cache:
       - $HOME/Library/Caches/pip
       osx_image: xcode10
       sudo: required
       language: generic
 install:
-  - pip3 install .[testing]
+  - pip3 install . && pip3 install .[testing]
 script:
   - pytest
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,12 +26,12 @@ matrix:
       sudo: required
       language: generic
 install:
-  - pip3 install .[dev]
+  - pip3 install .[testing]
 script:
   - pytest
-  - if [[ "$RUN_MYPY" == "yes" ]]; then pip3 install -r requirements.txt && ./tools/run-mypy; fi
-  - if [[ "$RUN_ISORT_CHECK" == "yes" ]]; then pip3 install -r requirements.txt && ./tools/run-isort-check; fi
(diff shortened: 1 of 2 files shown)


kaustubh-nair:
@neiljp Have you thought about using build stages to group jobs? Jobs in a group can run in parallel. This can improve CI time even more. https://docs.travis-ci.com/user/build-stages/

neiljp:
@kaustubh-nair Thanks for bringing that idea up; I've read about them but I'm not convinced they're useful for us (yet)?

The jobs do already run in parallel to a great extent in any case, with the fast linting processes first now. I agree that grouping them according to eg. linting vs install/test could be useful (including visually), but I'm not convinced yet that requiring one set to pass first helps. While contributors can run linting/tests locally, If they rely on CI then they might see only one set of failures first, then they need to re-push before seeing a potential other failure that they then need to fix and push again. I can see that knowing if tests fail fast can prompt action quicker, so that could be beneficial, but the tests run pretty quick already.

In any case the speed improvements from using just pip and the benefits of the other changes seem like independent improvements to using build stages, so I'd be happy to discuss it further and can certainly see it being a separate PR on top of this.

.travis.yml
```diff
@@ -28,13 +26,12 @@ matrix:
...
-  - pip install pipenv==11.10
-  - pipenv install --three --dev Pipfile
+  - pip3 install .[dev]
 script:
-  - pipenv run pytest
-  - if [[ "$RUN_MYPY" == "yes" ]]; then pipenv run pip install -r requirements.txt && pipenv run ./tools/run-mypy; fi
-  - if [[ "$RUN_ISORT_CHECK" == "yes" ]]; then pipenv run pip install -r requirements.txt && pipenv run ./tools/run-isort-check; fi
-  - if [[ "$RUN_PYCODESTYLE" == "yes" ]]; then pipenv run pip install -r requirements.txt && pipenv run pycodestyle; fi
+  - pytest
+  - if [[ "$RUN_MYPY" == "yes" ]]; then pip install -r requirements.txt && ./tools/run-mypy; fi
```
rht: Why is this line `pip` instead of `pip3`?

rht: Oh, this was fixed in the second commit. But that's confusing, should have been done in first commit.

.travis.yml
```diff
@@ -26,12 +26,12 @@ matrix:
...
       language: generic
 install:
-  - pip3 install .[dev]
+  - pip3 install .[testing]
 script:
   - pytest
-  - if [[ "$RUN_MYPY" == "yes" ]]; then pip install -r requirements.txt && ./tools/run-mypy; fi
-  - if [[ "$RUN_ISORT_CHECK" == "yes" ]]; then pip install -r requirements.txt && ./tools/run-isort-check; fi
-  - if [[ "$RUN_PYCODESTYLE" == "yes" ]]; then pip install -r requirements.txt && pycodestyle; fi
+  - if [[ "$RUN_MYPY" == "yes" ]]; then pip3 install .[linting] && ./tools/run-mypy; fi
```
rht: The deps on Travis are installed by copying Travis cache, which is fast. I doubt the speedup is substantial? How much is the speedup anyway?

.travis.yml
neiljp: Right; I missed this in trying to make clear the progression of changes. Will be updated in the next push.

.travis.yml
neiljp: Are you referring to the speedup from this particular change? I agree that they should be cached, but this is also for clarity of which packages are installed - I'll update that.

rht: Yes, for this particular change. the `pipenv` -> `pip3` is definitely a speedup. It's not a big issue though.

rht:
LGTM

neiljp:
@rht Thanks for the review; I've adjusted the intermediate `pip` usage and amended the commit text for the `travis/setup: ...` commit to indicate it is not just for speed reasons.

Regarding the approximate general performance improvements:
* before: ~7m30+ (total 17m+)
* after: ~4m30 (total 12m+)

The latter is also with split linting/testing, and doing an install separate from a test install on each non-linting run. So, there is a little extra work in the extra jobs, but this is still much faster than using `pipenv` as before.

PR #504, opened 2020-02-17
tools: Add tools/lint-all script to simplify local linting.

neiljp:
Adjust development section of README to indicate use of tools/lint-all for linting, rather than the separate tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,11 +290,13 @@ Once you have a development environment set up, you might find the following use
 | Run normally | `zulip-term` | `pipenv run zulip-term` |
 | Run in debug mode | `zulip-term -d` | `pipenv run zulip-term -d` |
 | Run with profiling | `zulip-term --profile` | `pipenv run zulip-term --profile` |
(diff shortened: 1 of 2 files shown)


sumanthvrao:
@neiljp The changes LGTM!

Further: What are your thoughts on having a `test-all` script similar to Zulip server repo, which tests both the pytest and the linting?

neiljp:
@sumanthvrao `test-all` sounds good, to just run everything. It might be useful to synchronize aspects with CI too. However, I'm slightly wary of how useful it is to have one command when we have only two categories to run.

neiljp:
@sumanthvrao I'm going to merge this for now, as I feel it has as a clear development benefit as it stands.

Separate to the aspect of having a script which 'just' runs two commands, we'll be explicitly using `pytest` for scripts and `tools/lint-all` for wrapping linting here - and so `test-all` could also be a confusing name (does it include linting? or jut testing?). We can definitely discuss this, though :)

PR #506, opened 2020-02-18
Type annotations: Improve keypress and mouse_event signatures

neiljp:
This PR attempts to improve the consistency of these methods, based on their types as defined in urwid.

Currently this defines `urwid_Size` rather loosely, but I wanted to set some kind of standard so that when we make additions then contributors can compare to the existing code and see this. The type alias can always be changed later if we want to.

@amanagr The `FIXME` in `ui.py` is the issue I was referring to you a while back.

### DIFF
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -145,6 +145,7 @@ class View(urwid.WidgetWrap):
         if visible:
             self.body.focus_position = 2
 
+    # FIXME: The type of size should be urwid_Size; this needs checking
     def keypress(self, size: Tuple[int, int], key: str) -> str:
         self.model.new_user_input = True
         if self.controller.editor_mode:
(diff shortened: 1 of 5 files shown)


zulipterminal/ui.py
```diff
@@ -13,6 +13,7 @@
 from zulipterminal.ui_tools.views import (
     LeftColumnView, MiddleColumnView, RightColumnView,
 )
+from zulipterminal.urwid_types import urwid_Size
```
sumanthvrao: do we still want to import it here, even if we don't use it yet?

neiljp: I did notice this, but if we can fix/check the usage near the FIXME, then this wouldn't be an issue, but I'm not sure if @amanagr is too busy to look right now. If so, I'll merge it quicker and we can address that remaining type difference later.

From memory the type hinting is not always more stringent, but just more closely matching the expectations described in the urwid documentation. That might inspire some review of the behavior of these methods.

neiljp:
@sumanthvrao I've removed the import but left the comment for now, and will merge this soon.

sumanthvrao:
@neiljp Sounds good. 👍

Issue #469, opened 2019-08-31
Migrate keypress tests to use keys_for_command


neiljp:
Current keypress tests use hardcoded tests against specific keys. It would seem cleaner to use `keys_for_command` to both ensure this works for multiple specified keys, and also to automatically update if the standard keys are updated in `config/keys.py`.

This might be able to be simplified by writing an appropriate parametrized fixture.

preetmishra:
@zulipbot claim

neiljp:
Thanks to the work of @preetmishra in #523 this issue has been resolved :+1:

zulipbot:
Hello @preetmishra, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

zee-bit:
@neiljp any particular reason why it's still partially resolved?
A quick `git grep` indicates we have covered almost all cases of hardcoded keys in keypress events.

neiljp:
I'm not entirely sure originally why this was left open. However, there are still explicit checks and/or calls to keypress with specific characters - that's not necessarily test-specific, though things (including tests) may break if we customized the keys.

zee-bit:
Right. There are few instances of hard-coded keys that are not specific to tests, do you want me to fix those?
I have a small UI improvement PR that I was about to propose. I can push this refactor into that PR -- though, this may not be a good practice.
Or, I can make a new PR, with a few more reactors if I could find?

neiljp:
Generally we want all the keys to be flexible to allow for customization, so explicitly calling with particular keys is quite likely to break at some point.

A separate PR is fine - it can often be quicker and easier to review, as long as it's straightforward. If there is another commit then they all need to be considered.

PR #518, opened 2020-02-25
Add support for navigation using keys 'j' and 'k' in help popup menu.

### LINKED ISSUE #517: Add support for navigation using keys 'j' and 'k' in help popup menu.
Keys 'j' and 'k' do not work for navigation in the help menu. We should add support for the same.

##### Steps to reproduce
1. Login to zulip-term.
2. Press '?' to popup the help menu.
3. Press 'j' or 'k' to navigate.

##### Screenshot
Screenshot from 2020-02-24 19-32-33

preetmishra:
@neiljp, I have added two separate commits.
- The former updates labels and descriptions for 'up/k' and 'down/j' keys. 
- The latter adds support for navigation using 'j' and 'k' keys in the help menu. I have also removed the redundant 'up/down scrolls' text from the menu's title.

Fixes #517.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1101,6 +1101,23 @@ class TestHelpMenu:
         self.help_view.keypress(size, key)
         assert self.controller.exit_popup.called
 
+    @pytest.mark.parametrize('key, expected_key', [
+        (key, expected_key)
+        for keys, expected_key in [
+            (keys_for_command('GO_UP'), 'up'),
+            (keys_for_command('GO_DOWN'), 'down'),
+            (keys_for_command('SCROLL_UP'), 'page up'),
+            (keys_for_command('SCROLL_DOWN'), 'page down'),
+            (keys_for_command('GO_TO_BOTTOM'), 'end'),
+        ]
+        for key in keys
+    ])
+    def test_keypress_navigation(self, mocker, key, expected_key):
+        size = (200, 20)
+        super_view = mocker.patch(VIEWS + '.urwid.ListBox.keypress')
+        self.help_view.keypress(size, key)
+        super_view.assert_called_once_with(size, expected_key)
+
 
 class TestPopUpConfirmationView:
     @pytest.fixture
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -24,14 +24,14 @@ KEY_BINDINGS = OrderedDict([
         'excluded_from_random_tips': False,
         'key_category': 'general',
     }),
-    ('PREVIOUS_MESSAGE', {
+    ('GO_UP', {
         'keys': {'k', 'up'},
-        'help_text': 'Previous message',
+        'help_text': 'Go up/Previous message',
         'key_category': 'navigation',
     }),
-    ('NEXT_MESSAGE', {
+    ('GO_DOWN', {
         'keys': {'j', 'down'},
-        'help_text': 'Next message',
+        'help_text': 'Go down/Next message',
         'key_category': 'navigation',
     }),
     ('GO_LEFT', {
@@ -44,19 +44,19 @@ KEY_BINDINGS = OrderedDict([
         'help_text': 'Go right',
         'key_category': 'navigation',
     }),
-    ('SCROLL_TO_TOP', {
+    ('SCROLL_UP', {
         'keys': {'K', 'page up'},
-        'help_text': 'Scroll to top',
+        'help_text': 'Scroll up',
         'key_category': 'navigation',
     }),
-    ('SCROLL_TO_BOTTOM', {
+    ('SCROLL_DOWN', {
         'keys': {'J', 'page down'},
-        'help_text': 'Scroll to bottom',
+        'help_text': 'Scroll down',
         'key_category': 'navigation',
     }),
-    ('END_MESSAGE', {
+    ('GO_TO_BOTTOM', {
         'keys': {'G', 'end'},
-        'help_text': 'Go to last message in view',
+        'help_text': 'Go to bottom/last message in view',
         'key_category': 'navigation',
     }),
     ('REPLY_MESSAGE', {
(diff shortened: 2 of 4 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -790,7 +790,11 @@ def __init__(self, controller: Any) -> None:
         super().__init__(self.log)
 
     def keypress(self, size: urwid_Size, key: str) -> str:
-        if is_command_key('GO_BACK', key) or is_command_key('HELP', key):
+        if is_command_key('SCROLL_UP', key):
```
neiljp: You may wish to include the other movement keys here too - ie. 'K'/'J' equivalents?

zulipterminal/config/keys.py
```diff
@@ -24,14 +24,14 @@
         'excluded_from_random_tips': False,
         'key_category': 'general',
     }),
-    ('PREVIOUS_MESSAGE', {
+    ('SCROLL_UP', {
```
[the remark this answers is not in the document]
preetmishra: I agree. I'll replace the tokens/labels with the suggestions you have made.

preetmishra:
@neiljp, Thanks for the review. I agree we should make the navigation consistent throughout the user interface which includes adding support for navigation using 'j/k' and 'J/K' in other popups as well. I will add a follow-up PR regarding this.

I have made the requested changes which include updating labels and description for the suggested keys and added support for other movement keys ('J/K'). It is ready for review.

zulipterminal/ui_tools/views.py
```diff
@@ -792,6 +792,12 @@ def __init__(self, controller: Any) -> None:
     def keypress(self, size: urwid_Size, key: str) -> str:
         if is_command_key('GO_BACK', key) or is_command_key('HELP', key):
             self.controller.exit_popup()
+        elif is_command_key('GO_UP', key) \
+                or is_command_key('SCROLL_UP', key):
+            key = 'up'
+        elif is_command_key('GO_DOWN', key) \
+                or is_command_key('SCROLL_DOWN', key):
+            key = 'down'
```
neiljp: I would think we'd want the `SCROLL_*` keys (and `GO_TO_BOTTOM`) to do what they would normally do, not just move up/down.

preetmishra:
@neiljp, I have updated the navigation keys and added the tests. However, I have not written much test code so I am not sure if the proposed implementation is what we are looking for. It'd be helpful to have your insight on this to improve it further.

tests/ui/test_ui_tools.py
```diff
@@ -1101,6 +1101,15 @@ def test_keypress_goback(self, key):
...
         assert self.controller.exit_popup.called
 
+    @pytest.mark.parametrize('key', [
+        'up', 'down', 'j', 'k', 'J', 'K',
+        'page up', 'page down', 'G', 'end',
+    ])
+    def test_keypress_navigation(self, key):
+        size = (200, 20)
+        self.controller.keypress(size, key)
+        assert self.controller.keypress.called
```
neiljp: As written, this test is always going to pass. `self.controller` is set to be a `Mock` in the first method, as the way these are currently set up (due to `autouse=True`) that method is a little like an `__init__` for each test call. So the `assert` is testing that an explicit function call you made on a mock was actually made...which it was, in the test itself :)

You likely want `self.help_view` instead of `self.controller` on line 1110. This is the object/function you're actually testing, and when reading tests you'll maybe see some kind of setup/mocking, always a function call of the thing that's being tested, and then the `assert` type checks which specify what the end result should be, given the setup and call.

You'd also want to re-think the form of the parametrize here and use  `keys_for_command` to generate lists of tuples, rather than hard-coding in all the keys in the parametrize - in particular since the `assert` code really needs to consider the 'input' key and the resulting key passed to urwid, which are not always the same.

preetmishra:
@neiljp, Thanks for the feedback that you have provided. I have updated the tests based on that. I have added `expected_key` to assert whether the correct key is being passed to urwid for each 'input key'.

tests/ui/test_ui_tools.py
```diff
@@ -1101,6 +1101,19 @@ def test_keypress_goback(self, key):
...
+        (keys_for_command('GO_DOWN'), 'down'),
+        (keys_for_command('SCROLL_UP'), 'page up'),
+        (keys_for_command('SCROLL_DOWN'), 'page down'),
+        (keys_for_command('GO_TO_BOTTOM'), 'end'),
+    ])
+    def test_keypress_navigation(self, keys, expected_key):
+        size = (200, 20)
+        for key in keys:
+            return_value = self.help_view.keypress(size, key)
+            assert return_value == expected_key
```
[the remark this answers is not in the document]
preetmishra: Thanks for the tip. Adding a link explaining pytest's options in the README would be helpful.

I have used list comprehension to generate parameters and removed the loop from the test code. Regarding the latter point, I have pushed an implementation without the call to the keypress method on the super class for now as urwid's implementation suggests that the value returned by the super's keypress function is the key which was passed to it. 
Urwid's link: https://github.com/urwid/urwid/blob/0b3bd90b2482e2d528aa37608304783ccadb43e2/urwid/listbox.py#L934

However, if this doesn't seem convincing, I have posted another implementation in #zulip-terminal. 
Conversation link: https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/.5Bhelp.20popup.5D.20Navigation.20using.20'j'.20and.20'k'.2E/near/825457

preetmishra:
@neiljp, Thanks. I have rewritten the commit messages and removed irrelevant files from the commit message. Gitlint is pretty handy. I have also amended the tests based on your input, more on that is here: https://github.com/zulip/zulip-terminal/pull/518#discussion_r387224356.

preetmishra:
@neiljp, I have amended the test to mock the super class and check how its `keypress` method was called. It is ready for review.

neiljp:
@preetmishra I just merged this - thanks for seeing this through to the end! :tada: 

I made a few very minor adjustments to the commit text and squashed the last two commits together as I indicated previously. The commits in master are be801c72a6e0296218c9e8d3302c57d412f35e54 and the preceding one.

It was good to see an alternative solution for the test in the end - I had used `itertools` in my exploration of how it could work. There are quite a few other cases of hard-coded keys that would benefit from the general approach with the helper function - see #469.

preetmishra:
Thanks, I got to learn more about the pytest library while working on this one. I enjoyed it.
I'll proceed with the mentioned issue next.

PR #519, opened 2020-03-04
Added feature: Show user info by pressing 'i'

### LINKED ISSUE #511: Feature: Enable view of user info
Just like the stream descriptions, on pressing `i`, while a user is in focus, the full description of the user should pop-up which must include the information available in the `View full profile` section on the web app.

SGeetansh:
Fixes #511 
This pull request implements a feature of displaying the information of the user in focus. All the information that is visible in the web app is present.
It doesn't pass the tests right now. Please suggest any required changes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,7 @@ gitlint==0.10.0
 mypy==0.740
 isort==4.3.21
 emoji==0.5.0
+pytz==2019.3
 urwid_readline==0.10
 beautifulsoup4==4.6.0
 lxml==4.2.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,5 +13,4 @@ emoji==0.5.0
 urwid_readline==0.10
 beautifulsoup4==4.6.0
 lxml==4.2.3
-
 zipp==1.0.0  # To support Python 3.5
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -200,6 +200,11 @@ KEY_BINDINGS = OrderedDict([
         'help_text': 'View stream description',
         'key_category': 'stream_list',
     }),
+    ('USER_INFO', {
+        'keys': {'i'},
+        'help_text': 'View user description',
+        'key_category': 'user_action',
+    }),
     ('REDRAW', {
         'keys': {'ctrl l'},
         'help_text': 'Redraw screen',
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -1,7 +1,9 @@
 import os
+import pytz
 import signal
 import sys
 import time
+from datetime import datetime
 from functools import partial
 from platform import platform
 from typing import Any, List, Optional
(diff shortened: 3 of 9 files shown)


zulipterminal/config/keys.py
```diff
@@ -200,6 +200,11 @@
         'help_text': 'View stream description',
         'key_category': 'stream_list',
     }),
+    ('USER_INFO', {
```
sumanthvrao: It would also benefit to have a note for this in `README`. (Maybe these two changes can be extracted into a separate commit)

zulipterminal/model.py
```diff
@@ -79,14 +79,14 @@ def __init__(self, controller: Any) -> None:
         self.stream_id = -1
         self.recipients = frozenset()  # type: FrozenSet[Any]
         self.index = initial_index
-
+
```
sumanthvrao: Some extra spaces ;)

zulipterminal/ui_tools/buttons.py
```diff
@@ -213,6 +215,11 @@ def _narrow_with_compose(self, button: Any) -> None:
         self._view.body.focus.original_widget.set_focus('footer')
         self._view.write_box.private_box_view(self)
 
+    def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
+        if is_command_key('USER_INFO', key):
+            self.model.controller.show_user_info(self.user_id)
```
sumanthvrao: alternatively, can we could just use `self.controller.show_user_info` here , and avoid the line on top? :)

zulipterminal/core.py
```diff
@@ -120,6 +122,65 @@ def show_stream_info(self, color: str, name: str, desc: str) -> None:
         show_stream_view = StreamInfoView(self, color, name, desc)
         self.show_pop_up(show_stream_view, "# {}".format(name))
 
+    def show_user_info(self, user_id:int) -> None:
+        user_id = user_id
+        url = 'users/' + str(user_id)
+        result = self.client.call_endpoint(
```
sumanthvrao: It's always good to check if the result was a success or not here. We do that in many places so you can have a look.

SGeetansh: I checked the other places. They just check if the response message was 'success' or not, with a return statement at the end of the function (`return response['result'] == success`). Do you want me to just write that only? Sorry, if I am wrong, but can you please elaborate on this? I didn't put a try-except here because I think it isn't required.

sumanthvrao: Yup, that's the correct way to do the validation. No need of a try-catch block.

zulipterminal/config/keys.py
```diff
@@ -200,6 +200,11 @@
...
+        'keys': {'i'},
```
sumanthvrao: You could extract the README and keys into a sepearte commit as well. As it stands, this commit has a lot of file changes.

requirements.txt
```diff
@@ -13,5 +13,4 @@ emoji==0.5.0
 urwid_readline==0.10
 beautifulsoup4==4.6.0
 lxml==4.2.3
-
-zipp==1.0.0  # To support Python 3.5
+zipp==1.0.0  # To support Python 3.5
```
sumanthvrao: Unwanted diff? :)

zulipterminal/ui_tools/views.py
```diff
@@ -1,5 +1,6 @@
 import threading
 import time
+import math
```
sumanthvrao: Isn't `import math` used in your third commit. If so, let's import it only in the commit where it is used.

zulipterminal/core.py
```diff
@@ -120,9 +122,64 @@ def show_stream_info(self, color: str, name: str, desc: str) -> None:
...
+        if res_data['is_admin']:
+            display_data['Role'] = "Admin"
+        elif res_data['is_guest']:
+            display_data['Role'] = "Guest"
+        else:
+            display_data['Role'] = "Member"
+
+        show_userinfo_view = UserInfoView(self, display_data)
+        self.show_pop_up(show_userinfo_view, "{}".format(res_data['full_name']))
+        return response['result'] == 'success'
```
sumanthvrao: Where do you validate whether the return value was true/false?

zulipterminal/ui_tools/views.py
```diff
@@ -850,13 +850,26 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
 
 
 class UserInfoView(urwid.ListBox):
-    def __init__(self, controller, display_data) -> None:
+    def __init__(self, controller, display_data:dict) -> None:
```
sumanthvrao: Maybe type-hint more?

zulipterminal/core.py
SGeetansh: Can you point me to an instance where this is done? I don't know what to do.

sumanthvrao: @SGeetansh I meant you could do this at the beginning of this function after you make the API call. Only of the response result is a success, you can parse and store the parameters as you have done.

neiljp:
@SGeetansh I hope you had a good Summer - just checking in if you planned to update this PR?

If you do plan to update this:
* Some timezone code has been added, so the dependency commit shouldn't be necessary now
* The popup code has been streamlined, which may well make this easier - see `PopUpView` and child classes
* Just having the available information from the existing presences data would be good in a working commit, then following up with custom profile fields in a later one - the titles need sourcing from somewhere as I think the values you're using are hard-coded from chat.zulip.org?

neiljp:
Closing, as this was replaced via #848.

@SGeetansh Thanks for filing the issue and taking a shot here :+1:

PR #520, opened 2020-03-05
Displays '(You)' after current user on rightColumn

souravtecken:
The PR prepends a "(You)" indicating that the user is the current user. I've chosen to prepend the indicator to avoid the issues faced in #468, where the indicator fell to the next line. 

Screenshot from 2020-03-05 09-48-08

This is how it looks on my terminal.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -966,7 +966,8 @@ class TestRightColumnView:
                 view=self.view,
                 width=width,
                 color=self.view.users[0]['status'],
-                count=1
+                count=1,
+                is_current_user=False
             )
         users_view.assert_called_once_with(right_col_view.users_btn_list)
         assert len(right_col_view.users_btn_list) == users_btn_len
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -20,13 +20,19 @@ class TopButton(urwid.Button):
                  show_function: Callable[..., Any], width: int,
                  prefix_character: Union[str, Tuple[Any, str]]='\N{BULLET}',
                  text_color: Optional[str]=None,
-                 count: int=0) -> None:
+                 count: int=0,
+                 postfix_caption: Union[str, Tuple[Any, str]]='') -> None:
         if isinstance(prefix_character, tuple):
             prefix = prefix_character[1]
         else:
             prefix = prefix_character
         assert len(prefix) in (0, 1)
+        if isinstance(postfix_caption, tuple):
+            self.postfix_text = postfix_caption[1]
+        else:
+            self.postfix_text = postfix_caption
         self._caption = caption
+        self.postfix_caption = postfix_caption
         self.prefix_character = prefix_character
         self.post_prefix_spacing = ' ' if prefix else ''
         self.count = count
@@ -53,7 +59,7 @@ class TopButton(urwid.Button):
     def update_widget(self, count_text: str) -> Any:
         # Note that we don't modify self._caption
         max_caption_length = (self.width_for_text_and_count -
-                              len(count_text))
+                              len(count_text) - len(self.postfix_text))
         if len(self._caption) > max_caption_length:
             caption = (self._caption[:max_caption_length-1] +
                        '\N{HORIZONTAL ELLIPSIS}')
@@ -61,13 +67,14 @@ class TopButton(urwid.Button):
             caption = self._caption
         num_extra_spaces = (
             self.width_for_text_and_count - len(count_text) - len(caption)
+            - len(self.postfix_text)
         )
 
         # NOTE: Generated text does not include space at end
         self._w = urwid.AttrMap(urwid.SelectableIcon(
             [' ', self.prefix_character, self.post_prefix_spacing,
              '{}{}'.format(caption, num_extra_spaces*' '),
-             ' ', ('idle',  count_text)],
+             ' ', self.postfix_caption, ('idle',  count_text)],
             self.width_for_text_and_count+5),  # cursor location
             self.text_color,
             'selected')
@@ -188,7 +195,8 @@ class StreamButton(TopButton):
 class UserButton(TopButton):
     def __init__(self, user: Dict[str, Any], controller: Any,
                  view: Any, width: int,
-                 color: Optional[str]=None, count: int=0) -> None:
+                 color: Optional[str]=None, count: int=0,
+                 is_current_user: bool=False) -> None:
         # Properties accessed externally
         self.email = user['email']
         self.user_id = user['user_id']
@@ -204,7 +212,9 @@ class UserButton(TopButton):
                          prefix_character=(color, '\N{BULLET}'),
                          text_color=color,
                          width=width,
-                         count=count)
+                         count=count,
+                         postfix_caption=('current_user_indicator', '(You)')
+                         if is_current_user else '')
 
     def _narrow_with_compose(self, button: Any) -> None:
         # Switches directly to composing with user
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -603,7 +603,10 @@ class RightColumnView(urwid.Frame):
                     view=self.view,
                     width=self.width,
                     color=user['status'],
-                    count=unread_count
+                    count=unread_count,
+                    is_current_user=(
+                        self.view.model.user_id == user['user_id']
+                    )
                 )
             )
         user_w = UsersView(users_btn_list)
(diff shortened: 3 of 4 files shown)


souravtecken:
@preetmishra, Thanks for your review! I've now fixed those isses.

sumanthvrao:
@souravtecken Thanks! I did look at your implementation (not the tests yet) briefly and it seems good.

More from a PoC point of view, this looks OK for me. @neiljp how do you think having `(you)` on the left looks. It certainly is a trade-off between additional complexity vs better look.

preetmishra:
Another workaround could be truncating the name and display something like 'first_name + (you)'.

zulipterminal/ui_tools/buttons.py
```diff
@@ -199,7 +200,8 @@ def __init__(self, user: Dict[str, Any], controller: Any,
         self.recipients = frozenset({self.user_id, view.model.user_id})
 
         super().__init__(controller,
-                         caption=user['full_name'],
+                         caption="(You) " + user['full_name'] if \
+                                  is_current_user else user['full_name'],
```
neiljp: This would be clearer with a separate local variable before the `super()` call; generally if a one-line conditional wraps like this, it's *probably* clearer defined elsewhere, particularly if variables are repeated.

zulipterminal/ui_tools/views.py
```diff
@@ -603,7 +603,9 @@ def users_view(self, users: Any=None) -> Any:
                     view=self.view,
                     width=self.width,
                     color=user['status'],
-                    count=unread_count
+                    count=unread_count,
+                    is_current_user=True if self.view.model.user_id == \
+                                    user['user_id'] else False
```
neiljp: This can be replaced with simply the expression itself.

souravtecken:
@neiljp I've incorporated your suggestions and have gone with leaving some space on the right in the burron for an optional text and reducing the user's name if need be. This is how it looks as of now
Screenshot from 2020-03-13 09-20-41

zulipterminal/ui_tools/buttons.py
```diff
@@ -197,14 +205,15 @@ def __init__(self, user: Dict[str, Any], controller: Any,
 
         # FIXME Is this still needed?
         self.recipients = frozenset({self.user_id, view.model.user_id})
-
+        postfix_caption = (color, '(You)') if is_current_user else ''
```
neiljp: You may have seen in the other work on this issue, that another style was added to the theme and used instead of `color` - I'd suggest we do that here, now that you have the layout working.

zulipterminal/ui_tools/buttons.py
```diff
@@ -52,22 +54,27 @@ def update_count(self, count: int) -> None:
...
             self.width_for_text_and_count - len(count_text) - len(caption)
+            - len(postfix_text)
         )
 
         # NOTE: Generated text does not include space at end
         self._w = urwid.AttrMap(urwid.SelectableIcon(
             [' ', self.prefix_character, self.post_prefix_spacing,
              '{}{}'.format(caption, num_extra_spaces*' '),
-             ' ', ('idle',  count_text)],
+             ' ', self.postfix_caption, ('idle',  count_text)],
```
neiljp: If we ever had both a `postfix_caption` and `count_text`, then I expect we'd want a space between them, which would have to be added manually right now. However, this is fine for now.

souravtecken:
> @souravtecken This works just fine now 
> 
> The main change I'd like to see is to move the tuple check logic into `__init__`, which may involve other logic changes too.
> 
> Currently this is probably just OK to fit in one commit, but you can see that the summary line sort of overflows into the commit text body, sort of like "it also does this", so be careful. Similarly, if you feel like you need a long commit title, or want to put commas or 'and' into a commit title, think whether all the code is sufficiently connected, or can be constructed in multiple commits that are independent and fit together like a story.
> 
> Do you plan to add tests for this new behavior? Let us know if you need advice with regard to that.

@neiljp Right, I get what you're saying about the commit messages, thanks for the review!
About tests, I'm not really familiar with writing tests, so I've been looking through, trying to understand what's happening. I think I'd need some help regarding that.

zulipterminal/ui_tools/buttons.py
```diff
@@ -204,7 +212,9 @@ def __init__(self, user: Dict[str, Any], controller: Any,
                          prefix_character=(color, '\N{BULLET}'),
                          text_color=color,
                          width=width,
-                         count=count)
+                         count=count,
+                         postfix_caption=('current_user_indicator', '(You)')
+                         if is_current_user else '')
```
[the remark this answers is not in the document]
souravtecken: I did, yes. I had some mypy errors thrown when it was a seperate variable, I'm not able to reproduce that error now, I'll put it back where it was.

zulipterminal/ui_tools/buttons.py
```diff
@@ -52,22 +54,27 @@ def update_count(self, count: int) -> None:
 
     def update_widget(self, count_text: str) -> Any:
         # Note that we don't modify self._caption
+        if isinstance(self.postfix_caption, tuple):
+            postfix_text = self.postfix_caption[1]
+        else:
+            postfix_text = self.postfix_caption
```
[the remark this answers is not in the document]
souravtecken: Got it, yes. Will work on it!

neiljp:
@souravtecken For now, if you'd like to try the tuple-check change, and we can discuss tests after that, potentially in a separate PR.

zulipterminal/ui_tools/buttons.py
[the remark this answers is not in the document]
neiljp: Hmm, I thought the linting was passing in your previous version I reviewed.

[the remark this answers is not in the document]
souravtecken: It did pass in my previous version, I'd kept the type then as `Union[Any, Tuple[Any, str]]`. That didn't throw any linting errors. But in this version, I've changed it to `Union[str, Tuple[Any, str]]`. So, I had to remove it as a seperate variable for it to pass.

tests/ui/test_ui_tools.py
```diff
@@ -966,7 +966,8 @@ def test_users_view(self, users, users_btn_len, editor_mode, status,
                 view=self.view,
                 width=width,
                 color=self.view.users[0]['status'],
-                count=1
+                count=1,
+                is_current_user=False
```
neiljp: I think it would be worth adding a `FIXME` comment here; this could be where we put the test later, or maybe nearby.

zulipterminal/ui_tools/buttons.py
```diff
@@ -53,21 +59,22 @@ def update_count(self, count: int) -> None:
...
-                              len(count_text))
+                              len(count_text) - len(self.postfix_text))
         if len(self._caption) > max_caption_length:
             caption = (self._caption[:max_caption_length-1] +
                        '\N{HORIZONTAL ELLIPSIS}')
         else:
             caption = self._caption
         num_extra_spaces = (
             self.width_for_text_and_count - len(count_text) - len(caption)
+            - len(self.postfix_text)
```
neiljp: Do we actually need to keep track of `self.postfix_text`? Can we include it in another variable? This would make this cleaner.

zulipterminal/ui_tools/buttons.py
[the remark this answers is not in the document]
neiljp: This gives a different error for me, but you should be able to use an explicit type comment here to avoid this.

zulipterminal/ui_tools/buttons.py
souravtecken: Since `self.postfix_caption` is a of type `Union[str, Tuple[Any, str]]`. Determining the length would require me to check instance. To avoid checking instance within the `update_widget` method, I had to keep track of `self.postfix_text`. Another approach would be to modify the parameters taken in by `update_count` and `update_widget`, or rather introducing an `update_postfix` method, which I thought might be a little messy. So, what exactly do I do here?

souravtecken: Or I could replace `self.postfix_text` with a `self.postfix_length`, would that be okay?

neiljp: Well, we definitely need the full (`self.`)`postfix_caption`, for rendering :+1: 

What I'm saying is that the `self.postfix_text` is only used for its length, and it's already known in the `__init__`, so it can be pre-computed. It could be as `self.postfix_length`, but I think it could be simplified further.

neiljp:
@souravtecken Thanks for working on this - I just merged an alternative version so am closing this now.

I hope you found this interesting to work on and will consider other contributions in future!

PR #523, opened 2020-03-05
tests: Migrate keypress tests to use keys_for_command.

### LINKED ISSUE #469: Migrate keypress tests to use keys_for_command

preetmishra:
This updates keypress tests to use `keys_for_command` to ensure that the tests are future-proof and they work for multiple specified keys. @neiljp

Fixes #469.

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -217,9 +217,10 @@ class TestView:
         view.model.controller.show_all_mentions.assert_called_once_with(view)
         assert view.body.focus_col == 1
 
+    @pytest.mark.parametrize('key', keys_for_command('SEARCH_PEOPLE'))
     @pytest.mark.parametrize('autohide', [True, False], ids=[
         'autohide', 'no_autohide'])
-    def test_keypress_w(self, view, mocker, autohide):
+    def test_keypress_autohide_users(self, view, mocker, autohide, key):
         view.users_view = mocker.Mock()
         view.body = mocker.Mock()
         view.controller.autohide = autohide
@@ -234,17 +235,17 @@ class TestView:
 
         view.body.focus_position = None
 
-        # Test "w" keypress
-        view.keypress(size, "w")
-        view.users_view.keypress.assert_called_once_with(size, "w")
+        view.keypress(size, key)
+        view.users_view.keypress.assert_called_once_with(size, key)
         assert view.body.focus_position == 2
         view.user_search.set_edit_text.assert_called_once_with("")
         assert view.controller.editor_mode is True
         assert view.controller.editor == view.user_search
 
+    @pytest.mark.parametrize('key', keys_for_command('SEARCH_STREAMS'))
     @pytest.mark.parametrize('autohide', [True, False], ids=[
         'autohide', 'no_autohide'])
-    def test_keypress_q(self, view, mocker, autohide):
+    def test_keypress_autohide_streams(self, view, mocker, autohide, key):
         view.stream_w = mocker.Mock()
         view.left_col_w = mocker.Mock()
         view.stream_w.stream_search_box = mocker.Mock()
@@ -261,10 +262,8 @@ class TestView:
 
         view.body.focus_position = None
 
-        # Test "q" keypress
-        view.keypress(size, "q")
-
-        view.left_panel.keypress.assert_called_once_with(size, "q")
+        view.keypress(size, key)
+        view.left_panel.keypress.assert_called_once_with(size, key)
         assert view.body.focus_position == 0
         view.stream_w.stream_search_box.set_edit_text.\
             assert_called_once_with("")
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -462,13 +462,17 @@ class TestStreamsView:
         stream_view.set_focus.assert_called_once_with("body")
         assert stream_view.log == self.streams_btn_list
 
+    @pytest.mark.parametrize('search_streams_key',
+                             keys_for_command('SEARCH_STREAMS'))
+    @pytest.mark.parametrize('go_back_key', keys_for_command('GO_BACK'))
     @pytest.mark.parametrize('current_focus, stream', [
         (0, 'FOO'),
         (2, 'fan'),
         (4, 'BOO'),
     ])
     def test_return_to_focus_after_search(self, mocker, stream_view,
-                                          current_focus, stream):
+                                          current_focus, stream,
+                                          search_streams_key, go_back_key):
         # Initialize log
         stream_view.streams_btn_list = [
             mocker.Mock(stream_name=stream_name) for stream_name in [
@@ -482,14 +486,12 @@ class TestStreamsView:
         previous_focus_stream_name = stream
 
         # Toggle Stream Search
-        key = "q"
         size = (20,)
-        stream_view.keypress(size, key)
+        stream_view.keypress(size, search_streams_key)
 
         # Exit Stream Search
-        key = "esc"
         size = (20,)
-        stream_view.keypress(size, key)
+        stream_view.keypress(size, go_back_key)
 
         # Obtain new stream focus
         new_focus = stream_view.log.get_focus()[1]
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1679,6 +1679,7 @@ class TestMessageBox:
             assert label[0].text == 'EDITED'
             assert label[1][1] == 7
 
+    @pytest.mark.parametrize('key', keys_for_command('EDIT_MESSAGE'))
     @pytest.mark.parametrize('to_vary_in_each_message, realm_editing_allowed,\
                              expect_editing_to_succeed', [
         ({'sender_id': 2, 'timestamp': 45}, True, False),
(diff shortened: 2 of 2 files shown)


tests/ui/test_ui.py
```diff
@@ -194,9 +198,17 @@ def test_keypress(self, view, mocker):
...
+            (autohide, key)
+            for autohide, keys in [
+                (True, keys_for_command('SEARCH_PEOPLE')),
+                (False, keys_for_command('SEARCH_PEOPLE')),
+            ]
+            for key in keys
+        ],
+        ids=['autohide', 'no_autohide']
+    )
+    def test_keypress_w(self, view, mocker, autohide, key):
```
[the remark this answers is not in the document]
preetmishra: @neiljp Certainly. It will pass the tests but the verbose won't be that much explicit.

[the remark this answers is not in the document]
preetmishra: @neiljp Certainly. If the number of keys changes for a particular command in the future, this will break.

preetmishra:
@neiljp Thanks. I couldn't get `$ git add -e` to work for me (I have posted my query [here](https://chat.zulip.org/#narrow/stream/49-development-help/topic/Git.20add's.20edit/near/827343)) so I had to do it the other way.

I have split the commit into five separate commits. The first one consists of all the keypress tests that have a simple parametrization structure and the rest of them have their own type of keypress tests. I have also renamed tests to be more descriptive of what they do and removed unnecessary comments and extended the test for the function mentioned in https://github.com/zulip/zulip-terminal/pull/523#discussion_r388736763.
Lastly, I have removed ids from the tests since the number of keys for a particular command can change and fixing a particular number of ids with tests can break the tests in the future.

neiljp:
@preetmishra Splitting this up has made it much easier to look at, particularly given the first commit has quite a lot of lines - they're mostly simple, but mixing them with non-simple changes made it more difficult to review. The first commit was almost good to go, so I've merged part of it as cd39f07273790467aab69bc83d223217945ef54a :tada:

The part I didn't merge of the first commit was the ui tests, as not all of the simple changes in that file had been made, whereas ui_tools looked to be fully covered by that first commit.

I've also merged a slightly adjusted version of the third commit, which was very similar to the test you already added elsewhere - I modified the commit text a little, needed to import key_for_command (which you had in an earlier commit) and appended `_navigation` to the test name: ed46b06945e6bb5e882dca858184b31b20db5dd9 :tada:

Regarding the ids, these would be useful to keep if we can - I've not used it before, but perhaps we could use something like a callable (eg. function) to provide the formatting for autohide? They also look rather complicated - it's possible something like the itertools approach I tried might be simpler?

Note that your 'areas' (ui, ui_tools) don't quite match what was updated in each test, which was one thing I adjusted when merging, but is easier if you can check beforehand in future :)

Given these as a whole will fix #469, I removed the 'fixes' text where you had it.

preetmishra:
@neiljp Thank you so much. I have covered all the simple keypress tests of `ui` in the first commit. I have split my comments into two sections:

**Ids**:
I have used a function for generating ids for autohide tests. However, I couldn't use a function for 'edit_message' test because I couldn't come up with a way to distinguish between the last two boolean parameters in each test case (apparently the id function only takes one parameter) so I had to use another way to do it.

**List comprehension vs Itertools**:
I tried to implement the parameterization with itertools' product function but it is nearly as complex. Plus, to bind a string together `[ ]` needs to be added.

```python
# Itertools
ls = [
    (['up', 'k'], ['up']),
    (['down', 'j'], ['down'])
]
ls = [
    (key, expected_key)
    for pair in ls
    for key, expected_key in it.product(*pair)
]

# List comprehension
ls = [
    (['up', 'k'], 'up'),
    (['down', 'j'], 'down')
]

ls = [
    (key, expected_key)
    for keys, expected_key in ls
    for key in keys
]
```


Lastly, I haven't added 'Fixes ...' in any commit due to the reason you stated and have fixed the 'areas' part of the commits.

tests/ui/test_ui.py
```diff
@@ -217,9 +217,21 @@ def test_keypress_ALL_MENTIONS(self, view, mocker, key):
...
-    def test_keypress_w(self, view, mocker, autohide):
+    def ids_autohide(param):
+        if isinstance(param, bool):
+            return 'autohide' if param else 'no_autohide'
+
+    @pytest.mark.parametrize('autohide, key', [
+            (autohide, key)
+            for autohide, keys in [
+                (True, keys_for_command('SEARCH_PEOPLE')),
+                (False, keys_for_command('SEARCH_PEOPLE')),
```
neiljp: Given the duplication, could we just loop this over the bool values? Same below.

tests/ui/test_ui.py
```diff
@@ -217,9 +217,21 @@ def test_keypress_ALL_MENTIONS(self, view, mocker, key):
         view.model.controller.show_all_mentions.assert_called_once_with(view)
         assert view.body.focus_col == 1
 
-    @pytest.mark.parametrize('autohide', [True, False], ids=[
-        'autohide', 'no_autohide'])
-    def test_keypress_w(self, view, mocker, autohide):
+    def ids_autohide(param):
+        if isinstance(param, bool):
+            return 'autohide' if param else 'no_autohide'
```
neiljp: This seems to work well :)

tests/ui/test_ui_tools.py
```diff
@@ -462,13 +462,23 @@ def test_keypress_GO_BACK(self, mocker, stream_view, key):
...
+    @pytest.mark.parametrize('current_focus, stream, \
+                             search_streams_key, go_back_key', [
+        (current_focus, stream, search_streams_key, go_back_key)
+        for current_focus, stream, search_streams_keys, go_back_keys in [
+            (0, 'FOO', keys_for_command('SEARCH_STREAMS'),
+                keys_for_command('GO_BACK')),
+            (2, 'fan', keys_for_command('SEARCH_STREAMS'),
+                keys_for_command('GO_BACK')),
+            (4, 'BOO', keys_for_command('SEARCH_STREAMS'),
+                keys_for_command('GO_BACK')),
```
neiljp: There's duplication here, as in the other commit, where I think we should just be able to loop over the same values each time.

tests/ui/test_ui_tools.py
```diff
@@ -1680,21 +1680,32 @@ def test_main_view_generates_EDITED_label(self, mocker,
...
-        ({'sender_id': 1, 'timestamp': 1}, True, False),
-        ({'sender_id': 1, 'timestamp': 45}, False, False),
-        ({'sender_id': 1, 'timestamp': 45}, True, True),
-    ], ids=['msg_sent_by_other_user',
-            'time_limit_esceeded',
-            'editing_not_allowed',
-            'all_conditions_met'])
+                             expect_editing_to_succeed, key', [
+        pytest.param(to_vary_in_each_message, realm_editing_allowed,
+                     expect_editing_to_succeed, key, id='-'.join([_id, key]))
```
neiljp: I think I remember seeing `pytest.param` somewhere, but this seems like an interesting usage - good find!

tests/ui/test_ui_tools.py
```diff
@@ -1680,21 +1680,32 @@ def test_main_view_generates_EDITED_label(self, mocker,
...
+                    keys_for_command('EDIT_MESSAGE'),
+                    'msg_sent_by_other_user'),
+                ({'sender_id': 1, 'timestamp': 1}, True, False,
+                    keys_for_command('EDIT_MESSAGE'),
+                    'time_limit_exceeded'),
+                ({'sender_id': 1, 'timestamp': 45}, False, False,
+                    keys_for_command('EDIT_MESSAGE'),
+                    'editing_not_allowed'),
+                ({'sender_id': 1, 'timestamp': 45}, True, True,
+                    keys_for_command('EDIT_MESSAGE'),
```
neiljp: As in the other cases, let's reduce this duplication of `keys_for_command` with the same parameter.

tests/ui/test_ui.py
preetmishra: Thanks. :)

tests/ui/test_ui_tools.py
preetmishra: Thanks. :)

preetmishra:
@neiljp Thanks for the suggestion, it makes the code simpler. I have adjusted the list comprehensions to eliminate duplications. It is ready for review.

amanagr:
@preetmishra this is great work! This will make keypress testing much easier. Thx.

preetmishra:
@amanagr @neiljp Thanks so much. It has been very exciting to work on this one. I have learned so much about the pytest library, the importance of clean code and commits, maintaining a clean PR so that it is easier to review and discuss upon.

Nested `parametrize` is a great suggestion. :+1: It eliminates the need for a callable (for generating ids for autohide tests) and `pytest.param`. It also reduces the code from 'XL' to 'M'. 

I have replaced the list comprehensions with nested parametrize calls and removed the unnecessary `ids_autohide` callable.

neiljp:
@preetmishra Thanks for working on this - I just manually merged the final commits squashed together :tada: (since while they are across files, the change is very similar). The commit is 3ff8a736dc822dd4f891a78fdaa12b9e4afb8cc4.

I agree with @amanagr that this really makes things cleaner :)

One thing I did notice is that some tests don't scale with this change, as some code does itself still rely on the hard-coded keys themselves. We should track that in an issue.

preetmishra:
@neiljp Thanks. It was a good learning experience. :) 

Yes, I have noticed them as well. I will gather more information on this and file an issue.

Issue #448, opened 2019-08-05
TRACKING: Typeahead improvements


sumanthvrao:
With the v1 implementation of #441 merged, here are a few followups in store :

- [x] Indicating the options one has to auto-complete in the footer (After user pressed `ctrl + f`) (fixed in #540)
- [x] Going back to the previous auto-complete. (fixed in #532)
- [x] Handling conflicts when 2 users have the same full name. (Possibly using `@**Name|User_id**` or email to distinguish) (fixed in #928)
- [x] Prioritizing mentions to users in the current narrow. (Eg: People who are subscribed to stream/narrow) (fixed in #730)
- [ ] Editing the 'starred' version (`#**test here**`) and autocompleting it further. (Place the cursor before last 2 stars
- [x] Auto complete emoji names in message content. (fixed in #525)
- [x] Support Group-autocomplete (Eg: `@*terminal*`)
- [x] Support topic typeahead in topic box (fixed in #, with spaces in #823)
- [x] Support topic stream+topic typeahead in message composition (fixed in #1004)
- [x] Allow spaces in autocomplete to work fully in composition (fixed in stream/topic boxes in #823) 
- [x] Support typeahead completion after eg. parentheses (fix #541)

Related ideas:
- [ ] Adjust layout of footer text to indicate text matching against results, separation between result parts, ...

sumanthvrao:
Update: We support group-autcomplete through e6972b0066946f35461e19bc1c238644767ba084 and 5 after

kaustubh-nair:
Update: Going back to previous autocomplete added through 74369b197bcce46e57266bde8d22550cf4cf47bc

sumanthvrao:
Update: We support auto-completing emoji names with 150a0fafd4f98700dfffce9ff3adfb69b379e049 and one after.

kaustubh-nair:
Mentioning the emoji download issue here https://github.com/zulip/zulip-terminal/issues/558

kaustubh-nair:
Follow-up: Add support to automatically show typeahead suggestions based on prefix string triggers.

neiljp:
Once #540 is merged, we can consider adjustments to the `[more]` text, to improve indications of how many there are (beyond 10) or maybe how to show the rest - if we don't find an alternative solution to this such as a scrolling footer.

kaustubh-nair:
#732 for improving ordering of group typeaheads

Ezio-Sarthak:
[...]

PR #525, opened 2020-03-06
Typeahead for emojis

kaustubh-nair:
This is a part of the typeahead follow-ups #448 

Here's a GIF
a


kaustubh-nair:
I chose to manually add the list of emojis ( from https://github.com/zulip/zulip/blob/master/tools/setup/emoji/emoji_names.py ) instead of getting it from the `emoji` package we already use because the web app does not support all the emojis provided by the package. If a user picks some unsupported emoji, it won't be rendered on the web-app. 
One problem with doing this is that the list of emojis won't be in sync with the web app. But I doubt it would be updated very frequently, so I guess it is okay?
One more thing to note is that I merged the dictionary into a list while omitting the emoji codes. I merged the canonical names and aliases together. We wouldn't be needing this data, I hope?

timabbott:
You could try downloading `static/generated/emoji/emoji_codes.json` from the server -- that I think will give you access to the mapping used by that Zulip server.

(This is the right model in general, since ideally the terminal app could be connected to multiple servers running different versions at the same time).

kaustubh-nair:
Oh that's awesome, thanks!

kaustubh-nair:
Thanks for the review @neiljp I'll make the changes and soon
Added a new gif; ordering is done by finding the first matching emoji from the list

tests/ui_tools/test_boxes.py
```diff
@@ -124,3 +124,23 @@ def test_autocomplete_streams(self, write_box, streams_fixture,
             streams_fixture[len(streams_fixture)//2:]]
         typeahead_string = write_box.autocomplete_streams(text, state)
         assert typeahead_string == required_typeahead
+
+    @pytest.mark.parametrize('text, state, required_typeahead', [
+        (':rock_o', 0, ':rock_on:'),
+        (':rock_o', 1, ':rock_o'),
+        (':smi', 0, ':smiley:'),
+        (':smi', 1, ':smirk:'),
+        (':smi', 2, ':smile:'),
```
neiljp: This was what I meant by the order of cycling - it's not too bad as it is, but my expectation was that it would be in alphabetical order.

neiljp:
@kaustubh-nair I'll wait to review again until you have this passing tests, unless you have a question, but just wanted to clarify the ordering issue I mentioned.

Note that you can add text in the commit other than the title; at least in ZT we note if we're amending tests (to pass changed code/behavior) or adding them (for a new function/behavior). Also, being descriptive in commit titles/text is useful, as long as it's not duplicating text elsewhere :)

Lastly, this PR is likely a good foundation to try adding different reactions to messages using `:`, as in the webapp. We've supported `+` for some time, but general ones would be good, if potentially more tricky as it's not in the message window itself.

tests/ui_tools/test_boxes.py
kaustubh-nair: If we want to keep it alphabetical, we could perhaps sort the entire list of emojis after retrieving it. This would be a one time thing, so it shouldn't be much of a performance issue.

kaustubh-nair:
Whoops, didn't realize tests were failing :sweat_smile: 
I've added more descriptive commit messages. ZT has a very organized commit system. It's a little hard to get used to, but I can see why it's useful :) 
Also, do you think I should explore the solution Tim suggested in this PR, or maybe make a follow-up issue for it?

neiljp:
@kaustubh-nair From what I can tell of Tim's comment on #zulip-terminal, we probably should get a proper API for this from the server.

Did you automate the conversion process from the server file to the python list of emoji added in the first commit? If that changes, or we use an API call, then having stages of that conversion might be useful to retain - and at minimum to use as documentation, if our data gets updated manually in future.

kaustubh-nair:
I've added a script that fetches the dictionary from github and converts it into the list that we want.

kaustubh-nair:
@neiljp I've added a python script instead of using 'sed' since it's easier to maintain. Also, sorted the list of emojis so the suggestions should now be displayed in alphabetical order.

kaustubh-nair:
I can also open an issue for extending this to use an API endpoint, unless you want to wait to merge this until we get the endpoint?

kaustubh-nair:
Added some tests for cycling backwards.
This PR is ready for a review.

tools/fetch-emoji-names-list
```diff
@@ -0,0 +1,13 @@
...
+set -x
+
+FILENAME="zulipterminal/emoji_names.py"
+SCRIPT="tools/convert-emoji-names-list"
+
+curl https://raw.githubusercontent.com/zulip/zulip/master/tools/setup/emoji/emoji_names.py > $FILENAME
+
+sed -i '1d' $FILENAME  # delete first line containing imports
+sed -i 's/EMOJI_NAME_MAPS = //g' $FILENAME  # delete variable name
+python3 $SCRIPT $FILENAME
```
neiljp: I see you're using the script here, but but was a little confused as I thought you were using python instead of `sed`?

kaustubh-nair:
https://travis-ci.org/github/zulip/zulip-terminal/jobs/664746621?utm_medium=notification&utm_source=github_status This build failed, without any specific error :thinking: 
Let me try restarting

kaustubh-nair:
Thanks for the review @neiljp `eval` was returning an error because the file had a line with imports, hence I had to use `sed` to delete the first line. I'm importing the file directly using python now, so it should be much easier now. 

I've added a test for ensuring the list is sorted in `test_boxes.py`. I'm not sure if it belongs there, but the list is being used in `boxes.py` so I thought it made sense.

tools/fetch-emoji-names
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python3
+import requests
+
+
+OUTPUT_FILE = "zulipterminal/emoji_names.py"
```
neiljp: I'm not too keen on this overwriting the previous version.

tools/fetch-emoji-names
```diff
@@ -0,0 +1,12 @@
...
+
+# download emojis dictionary
+URL = "https://raw.githubusercontent.com/zulip/zulip/master/tools/setup/emoji/emoji_names.py"
+r = requests.get(URL)
+
+with open(OUTPUT_FILE, 'w') as f:
+    f.write(r.text)
```
neiljp: On success, output a message to run the conversion script?

tests/ui_tools/test_boxes.py
```diff
@@ -136,3 +137,6 @@ def test_autocomplete_streams(self, write_box, streams_fixture,
             streams_fixture[len(streams_fixture)//2:]]
         typeahead_string = write_box.autocomplete_streams(text, state)
         assert typeahead_string == required_typeahead
+
+    def test_autocomplete_emojis_default_list_sorted(self):
+        assert EMOJI_NAMES == sorted(EMOJI_NAMES)
```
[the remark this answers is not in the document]
neiljp: We could have a test which ensures the small fixture supports this too?

neiljp:
@kaustubh-nair I just remembered, you should set the execute mode on your scripts too.

kaustubh-nair:
@neiljp This is ready!
I'll open the issue for the api call

neiljp:
@kaustubh-nair I just manually merged this :tada:

The only change I made was regarding the emoji tests. I moved them to fit the structure for other files, and since they were not named correctly they weren't running, and were actually failing for a simple reason which I fixed. See 2c30caf34e8b65d1d1ad8f0a61e27068089bd941 and the prior commit.

neiljp:
If you could reference the API call issue in the tracking issue that'd be useful :)

PR #510, opened 2020-02-22
views: Fix stream and topic search.

preetmishra:
This commit fixes the stream and topic search to find if the search query is a substring of any stream/topic.

Fixes #509.

I'll proceed to add the tests if the implementation looks good to you, @neiljp.

PR #530, opened 2020-03-09
Improve typeahead suggestion and search for streams

preetmishra:
@neiljp This is a follow up to https://github.com/zulip/zulip-terminal/pull/510#pullrequestreview-363060680. 

The previous implementation only did a `startswith` lookup which caused it to miss out on the suggestions that were expected. I have fixed that with a substring lookup. I have also amended the test cases as now the suggestions are being given in an sorted (non-decreasing) order.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -409,23 +409,29 @@ class TestStreamsView:
         self.stream_search_box.assert_called_once_with(
             stream_view, 'SEARCH_STREAMS', stream_view.update_streams)
 
-    @pytest.mark.parametrize('new_text, expected_log', [
-        ('f', ['fan', 'FOO', 'foo', 'FOOBAR']),
-        ('a', ['bar', 'fan', 'FOOBAR']),
-        ('bar', ['bar', 'FOOBAR']),
-        ('foo', ['FOO', 'foo', 'FOOBAR']),
-        ('FOO', ['FOO', 'foo', 'FOOBAR']),
-        ('test', ['test here']),
-        ('here', ['test here']),
+    @pytest.mark.parametrize('new_text, expected_log, to_pin', [
+        ('f', ['fan', 'FOO', 'foo', 'FOOBAR'], []),
+        ('bar', ['bar'], []),
+        ('foo', ['FOO', 'foo', 'FOOBAR'], []),
+        ('FOO', ['FOO', 'foo', 'FOOBAR'], []),
+        ('test', ['test here'], []),
+        ('here', ['test here'], []),
+        ('test here', ['test here'], []),
+        # With 'foo' pinned.
+        ('f', ['foo', 'fan', 'FOO', 'FOOBAR'], [['foo'], ]),
+        ('FOO', ['foo', 'FOO', 'FOOBAR'], [['foo'], ]),
     ])
-    def test_update_streams(self, mocker, stream_view, new_text, expected_log):
+    def test_update_streams(self, mocker, stream_view, new_text, expected_log,
+                            to_pin):
         stream_names = [
             'FOO', 'FOOBAR', 'foo', 'fan',
             'boo', 'BOO', 'bar', 'test here',
         ]
         stream_names.sort(key=lambda stream_name: stream_name.lower())
+        self.view.pinned_streams = to_pin
+        stream_names.sort(key=lambda stream_name: stream_name in [
+            stream[0] for stream in to_pin], reverse=True)
         self.view.controller.editor_mode = True
-        new_text = new_text
         search_box = "SEARCH_BOX"
         stream_view.streams_btn_list = [
             mocker.Mock(stream_name=stream_name)
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -410,11 +410,11 @@ class TestStreamsView:
             stream_view, 'SEARCH_STREAMS', stream_view.update_streams)
 
     @pytest.mark.parametrize('new_text, expected_log', [
-        ('f', ['FOO', 'FOOBAR', 'foo', 'fan']),
-        ('a', ['FOOBAR', 'fan', 'bar']),
-        ('bar', ['FOOBAR', 'bar']),
-        ('foo', ['FOO', 'FOOBAR', 'foo']),
-        ('FOO', ['FOO', 'FOOBAR', 'foo']),
+        ('f', ['fan', 'FOO', 'foo', 'FOOBAR']),
+        ('a', ['bar', 'fan', 'FOOBAR']),
+        ('bar', ['bar', 'FOOBAR']),
+        ('foo', ['FOO', 'foo', 'FOOBAR']),
+        ('FOO', ['FOO', 'foo', 'FOOBAR']),
         ('test', ['test here']),
         ('here', ['test here']),
     ])
@@ -423,6 +423,7 @@ class TestStreamsView:
             'FOO', 'FOOBAR', 'foo', 'fan',
             'boo', 'BOO', 'bar', 'test here',
         ]
+        stream_names.sort(key=lambda stream_name: stream_name.lower())
         self.view.controller.editor_mode = True
         new_text = new_text
         search_box = "SEARCH_BOX"
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -14,8 +14,20 @@ class TestWriteBox:
         self.view.model = mocker.Mock()
 
     @pytest.fixture()
-    def write_box(self, mocker):
+    def write_box(self, mocker, users_fixture, user_groups_fixture,
+                  streams_fixture, emojis_fixture):
         write_box = WriteBox(self.view)
+        write_box.view.users = users_fixture
+        write_box.model.user_group_names = [
+            groups['name'] for groups in user_groups_fixture]
+
+        write_box.view.pinned_streams = []
+        write_box.view.unpinned_streams = [
+            [stream['name']] for stream in
+            streams_fixture]
+
+        mocker.patch('zulipterminal.ui_tools.boxes.emoji_names',
+                     EMOJI_NAMES=emojis_fixture)
         return write_box
 
     def test_init(self, write_box):
@@ -83,12 +95,8 @@ class TestWriteBox:
         ('@_', 4, None),  # Beyond end
         ('@_', -1, '@_**Human 2**'),
     ])
-    def test_generic_autocomplete_mentions(self, write_box, users_fixture,
-                                           text, state, required_typeahead,
-                                           user_groups_fixture):
-        write_box.view.users = users_fixture
-        write_box.model.user_group_names = [
-            groups['name'] for groups in user_groups_fixture]
+    def test_generic_autocomplete_mentions(self, write_box, text,
+                                           required_typeahead, state):
         typeahead_string = write_box.generic_autocomplete(text, state)
         assert typeahead_string == required_typeahead
 
@@ -112,14 +120,8 @@ class TestWriteBox:
         ('#St', 1, '#**Stream 2**'),
         ('#Stream 1', 0, '#**Stream 1**'),
     ])
-    def test_generic_autocomplete_streams(self, write_box, streams_fixture,
-                                          text, state, required_typeahead):
-        write_box.view.pinned_streams = [
-            [stream['name']] for stream in
-            streams_fixture[:len(streams_fixture) // 2]]
-        write_box.view.unpinned_streams = [
-            [stream['name']] for stream in
-            streams_fixture[len(streams_fixture) // 2:]]
+    def test_generic_autocomplete_streams(self, write_box, text,
+                                          state, required_typeahead):
         typeahead_string = write_box.generic_autocomplete(text, state)
         assert typeahead_string == required_typeahead
 
@@ -141,10 +143,8 @@ class TestWriteBox:
         (':nomatch', 0, None),
         (':nomatch', -1, None),
         ])
-    def test_generic_autocomplete_emojis(self, write_box, emojis_fixture, text,
+    def test_generic_autocomplete_emojis(self, write_box, text,
                                          mocker, state, required_typeahead):
-        mocker.patch('zulipterminal.ui_tools.boxes.emoji_names',
-                     EMOJI_NAMES=emojis_fixture)
         typeahead_string = write_box.generic_autocomplete(text, state)
         assert typeahead_string == required_typeahead
 
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -22,9 +22,9 @@ class TestWriteBox:
             groups['name'] for groups in user_groups_fixture]
 
         write_box.view.pinned_streams = []
-        write_box.view.unpinned_streams = [
+        write_box.view.unpinned_streams = sorted([
             [stream['name']] for stream in
-            streams_fixture]
+            streams_fixture], key=lambda stream: stream[0].lower())
 
         mocker.patch('zulipterminal.ui_tools.boxes.emoji_names',
                      EMOJI_NAMES=emojis_fixture)
@@ -103,14 +103,14 @@ class TestWriteBox:
     @pytest.mark.parametrize('text, state, required_typeahead', [
         ('#Stream', 0, '#**Stream 1**'),
         ('#Stream', 1, '#**Stream 2**'),
-        ('#S', 0, '#**Some general stream**'),
-        ('#S', 1, '#**Secret stream**'),
+        ('#S', 0, '#**Secret stream**'),
+        ('#S', 1, '#**Some general stream**'),
         ('#S', 2, '#**Stream 1**'),
         ('#S', 3, '#**Stream 2**'),
         ('#S', -1, '#**Stream 2**'),
         ('#S', -2, '#**Stream 1**'),
-        ('#S', -3, '#**Secret stream**'),
-        ('#S', -4, '#**Some general stream**'),
+        ('#S', -3, '#**Some general stream**'),
+        ('#S', -4, '#**Secret stream**'),
         ('#S', -5, None),
         ('#So', 0, '#**Some general stream**'),
         ('#So', 1, None),
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -100,28 +100,54 @@ class TestWriteBox:
         typeahead_string = write_box.generic_autocomplete(text, state)
         assert typeahead_string == required_typeahead
 
-    @pytest.mark.parametrize('text, state, required_typeahead', [
-        ('#Stream', 0, '#**Stream 1**'),
-        ('#Stream', 1, '#**Stream 2**'),
-        ('#S', 0, '#**Secret stream**'),
-        ('#S', 1, '#**Some general stream**'),
-        ('#S', 2, '#**Stream 1**'),
-        ('#S', 3, '#**Stream 2**'),
-        ('#S', -1, '#**Stream 2**'),
-        ('#S', -2, '#**Stream 1**'),
-        ('#S', -3, '#**Some general stream**'),
-        ('#S', -4, '#**Secret stream**'),
-        ('#S', -5, None),
-        ('#So', 0, '#**Some general stream**'),
-        ('#So', 1, None),
-        ('#Se', 0, '#**Secret stream**'),
-        ('#Se', 1, None),
-        ('#St', 0, '#**Stream 1**'),
-        ('#St', 1, '#**Stream 2**'),
-        ('#Stream 1', 0, '#**Stream 1**'),
+    @pytest.mark.parametrize('text, state, required_typeahead, to_pin', [
+        # With no streams pinned.
+        ('#Stream', 0, '#**Stream 1**', []),  # 1st-word startswith match.
+        ('#Stream', 1, '#**Stream 2**', []),  # 1st-word startswith match.
+        ('#Stream', 2, '#**Secret stream**', []),  # 2nd-word startswith match.
+        ('#Stream', 3, '#**Some general stream**', []),  # 3rd-word startswith.
+        ('#S', 0, '#**Secret stream**', []),  # 1st-word startswith match.
+        ('#S', 1, '#**Some general stream**', []),  # 1st-word startswith.
+        ('#S', 2, '#**Stream 1**', []),  # 1st-word startswith match.
+        ('#S', 3, '#**Stream 2**', []),  # 1st-word startswith match.
+        ('#S', -1, '#**Stream 2**', []),
+        ('#S', -2, '#**Stream 1**', []),
+        ('#S', -3, '#**Some general stream**', []),
+        ('#S', -4, '#**Secret stream**', []),
+        ('#S', -5, None, []),
+        ('#So', 0, '#**Some general stream**', []),
+        ('#So', 1, None, []),
+        ('#Se', 0, '#**Secret stream**', []),
+        ('#Se', 1, None, []),
+        ('#St', 0, '#**Stream 1**', []),
+        ('#St', 1, '#**Stream 2**', []),
+        ('#g', 0, '#**Some general stream**', []),
+        ('#g', 1, None, []),
+        ('#Stream 1', 0, '#**Stream 1**', []),  # Complete match.
+        ('#nomatch', 0, None, []),
+        ('#ene', 0, None, []),
+        # With 'Secret stream' pinned.
+        ('#Stream', 0, '#**Secret stream**',
+         [['Secret stream'], ]),  # 2nd-word startswith match (pinned).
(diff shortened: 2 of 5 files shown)


tests/ui_tools/test_boxes.py
```diff
@@ -99,8 +99,10 @@ def test_autocomplete_mentions(self, write_box, users_fixture,
...
 
     @pytest.mark.parametrize('text, state, required_typeahead', [
-        ('#Stream', 0, '#**Stream 1**'),
-        ('#Stream', 1, '#**Stream 2**'),
+        ('#Stream', 0, '#**Some general stream**'),
+        ('#Stream', 1, '#**Secret stream**'),
+        ('#Stream', 2, '#**Stream 1**'),
+        ('#Stream', 3, '#**Stream 2**'),
         ('#S', 0, '#**Some general stream**'),
         ('#S', 1, '#**Secret stream**'),
```
kaustubh-nair: Wouldn't it be more desirable if it matches a stream starting with some text, and only then do a full text search?

preetmishra:
@kaustubh-nair Thanks. We mostly try to imitate the Zulip's web app and it has a full string search so I think the changes should be incorporated.

Regarding mentions (or `match_user`), we have that already implemented in ZT (this thing has been mentioned in the https://github.com/zulip/zulip-terminal/pull/510#pullrequestreview-363060680 as well).

tests/ui_tools/test_boxes.py
preetmishra: I'd like to know what @neiljp thinks about this.

sumanthvrao: Intuitively, I would expect to get search results which match with the starting text first and later on those which match with intermediate text. Maybe a full text search can be implemented as a secondary sort.

preetmishra:
@kaustubh-nair @sumanthvrao Thanks for the feedback. I redid the implementation. Now the `autocomplete_streams` suggests the startswith first followed by full-text.

preetmishra:
@kaustubh-nair Thanks. Sorting is required to maintain the order of suggestions in this particular way. Moreover, I think it's unlikely for organizations to have more than 500 streams.

I have updated the commit title from _Improve suggestions from autocomplete_streams_ to _Improve typeahaed suggestions from autocomplete_streams_.

sumanthvrao:
As a datapoint, `czo` has 138 streams (as of today) and we are one of the very big zulip community out there :)

neiljp:
@sumanthvrao It's possible the recurse center has more streams, but I think we can optimize this if it becomes necessary.

@preetmishra This looks good; my only concern at this point is that searching in the streams box gives a different order than in the autocomplete (search for `ter`, have `terminal_clients`, `zulip-terminal` and `twitter` streams). Maybe you can find another test which triggers this.

There's also a typo in the commit text.

preetmishra:
@neiljp Thanks, I have fixed the typo and the order for the stream search box. 

The reason for that order in the stream search box was the way we build `streams_btn_list` (we put `pinned_stream` first in the list to show them at the top in the streams view), sorting this helps in showing the streams in alphabetical order while search.

neiljp:
@preetmishra I'm not sure if changing the order of the search box filtering is the right approach - depending on my pinning, I might want one response to always be at the bottom?

Competing factors could be:
* matches first word
* [matches n'th word?]
* matches text anywhere
* takes pinning order into account

I think we need to take pinning into account, rather than resort the list for stream searches - I almost wonder if we want to include a divider in the sorted list, to show which are pinned?

preetmishra:
@neiljp I have updated the PR as per our [conversation](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Stream.20typeahead.20.23T530) in **#zulip-terminal**. It is ready for review.

tests/ui_tools/test_boxes.py
```diff
@@ -105,23 +105,25 @@ def test_autocomplete_mentions(self, write_box, users_fixture,
...
 
     @pytest.mark.parametrize('text, state, required_typeahead', [
-        ('#Stream', 0, '#**Stream 1**'),
-        ('#Stream', 1, '#**Stream 2**'),
-        ('#S', 0, '#**Some general stream**'),
-        ('#S', 1, '#**Secret stream**'),
+        ('#Stream', 0, '#**Secret stream**'),
+        ('#Stream', 1, '#**Some general stream**'),
+        ('#Stream', 2, '#**Stream 1**'),
+        ('#Stream', 3, '#**Stream 2**'),
```
neiljp, line 111: Shouldn't this give the order of "Stream 1", "Stream 2", "Secret stream", "Some general stream" ?

Are there pinning considerations? If so, can we make that explicit in some way?

zulipterminal/helper.py
```diff
@@ -456,9 +456,7 @@ def match_stream(stream: Any, text: str) -> bool:
     False otherwise.
     """
     stream_name = stream[0].lower()
-    if stream_name.startswith(text.lower()):
-        return True
-    return False
+    return text in stream_name
```
neiljp, line 459: `text.lower()`?

zulipterminal/ui_tools/boxes.py
```diff
@@ -126,9 +126,21 @@ def autocomplete_mentions(self, text: str, state: int,
...
-                            if match_stream(stream, text[1:])]
+                            if match_stream(stream, query)]
+        stream_typeahead.sort(key=lambda stream_name: stream_name.lower())
+        # Give priority to the streams which start with the `query`.
+        stream_typeahead.sort(
+            key=lambda stream_name: stream_name.lower().startswith(query),
+            reverse=True)
+        # Order by `pin_to_top` status.
+        stream_typeahead.sort(key=lambda stream_name: stream_name in map(
+            lambda stream: stream[0], self.view.pinned_streams), reverse=True)
```
neiljp, line 140: Could we simplify this multiple sort by sorting a list of tuples?

zulipterminal/ui_tools/views.py
```diff
@@ -276,6 +276,16 @@ def update_streams(self, search_box: Any, new_text: str) -> None:
...
+            streams_display.sort(key=lambda stream: stream.stream_name.lower())
+            # Give priority to the streams which start with the `new_text`.
+            streams_display.sort(
+                key=(lambda stream:
+                     stream.stream_name.lower().startswith(new_text)),
+                reverse=True)
+            # Order by `pin_to_top` status.
+            streams_display.sort(key=lambda stream: stream.stream_name in map(
+                lambda stream: stream[0], self.view.pinned_streams),
+                reverse=True)
```
neiljp, line 288: This code looks very familiar to the other? Is there a reason you didn't refactor it first?

zulipterminal/helper.py
preetmishra: @neiljp I pass `query`, which is already in lowercase, to this function. I extracted `text[1:].lower()` as `query`.

tests/ui_tools/test_boxes.py
preetmishra: @neiljp Yes, there are pinning considerations.
The 'pinned_streams' are defined, in `test_autocomplete_streams`, through the following:
```
write_box.view.pinned_streams = [
        [stream['name']] for stream in
        streams_fixture[:len(streams_fixture)//2]
```
Would you like the function to be refactored?

zulipterminal/ui_tools/views.py
preetmishra: @neiljp I had thought of introducing `order_stream`, to sort the streams as per these criteria, but that required some pre-refactoring and structural changes in `update_streams` and `autocomplete_streams`.

Should I go ahead with this? Also, what could be the commit message for pre-refactoring commits like these (where some refactoring is required only for getting it ready for a 'new' feature)?

preetmishra:
@neiljp Thanks for pointing 'multiple word searches' out, I missed out on it.

If you could address a few queries that I have left in the inline-comments, it would assist me to move ahead with this. :)

tests/ui_tools/test_boxes.py
neiljp: My concern is that these parametrizations don't demonstrate what happens very clearly.

I would expect test cases showing each of the features we claim to support, if that's possible with the stream names we have. So `#stream` would (without pinning) prefer stream-1, stream-2, secret, then general; `#S` would just be alphabetical (?); more matches limit things (like `#so`, `#se` do already :+1:).

A comment may be sufficient here, though we may need a pinning of specific named streams in the test to make it clearer too?

zulipterminal/helper.py
neiljp: I did see that. Let's leave this pending until the sorting is simplified - that may clarify this.

zulipterminal/ui_tools/views.py
neiljp: Refactoring is generally fine to do first. Otherwise you can often end up duplicating and slightly altering code, only to shrink it in a refactoring commit afterwards.

Would it make sense to move this code into `match_stream`, or perhaps once you've rebased against the refactoring I merged from @kaustubh-nair this will be more straight-forward in any case?

preetmishra:
@neiljp I have made the amendments as per your suggestions and included a test case for 'multiple word searches'. I have introduced `order_streams` function to order the streams in the order that we discussed and used it in both the places. I have combined the last two sorts into one.

However, the test cases for `autocomplete_streams` have not been made more explicit (Kaustubh is working on the tests in #540; https://github.com/zulip/zulip-terminal/pull/540#discussion_r406050250), I'll improve those as soon as Kaustubh is done.

kaustubh-nair:
@preetmishra you can cherry-pick 3e9c627f969e873d80dada726ca5af2ec35ef61b here. I've removed pinned streams from the fixture. You can now set pinned streams in the parameterize block and assign it to `write_box.view.pinned_streams` which will appropriately affect the typeahead suggestion. Whether you want to modify the existing test to do so or create a new test set is up to you.

preetmishra:
@kaustubh-nair Thanks for the update. :+1: 

I have amended the tests to show different arrangements explicitly.

preetmishra:
@neiljp Thanks for the review. I have updated the PR as per the discussion we had in **#zulip-terminal**.

The streams are matched in, for both cases, in the following order:
* startswith match (pinned + unpinned), if any
* nth word match (pinned + unpinned), if any

I have updated the `match_stream` function that to order the streams (using the `dict` approach that you suggested :+1:) and used it for `update_streams` and `autocomplete_streams`.

tests/ui/test_ui_tools.py
```diff
@@ -409,22 +409,26 @@ def test_init(self, mocker, stream_view):
...
-    def test_update_streams(self, mocker, stream_view, new_text, expected_log):
+    def test_update_streams(self, mocker, stream_view, new_text, expected_log,
+                            to_pin):
         stream_names = [
             'FOO', 'FOOBAR', 'foo', 'fan',
             'boo', 'BOO', 'bar', 'test here',
         ]
+        self.view.pinned_streams = to_pin
         self.view.controller.editor_mode = True
-        new_text = new_text
```
preetmishra, line 427: This was redundant - we already have it as a parameter (line 421).

preetmishra, line 427: This was redundant.

zulipterminal/helper.py
```diff
@@ -459,15 +459,48 @@ def match_emoji(emoji: str, text: str) -> bool:
...
+
+    delimiters = [' ', '-', '_', '/']
+    for stream in streams:
+        stream_name = streams[stream][0]
+        for delimiter in delimiters:
+            if delimiter in stream_name:
+                streams[stream].extend(stream_name.split(delimiter)[1:])
+                break
+
+    matches = OrderedDict([
```
preetmishra, line 483: For `matches`, an alternative could be to use a `list` (where index 0: 'pinned' and index 1: 'unpinned').

I went with an `OrderedDict` for 'readability' (see line 490).

preetmishra:
@neiljp I have updated the PR to put pinned streams first based on your [suggestion](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Stream.20typeahead.20.23T530/near/864902). 

Now, the order is:
* subsequent startswith match up till the nth word (pinned), if any
* subsequent startswith match up till the nth word (unpinned), if any

zulipterminal/helper.py
```diff
@@ -459,15 +459,48 @@ def match_emoji(emoji: str, text: str) -> bool:
...
+        ('pinned', defaultdict(list)),
+        ('unpinned', defaultdict(list)),
+    ])  # type: Dict[str, Dict[int, List[str]]]
```
neiljp, line 486: DefaultDict?

zulipterminal/helper.py
```diff
@@ -459,15 +459,48 @@ def match_emoji(emoji: str, text: str) -> bool:
...
+    ordered_matches = []
+    for kind in matches:
+        if not matches[kind]:
+            continue
+        for match_position in range(max(dict(matches[kind])) + 1):
+            if match_position in matches[kind]:
+                for stream in matches[kind][match_position]:
+                    if stream not in ordered_matches:
+                        ordered_matches.append(stream)
+    return ordered_matches
```
neiljp, line 503: Is this a `List[str]`, or `List[Any]`, where it matches the `Any` passed in? If so, you could use a `TypeVar`, both for type-checking and making this clearer.

zulipterminal/ui_tools/boxes.py
```diff
@@ -129,10 +129,12 @@ def autocomplete_mentions(self, text: str, prefix_string: str
 
     def autocomplete_streams(self, text: str) -> List[str]:
         streams_list = self.view.pinned_streams + self.view.unpinned_streams
-        stream_typeahead = ['#**{}**'.format(stream[0])
-                            for stream in streams_list
-                            if match_stream(stream, text[1:])]
-        return stream_typeahead
+        stream_typeahead = OrderedDict(
```
neiljp, line 132: Is passing in a `List[Tuple[T, List[str]]]` sufficient?

zulipterminal/helper.py
```diff
@@ -459,15 +459,48 @@ def match_emoji(emoji: str, text: str) -> bool:
...
+    Note: This function expects `streams` to be sorted in a non-decreasing
+    order.
     """
-    stream_name = stream[0].lower()
-    if stream_name.startswith(text.lower()):
-        return True
-    return False
+    pinned_stream_names = [stream[0] for stream in pinned_streams]
+
+    delimiters = [' ', '-', '_', '/']
```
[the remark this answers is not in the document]
preetmishra: I am not sure it does. For instance, 'help' finds streams that have 'help' as a second word and 'term' doesn't find 'zulip-terminal' (in stream search).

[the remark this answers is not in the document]
neiljp: Right, well I feel this is an improvement; not sure if this is a conscious choice @timabbott ?

zulipterminal/helper.py
```diff
@@ -459,15 +459,48 @@ def match_emoji(emoji: str, text: str) -> bool:
...
+    for stream in streams:
+        stream_name = streams[stream][0]
+        for delimiter in delimiters:
+            if delimiter in stream_name:
+                streams[stream].extend(stream_name.split(delimiter)[1:])
+                break
```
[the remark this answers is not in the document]
preetmishra: I had assumed that the user would use only one kind of delimiter. Thanks for bringing this up. :+1:

zulipterminal/helper.py
preetmishra: My bad. It is `List[Any]`.

Are you suggesting using something like the following?
```python
T = TypeVar('T', Any, str)
```

preetmishra:
@neiljp Thanks for the review and the suggestions. :+1:

I have moved from using `OrderedDict` to using `List[Tuple[Any, str]]`, simplified and generalized the 'delimiters' part along with other typing improvements. This is ready for another pass.

zulipterminal/helper.py
neiljp: You can constrain the types in that way, though the one you mention is not very useful - it's saying that `T` can be consistently `Any` or `str`. You may be thinking of `AnyStr` as an example of this.

Really I was indicating how you can relate the function parameter types and return type(s) - if they *are* supposed to be related - by using a `TypeVar`.

The mypy pages describe this quite well I think, and there are examples in zulip/zulip, but I was querying whether this kind of approach was applicable:
```
X = TypeVar('X')
def foo(x: X) -> X:
  ...

foo(5)  # must return int
foo(2.5)  # must return float
foo("y")  # must return str
```

preetmishra: Thanks for the explanation. :)

preetmishra:
I have updated the PR to use `TypeVar` and included notes about the data structures that are being used in `match_stream` in the docstring.

zulipterminal/helper.py
```diff
@@ -459,15 +460,71 @@ def match_emoji(emoji: str, text: str) -> bool:
...
+    for stream, splits in stream_splits:
+        stream_name = splits[0]
+        for match_position, word in enumerate(splits):
+            if word.lower().startswith(text.lower()):
+                kind = ('pinned' if stream_name in pinned_stream_names
+                        else 'unpinned')
+                matches[kind][match_position].append(stream)
+
+    ordered_matches = []
+    for kind in matches:
```
neiljp, line 520: You're iterating over the keys here (implicitly, via `in`), but we don't ever use `kind` directly, and only ever the values? So we can just iterate over what we use instead? This basically emphasizes what you said, that the use of `pinned` as a key is descriptive, but we don't directly use it.

zulipterminal/helper.py
```diff
@@ -459,15 +460,71 @@ def match_emoji(emoji: str, text: str) -> bool:
...
+
+    # Note that 'space' is put in `delimiters` only to show the types of
+    # delimiters that are supported. Otherwise, it is redundant (see the
+    # following .replace method).
+    delimiters = [' ', '-', '_', '/']
+    stream_splits = []
+    for stream, stream_name in streams:
+        delimit_with_spaces = stream_name
+        for delimiter in delimiters:
+            delimit_with_spaces = delimit_with_spaces.replace(delimiter, ' ')
```
neiljp, line 501: This might be clearer with maketrans or a regex? You might then be able to shrink this into a list comprehension?

zulipterminal/helper.py
```diff
@@ -459,15 +460,71 @@ def match_emoji(emoji: str, text: str) -> bool:
...
+        ('pinned', defaultdict(list)),
+        ('unpinned', defaultdict(list)),
+    ])  # type: OrderedDict[str, DefaultDict[int, List[StreamT]]]
+
+    for stream, splits in stream_splits:
+        stream_name = splits[0]
+        for match_position, word in enumerate(splits):
+            if word.lower().startswith(text.lower()):
+                kind = ('pinned' if stream_name in pinned_stream_names
+                        else 'unpinned')
```
neiljp, line 516: We generate `kind` here, but it only depends on the `stream_name` at the top of the first loop. Perhaps you were wanting to avoid calculating `kind` until it was definitely used? If we move it up it should make it clearer that kind doesn't change depending upon what's in the lower loop/condition.

The pinned logic could be taken further and we could pre-sort the pinned and unpinned, but let's skip that for now, unless you find it looks even tidier!

tests/ui_tools/test_boxes.py
```diff
@@ -99,28 +99,54 @@ def test_generic_autocomplete_mentions(self, write_box, text,
...
-        ('#St', 1, '#**Stream 2**'),
-        ('#Stream 1', 0, '#**Stream 1**'),
+    @pytest.mark.parametrize('text, state, required_typeahead, to_pin', [
+        # With no streams pinned.
+        ('#Stream', 0, '#**Stream 1**', []),  # 1st-word startswith match.
+        ('#Stream', 1, '#**Stream 2**', []),  # 1st-word startswith match.
+        ('#Stream', 2, '#**Secret stream**', []),  # 2nd-word startswith match.
+        ('#Stream', 3, '#**Some general stream**', []),  # 3rd-word startswith.
+        ('#S', 0, '#**Some general stream**', []),  # 1st-word startswith.
+        ('#S', 1, '#**Secret stream**', []),  # 1st-word startswith match.
```
neiljp, line 109: Should the above two be reversed? ('Se' before 'So')

So, an equal match is sorted. This is relatively minor, but it's unclear how it's sorting otherwise?

zulipterminal/helper.py
```diff
@@ -459,15 +460,71 @@ def match_emoji(emoji: str, text: str) -> bool:
     return emoji.lower().startswith(text.lower())
 
 
-def match_stream(stream: Any, text: str) -> bool:
+StreamT = TypeVar('StreamT')
+
+
+def match_stream(streams: List[Tuple[StreamT, str]], text: str,
```
neiljp, line 466: Consider `text` -> `search_text`?

tests/ui/test_ui_tools.py
```diff
@@ -409,22 +409,26 @@ def test_init(self, mocker, stream_view):
...
+        ('f', ['FOO', 'FOOBAR', 'foo', 'fan'], []),
+        ('bar', ['bar'], []),
+        ('foo', ['FOO', 'FOOBAR', 'foo'], []),
+        ('FOO', ['FOO', 'FOOBAR', 'foo'], []),
+        ('test', ['test here'], []),
+        ('here', ['test here'], []),
+        ('test here', ['test here'], []),
+        # With 'foo' pinned.
+        ('f', ['foo', 'FOO', 'FOOBAR', 'fan'], [['foo'], ]),
+        ('FOO', ['foo', 'FOO', 'FOOBAR'], [['foo'], ]),
```
neiljp, line 422: These test cases are OK, but it feels like duplication of the other test parameters, and don't go into as much depth - and we could also test `match_stream`, fundamentally, perhaps instead of these two.

tests/ui_tools/test_boxes.py
preetmishra: To clarify, `match_stream` doesn't sort; it expects `streams` to be sorted (I have mentioned this in the docstring).

Re the 'Se' before 'So', 'Some general stream' has been put before 'Secret stream' in `streams_fixture` which is why the suggestions are being given in this order.

```python
@pytest.fixture
def streams_fixture():
    streams = [general_stream, secret_stream]
    ...
```

zulipterminal/helper.py
preetmishra: Indeed, I did that to avoid calculating `kind` every time. However, the suggested approach is much cleaner. :+1:

zulipterminal/helper.py
preetmishra: Wow, I didn't know something like `maketrans` existed. Thanks!

preetmishra:
@neiljp Thanks for the review and in-line suggestions. I have incorporated the suggested changes; `match_stream` should be cleaner now. :)

Re tests, we could take another approach i.e. test `match_stream` instead or tidy up the existing tests, though, in my opinion, the tests that are present currently are decent enough to work with until we are done with other PRs that are awaiting.

tests/ui_tools/test_boxes.py
preetmishra: I agree. We could either change the name from 'Some general stream' to 'General Stream' (that'd fix the order) or sort the streams in the fixture itself. 

The assert sounds appealing. It would require the input to be sorted while maintaining the pinning order (e.g. `assert input = sorted(input, key=is_pinned)`), right?. In that case, would the assert be better than actually sorting the input?

neiljp: The fixture is supposed to represent what is present in the running code, so it should probably be in the right order in any case, given that's what the function expects. I think that's probably the minimum, whether we adjust the fixture or potentially change the name - though I don't know if the latter approach would affect the scope of other tests that use the same fixture.

The assert would be a code-based representation of what the range of inputs are supposed to be, a little like the +1/-1 in set_counts. Sometimes that can be expressed more cleanly in terms of the data type used, but failing that then it's useful to ensure that eg. in that case we don't pass +5 to set _count!

preetmishra:
@neiljp Thanks for the review and the suggestions.

I have incorporated the suggestions that you left, added a separate commit to adjust the fixture and the method to return the data as it is expected and is present in the running code and introduced an assert statement in the method.

Re tests, I have added it to my To-Do. I'll get back to it as soon as we're done with the PRs that are related.

I agree. We'd need quite a similar approach for other components (e.g. emojis).

preetmishra:
@neiljp Thanks for the review and the suggestions.

I have incorporated the suggestions that you left, added a separate commit that adjusts the fixture and the method to return streams as they're expected and are present in the running code and introduced an assert statement in the method.

Re tests, I have added it to my To-Do. I'll get back to it as soon as we're done with the PRs that are related.

I agree. We'd need quite a similar approach for other components (e.g. emojis).

neiljp:
@preetmishra I've just re-examined this and merged it manually with I think just a typo fixed in a commit message :tada: We can explore the other work (eg. test merging, generalization to other searches/matches) later, maybe add it to a list :) That is, once Kaustubh is done with typeahead, or else wants to pick up these aspects.

Thanks to @kaustubh-nair for the fixture :+1: This should now unblock #540?

PR #531, opened 2020-03-10
run: Support autohide flags during startup.

### LINKED ISSUE #445: Support --autohide flag argument during zt startup.
The current optional arguments ZT supports are the following
```
optional arguments:
  -h, --help            show this help message and exit
  --config-file CONFIG_FILE, -c CONFIG_FILE
                        config file downloaded from your zulip
                        organization.(e.g. ~/zuliprc)
  --theme THEME, -t THEME
                        choose color theme. (e.g. blue, light)
  -d, --debug           Start zulip terminal in debug mode.
  --profile             Profile runtime.
  -v, --version         Print zulip-terminal version and exit
```
Particularly now that we made `autohide` setting the default standard, we could benefit from a flag which can switch the setting between the two options during startup, rather than having to tweak the zuliprc.

kaustubh-nair:
Fixes #445

### DIFF
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -1,6 +1,8 @@
 import pytest
 
-from zulipterminal.cli.run import THEMES, get_login_id, in_color, main
+from zulipterminal.cli.run import (
+    THEMES, get_login_id, in_color, main, parse_args,
+)
 from zulipterminal.model import ServerConnectionFailure
 from zulipterminal.version import ZT_VERSION
 
@@ -57,6 +59,8 @@ def test_main_help(capsys, options):
         '-d, --debug',
         '--profile',
         '--config-file CONFIG_FILE, -c CONFIG_FILE',
+        '--autohide',
+        '--no-autohide',
         '-v, --version'
     }
     optional_argument_lines = {line[2:] for line in lines
@@ -150,3 +154,30 @@ def test_zt_version(capsys, options):
     assert lines == expected
 
     assert captured.err == ""
+
+
+@pytest.mark.parametrize('option, autohide', [
+        ('--autohide', 'autohide'),
+        ('--no-autohide', 'no_autohide'),
+        ('--debug', None),  # no-autohide by default
+        ]
+)
+def test_parse_args_valid_autohide_option(option, autohide):
+    args = parse_args([option])
+    assert args.autohide == autohide
+
+
+@pytest.mark.parametrize('options', [
+        ['--autohide', '--no-autohide'],
+        ['--no-autohide', '--autohide']]
+)
+def test_main_multiple_autohide_options(capsys, options):
+    with pytest.raises(SystemExit) as e:
+        main(options)
+        assert str(e.value) == "2"
+    captured = capsys.readouterr()
+    lines = captured.err.strip('\n')
+    lines = lines.split("pytest: ", 1)[1]
+    expected = ("error: argument {}: not allowed "
+                "with argument {}".format(options[1], options[0]))
+    assert lines == expected
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -58,6 +58,15 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
                         action="store_true",
                         default=False, help='Profile runtime.')
 
+    autohide_group = parser.add_mutually_exclusive_group()
+    autohide_group.add_argument('--autohide', dest='autohide', default=None,
+                                action="store_const", const='autohide',
+                                help='Autohide list of users and streams.')
+
+    autohide_group.add_argument('--no-autohide', dest='autohide', default=None,
+                                action="store_const", const='no_autohide',
+                                help='Don\'t autohide list of users and streams.')
+
     parser.add_argument('-v',
                         '--version',
                         action='store_true',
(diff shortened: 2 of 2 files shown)


kaustubh-nair:
This issue has been open for quite a while, so I thought I'd give it a go.
It build on @TaranSinghania 's https://github.com/zulip/zulip-terminal/pull/475 by using a mutually exclusive group.

neiljp:
@kaustubh-nair Thanks for working on this. This seems fairly complete :+1: 

Could you maybe add some tests for the actual autohide and no-autohide cases, in addition to the invalid cases?

@amanagr This may interact with #457, but I think could be worthwhile as it would allow temporarily overriding the value on the command-line if the tutorial/menu was disabled, or setting a different default temporarily (default < config < command-line < menu ?).

kaustubh-nair:
Added tests for autohide and no_autohide

tests/cli/test_run.py
```diff
@@ -150,3 +154,31 @@ def test_zt_version(capsys, options):
...
+@pytest.mark.parametrize('option, autohide, no_autohide', [
+        ('--autohide', True, False),
+        ('--no-autohide', False, True),
+        ('--debug', False, False),
+        ]
+)
+def test_autohide_option(capsys, option, autohide, no_autohide):
+    args = parse_args([option])
+    assert args.autohide == autohide
+    assert args.no_autohide == no_autohide
```
neiljp: I was thinking with the testing that we might be able to check that these options get passed through, rather than testing at this level - but the code perhaps isn't really structured for that right now.

tests/cli/test_run.py
```diff
@@ -150,3 +154,31 @@ def test_zt_version(capsys, options):
     assert lines == expected
 
     assert captured.err == ""
+
+
+@pytest.mark.parametrize('option, autohide, no_autohide', [
+        ('--autohide', True, False),
+        ('--no-autohide', False, True),
+        ('--debug', False, False),
```
neiljp: I can see why you included another valid option here, to cover more autohide values.

However, along with the explicit mutually exclusive code, this makes me wonder whether we can use one 'dest' value instead, to simplify this further?

tests/cli/test_run.py
```diff
@@ -150,3 +154,31 @@ def test_zt_version(capsys, options):
...
+    args = parse_args([option])
+    assert args.autohide == autohide
+    assert args.no_autohide == no_autohide
+
+
+@pytest.mark.parametrize('options', [
+        ['--autohide', '--no-autohide'],
+        ['--no-autohide', '--autohide']]
+)
+def test_multiple_autohide_options(capsys, options):
```
neiljp: For test names, it's often useful to start them with the function you're testing (eg. see `test_main_...` or `test_get_login_id`), though I note we've not been completely consistent even in this file.

tests/cli/test_run.py
kaustubh-nair: Hmm, yes it is possible to do that, but it'll get a little messy. I'll have to call `main` and check `stdout`. But the problem is `main` won't exit on it's own. So i'll have to use some kind of manual interrupt to kill main.

neiljp: To be clear, I think this approach is fine for now.

tests/cli/test_run.py
kaustubh-nair: @neiljp can you elaborate on this?

neiljp: I was referring to the use of `dest` in `argparse`, which we might be able to set to the same for each related command-line parameter.

kaustubh-nair: That's a nice idea :+1:, added that too.

tests/cli/test_run.py
```diff
@@ -150,3 +154,30 @@ def test_zt_version(capsys, options):
...
+def test_parse_args_valid_autohide_option(capsys, option, autohide):
+    args = parse_args([option])
+    assert args.autohide == autohide
+
+
+@pytest.mark.parametrize('options', [
+        ['--autohide', '--no-autohide'],
+        ['--no-autohide', '--autohide']]
+)
+def test_parse_args_multiple_autohide_options(capsys, options):
```
neiljp: Is this testing just `parse_args`? Which function is being called? :)

zulipterminal/cli/run.py
```diff
@@ -58,6 +58,17 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
...
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--autohide', dest='autohide',
+                       action="store_const", const='autohide',
+                       default=False, help='Autohide list of users\
+                               and streams.')
+
+    group.add_argument('--no-autohide', dest='autohide',
+                       action="store_const", const='no_autohide',
+                       default=False, help='Don\'t autohide list of\
+                               users and streams.')
```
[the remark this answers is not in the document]
kaustubh-nair: The line was too long for `pycodestyle`, but I moved the arguments around to avoid a `\`

kaustubh-nair:
Thanks for the review :)
I've pushed the changes.

neiljp:
@kaustubh-nair Thanks for this - merged as 45c71922e60b528a56afae8e9ea2783d38f4c789 :tada:

You had this as two commits in the branch, perhaps you forgot to `--amend`? Anyhow, I squashed them, and also did some slight changes to the parametrize formatting to save space and match how we have done it in other tests.

zulipterminal/cli/run.py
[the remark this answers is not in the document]
neiljp: I noticed we did have some lines with those characters to break lines, so this seems OK in context for now, particularly as you had the `autohide_group`, making the line longer.

PR #535, opened 2020-03-10
Handle read events from server

kaustubh-nair:
TODO:
- [ ] Limit number of unread msgs that are stored to ~10k

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -39,6 +39,7 @@ Index = TypedDict('Index', {
     'search': Set[int],  # {message_id, ...}
     # Downloaded message data
     'messages': Dict[int, Message],  # message_id: Message
+    'unread_msgs': Dict[int, Dict[str, Any]],  # message_id: Dict
 })
 
 initial_index = Index(
@@ -54,6 +55,7 @@ initial_index = Index(
     topics=defaultdict(list),
     search=set(),
     messages=defaultdict(dict),
+    unread_msgs=defaultdict(dict),
 )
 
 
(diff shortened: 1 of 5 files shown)


neiljp:
@kaustubh-nair I know this is WIP, but continuing this work would be helpful to reliably track unread counts, depending on how busy you are with other things right now.

kaustubh-nair:
@neiljp I was just finishing up with the typeahead footer PR, I'll be working on this one now.
Added some tests for `set_counts` in the PR referenced above.

zulipterminal/helper.py
```diff
@@ -133,14 +132,16 @@ def set_count(id_list: List[int], controller: Any, new_count: int) -> None:
...
-
         msg_type = messages[id]['type']
         add_to_counts = True
+        # FIXME no user_id for streams?
+        if msg_type != 'stream':
+            user_id = messages[id]['sender_id']
+
+            # If we sent this message, don't increase the count
+            if user_id == controller.model.user_id:
+                continue
```
kaustubh-nair, line 143: This is kind of hacky - we don't have `sender_id` for stream messages in `initial_data['unread_msgs']`
Though on close inspection this piece of code doesn't seem to be doing anything?
Refer to conversation on czo https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Handling.20read.20events/near/845434

neiljp (commented):
@kaustubh-nair The PR title refers to 'handle read events', and in this sense it seems to *work* well :+1:

I didn't anticipate the approach you've taken - rather messages+unread being handled separately in each case where they would be updated.

Functionality while it's not strictly part of the PR intent, it would seem good to handle the addition (append'ing) of messages too.

I'm wary that we're ending up with two datastructures which hold overlapping information, and in some cases lots of duplicate information.

I also get a bug where `changed_messages` (helper.py, L191) has a `KeyError` (id not in messages, presumably) in some cases.

I'm concerned by the change of the `elif` statement for `1-1 pms` etc, but we can look into that - I've not dug deep into that.

kaustubh-nair:
I've dealt with handling appending of messages as well. I'm pretty sure the keyerror you came across was due to this, since `set_count` was being called while new messages was not added to the index. If you still come across the exception, let me know! 

Last two commits can be squashed later; I've kept them separate for now.

neiljp:
@kaustubh-nair I'm still getting tracebacks in this PR; I've not read through it again in detail though, so can't make further comments at this stage.

mkp6781:
@kaustubh-nair can I take this up if you are not planning to continue work on this?

PR #539, opened 2020-03-11
Migrate mouse_event(s) to use keys_for_command.

### LINKED ISSUE #533: TRACKING: Migrate tests using hard-coded keys to use keys_for_command.

preetmishra:
This migrates mouse_event(s) and the associated tests to use `keys_for_command`.
I have split the commit into four commits for making it easier to review.

Fixes #533 partially.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -443,19 +443,20 @@ class TestStreamsView:
                 ] == expected_log
         self.view.controller.update_screen.assert_called_once_with()
 
-    def test_mouse_event(self, mocker, stream_view):
-        mocker.patch.object(stream_view, 'keypress')
+    @pytest.mark.parametrize('button, key', [
+            (4, 'up'),
+            (5, 'down'),
+        ],
+        ids=['scroll_wheel_up', 'scroll_wheel_down']
+    )
+    def test_mouse_event(self, mocker, stream_view, button, key):
         size = (200, 20)
         col = 1
         row = 1
         focus = "WIDGET"
-        # Left click
-        stream_view.mouse_event(size, "mouse press", 4, col, row, focus)
-        stream_view.keypress.assert_called_once_with(size, "up")
-
-        # Right click
-        stream_view.mouse_event(size, "mouse press", 5, col, row, focus)
-        stream_view.keypress.assert_called_with(size, "down")
+        mocker.patch.object(stream_view, 'keypress')
+        stream_view.mouse_event(size, 'mouse press', button, col, row, focus)
+        stream_view.keypress.assert_called_once_with(size, key)
 
     @pytest.mark.parametrize('key', keys_for_command('SEARCH_STREAMS'))
     def test_keypress_SEARCH_STREAMS(self, mocker, stream_view, key):
@@ -674,19 +675,20 @@ class TestUsersView:
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
         return UsersView("USER_BTN_LIST")
 
-    def test_mouse_event(self, mocker, user_view):
-        mocker.patch.object(user_view, 'keypress')
+    @pytest.mark.parametrize('button, key', [
+            (4, 'up'),
+            (5, 'down'),
+        ],
+        ids=['scroll_wheel_up', 'scroll_wheel_down']
+    )
+    def test_mouse_event(self, mocker, user_view, button, key):
         size = (200, 20)
         col = 1
         row = 1
         focus = "WIDGET"
-        # Left click
-        user_view.mouse_event(size, "mouse press", 4, col, row, focus)
-        user_view.keypress.assert_called_with(size, "up")
-
-        # Right click
-        user_view.mouse_event(size, "mouse press", 5, col, row, focus)
-        user_view.keypress.assert_called_with(size, "down")
+        mocker.patch.object(user_view, 'keypress')
+        user_view.mouse_event(size, 'mouse press', button, col, row, focus)
+        user_view.keypress.assert_called_with(size, key)
 
     @pytest.mark.parametrize('event, button', [
             ('mouse release', 0),
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -443,6 +443,14 @@ class TestStreamsView:
                 ] == expected_log
         self.view.controller.update_screen.assert_called_once_with()
 
+    def test_keypress_navigation(self, mocker, stream_view,
+                                 mouse_event_navigation_key_expected_key_pair):
+        key, expected_key = mouse_event_navigation_key_expected_key_pair
+        size = (200, 20)
+        super_keypress = mocker.patch(VIEWS + '.urwid.Frame.keypress')
+        stream_view.keypress(size, key)
+        super_keypress.assert_called_once_with(size, expected_key)
+
     @pytest.mark.parametrize('button, key', [
             (4, 'up'),
             (5, 'down'),
@@ -667,6 +675,14 @@ class TestTopicsView:
         assert new_focus == previous_focus
         assert previous_focus_topic_name == new_focus_topic_name
 
+    def test_keypress_navigation(self, mocker, topic_view,
+                                 mouse_event_navigation_key_expected_key_pair):
+        key, expected_key = mouse_event_navigation_key_expected_key_pair
+        size = (200, 20)
+        super_keypress = mocker.patch(VIEWS + '.urwid.Frame.keypress')
+        topic_view.keypress(size, key)
+        super_keypress.assert_called_once_with(size, expected_key)
+
 
 class TestUsersView:
 
@@ -674,6 +690,14 @@ class TestUsersView:
     def user_view(self, mocker):
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
         return UsersView("USER_BTN_LIST")
+
+    def test_keypress_navigation(self, mocker, user_view,
+                                 mouse_event_navigation_key_expected_key_pair):
+        key, expected_key = mouse_event_navigation_key_expected_key_pair
+        size = (200, 20)
+        super_keypress = mocker.patch(VIEWS + '.urwid.ListBox.keypress')
+        user_view.keypress(size, key)
+        super_keypress.assert_called_once_with(size, expected_key)
 
     @pytest.mark.parametrize('button, key', [
             (4, 'up'),
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -235,14 +235,13 @@ class TestMessageView:
                                                         num_after=30,
                                                         anchor=0)
 
-    @pytest.mark.parametrize("event, button, keypress", [
-        ("mouse press", 4, "up"),
-        ("mouse press", 5, "down"),
-    ])
-    def test_mouse_event(self, mocker, msg_view, event, button, keypress):
+    def test_mouse_event(self, mocker, msg_view, mouse_scroll_key_button_pair):
+        key, button = mouse_scroll_key_button_pair
+        size = (20, )
+        mocker.patch(VIEWS + '.keys_for_command', return_value=[key])
         mocker.patch.object(msg_view, "keypress")
-        msg_view.mouse_event((20,), event, button, 0, 0, mocker.Mock())
-        msg_view.keypress.assert_called_once_with((20,), keypress)
+        msg_view.mouse_event(size, 'mouse press', button, 0, 0, mocker.Mock())
+        msg_view.keypress.assert_called_once_with(size, key)
 
     @pytest.mark.parametrize('key', keys_for_command('GO_DOWN'))
     def test_keypress_GO_DOWN(self, mocker, msg_view, key):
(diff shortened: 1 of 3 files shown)


preetmishra:
More on the implementation: I have used `next(iter())` to get a key from the set returned by `keys_for_command`. Other options to do this were: 
- Using a for loop
- Converting the set to a list and then indexing it.

While the for loop would have been the fastest, the implementation would have added 2 more lines and a variable per `keys_for_command` and the latter option is expensive. 
`next(iter())` comes after for loop in terms of cost.

I had to add `GO_UP/DOWN` presses for `Streams/Topics/UsersView` views. They already support navigation through alternative navigation keys (this is defined in `View` class) but since mouse_event directly calls its keypress method, the addition was necessary to make it work.

zulipterminal/ui_tools/views.py
```diff
@@ -135,10 +135,10 @@ def mouse_event(self, size: urwid_Size, event: str, button: int, col: int,
                     row: int, focus: bool) -> bool:
         if event == 'mouse press':
             if button == 4:
-                self.keypress(size, 'up')
+                self.keypress(size, next(iter(keys_for_command('GO_UP'))))
```
neiljp: While this doesn't modify the `set` returned by `keys_for_command`, you might note that the latter returns a copy of the set - both to make this kind of situation easier and to avoid issues with accidentally setting keys!

This means we can use here a simpler `pop` approach, as currently used elsewhere in the code.

We could actually add a test for this, to ensure that each call to `keys_for_command` returns the original data in a new set, to ensure this assumption continues to be validated.

tests/ui/test_ui_tools.py
```diff
@@ -233,8 +233,8 @@ def test_load_new_messages_mocked_log(self, mocker, msg_view,
                                                         anchor=0)
 
     @pytest.mark.parametrize("event, button, keypress", [
-        ("mouse press", 4, "up"),
-        ("mouse press", 5, "down"),
+        ("mouse press", 4, next(iter(keys_for_command('GO_UP')))),
+        ("mouse press", 5, next(iter(keys_for_command('GO_DOWN')))),
```
neiljp: Since these calls (or the `pop` I recommend elsewhere) are non-deterministic, we should probably take a different testing approach - we want to know that one of the valid up or down keypress keys was used, but that any should work. We could maybe parametrize further, or take another approach.

zulipterminal/ui_tools/boxes.py
```diff
@@ -652,7 +652,7 @@ def mouse_event(self, size: urwid_Size, event: str, button: int,
                     col: int, row: int, focus: bool) -> bool:
         if event == 'mouse press':
             if button == 1:
-                self.keypress(size, 'enter')
+                self.keypress(size, next(iter(keys_for_command('ENTER'))))
```
neiljp: You don't have a connected test change for this?

tests/ui/test_ui_tools.py
```diff
@@ -439,13 +439,31 @@ def test_mouse_event(self, mocker, stream_view):
...
+            size, next(iter(keys_for_command('GO_UP')))
+        )
 
-        # Right click
+        # Scroll wheel down: 5.
         stream_view.mouse_event(size, "mouse press", 5, col, row, focus)
-        stream_view.keypress.assert_called_with(size, "down")
+        stream_view.keypress.assert_called_with(
+            size, next(iter(keys_for_command('GO_DOWN')))
+        )
```
neiljp: There's some independent copy/pasted code in this test - can you see how to refactor it? This would make an actual followup test-commit easier. Do you see what the comments could translate to?

zulipterminal/ui_tools/views.py
```diff
@@ -302,6 +302,10 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             self.log.set_focus(self.focus_index_before_search)
             self.view.controller.update_screen()
             return key
+        elif is_command_key('GO_UP', key):
+            key = 'up'
+        elif is_command_key('GO_DOWN', key):
+            key = 'down'
```
neiljp: One aspect which makes me wonder if we should approach this differently is that scrolling with the extra keys (vim-like) already works in the stream-view, from the base `View`? So it would be simpler if we could avoid adding extra code and tests just for this case.

I can follow why you've added this, but this and the associated test are separate code which are needed before the actual mouse_event is changed - so if we do take this route, I think we should put this in a separate commit, at least for now.

tests/ui/test_ui_tools.py
```diff
@@ -635,6 +635,39 @@ def test_keypress_GO_BACK(self, mocker, topic_view, key):
...
+        topic_view.mouse_event(size, "mouse press", 4, col, row, focus)
+        topic_view.keypress.assert_called_once_with(
+            size, next(iter(keys_for_command('GO_UP')))
+        )
+
+        # Scroll wheel down: 5.
+        topic_view.mouse_event(size, "mouse press", 5, col, row, focus)
+        topic_view.keypress.assert_called_with(
+            size, next(iter(keys_for_command('GO_DOWN')))
+        )
```
neiljp: Same response here as for the other two-part test you altered, except you're adding it from scratch here, so there's no need to refactor.

zulipterminal/ui_tools/views.py
```diff
@@ -403,6 +403,10 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             self.set_focus('body')
             self.view.controller.update_screen()
             return key
+        elif is_command_key('GO_UP', key):
+            key = 'up'
+        elif is_command_key('GO_DOWN', key):
+            key = 'down'
```
neiljp: Same point here re duplicating work from the main `View`.

zulipterminal/ui_tools/views.py
preetmishra: I have replaced the `next(iter())` calls with `pop` as it is simpler and doesn't harm the original keys set.

I have added a test (148e2c98e2dca43d9c48eb7aad4d8d2cfa2254c4) for validating that each call to `keys_for_command` returns the original data in a new set

tests/ui/test_ui_tools.py
preetmishra: I could only validate the first bit (one of the valid up or down keypress keys was used). Could you assist me with how to approach the second bit (but any should work)?

zulipterminal/ui_tools/boxes.py
preetmishra: My bad, we didn't have any test for this to change. I have added a test now.

tests/ui/test_ui_tools.py
preetmishra: I have refactored this and a similar test in `UsersView` in a3ee7387670a70d9a5c7082f031c742c5134e0ed. 

I am afraid, I am not sure about what the comments could translate to mean?

zulipterminal/ui_tools/views.py
preetmishra: > I had to add GO_UP/DOWN presses for Streams/Topics/UsersView views. They already support navigation through alternative navigation keys (this is defined in View class) but since mouse_event directly calls its keypress method, the addition was necessary to make it work. 

I think it is necessary to add these to make this work. I have separated these changes in another commit 381908c5f1a1fa5898a46017900fe2a66cd3e327 with their tests.

preetmishra:
@neiljp Thanks. I had to close and reopen the PR in order to restart the build check, it was failing with 'An error occurred while generating the build script.' for `Install & pytest - Python 3 (OS X; xcode10)` even though it was passing all the tests locally.

I have made my remarks in the respective inline-comments.

tests/ui/test_ui_tools.py
neiljp: Well, what does 'mouse press, 4, up' mean? In a parametrized test, you can use custom ids, which I think you've already seen.

It may be worth checking that the comments are correct, as I'm not sure if 4 means 'left click'.

tests/ui/test_ui_tools.py
neiljp: To clarify, I mean by 'any should work', that any key extracted using pop should work, so you should be able to work through those in a test parametrization, or another similar approach.

tests/ui/test_ui_tools.py
preetmishra: I have added the appropriate test ids, the comments weren't correct. I referred to urwid's docs to find out the meanings.

preetmishra:
@neiljp I restructured the order of commits and made a few changes that eliminate code duplication.

- The first commit has a test for validating that each call to `keys_for_command` returns the original data in a new set.
- The second commit is a refactor commit which reduces duplication in `mouse_event` tests for `Streams/UsersView`.
- The third commit introduces `key_for_command` in 'simple' `mouse_event` functions to remove hard-coded keys. For the tests, I have added a fixture as we have a bunch of similar tests in `ui_tools`.
- The fourth introduces extra navigation keys to get `mouse_event` ready for `keys_for_command` and the fifth commit has code for introducing `keys_for_command` in 'non-trivial' `mouse_event` functions and tests.

tests/ui/test_ui_tools.py
```diff
@@ -629,28 +627,28 @@ def user_view(self, mocker):
...
+        size = (200, 20)
         mocker.patch.object(user_view, 'keypress')
+        user_view.mouse_event(size, 'mouse press', button, 1, 1, 'WIDGET')
+        user_view.keypress.assert_called_with(size, key)
+
+    @pytest.mark.parametrize('event, button', [
+            ('mouse release', 4),
+            ('mouse press', 1),
+        ],
+        ids=['invalid action', 'invalid button']
```
neiljp, line 708: We could cover extra cases here?

tests/ui/test_ui_tools.py
```diff
@@ -433,19 +433,17 @@ def test_update_streams(self, mocker, stream_view, new_text, expected_log):
...
-        # Left click
-        stream_view.mouse_event(size, "mouse press", 4, col, row, focus)
-        stream_view.keypress.assert_called_once_with(size, "up")
-
-        # Right click
-        stream_view.mouse_event(size, "mouse press", 5, col, row, focus)
-        stream_view.keypress.assert_called_with(size, "down")
+        mocker.patch.object(stream_view, 'keypress')
+        stream_view.mouse_event(size, 'mouse press', button, 1, 1, 'WIDGET')
+        stream_view.keypress.assert_called_once_with(size, key)
```
neiljp, line 487: These tests are looking a lot better :+1: 

Would it be logical to introduce the fixture in the next commit in this one? If so, you could focus on one commit introducing and using the new fixture, and put the invalid test changes into another for now?

tests/ui/test_ui_tools.py
```diff
@@ -26,6 +26,24 @@
...
+        for keys, button in [
+            (keys_for_command('GO_UP'), 4),
+            (keys_for_command('GO_DOWN'), 5),
+        ]
+        for key in keys
+    ],
+    ids=lambda param: 'scroll_wheel_' + ('up' if param[1] == 4 else 'down')
+                      + '-key:{}-button:{}'.format(*param)
+)
+def mouse_press_key_button_pair(request):
```
neiljp, line 40: This is specifically for a mouse scrolling action, right? If so, maybe adjust the name accordingly.

Did you keep this out of `conftest.py` for a specific reason? (not that all fixtures have to be in there)

preetmishra: I kept this here as it was only being used in `ui_tools.py`. Thanks for pointing this out; I have moved this to `conftest.py` as we could need it in other files as well.

preetmishra:
@neiljp Thanks for the review. I have restructured the commits, to better convey the flow, and incorporated the suggestions that you left in the review.

* The first two commits are refactoring commits. The first one extracts `mouse_event_invalid` and
  the second one reduces code duplication.
* The third one adds support for `j/k` keypresses in other views (`MessageView` already has the 
   support). After which the fourth commit migrates the `mouse_event` functions that we've in 
   `views.py` altogether.
* The last one introduces `keys_for_command` for left click in the `MessageBox`.

Hopefully, this will be clearer to review. :)

neiljp:
@preetmishra I just merged the first and last commits, as these are independent, with a few small changes. I held off on the second as it might be related to the others? In any case, clearer to review these related commits together.

preetmishra:
@neiljp I have resolved the conflicts and rebased against `master`. Thanks for the refinements. :+1:

preetmishra:
Updated to resolve conflicts.

Ezio-Sarthak:
This PR might close tracking issue #533. Are you planning to update/rebase this in near future?

preetmishra:
Hi, @Ezio-Sarthak. Thanks for taking an interest! Feel free to take this further.

neiljp:
@preetmishra #953 appears to cover the first commit here, and replaces the last commit; I'm not sure about the middle commit?

It'd be good to close this out, so if we don't close outright, perhaps @zee-bit may want to take this on given previous work, or @Ezio-Sarthak ?

zee-bit:
@neiljp The middle commit here looks unnecessary after #953. If I understand correctly, it handles 'j/k' keys for scrolling because `keys_for_command` includes those along with the directional keys (up/down etc), so adding that middle commit before migrating to `keys_for_command` makes sense.
But, in #953, we don't use `keys_for_command`, but rather `primary_keys_for_command`, which returns only the directional keys, so there's no need to handle other navigational keys such as j/k.

Perhaps @preetmishra can confirm/validate my above hypothesis before closing this?

preetmishra:
@neiljp As @zee-bit mentions, this was needed earlier as I was relying on `keys_for_command` to add the support. However, using `primary_keys_for_command` makes more sense and helps avoid unnecessary changes.

Thanks once again, @zee-bit, for taking the PR further. I am closing this now.

PR #540, opened 2020-03-12
ui/footer: Display typeahead suggestions in footer.

kaustubh-nair:
Part of typeahead followups #448 
GIF: 
footer_new

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -81,6 +81,29 @@ class TestView:
         mock_sleep.assert_called_once_with(duration)
         assert view.controller.update_screen.call_count == 2
 
+    @pytest.mark.parametrize('suggestions, state, truncated, footer_text', [
+        ([], None, False, [' [No matches found]']),
+        (['some', 'text'], None, False, [[' '], ' some ', ' text ']),
+        (['some', 'text'], None, True,
+         [[' '], ' some ', ' text ', ' [more] ']),
+        (['some', 'text'], 0, False, [[' '], ('code', ' some '), ' text ']),
+        (['some', 'text'], 0, True,
+         [[' '], ('code', ' some '), ' text ', ' [more] ']),
+        (['some', 'text'], -1, False, [[' '], ' some ', ('code', ' text ')]),
+    ], ids=[
+        'no_matches',
+        'no_highlight',
+        'no_highlight_truncated',
+        'first_suggestion_highlighted',
+        'first_suggestion_highlighted_truncated',
+        'last_suggestion_highlighted',
+        ])
+    def test_set_typeahead_footer(self, mocker, view, state, suggestions,
+                                  truncated, footer_text):
+        set_footer_text = mocker.patch('zulipterminal.ui.View.set_footer_text')
+        view.set_typeahead_footer(suggestions, state, truncated)
+        set_footer_text.assert_called_once_with(footer_text, duration=3)
+
     def test_footer_view(self, mocker, view):
         footer = view.footer_view()
         assert isinstance(footer.text, str)
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -102,7 +102,7 @@ class TestView:
                                   truncated, footer_text):
         set_footer_text = mocker.patch('zulipterminal.ui.View.set_footer_text')
         view.set_typeahead_footer(suggestions, state, truncated)
-        set_footer_text.assert_called_once_with(footer_text, duration=3)
+        set_footer_text.assert_called_once_with(footer_text)
 
     def test_footer_view(self, mocker, view):
         footer = view.footer_view()
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -203,6 +203,35 @@ class TestWriteBox:
         typeahead_string = write_box.generic_autocomplete(text, state)
         assert typeahead_string == required_typeahead
 
+    @pytest.mark.parametrize(['key', 'current_typeahead_mode',
+                              'expected_typeahead_mode',
+                              'expect_footer_was_reset'], [
+        # footer does not reset
+        (keys_for_command('AUTOCOMPLETE').pop(), False, False, False),
+        (keys_for_command('AUTOCOMPLETE_REVERSE').pop(), False, False, False),
+        (keys_for_command('AUTOCOMPLETE').pop(), True, True, False),
+        (keys_for_command('AUTOCOMPLETE_REVERSE').pop(), True, True, False),
+        # footer resets
+        (keys_for_command('GO_BACK').pop(), True, False, True),
+        ('space', True, False, True),
+        ('k', True, False, True),
+    ])
+    def test_keypress_typeahead_mode_autocomplete_key(self, mocker, write_box,
+                                                      current_typeahead_mode,
+                                                      expected_typeahead_mode,
+                                                      expect_footer_was_reset,
+                                                      key):
+        write_box.is_in_typeahead_mode = current_typeahead_mode
+        size = (20,)
+
+        write_box.keypress(size, key)
+
+        assert write_box.is_in_typeahead_mode == expected_typeahead_mode
+        if expect_footer_was_reset:
+            self.view.set_footer_text.assert_called_once_with()
+        else:
+            self.view.set_footer_text.assert_not_called()
+
 
 class TestPanelSearchBox:
     search_caption = "Search Results "
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -43,6 +43,35 @@ class TestWriteBox:
                                             state):
         return_val = write_box.generic_autocomplete(text, state)
         assert return_val == text
+        write_box.view.set_typeahead_footer.assert_not_called()
+
+    @pytest.mark.parametrize('text, state, footer_text', [
+        # mentions
+        ('@Human', 0, ['Human Myself', 'Human 1', 'Human 2']),
+        ('@_Human', 0, ['Human Myself', 'Human 1', 'Human 2']),
+        ('@Human', None, ['Human Myself', 'Human 1', 'Human 2']),
+        ('@NoMatch', None, []),
+        # streams
+        ('#Stream', 0, ['Stream 1', 'Stream 2', 'Secret stream',
+                        'Some general stream']),
+        ('#Stream', None, ['Stream 1', 'Stream 2', 'Secret stream',
+                           'Some general stream']),
+        ('#NoMatch', None, []),
+        # emojis
+        (':smi', 0, ['smile', 'smiley', 'smirk']),
+        (':smi', None, ['smile', 'smiley', 'smirk']),
+        (':NoMatch', None, []),
+    ])
+    def test_generic_autocomplete_set_footer(self, mocker, write_box,
+                                             state, footer_text, text):
+        write_box.view.set_typeahead_footer = mocker.patch(
+                                'zulipterminal.ui.View.set_typeahead_footer')
+        write_box.generic_autocomplete(text, state)
+
+        write_box.view.set_typeahead_footer.assert_called_once_with(
+                                                    footer_text,
+                                                    state,
+                                                    False)
 
     @pytest.mark.parametrize('text, state, required_typeahead', [
         ('@Human', 0, '@**Human Myself**'),
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -463,10 +463,11 @@ DataT = TypeVar('DataT')
 
 
 def match_stream(data: List[Tuple[DataT, str]], search_text: str,
-                 pinned_streams: List[List[Any]]) -> List[DataT]:
+                 pinned_streams: List[List[Any]]
+                 ) -> Tuple[List[DataT], List[str]]:
     """
-    Returns a list of DataT (streams) whose words match with the 'text' in the
-    following order:
+    Returns a list of DataT (streams) and a list of their corresponding names
+    whose words match with the 'text' in the following order:
     * 1st-word startswith match > 2nd-word startswith match > ... (pinned)
     * 1st-word startswith match > 2nd-word startswith match > ... (unpinned)
 
@@ -491,23 +492,26 @@ def match_stream(data: List[Tuple[DataT, str]], search_text: str,
     matches = OrderedDict([
         ('pinned', defaultdict(list)),
         ('unpinned', defaultdict(list)),
-    ])  # type: OrderedDict[str, DefaultDict[int, List[DataT]]]
+    ])  # type: OrderedDict[str, DefaultDict[int, List[Tuple[DataT, str]]]]
 
     for datum, splits in stream_splits:
-        kind = 'pinned' if splits[0] in pinned_stream_names else 'unpinned'
+        stream_name = splits[0]
+        kind = 'pinned' if stream_name in pinned_stream_names else 'unpinned'
         for match_position, word in enumerate(splits):
             if word.lower().startswith(search_text.lower()):
-                matches[kind][match_position].append(datum)
+                matches[kind][match_position].append((datum, stream_name))
 
     ordered_matches = []
+    ordered_names = []
     for matched_data in matches.values():
         if not matched_data:
             continue
         for match_position in range(max(matched_data.keys()) + 1):
-            for datum in matched_data.get(match_position, []):
+            for datum, name in matched_data.get(match_position, []):
                 if datum not in ordered_matches:
                     ordered_matches.append(datum)
-    return ordered_matches
+                    ordered_names.append(name)
+    return ordered_matches, ordered_names
 
 
 def match_group(group_name: str, text: str) -> bool:
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -518,6 +518,13 @@ def match_group(group_name: str, text: str) -> bool:
     return group_name.lower().startswith(text.lower())
 
 
+def format_string(names: List[str], wrapping_text: str) -> List[str]:
+    """
+    Wrap a list of names using the wrapping characters for typeahead
+    """
+    return [wrapping_text.format(name) for name in names]
+
+
 def powerset(iterable: Iterable[Any],
              map_func: Callable[[Any], Any]=set) -> List[Any]:
     """
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -77,6 +77,24 @@ class View(urwid.WidgetWrap):
             time.sleep(duration)
             self.set_footer_text()
 
+    @asynch
+    def set_typeahead_footer(self, suggestions: List[str],
+                             state: Optional[int],
+                             is_truncated: bool) -> None:
+        if suggestions:
+            # Wrap by space.
+            footer_text = [' ' + s + ' '
+                           for s in suggestions]  # type: List[Any]
+            if state is not None:
+                footer_text[state] = ('code', footer_text[state])
+            if is_truncated:
+                footer_text += [' [more] ']
+            footer_text.insert(0, [' '])  # Add leading space.
+        else:
+            footer_text = [' [No matches found]']
+
+        self.set_footer_text(footer_text, duration=3)
+
     def footer_view(self) -> Any:
         text_header = self.get_random_help()
         return urwid.AttrWrap(urwid.Text(text_header), 'footer')
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -93,7 +93,7 @@ class View(urwid.WidgetWrap):
         else:
             footer_text = [' [No matches found]']
 
-        self.set_footer_text(footer_text, duration=3)
+        self.set_footer_text(footer_text)
 
     def footer_view(self) -> Any:
         text_header = self.get_random_help()
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -13,7 +13,7 @@ from urwid_readline import ReadlineEdit
 from zulipterminal import emoji_names
 from zulipterminal.config.keys import is_command_key, keys_for_command
 from zulipterminal.helper import (
-    Message, match_emoji, match_group, match_stream, match_user,
+    Message, format_string, match_emoji, match_group, match_stream, match_user,
 )
 from zulipterminal.ui_tools.tables import render_table
 from zulipterminal.urwid_types import urwid_Size
@@ -115,14 +115,17 @@ class WriteBox(urwid.Pile):
                               ) -> List[str]:
         # Handles user mentions (@ mentions and silent mentions)
         # and group mentions.
-        group_typeahead = ['@*{}*'.format(group_name)
-                           for group_name in self.model.user_group_names
-                           if match_group(group_name, text[1:])]
+        groups = [group_name
+                  for group_name in self.model.user_group_names
+                  if match_group(group_name, text[1:])]
+        group_typeahead = format_string(groups, '@*{}*')
 
         users_list = self.view.users
-        user_typeahead = [prefix_string + '**{}**'.format(user['full_name'])
-                          for user in users_list
-                          if match_user(user, text[len(prefix_string):])]
+        users = [user['full_name']
+                 for user in users_list
+                 if match_user(user, text[len(prefix_string):])]
+        user_typeahead = format_string(users, prefix_string + '**{}**')
+
         combined_typeahead = group_typeahead + user_typeahead
 
         return combined_typeahead
@@ -130,17 +133,22 @@ class WriteBox(urwid.Pile):
     def autocomplete_streams(self, text: str, prefix_string: str
                              ) -> List[str]:
         streams_list = self.view.pinned_streams + self.view.unpinned_streams
-        stream_typeahead = [('#**{}**'.format(stream[0]), stream[0])
-                            for stream in streams_list]
-        return match_stream(stream_typeahead, text[1:],
+        streams = [stream[0]
+                   for stream in streams_list]
+        stream_typeahead = format_string(streams, '#**{}**')
+        stream_data = list(zip(stream_typeahead, streams))
+
+        return match_stream(stream_data, text[1:],
                             self.view.pinned_streams)
 
     def autocomplete_emojis(self, text: str, prefix_string: str
                             ) -> List[str]:
         emoji_list = emoji_names.EMOJI_NAMES
-        emoji_typeahead = [':{}:'.format(emoji)
-                           for emoji in emoji_list
-                           if match_emoji(emoji, text[1:])]
+        emojis = [emoji
+                  for emoji in emoji_list
+                  if match_emoji(emoji, text[1:])]
+        emoji_typeahead = format_string(emojis, ':{}:')
+
         return emoji_typeahead
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -103,19 +103,25 @@ class WriteBox(urwid.Pile):
 
         for prefix, autocomplete_func in autocomplete_map.items():
             if text.startswith(prefix):
-                typeaheads = autocomplete_func(text, prefix)
+                typeaheads, suggestions = autocomplete_func(text, prefix)
                 fewer_typeaheads = typeaheads[:num_suggestions]
+                reduced_suggestions = suggestions[:num_suggestions]
+                is_truncated = len(fewer_typeaheads) != len(typeaheads)
 
                 if (state is not None and state < len(fewer_typeaheads)
                         and state >= -len(fewer_typeaheads)):
-                    return fewer_typeaheads[state]
+                    typeahead = fewer_typeaheads[state]  # type: Optional[str]
                 else:
-                    return None
+                    typeahead = None
+                    state = None
+                self.view.set_typeahead_footer(reduced_suggestions,
+                                               state, is_truncated)
+                return typeahead
 
         return text
 
     def autocomplete_mentions(self, text: str, prefix_string: str
-                              ) -> List[str]:
+                              ) -> Tuple[List[str], List[str]]:
         # Handles user mentions (@ mentions and silent mentions)
         # and group mentions.
         groups = [group_name
@@ -130,29 +136,31 @@ class WriteBox(urwid.Pile):
         user_typeahead = format_string(users, prefix_string + '**{}**')
 
         combined_typeahead = group_typeahead + user_typeahead
+        combined_names = groups + users
 
-        return combined_typeahead
+        return combined_typeahead, combined_names
 
     def autocomplete_streams(self, text: str, prefix_string: str
-                             ) -> List[str]:
+                             ) -> Tuple[List[str], List[str]]:
         streams_list = self.view.pinned_streams + self.view.unpinned_streams
         streams = [stream[0]
                    for stream in streams_list]
         stream_typeahead = format_string(streams, '#**{}**')
         stream_data = list(zip(stream_typeahead, streams))
 
-        return match_stream(stream_data, text[1:],
-                            self.view.pinned_streams)[0]
+        matched_data = match_stream(stream_data, text[1:],
+                                    self.view.pinned_streams)
+        return matched_data
 
     def autocomplete_emojis(self, text: str, prefix_string: str
-                            ) -> List[str]:
+                            ) -> Tuple[List[str], List[str]]:
         emoji_list = emoji_names.EMOJI_NAMES
         emojis = [emoji
                   for emoji in emoji_list
                   if match_emoji(emoji, text[1:])]
         emoji_typeahead = format_string(emojis, ':{}:')
 
-        return emoji_typeahead
+        return emoji_typeahead, emojis
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         if is_command_key('SEND_MESSAGE', key):
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -25,6 +25,7 @@ class WriteBox(urwid.Pile):
         self.model = view.model
         self.view = view
         self.msg_edit_id = None  # type: Optional[int]
+        self.is_in_typeahead_mode = False
 
     def main_view(self, new: bool) -> Any:
         if new:
@@ -103,6 +104,7 @@ class WriteBox(urwid.Pile):
 
         for prefix, autocomplete_func in autocomplete_map.items():
             if text.startswith(prefix):
+                self.is_in_typeahead_mode = True
                 typeaheads, suggestions = autocomplete_func(text, prefix)
                 fewer_typeaheads = typeaheads[:num_suggestions]
                 reduced_suggestions = suggestions[:num_suggestions]
@@ -163,6 +165,13 @@ class WriteBox(urwid.Pile):
         return emoji_typeahead, emojis
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
+        if self.is_in_typeahead_mode:
+            if not (is_command_key('AUTOCOMPLETE', key)
+                    or is_command_key('AUTOCOMPLETE_REVERSE', key)):
+                # set default footer when done with autocomplete
+                self.is_in_typeahead_mode = False
+                self.view.set_footer_text()
+
         if is_command_key('SEND_MESSAGE', key):
             if self.msg_edit_id:
                 if not self.to_write_box:
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -91,7 +91,8 @@ class WriteBox(urwid.Pile):
         ]
         self.contents = write_box
 
-    def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
+    def generic_autocomplete(self, text: str, state: Optional[int]
+                             ) -> Optional[str]:
         num_suggestions = 10
         autocomplete_map = OrderedDict([
                 ('@_', self.autocomplete_mentions),
@@ -104,9 +105,11 @@ class WriteBox(urwid.Pile):
             if text.startswith(prefix):
                 typeaheads = autocomplete_func(text, prefix)
                 fewer_typeaheads = typeaheads[:num_suggestions]
-                try:
+
+                if (state is not None and state < len(fewer_typeaheads)
+                        and state >= -len(fewer_typeaheads)):
                     return fewer_typeaheads[state]
-                except (IndexError, TypeError):
+                else:
                     return None
 
         return text
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -92,6 +92,7 @@ class WriteBox(urwid.Pile):
         self.contents = write_box
 
     def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
+        num_suggestions = 10
         autocomplete_map = OrderedDict([
                 ('@_', self.autocomplete_mentions),
                 ('@', self.autocomplete_mentions),
@@ -102,8 +103,9 @@ class WriteBox(urwid.Pile):
         for prefix, autocomplete_func in autocomplete_map.items():
             if text.startswith(prefix):
                 typeaheads = autocomplete_func(text, prefix)
+                fewer_typeaheads = typeaheads[:num_suggestions]
                 try:
-                    return typeaheads[state]
+                    return fewer_typeaheads[state]
                 except (IndexError, TypeError):
                     return None
 
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -142,7 +142,7 @@ class WriteBox(urwid.Pile):
         stream_data = list(zip(stream_typeahead, streams))
 
         return match_stream(stream_data, text[1:],
-                            self.view.pinned_streams)
+                            self.view.pinned_streams)[0]
 
     def autocomplete_emojis(self, text: str, prefix_string: str
                             ) -> List[str]:
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -274,7 +274,7 @@ class StreamsView(urwid.Frame):
                 for stream in self.streams_btn_list.copy()
             ]
             streams_display = match_stream(stream_buttons, new_text,
-                                           self.view.pinned_streams)
+                                           self.view.pinned_streams)[0]
             self.log.clear()
             self.log.extend(streams_display)
             self.view.controller.update_screen()


kaustubh-nair:
Adding tests soon...

zulipterminal/ui_tools/boxes.py
```diff
@@ -107,26 +107,40 @@ def autocomplete_mentions(self, text: str, state: int,
...
-                           for group_name in self.model.user_group_names
-                           if match_groups(group_name, text[1:])]
-
-        users_list = self.view.users
-        user_typeahead = [prefix_string+'**{}**'.format(user['full_name'])
-                          for user in users_list
-                          if match_user(user, text[len(prefix_string):])]
+
+        group_typeahead = []
+        groups_list = []
```
preetmishra: I think the names `*_list` aren't as descriptive of what the variables hold. Maybe something like `*_suggestions`  would be more appropriate?
`group_suggestions`, `combined_suggestions`, for instance.

sumanthvrao:
@kaustubh-nair I have been running with this PR for a day now, and I find it incredibly useful! 🎉  

Here are some of my thoughts purely from a user experience POV.

* As preet mentioned in his review, we can set the footer back to default, once he escapes from the write box. We could even reset it when user cycles through all the type aheads and reaches the raw text they began with.
* Another doubt I had was whether to set `set_typeahead_footer` function as `@asynch` since it intern calls an asynchronous function(`set_footer_text`) and may go to blocking if the other function isn't executed immediately

I found a bug when testing your PR. Steps to reproduce:

* Search for `#zulip-` and press `ctrl f`
* The footer shows 2 options. Press `ctrl f` thrice until you cycle back to the raw search text, now press `ctrl r` twice.
* The footer shows `#zulip-term` twice. 
* I tried this with other streams and this forward and backward combination seems to throw an issue.

Nice work @kaustubh-nair!

zulipterminal/ui_tools/boxes.py
kaustubh-nair: We already have a list called `group_typeahead` so `group_suggestions` would be a bit confusing. I've added `group_names` instead

preetmishra: I see what you are saying but since the purpose of the list is to hold 'suggestions', `*_suggestions` would be a good alternative. Also, in my opinion, typeahead and suggestions won't be as confusing.  :)

kaustubh-nair: Yeah, on second-thought, `_suggestions` sounds better.

kaustubh-nair:
Thanks a lot for the reviews @preetmishra and @sumanthvrao 
Both of you raised some really nice points.
* Not highlighting the footer when the suggestion is reset is a good idea. I've added that feature.
* Calling `set_typeahead_footer` asynchronously should not cause any potential lags when displaying typeahead. Added that too.
* @sumanthvrao  I think I've fixed the bug that you are facing. Please check and let me know if you're still facing it.

I think there is some refactoring which can be done to decrease a bit of redundancy, I'll be doing that after adding the tests.

kaustubh-nair:
I've found a good fix for my original problem of trying to reset the footer to it's default text after a user finishes using typeahead.

I added a boolean variable `typeahead_mode` to determine when the user is using typeahead. Pressing any key when in this mode will reset the footer text. This is being done for any key, not only `space`, to support brackets, braces, etc.

This approach can be extended for a lot of useful followups. One of them would be to dynamically display the suggestions as the user modifies the text. (Similar to what is done on the web-app)
This can be achieved by: 
1. Displaying the suggestions in the footer on pressing `Ctrl + f`.
2. Add the wrapping characters `**{}** only on pressing a different key. ( Something similar to `tab`/`enter` )

kaustubh-nair:
Updated GIF and added tests.

sumanthvrao:
@kaustubh-nair Is this ready for another review?

kaustubh-nair:
@sumnthrao yes this is ready for review

neiljp:
@kaustubh-nair This is a great useful working feature :+1: I'll rely on @sumanthvrao and others to continue reviewing for the moment since they are familiar with your work on this.

Potential issues:
* If the screen is narrow then it can overflow - so limiting by space might work, or maybe we could just add another explicit bar, or scroll the suggestions?
* This PR restricts the suggestions (not just those displayed) to only 10, compared to many before? The webapp only shows 5 at once, so I can see how this limited/fixed-number approach can work (with more here!), however the webapp keeps filtering as you type. We *don't* show the possibilities until after ctrl-f/r, and then if it's too many (beyond the 10-limit) or we otherwise can't see it then we need to go back to the raw text to try again, leading to cycles of ctrl-f/ctrl-r/<edit>/ctrl-f/ctrl-r/<edit>/ctrl-f etc.

kaustubh-nair:
Rebased to include emoji typeaheads; ready for review

@neiljp I had thought of those issues as well, I will address them in followup issues. 
We could discuss the best possible way to approach the first one.
And I have mentioned a solution to the second one the the comment above ( https://github.com/zulip/zulip-terminal/pull/540#issuecomment-600000433 )

kaustubh-nair:
I split the third commit for easy reviewing

zulipterminal/ui.py
```diff
@@ -79,6 +79,29 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
...
+            # FIXME deal with suggestions having same names
+            footer_text = [('code', x)
+                           if x == ' ' + suggestion + ' '
+                           else x
+                           for x in suggestions]
+        else:
+            footer_text = suggestions
+
+        # no suggestions
+        if footer_text == []:
```
preetmishra: You can use `if not footer_text` here or can even make the block a one-liner.

tests/ui/test_ui.py
```diff
@@ -81,6 +81,29 @@ def test_set_footer_text_with_duration(self, view, mocker,
...
+        (1, [' help ', ('code', ' community '), ' issues ']),
+        (2, [' help ', ' community ', ('code', ' issues ')]),
+        (3, [' help ', ' community ', ' issues ']),
+        (None, [' help ', ' community ', ' issues ']),
+        (30, [' help ', ' community ', ' issues ']),
+    ])
+    def test_set_typeahead_footer(self, mocker, view, suggestions,
+                                  state, footer_text):
+        set_footer_text = mocker.patch('zulipterminal.ui.View.set_footer_text')
+        view.set_typeahead_footer(suggestions[:3], state)
```
preetmishra: You can parameterize the number of suggestions rather than hard-coding (`3`) it here to make it more readable. Also, you might want to add a little variation in the number of suggestions along with that.

tests/ui_tools/test_boxes.py
```diff
@@ -17,6 +17,25 @@ def write_box(self, mocker):
         write_box = WriteBox(self.view)
         return write_box
 
+    @pytest.fixture()
+    def write_box_users(self, mocker, users_fixture, user_groups_fixture):
+        write_box = WriteBox(self.view)
```
preetmishra: You might want to use the `write_box` fixture that is already defined.

tests/ui_tools/test_boxes.py
```diff
@@ -130,6 +130,22 @@ def test_autocomplete_mentions_first_n(self, write_box_users, n, text,
             text, state, '@', n)
         assert typeahead_string == required_typeahead
 
+    @pytest.mark.parametrize('text, state, users_suggestions', [
+        ('@Human', 0, ['Human Myself', 'Human 1', 'Human 2']),
+        ('@Human', 1, ['Human Myself', 'Human 1', 'Human 2']),
+        ('@Group', 0, ['Group 1', 'Group 2', 'Group 3', 'Group 4']),
+        ('@Group', 1, ['Group 1', 'Group 2', 'Group 3', 'Group 4']),
+        ('@Human', None, ['Human Myself', 'Human 1', 'Human 2']),
```
preetmishra: We could use a test case for silent mentions.

zulipterminal/ui_tools/boxes.py
```diff
@@ -25,6 +25,7 @@ def __init__(self, view: Any) -> None:
         self.model = view.model
         self.view = view
         self.msg_edit_id = None  # type: Optional[int]
+        self.typeahead_mode = False
```
preetmishra: Nice work! :+1:

preetmishra: Nice work! :+1:

preetmishra: Nice work! :+1:

zulipterminal/ui_tools/boxes.py
```diff
@@ -176,6 +180,13 @@ def autocomplete_emojis(self, text: str, state: int,
             return None
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
+        if self.typeahead_mode:
+            if not (is_command_key('AUTOCOMPLETE', key) or
+                    is_command_key('AUTOCOMPLETE_REVERSE', key)):
+                # set default footer when done with autocomplete
+                self.typeahead_mode = False
+                self.view.set_footer_text()
+
```
preetmishra: We could use a test for this.

kaustubh-nair:
Thanks a lot for the review @preetmishra!

To answer some of your comments,
> However, as mentioned by @neiljp, the num_suggestions sets a limitation on how many suggestions I can go back and forth to. Try Ctrl + f for '#'.
Well that's the whole point. If all suggestions are displayed, the footer will consume the whole page. Also, as I've mentioned in the above comments, the last commit can be extended to display typeaheads in real-time. We can have a follow-up PR for this, so this issue would not be a problem.

> The *_first_n tests that you have written are like the autocomplete_mentions/streams tests that we already have, what we might want here to test is whether or not the function(s) returns num_suggestions number of suggestions and in the expected order.
`autocomplete_mentions()` returns only a single suggestion, not the entire list. I've tested this by ensuring it returns a suggestion only if `state < num_suggestions`. With regards to the order, that is taken care of by the other tests.

> The commit feels out of place, it says 'Display only first 10 suggestions' but note that we don't have any displaying mechanism at this point (you introduce it in the next commits).
I've renamed the commit to 'Restrict typeahead suggestions to 10`. I think that makes more sense?

> The fifth commit shouldn't be a stand-alone commit or get introduced after the 'new' feature you're adding as the feature doesn't work properly without it (i.e. footer text changes on every keystroke).
I agree with this. But I wanted to keep it separate for the sake of reviewing, since it's a completely different feature. We can probable squash the fourth and fifth commit while merging ( third commit too maybe? ) so I think it should be okay.

kaustubh-nair:
Also I've made the changes and pushed the modified commits

tests/ui_tools/test_boxes.py
```diff
@@ -17,6 +17,24 @@ def write_box(self, mocker):
...
+    @pytest.fixture()
+    def write_box_users(self, mocker, users_fixture, user_groups_fixture,
+                        write_box):
+        write_box.view.users = users_fixture
+        write_box.model.user_group_names = [
+            groups['name'] for groups in user_groups_fixture]
+        return write_box
+
+    @pytest.fixture()
+    def write_box_streams(self, mocker, streams_fixture, write_box):
```
neiljp: These new local fixtures seem OK, but it may be worth investigating if it would be more straightforward to extract common fixtures related to the groups and pinned streams.

zulipterminal/ui_tools/boxes.py
```diff
@@ -119,27 +122,32 @@ def autocomplete_mentions(self, text: str, state: int,
                           for user in users_list
                           if match_user(user, text[len(prefix_string):])]
         combined_typeahead = group_typeahead + user_typeahead
+        combined_typeahead = combined_typeahead[:num_suggestions]
```
neiljp: While this reuses a variable name, my preference would be to give it a different name (`limited_typeahead`?) or to simply place the `:num_suggestions` into the try clause.

zulipterminal/ui_tools/boxes.py
```diff
@@ -113,14 +113,17 @@ def autocomplete_mentions(self, text: str, state: int, prefix_string: str,
...
-                           if match_groups(group_name, text[1:])]
-
-        users_list = self.view.users
-        user_typeahead = [prefix_string+'**{}**'.format(user['full_name'])
-                          for user in users_list
-                          if match_user(user, text[len(prefix_string):])]
+        group_typeahead = []
+        for group_name in self.model.user_group_names:
+            if match_groups(group_name, text[1:]):
+                group_typeahead.append('@*{}*'.format(group_name))
```
neiljp: I'm wary of this refactoring. While I'm not averse to for loops, I think this could be simpler by constructing a filtered name list, and then a 'styled' list for the typeahead.

See main review note too.

tests/ui_tools/test_boxes.py
kaustubh-nair: You mean combine the fixtures for streams and groups in the writebox fixture?

neiljp: Well, I recall pinning tests doing something similar to what you have here.

zulipterminal/ui_tools/boxes.py
```diff
@@ -130,24 +132,25 @@ def autocomplete_mentions(self, text: str, prefix_string: str
...
 
-    def autocomplete_streams(self, text: str) -> List[str]:
+    def autocomplete_streams(self, text: str) -> Tuple[List[str], List[str]]:
         streams_list = self.view.pinned_streams + self.view.unpinned_streams
         streams = [stream[0]
                    for stream in streams_list
                    if match_stream(stream, text[1:])]
         stream_typeahead = style_typeahead(streams, '#**{}**')
-        return stream_typeahead
+        return streams, stream_typeahead
```
kaustubh-nair, line 147: I wanted the `autocomplete_*` functions to return only the list of names, and do the styling in `generic_autocomplete`. But this becomes complicated for mentions, since groups and users have different styling

zulipterminal/ui_tools/boxes.py
```diff
@@ -107,8 +108,9 @@ def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
         else:
             return text
 
+        reduced_typeahead = typeahead[:num_suggestions]
```
kaustubh-nair, line 114: Writing tests is a little difficult, so I've let this be since its a small change.

zulipterminal/ui_tools/boxes.py
neiljp: I can understand this, but am wondering if instead of`Tuple[List[str], List[str]]` it might be cleaner with `List[Tuple[str, str]]`; then you would just return one list and index into the tuple? This means you can build it at the same time in the comprehensions.

tests/ui/test_ui.py
```diff
@@ -81,6 +81,35 @@ def test_set_footer_text_with_duration(self, view, mocker,
         mock_sleep.assert_called_once_with(duration)
         assert view.controller.update_screen.call_count == 2
 
+    @pytest.mark.parametrize('state, footer_text, num_suggestions', [
+        (0, [('code', ' help '), ' community ', ' issues '], 3),
```
neiljp, line 85: If you're going to test the actual text, can we extract the style as a parameter at least?

zulipterminal/ui.py
```diff
@@ -79,6 +79,28 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
             time.sleep(duration)
             self.set_footer_text()
 
+    @asynch
+    def set_typeahead_footer(self, suggestions: List[Any], state: int) -> None:
+        try:
+            suggestion = suggestions[state]
+        except (IndexError, TypeError):
+            suggestion = None
```
neiljp, line 87: Is this a duplication of, or replacing, the code you just refactored in the other PR?

zulipterminal/ui_tools/boxes.py
```diff
@@ -153,6 +155,13 @@ def autocomplete_emojis(self, text: str) -> Tuple[List[str], List[str]]:
         return emojis, emoji_typeahead
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
+        if self.typeahead_mode:
```
neiljp, line 158: Without this commit, what is the expected behavior? The commit describes the new behavior, but not the previous - is this an enhancement on the footer code, or a requirement for it to work 'properly'?

kaustubh-nair: I've explained this in a comment in the discussion above, but I think I'll add it to the commit text too.
This commit fixes the problem of the footer continuing to display the suggestions once user is done with auto-complete. Once the footer is set, there's no way of knowing 'when' to reset the footer. Ideally this should be squashed with the previous commit while merging, but I've kept them seperate so there is a little distinction while reviewing.

tests/ui_tools/test_boxes.py
```diff
@@ -13,8 +13,22 @@ def mock_external_classes(self, mocker, initial_index):
...
+        write_box.view.users = users_fixture
+        write_box.model.user_group_names = [
+            groups['name'] for groups in user_groups_fixture]
+
+        write_box.view.pinned_streams = [
+            [stream['name']] for stream in
+            streams_fixture[:len(streams_fixture)//2]]
+        write_box.view.unpinned_streams = [
+            [stream['name']] for stream in
+            streams_fixture[len(streams_fixture)//2:]]
```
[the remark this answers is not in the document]
kaustubh-nair: Would it be enough to keep the elements for streams, mentions, and emojis in different fixtures? I would not prefer to keep things as they are - there would be considerable duplication in `test_generic_autocomplete_set_footer`. 
@preetmishra Would that work for your PR?

tests/ui_tools/test_boxes.py
```diff
@@ -82,12 +96,8 @@ def test_generic_autocomplete_no_prefix(self, mocker, write_box, text,
         ('@_', 4, None),  # Beyond end
         ('@_', -1, '@_**Human 2**'),
     ])
-    def test_generic_autocomplete_mentions(self, write_box, users_fixture,
-                                           text, state, required_typeahead,
-                                           user_groups_fixture):
-        write_box.view.users = users_fixture
-        write_box.model.user_group_names = [
-            groups['name'] for groups in user_groups_fixture]
+    def test_autocomplete_mentions(self, write_box, required_typeahead,
```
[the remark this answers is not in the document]
kaustubh-nair: Good catch

tests/ui/test_ui.py
kaustubh-nair: Good idea; I can use the fixture from `conftest.py` too.

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
preetmishra: @kaustubh-nair I would need a way to define what streams I want to be pinned (in parametrize) for every test case. Given the way we construct `pinned_streams`, it seems that we would need further refactoring just to accomplish that. I am thinking of _not_ using `streams_fixture` in https://github.com/zulip/zulip-terminal/pull/530#discussion_r403516773. Thoughts?

zulipterminal/ui.py
kaustubh-nair: I've dealt with this in the new refactor for `generic_autocomplete`

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
kaustubh-nair: Would it be a better idea to write a new set of test for testing pinned streams? The current streams test can be modified to not take pinning into consideration. This should decrease the fixture set but it will require very close inspection of the existing fixtures.

kaustubh-nair:
@preetmishra I've pushed your 9867896eb122bd76b8f3a1fd9a3f99c58eca2b82 commit from https://github.com/zulip/zulip-terminal/pull/542/ here as well since I'm going to be using those changes later on. I hope that's okay :)

@neiljp I've updated the PR - also refactored `generic_autocomplete`. I think it looks a little cleaner now.

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
preetmishra: @kaustubh-nair Could you explain how that would address my concern (that I mentioned in the previous comment)?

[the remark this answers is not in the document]
kaustubh-nair: Since the objective is to make pinned tests more explicit, one way to do it would be to test them seperately - this way you get more control over the way pinned streams are generated without affecting the current tests a lot. Does that make sense?
In any case, the `write_box` attributes for streams will require refactoring, so let me know what works best for you.

kaustubh-nair:
Hmm, Travis passed but it is not returning the status. I tried restarting the build but no luck

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
preetmishra: Yes, "more control over the way pinned streams are generated" would be better. :+1:

kaustubh-nair:
@zulipbot add "needs review"

zulipterminal/ui_tools/boxes.py
```diff
@@ -118,31 +119,39 @@ def autocomplete_mentions(self, text: str, prefix_string: str
...
     def autocomplete_streams(self, text: str,
                              prefix_string: str) -> List[str]:
         streams_list = self.view.pinned_streams + self.view.unpinned_streams
-        stream_typeahead = [('#**{}**'.format(stream[0]), stream[0])
-                            for stream in streams_list]
-        return match_stream(stream_typeahead, text[1:],
+        streams = [stream[0]
+                   for stream in streams_list]
+        stream_typeahead = generate_typeaheads(streams, '#**{}**')
+        stream_data = list(zip(stream_typeahead, streams))
```
kaustubh-nair, line 150: This is a temp fix. @preetmishra what do you think about refactoring `match_streams` to match only using stream names? I don't mind doing it myself, asking for inputs.

preetmishra: It could work for `autocomplete_streams` as the 'formatting' that is applied to each stream is simple. However, I think formatting/rebuilding buttons in `update_streams`, after every search,  would be an overhead.

zulipterminal/ui_tools/boxes.py
```diff
@@ -109,14 +109,17 @@ def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
                 typeaheads = autocomplete_func(text, prefix)
                 reduced_typeaheads = typeaheads[:num_suggestions]
                 try:
-                    return reduced_typeaheads[state]
+                    typeahead = reduced_typeaheads[state]
                 except (IndexError, TypeError):
-                    return None
+                    typeahead = [None, None]  # type: ignore
```
neiljp, line 117: Is the type ignore necessary due to the nested tuples?

Can we just return `None` here as before and handle the success below the try block?

zulipterminal/ui_tools/boxes.py
```diff
@@ -135,7 +138,7 @@ def autocomplete_mentions(self, text: str, prefix_string: str
         return combined_typeahead
 
     def autocomplete_streams(self, text: str,
-                             prefix_string: str) -> List[str]:
+                             prefix_string: str) -> List[Tuple[str, str]]:
```
neiljp, line 145: Can we be consistent with the style here, between the autocomplete functions?

tests/ui_tools/test_boxes.py
```diff
@@ -41,8 +42,48 @@ def test_init(self, write_box):
...
+        (':smi', None, ['smile', 'smiley', 'smirk'], ':{}:'),
+        (':NoMatch', None, [], ':{}:'),
+        ])
+    def test_generic_autocomplete_set_footer(self, mocker, write_box,
+                                             wrapping_text, text,
+                                             state, footer_text):
+        write_box.view.set_typeahead_footer = mocker.patch(
+                                'zulipterminal.ui.View.set_typeahead_footer')
+        suggestion = footer_text[state] if state is not None else None
+        footer_text = generate_typeaheads(footer_text, wrapping_text)
```
neiljp, line 82: These two lines are to enable calculation of the expected result? If so, maybe move them into a block with the assert?

zulipterminal/ui.py
```diff
@@ -77,6 +77,26 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
...
             self.set_footer_text()
 
+    @asynch
+    def set_typeahead_footer(self, suggestions: List[Tuple[Any]],
+                             suggestion: Optional[str]) -> None:
+        # wrap by space and use only name from tuple
+        suggestions = list(map(lambda x: ' ' + x[0] + ' ', suggestions))
+
+        if suggestion is not None:
+            # FIXME deal with suggestions having same names
```
neiljp, line 87: This is likely where we'll start to include ids, and those can be included in at least mentions. I'm not sure if there is ambiguity in other cases?

zulipterminal/ui_tools/boxes.py
neiljp: This does make the commit rather big; I'm not sure at this point how to split it, as it seems non-ideal right now to have it stay in the autocomplete state between commits.

zulipterminal/ui_tools/boxes.py
```diff
@@ -95,6 +95,7 @@ def stream_box_view(self, caption: str='', title: str='') -> None:
         self.contents = write_box
 
     def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
+        num_suggestions = 10
```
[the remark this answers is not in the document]
kaustubh-nair: I had actually implemented this before. The initial version could cycle through all possible suggestions, but I changed it to something simpler on Aman's suggestion.
If you want to have a look I have a working version of that on my fork ( https://github.com/kaustubh-nair/zulip-terminal/tree/autocomplete_footer_v1 )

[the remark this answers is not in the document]
kaustubh-nair: Another option to `[more]` would be something like `....` ? Could use unicode characters for dot too.

zulipterminal/ui_tools/boxes.py
kaustubh-nair: @neiljp I don't see where the nested tuples are being generated?

zulipterminal/ui.py
```diff
@@ -77,6 +77,26 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
...
+            footer_text = [('code', x)
+                           if x == ' ' + suggestion + ' '
+                           else x
+                           for x in suggestions]  # type: Any
```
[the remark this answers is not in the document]
kaustubh-nair: There's already 1 space before and after each entry. Should I add another one before?

zulipterminal/ui_tools/boxes.py
kaustubh-nair: We can't return `None` here because we still need to set the footer when the exception is raised. The exception is raised when a user has cycled through all the suggestions and returns back to the non-completed state.(Or goes back using `ctrl+r`)

kaustubh-nair: The tuples are of the form `('#**announce**', 'announce')`. I need to set the type to `Tuple[Optional[str], Optional[str]]` which doesn't fit the line so I added type ignore. mypy doesn't allow line breaks for type declarations(https://github.com/python/mypy/issues/4511)

But I changed it from list to tuple for consistency.

zulipterminal/ui_tools/boxes.py
kaustubh-nair: My bad, found it. I had forgotten `generate_typeaheads` returns a list of tuples.

zulipterminal/ui.py
kaustubh-nair: There isn't any in emojis, and I doubt there would be much in streams either.

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
kaustubh-nair: Added `[more]` to the end of footer text

zulipterminal/ui_tools/boxes.py
```diff
@@ -95,53 +96,83 @@ def stream_box_view(self, caption: str='', title: str='') -> None:
...
-                            for stream in streams_list]
-        return match_stream(stream_typeahead, text[1:],
-                            self.view.pinned_streams)
+        streams = [stream[0]
+                   for stream in streams_list]
+        stream_typeahead = format_string(streams, '#**{}**')
+
+        matched_typeaheads = match_stream(stream_typeahead, text[1:],
+                                          self.view.pinned_streams)
+        typeahead_data = list(map(lambda x: (x, x[3:-2]), matched_typeaheads))
```
kaustubh-nair, line 154: I still have to do this hacky stuff to generate the structure again.
It would be a lot more useful if `match_stream` only took a list of stream names and returned the matched list. Styling of the list could be done after that. But I guess, there would be problems in changing `match_streams` this due to the buttons?

neiljp: I think it would be reasonable to adapt `match_stream` return the text it is matching against too. That's a pretty minimal change, particularly if we later generalize `match_stream` to be useful for other matching, which seems likely?

zulipterminal/ui_tools/boxes.py
```diff
@@ -107,15 +107,20 @@ def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
             if text.startswith(prefix):
                 typeaheads = autocomplete_func(text, prefix)
                 reduced_typeaheads = typeaheads[:num_suggestions]
+                truncated = len(reduced_typeaheads) != len(typeaheads)
```
neiljp, line 113: It'd be useful to prefix this and similar boolean values with a `is_` (or sometimes `in_` or similar)

tests/ui_tools/test_boxes.py
```diff
@@ -217,6 +217,34 @@ def test_generic_autocomplete_emojis(self, write_box, text,
         typeahead_string = write_box.generic_autocomplete(text, state)
         assert typeahead_string == required_typeahead
 
+    @pytest.mark.parametrize(['key', 'current_typeahead_mode',
+                             'expected_typeahead_mode', 'set_footer_called'], [
```
neiljp, line 221: Doesn't `set_footer_called` represent `expect_footer_was_reset` or similar?
(partly that's since `set_footer` is called elsewhere and doesn't just reset it, so this is confusing?)

zulipterminal/ui_tools/boxes.py
neiljp: I'm mostly happy with this separation at this point, though leaving the UI as it is in the previous commit just seems a bit messy.

kaustubh-nair:
Merging the first few commits sounds good.

neiljp:
@kaustubh-nair Pushed those two commits with minor tweaks.

zulipterminal/ui_tools/boxes.py
kaustubh-nair: I could squash before merging?

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Ah, I see. So you want to call `set_footer_text( )` from this method itself instead of going to `set_typeahead_footer( )`? Well, that's certainly not a problem - only thing would be the `  [more]  ` addition would be duplicated.

zulipterminal/ui_tools/boxes.py
```diff
@@ -91,7 +91,7 @@ def stream_box_view(self, caption: str='', title: str='') -> None:
         ]
         self.contents = write_box
 
-    def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
+    def generic_autocomplete(self, text: str, state: Any) -> Optional[str]:
```
neiljp, line 95: Tighten the type down here?

zulipterminal/ui.py
```diff
@@ -77,6 +77,26 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
             time.sleep(duration)
             self.set_footer_text()
 
+    @asynch
+    def set_typeahead_footer(self, suggestions: List[Any],
```
neiljp, line 81: You loosened this type compared to the last iteration? What's the reasoning?

zulipterminal/ui.py
```diff
@@ -77,6 +77,26 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
...
+                             state: Optional[int],
+                             is_truncated: bool) -> None:
+        # wrap by space and use only name from tuple
+        suggestions = list(map(lambda x: ' ' + x[1] + ' ', suggestions))
```
neiljp, line 85: You only use the second element of the tuple? Why are you passing in both?

See another comment, where I suggest unzipping the list of tuples first.

zulipterminal/ui_tools/boxes.py
neiljp: It's fine to leave for now.

zulipterminal/ui_tools/boxes.py
```diff
@@ -102,6 +103,7 @@ def generic_autocomplete(self, text: str, state: Any) -> Optional[str]:
 
         for prefix, autocomplete_func in autocomplete_map.items():
             if text.startswith(prefix):
+                self.typeahead_mode = True
```
neiljp, line 106: minor: as per a comment from an earlier review cycle, this could have a `in_` prefix.

That would make it read `if self.in_typeahead_mode:` and similar elsewhere, so read more clearly.

tests/ui/test_ui.py
```diff
@@ -81,6 +82,32 @@ def test_set_footer_text_with_duration(self, view, mocker,
         mock_sleep.assert_called_once_with(duration)
         assert view.controller.update_screen.call_count == 2
 
+    @pytest.mark.parametrize('suggestions, state, truncated, footer_text', [
+        ([], None, False, [' [No matches found]']),
+        (['some', 'text'], None, False, [[' '], ' some ', ' text ']),
+        (['some', 'text'], None, True, [[' '], ' some ', ' text ',
+                                        ' [more] ']),
```
neiljp, line 89: minor: break the entire list onto the next line in these cases.

tests/ui_tools/test_boxes.py
```diff
@@ -41,8 +42,50 @@ def test_init(self, write_box):
     ])
     def test_generic_autocomplete_no_prefix(self, mocker, write_box, text,
                                             state):
+        write_box.view.set_typeahead_footer = mocker.patch(
+                                'zulipterminal.ui.View.set_typeahead_footer')
```
neiljp, line 46: If this patch works, this shouldn't need such a complex assignment, as per the other test in this commit.

zulipterminal/ui.py
```diff
@@ -77,6 +77,26 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
...
+
+        if state is not None:
+            suggestions[state] = ('code', suggestions[state])
+
+        if is_truncated:
+            suggestions += [' [more] ']
+
+        if not suggestions:
+            suggestions = [' [No matches found]']
```
neiljp, line 94: This is new compared to the last version?

zulipterminal/ui.py
```diff
@@ -77,6 +77,26 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
...
+        else:
+            suggestions.insert(0, [' '])  # Add leading space.
```
neiljp, line 96: It would be clearer if the logic above this conditional was in this else statement?

The leading space looks good!

zulipterminal/ui_tools/boxes.py
```diff
@@ -104,17 +104,22 @@ def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
             if text.startswith(prefix):
                 typeaheads = autocomplete_func(text, prefix)
                 reduced_typeaheads = typeaheads[:num_suggestions]
+                is_truncated = len(reduced_typeaheads) != len(typeaheads)
 
                 if state in range(-len(reduced_typeaheads),
                                   len(reduced_typeaheads)):
-                    return reduced_typeaheads[state]
+                    typeahead = reduced_typeaheads[state][0]
```
neiljp, line 113: Would it make sense to unzip the list of tuples here, so you could set typeahead to one element without the `[0]`, and pass the other unzipped list to `set_typeahead_footer` not as a list of tuples?

zulipterminal/ui_tools/boxes.py
```diff
@@ -104,17 +104,22 @@ def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
...
                 else:
-                    return None
+                    typeahead = None  # type: ignore
```
neiljp, line 115: We return an `Optional`, so what's the problem here?

zulipterminal/ui.py
kaustubh-nair: Yes, this is new.

zulipterminal/ui_tools/boxes.py
```diff
@@ -104,9 +104,11 @@ def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
             if text.startswith(prefix):
                 typeaheads = autocomplete_func(text, prefix)
                 reduced_typeaheads = typeaheads[:num_suggestions]
-                try:
+
+                if state in range(-len(reduced_typeaheads),
+                                  len(reduced_typeaheads)):
```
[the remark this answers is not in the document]
kaustubh-nair: I tried using `range` for a huge 20-digit number and it still works without any time delay.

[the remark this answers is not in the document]
kaustubh-nair: I used `range` because it's easier to write, than specifying bounds explicitly.
And the range is not `[-n , n]`, but rather `[-n, n)`. So I'd have to check for negative/positive states and compare those separately. This is a lot more concise.

tests/ui_tools/test_boxes.py
kaustubh-nair: What do you mean by this? Complex assignment of the patch?

tests/ui/test_ui.py
```diff
@@ -81,6 +82,32 @@ def test_set_footer_text_with_duration(self, view, mocker,
...
+        (['some', 'text'], -1, False, [[' '], ' some ', ('code', ' text ')]),
+    ], ids=[
+        'no_matches',
+        'no_highlight',
+        'no_hightlight_truncated',
+        'first_suggestion_hightlighted',
+        'first_suggestion_hightlighted_truncated',
+        'last_suggestion_hightlighted',
+        ])
+    def test_set_typeahead_footer(self, mocker, view, state, suggestions,
```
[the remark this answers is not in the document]
kaustubh-nair: Oops the fixture was being used before I pushed the refactor. Nothing misses your eye, does it? ;)

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
neiljp: Re a large range, that likely depends on `state`?

Also, I was leaning towards the more explicit `state is not None and -n <= state < n`, which is pretty clear and involves no data-structure.

tests/ui_tools/test_boxes.py
neiljp: As per the other test, you have eg. `set_typeahead_footer = mocker.patch(...`, which is enough to patch the method, and then you can test that local variable later.

zulipterminal/ui.py
kaustubh-nair: I can't make this `List[str]` because I need to do assign a tuple for highlighting
```
suggestions[state] = ('code', suggestions[state])
```

kaustubh-nair: The older version was `List[Tuple[str, str]]` which didn't cause problems here.

tests/ui_tools/test_boxes.py
kaustubh-nair: I have the same assignment in the other test too. I tried this and a local assignment does not work. Probably because `set_typeahead_footer` is called using `write_box.view`?

zulipterminal/ui.py
kaustubh-nair: Fixed by refactoring `match_stream`

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Fixed by refactoring `match_stream`

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
kaustubh-nair: Oh right, that can work too

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Fixed.

kaustubh-nair: The type is not determined by the return type, but by 
```
typeahead = reduced_typeaheads[state]
```

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Fixed

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
kaustubh-nair: Any suggestions for this?
Though, I think its good enough - as a user gets used to using ZT they would have figured out `[more]` is not a match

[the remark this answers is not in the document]
neiljp: This is a good start, certainly. This indicates to the user that there are more than those listed, which is important. However, it doesn't indicate aspects such as how to get the rest, how many more there are, etc. How best to show that may be debatable, so this seems like a good aspect to add to the general improvements to make to typeahead (maybe in the tracking issue).

zulipterminal/helper.py
```diff
@@ -491,23 +492,25 @@ def match_stream(data: List[Tuple[DataT, str]], search_text: str,
...
         ('unpinned', defaultdict(list)),
-    ])  # type: OrderedDict[str, DefaultDict[int, List[DataT]]]
+    ])  # type: OrderedDict[str, DefaultDict[int, List[Tuple[DataT, str]]]]
 
     for datum, splits in stream_splits:
         kind = 'pinned' if splits[0] in pinned_stream_names else 'unpinned'
         for match_position, word in enumerate(splits):
             if word.lower().startswith(search_text.lower()):
-                matches[kind][match_position].append(datum)
+                matches[kind][match_position].append((datum, splits[0]))
```
preetmishra, line 501: As `splits[0]` has two occurrences now (line 498 and line 501), it might be clearer to extract it as a variable - `stream_name`.

kaustubh-nair: Updated - thanks for the review :+1:

zulipterminal/ui.py
neiljp: Right, but from a logical perspective, we just know we have a `List[str]` coming in. You don't *have* to assign to the same name, and perhaps it would be clearer if you didn't? It doesn't really make much sense to have the not-found case be 'suggestions' by name, either.

zulipterminal/ui.py
```diff
@@ -77,6 +77,24 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
             time.sleep(duration)
             self.set_footer_text()
 
+    @asynch
+    def set_typeahead_footer(self, suggestions: List[Any],
+                             state: Optional[int],
+                             is_truncated: bool) -> None:
+        # Wrap by space.
+        suggestions = list(map(lambda x: ' ' + x + ' ', suggestions))
```
neiljp, line 85: I see you moved a lot of the suggestions code into the conditional block (if/else) - it looks easier to read :+1:

However, we only need this line if there are suggestions, so can it move into the block too?

zulipterminal/ui.py
```diff
@@ -77,6 +77,24 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
...
+        if suggestions:
+            if state is not None:
+                suggestions[state] = ('code', suggestions[state])
+            if is_truncated:
+                suggestions += [' [more] ']
+            suggestions.insert(0, [' '])  # Add leading space.
+        else:
+            suggestions = [' [No matches found]']
+
+        self.set_footer_text(suggestions)
```
neiljp, line 96: minor: To avoid the footer getting 'stuck' before the next commit, we could just use the time parameter here?

zulipterminal/ui_tools/boxes.py
```diff
@@ -163,6 +165,13 @@ def autocomplete_emojis(self, text: str, prefix_string: str
         return emoji_typeahead, emojis
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
+        if self.is_typeahead_mode:
```
neiljp, line 168: minor: I think I mentioned both `is_` and `in_`, but consider the commit text you have: "If the user is not in typeahead mode". Based on expressing that in code,  `in_` probably works better, right? You could even have `is_in_` since you have space to be verbose - assuming we leave the `not` from the commit text, to be a logical operator.

tests/ui_tools/test_boxes.py
```diff
@@ -215,6 +215,36 @@ def test_generic_autocomplete_emojis(self, write_box, text,
...
+        ('k', True, False, True),
+    ])
+    def test_keypress_typeahead_mode_autocomplete_key(self, mocker, write_box,
+                                                      current_typeahead_mode,
+                                                      expected_typeahead_mode,
+                                                      expect_footer_was_reset,
+                                                      key):
+
+        self.view.set_footer_text = mocker.patch(
+                                    'zulipterminal.ui.View.set_footer_text')
```
neiljp, line 238: Is the same simplification possible here, as discussed elsewhere? Can we just remove this line, since the entire view is mocked already?

tests/ui_tools/test_boxes.py
```diff
@@ -215,6 +215,36 @@ def test_generic_autocomplete_emojis(self, write_box, text,
...
+        write_box.is_typeahead_mode = current_typeahead_mode
+        size = (20,)
+        write_box.keypress(size, key)
```
neiljp, line 241: Please have blank lines around the actual test method to emphasize it. We can skip the blank L236 too.

zulipterminal/ui_tools/boxes.py
```diff
@@ -104,9 +104,11 @@ def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
             if text.startswith(prefix):
                 typeaheads = autocomplete_func(text, prefix)
                 reduced_typeaheads = typeaheads[:num_suggestions]
-                try:
+
+                if(state is not None and state < len(reduced_typeaheads)
+                        and state >= -len(reduced_typeaheads)):
```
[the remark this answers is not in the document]
kaustubh-nair: Added. 

No, the whole thing doesn't fit in one line.

tests/ui_tools/test_boxes.py
kaustubh-nair: Ah, right, this doesn't need to be patched at all.

zulipterminal/ui.py
kaustubh-nair: Did this, but if `ctrl f` is pressed 10 times, the help text will change 10 times. I guess we could let this slide since it's fixed in the next commit.

zulipterminal/ui.py
kaustubh-nair: When I assign it to a new name, I would still need a `List[Any]` type annotation?

So the best place to do it would be 
```python3
footer_text = list(map(lambda x: ' ' + x + ' ', suggestions))
```
But the type annotation will not fit in this line, so I'd end up doing
```python3
footer_text = suggestions  # type: List[Any]
```
Is it really worth adding a new line to copy a variable just to typecast it?

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Yeah, adding a type for `typeahead = reduced_typeaheads[state]` would be ideal but it does not fit the line

zulipterminal/ui_tools/boxes.py
```diff
@@ -91,7 +91,8 @@ def stream_box_view(self, caption: str='', title: str='') -> None:
         ]
         self.contents = write_box
 
-    def generic_autocomplete(self, text: str, state: int) -> Optional[str]:
+    def generic_autocomplete(self, text: str, state: Optional[int]
```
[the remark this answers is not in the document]
kaustubh-nair: It belongs in this commit because I'm doing an explicit `state = None` in this commit which wasn't required before.

kaustubh-nair:
@neiljp thanks for the review, this is updated

tests/ui_tools/test_boxes.py
```diff
@@ -43,6 +43,45 @@ def test_generic_autocomplete_no_prefix(self, mocker, write_box, text,
...
+        (':NoMatch', None, [], ':{}:'),
+    ])
+    def test_generic_autocomplete_set_footer(self, mocker, write_box,
+                                             wrapping_text, text,
+                                             state, footer_text):
+        write_box.view.set_typeahead_footer = mocker.patch(
+                                'zulipterminal.ui.View.set_typeahead_footer')
+        write_box.generic_autocomplete(text, state)
+
+        suggestion = footer_text[state] if state is not None else None
```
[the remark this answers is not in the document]
kaustubh-nair: Oops forgot to update this test after the recent refactor. Fixed.

neiljp:
@kaustubh-nair Thanks for your perseverance on this, I've just merged it manually with only minor commit text adjustments :tada:

PR #542, opened 2020-03-13
boxes: Improve trigger sequence for generic_autocomplete

### LINKED ISSUE #541: Typeahead doesn't initiate with some characters before trigger character
For example, it may be common to have `(#**some stream**)` in a conversation, or similarly with users. Currently this does not work, presumably because the `(` excludes the text from matching autocomplete?

The webapp allows this in some cases, but for example doesn't with emails, as you might expect (`foo@ba<ctrl-f>`, for ZT) since that could be an email, presumably.

We should probably check the webapp and/or mobile codebases for this, and maybe discuss with those developers the strategies they took and why, if that is not clear from their code.

preetmishra:
Updated generic_autocomplete along with its autocomplete_mentions/streams to provide typeahead even when there are some characters before their respective trigger character.

Fixes #541 and #448 (partially).

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -123,6 +123,24 @@ class TestWriteBox:
         ('@_', 3, None),  # Reached last match
         ('@_', 4, None),  # Beyond end
         ('@_', -1, '@_**Human 2**'),
+        # Complex autocomplete prefixes.
+        ('(@H', 0, '(@**Human Myself**'),
+        ('(@H', 1, '(@**Human 1**'),
+        ('-@G', 0, '-@*Group 1*'),
+        ('-@G', 1, '-@*Group 2*'),
+        ('_@H', 0, '_@**Human Myself**'),
+        ('_@G', 0, '_@*Group 1*'),
+        ('@@H', 0, '@@**Human Myself**'),
+        (':@H', 0, ':@**Human Myself**'),
+        ('#@H', 0, '#@**Human Myself**'),
+        ('@_@H', 0, '@_@**Human Myself**'),
+        ('>@_H', 0, '>@_**Human Myself**'),
+        ('>@_H', 1, '>@_**Human 1**'),
+        ('@_@_H', 0, '@_@_**Human Myself**'),
+        ('@@_H', 0, '@@_**Human Myself**'),
+        (':@_H', 0, ':@_**Human Myself**'),
+        ('#@_H', 0, '#@_**Human Myself**'),
+        ('@@_H', 0, '@@_**Human Myself**'),
     ])
     def test_generic_autocomplete_mentions(self, write_box, text,
                                            required_typeahead, state):
@@ -155,6 +173,13 @@ class TestWriteBox:
         ('#Stream 1', 0, '#**Stream 1**', []),  # Complete match.
         ('#nomatch', 0, None, []),
         ('#ene', 0, None, []),
+        # Complex autocomplete prefixes.
+        ('[#Stream', 0, '[#**Stream 1**', []),
+        ('(#Stream', 1, '(#**Stream 2**', []),
+        ('@#Stream', 0, '@#**Stream 1**', []),
+        ('@_#Stream', 0, '@_#**Stream 1**', []),
+        (':#Stream', 0, ':#**Stream 1**', []),
+        ('##Stream', 0, '##**Stream 1**', []),
         # With 'Secret stream' pinned.
         ('#Stream', 0, '#**Secret stream**',
          [['Secret stream'], ]),  # 2nd-word startswith match (pinned).
@@ -197,6 +222,12 @@ class TestWriteBox:
         (':', -1, ':smirk:'),
         (':nomatch', 0, None),
         (':nomatch', -1, None),
+        # Complex autocomplete prefixes.
+        ('(:smi', 0, '(:smile:'),
+        ('&:smi', 1, '&:smiley:'),
+        ('@:smi', 0, '@:smile:'),
+        ('@_:smi', 0, '@_:smile:'),
+        ('#:smi', 0, '#:smile:'),
         ])
     def test_generic_autocomplete_emojis(self, write_box, text,
                                          mocker, state, required_typeahead):
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -102,25 +102,45 @@ class WriteBox(urwid.Pile):
                 (':', self.autocomplete_emojis),
             ])
 
-        for prefix, autocomplete_func in autocomplete_map.items():
-            if text.startswith(prefix):
-                self.is_in_typeahead_mode = True
-                typeaheads, suggestions = autocomplete_func(text, prefix)
-                fewer_typeaheads = typeaheads[:num_suggestions]
-                reduced_suggestions = suggestions[:num_suggestions]
-                is_truncated = len(fewer_typeaheads) != len(typeaheads)
+        # Look in a reverse order to find the last autocomplete prefix used in
+        # the text. For instance, if text='@#example', use '#' as the prefix.
+        reversed_text = text[::-1]
+        for reverse_index, char in enumerate(reversed_text):
+            # Patch for silent mentions.
+            if (char == '_' and reverse_index + 1 < len(reversed_text)
+                    and reversed_text[reverse_index + 1] == '@'):
+                char = '@_'
 
-                if (state is not None and state < len(fewer_typeaheads)
-                        and state >= -len(fewer_typeaheads)):
-                    typeahead = fewer_typeaheads[state]  # type: Optional[str]
-                else:
-                    typeahead = None
-                    state = None
-                self.view.set_typeahead_footer(reduced_suggestions,
-                                               state, is_truncated)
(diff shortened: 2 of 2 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -107,27 +107,30 @@ def autocomplete_mentions(self, text: str, state: int,
                               prefix_string: str) -> Optional[str]:
         # Handles user mentions (@ mentions and silent mentions)
         # and group mentions.
+        at_index = max(text.find(prefix_string), 0)
```
kaustubh-nair: A more generic name like `prefix_index` would be more descriptive here

zulipterminal/ui_tools/boxes.py
```diff
@@ -107,27 +107,30 @@ def autocomplete_mentions(self, text: str, state: int,
...
+                                        text[len(prefix_string)+at_index:])]
         combined_typeahead = group_typeahead + user_typeahead
         try:
-            return combined_typeahead[state]
+            return text[:at_index] + combined_typeahead[state]
         except (IndexError, TypeError):
             return None
 
     def autocomplete_streams(self, text: str, state: int) -> Optional[str]:
+        hashtag_index = max(text.find('#'), 0)
```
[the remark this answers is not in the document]
preetmishra: Changing the `autocomplete_streams`' signature would require its test to be amended as well which is not related to this PR (if that makes sense). I think this can be a follow-up though. 
@neiljp What are your thoughts on this?

zulipterminal/ui_tools/boxes.py
preetmishra: Thanks, `prefix_index` would be more appropriate.

preetmishra:
@kaustubh-nair Thanks for the review. I have pushed the amended tests now.

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
kaustubh-nair: Hmm, since it's a minor change, maybe you could add it in a separate commit before this one? 
This will not only separate unrelated changes but will also ensure that the current commit is of better quality. Just throwing out suggestions, feel free to do a follow-up if you like :)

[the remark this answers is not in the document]
preetmishra: I was suggesting a follow-up for the sake of easiness in terms of reviewing.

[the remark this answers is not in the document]
kaustubh-nair: Well, PRs are generally reviewed commit-wise, so I don't think that should be a problem either :)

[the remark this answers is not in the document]
preetmishra: Alright, I will push the changes.

preetmishra:
Update: I have added a refactor commit to make `autocomplete_streams`' signature consistent with `autocomplete_mentions` as per the suggestion of @kaustubh-nair. (https://github.com/zulip/zulip-terminal/pull/542#pullrequestreview-374700869)

preetmishra:
@sumanthvrao Thanks. I have pushed a new commit to address the latter point.

preetmishra:
The Travis CI build is failing with 'An error occurred while generating the build script.' for `Install & pytest - Python 3 (OS X; xcode10)`. I tried to restart it with force push but it didn't work again.

sumanthvrao:
@preetmishra thanks for addressing the minor feedback quickly. the additional commit LGTM! 

@neiljp FYI :)

tests/ui_tools/test_boxes.py
```diff
@@ -104,35 +104,38 @@ def test_autocomplete_mentions(self, write_box, users_fixture,
...
+        ('#S', -5, '#', None),
+        ('#So', 0, '#', '#**Some general stream**'),
+        ('#So', 1, '#', None),
+        ('#Se', 0, '#', '#**Secret stream**'),
+        ('#Se', 1, '#', None),
+        ('#St', 0, '#', '#**Stream 1**'),
+        ('#St', 1, '#', '#**Stream 2**'),
+        ('#Stream 1', 0, '#', '#**Stream 1**'),
+        ('No match', 0, '#', None),
+        ('No match', -1, '#', None)
```
kaustubh-nair: @preetmishra  It would be better to pass the prefix string in a separate `parameterize` block, since the value is the same for all the fixtures.

preetmishra: @kaustubh-nair Indeed. Thanks for pointing that out. :)

preetmishra:
Update: Rebased to include `autocomplete_emojis` in the PR as well.

preetmishra:
Update: Rebased to include `autocomplete_emojis`, that was added in 2c30caf, in the PR as well.

neiljp:
@preetmishra Sorry for not getting back to this; could you rebase?

zulipterminal/ui_tools/boxes.py
```diff
@@ -110,7 +110,7 @@ def autocomplete_mentions(self, text: str, state: int,
                               prefix_string: str) -> Optional[str]:
         # Handles user mentions (@ mentions and silent mentions)
         # and group mentions.
-        group_typeahead = ['@*{}*'.format(group_name)
+        group_typeahead = [prefix_string+'*{}*'.format(group_name)
```
kaustubh-nair, line 114: Group typeaheads don't support silent mentions I think?

preetmishra: @kaustubh-nair Indeed, they don't. The `match_group(s)` function takes `text[1:]` as an argument which means that it'd never return `True` for silent mentions ('@_').

preetmishra:
@neiljp No problem; I have rebased the PR. :)

The first commit was a suggestion from @sumanthvrao to improve consistency. The next two are self-explanatory refactor commits that facilitate the new change.

kaustubh-nair:
@preetmishra I just noticed that you recently pushed a07c87d4c32735d0c4a7031deb0231a3816b4e4f for refactoring `generic_autocomplete`. While it looks good but there is still a little bit of duplication. I had pushed similar changes to the footer PR.

I'd recommend holding off refactors to autocomplete methods for a while, since there are already so many of them in the footer PR. Adding them in different PRs will cause a lot of confusion. 
That being said, if you want to make any changes, feel free to push to the `autocomplete_footer` branch on my fork. It would be better if we keep autocomplete refactors in the same place :)

preetmishra:
@kaustubh-nair Thanks for taking a look. :+1:  

I see what you're suggesting. While keeping the refactors in the same PR sounds valid, the fact that this PR is comparatively small and easier to review makes it likely to be merged quickly. If that's the case, you can always rebase against `master` to incorporate the new changes.

I have updated the commit to make `generic_autocomplete` less redundant and more pythonic.

kaustubh-nair:
@preetmishra Sorry I did not make myself more clear - One of the major reasons to avoid the refactors at different places was so that both of us don't waste our time working on the same thing. I have a similar commit [here](https://github.com/zulip/zulip-terminal/pull/540/commits/c01d4cd389afbb4fe88705d6880d3403ec37a83b) which is used for the footer PR

preetmishra:
I have rebased the PR against the recent refactors that were pushed in #540. This should be independent and comparatively smaller to review now.

neiljp:
@preetmishra Could you rebase? Also, the first commit is small now? I'll be interested to confirm how the code works for silent group mentions (which don't exist, by my understanding), and whether there are any strange cases where `in` vs `startswith` causes strange behavior

zulipterminal/ui_tools/boxes.py
```diff
@@ -113,7 +113,7 @@ def autocomplete_mentions(self, text: str, prefix_string: str
                               ) -> List[str]:
         # Handles user mentions (@ mentions and silent mentions)
         # and group mentions.
-        group_typeahead = ['@*{}*'.format(group_name)
+        group_typeahead = [prefix_string + '*{}*'.format(group_name)
                            for group_name in self.model.user_group_names
                            if match_group(group_name, text[1:])]
```
preetmishra, line 119: This worked and didn't show any suggestion for `@_` or silent group mentions because `match_group` received `text[1:]` instead of `text[len(prefix_string):]`.

However, on second thought, the existing code, with hard-coded `@`, seems clearer and doesn't have any potential anomaly. I'll shelve this commit.

preetmishra:
@neiljp Thanks for the review. I have rebased the changes and shelved the first commit (see https://github.com/zulip/zulip-terminal/pull/542#discussion_r449891100).

Re strange cases, I didn't come across any strange case while testing it manually or analyzing its logic. This shall work well.

preetmishra:
@neiljp Thanks for the catching that out! I have reworked the approach to find the last autocomplete prefix using a reverse lookup.

This should fix the inconsistencies that you reported (test cases are added) and should be more reliable.

neiljp:
@preetmishra This looks good - and if we refactor the code later the tests should ensure things keep working :+1: Merging this now :tada:

We could maybe make the code simpler and without the specific case for the silent mention, but we have a limited subset of autocomplete so this is fine for now. I'm not sure if eg. a sorted rfind of prefixes might work?

preetmishra:
Thanks for the merge! The sorted rfind of prefixes sounds appealing for refactoring. :raised_hands:

PR #548, opened 2020-03-18
WIP: tests: helper: Add tests for set_count.

kaustubh-nair:
- [x] Test streams
- [x] Test pms
- [x] Test self-pms
- [x] Test group pms
- [x] ~~Test messages sent by user~~ ( `set_count` not called)
- [x] Test muted streams
- [x] Test muted topics
- [ ] ~~Test topics whose view is open~~ ( UI test, fix in different PR )
- [x] Test screen updated
- [ ] Test mentions

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -51,6 +51,36 @@ def test_set_count_stream(controller, initial_unread_counts,
     assert controller.model.unread_counts == expected_unread_counts
 
 
+@pytest.mark.parametrize('all_msg, all_pms, unread_pms, user_id', [
+    (4, 3, {200: 2, 201: 1}, 202)
+    ])
+@pytest.mark.parametrize(('new_count, id_list, expected_all_msg,'
+                         'expected_all_pms, expected_unread_pms'), [
+    (-1, [10, 13], 2, 1, {201: 1}),
+    (-1, [10, 13, 11], 1, 0, {}),
+    (-1, [], 4, 3, {200: 2, 201: 1}),
+    (-1, [15], 4, 3, {200: 2, 201: 1}),   # self-pm
+    (1, [16, 10], 6, 5, {200: 4, 201: 1}),
+    (1, [], 4, 3, {200: 2, 201: 1}),
+    ])
+def test_set_count_pms(controller, initial_unread_counts, all_msg,
+                       all_pms, unread_pms, user_id, id_list, new_count,
+                       expected_unread_pms, expected_all_pms,
+                       expected_all_msg):
+    controller.model.user_id = user_id
+    unread_counts = deepcopy(dict(initial_unread_counts,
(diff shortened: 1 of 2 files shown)


kaustubh-nair:
This is an important stepping stone to fixing #535.
Handling read events will require modifications to `set_count`. Since, `set_count` is not being tested at all, I'm adding tests for the same. It might be difficult to cover every faucet of `set_counts`, but the goal is to test atleast the major features, so that significant bugs do not arise.

kaustubh-nair:
There's will be a little bit of duplication across the tests for setting up and initializing variables. I'm wondering if it would be viable to abstract out the setup, or would it make the tests too complex?

tests/helper/test_helper.py
```diff
@@ -1,13 +1,85 @@
...
+    ])
+@pytest.mark.parametrize(('new_count, id_list, expected_all_msg,'
+                         'expected_all_pms, expected_unread_pms'), [
+    (-1, [10, 13], 2, 1, {201: 1}),
+    (-1, [10, 13, 11], 1, 0, {}),
+    (-1, [], 4, 3, {200: 2, 201: 1}),
+    (-1, [15], 4, 3, {200: 2, 201: 1}),   # self-pm
+    (1, [16, 10], 6, 5, {200: 4, 201: 1}),
+    (1, [], 4, 3, {200: 2, 201: 1}),
+    # (1, [15], 4, 3, {200: 2, 201: 1}),   # self-pm  # FIXME
```
kaustubh-nair: The test fails when a pm sent by the user is to be marked as unread. I don't know if such a case will ever arise, but `set_count` should probably handle it anyway.

kaustubh-nair:
@neiljp @sumanthvrao, if you have time to give this a review please let me know if these tests look okay. :) 
And also the extent to which I should be testing `set_count`, so I can go ahead with the rest of the tests?

tests/helper/test_helper.py
sumanthvrao: I vaguely remember discovering the same fact a while ago. I think the reason I concluded was that the server does not set the unread for self PM's hence there is no point setting the counts from ZT side

sumanthvrao:
This looks like a great start @kaustubh-nair .  If possible we can try reusing some fixtures in `conftest.py` which may help us avoid the duplication (not entirely I suppose).

The list you have mentioned above seems comprehensive enough! Let's go ahead with that.

kaustubh-nair:
The fixtures in `conftest.py` don't contain enough messages to properly test the counts, so I ended up adding new ones

tests/helper/test_helper.py
kaustubh-nair: Okay, I'll remove this then

kaustubh-nair:
@sumanthvrao @neiljp This is ready for an initial review. I do realize there is a LOT of duplication going on here. I will push some refactor commits sometime later tomorrow.

PR #549, opened 2020-03-18
Fixed Focus in Search Stream

Ar9av:
- Stream maintains order when text is removed after entering in the search stream

- Added pytest test case for the same

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -415,6 +415,8 @@ class TestStreamsView:
         ('FOO', ['FOO', 'FOOBAR', 'foo']),
         ('test', ['test here']),
         ('here', ['test here']),
+        ('', ['FOO', 'FOOBAR', 'foo', 'fan',
+            'boo', 'BOO', 'bar', 'test here'])
     ])
     def test_update_streams(self, mocker, stream_view, new_text, expected_log):
         stream_names = [
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -278,6 +278,7 @@ class StreamsView(urwid.Frame):
             ]
             self.log.clear()
             self.log.extend(streams_display)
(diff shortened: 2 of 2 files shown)


preetmishra:
@Ar9av Hey, thanks for your initiative to contribute to Zulip. It looks like your code is failing a linting test, `pycodestyle` specifically.

It is recommended to run all the linting tests along with `pytest` locally before you push your code for review (You can read more about it [here](https://github.com/zulip/zulip-terminal#development-tasks)).

However, if you would like to discuss why the linting test is failing, you can post your query here or in the [#zulip-terminal](https://chat.zulip.org/#narrow/stream/206-zulip-terminal) stream. :)

tests/ui/test_ui_tools.py
```diff
@@ -415,6 +415,8 @@ def test_init(self, mocker, stream_view):
         ('FOO', ['FOO', 'FOOBAR', 'foo']),
         ('test', ['test here']),
         ('here', ['test here']),
+        ('', ['FOO', 'FOOBAR', 'foo', 'fan',
+            'boo', 'BOO', 'bar', 'test here'])
     ])
     def test_update_streams(self, mocker, stream_view, new_text, expected_log):
         stream_names = [
```
amanagr: @Ar9av I think you just need test if `log.set_focus` was called. No other change in needed I think.

neiljp:
@Ar9av Since you've not referenced an issue, could you describe the bug this is fixing in a little more detail?

neiljp:
@Ar9av Hey, just checking in with you on this; we'd love to have your contribution, but it'd be good to hear more from you regarding this :+1:

We did fix some bugs recently for the stream searching, so this may not be an issue any more, but please let us know :)

neiljp:
@Ar9av Since we've not heard from you in a while I'm going to close this pull request. If you want to continue this work, you're welcome to do so, though please identify which problem you're trying to address up-front.

PR #550, opened 2020-03-18
boxes/helper: Add support for rendering tables.

### LINKED ISSUE #452: TRACKING: Message rendering bugs/enhancements
- [ ] #402 : quoting (other cases may be present)
- [ ] Render horizontal rulers
- [ ] Render tables
- [ ] #415 Ascii images
- [ ] #359 PR: Images in the default browser
- [ ] #424 PR: Messages links popup menu
- [ ] #397 PR: Viewing a message in the browser

preetmishra:
I have added a helper function, render_table, to process a table and return its content as a list of strings which can then be rendered in the MessageBox and amended the test.

This implementation is v1. Feedbacks are welcomed. :) 
There can certainly be follow-ups to improve this further.

Partially fixes #452.

@neiljp

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1397,7 +1397,59 @@ class TestMessageBox:
         ('<hr/>', ['[RULER NOT RENDERED]']),
         ('<img>', ['[IMAGE NOT RENDERED]']),
         ('<img/>', ['[IMAGE NOT RENDERED]']),
-        ('<table>stuff</table>', ['[TABLE NOT RENDERED]']),
+        ('<table><thead><tr><th>Firstname</th><th>Lastname</th></tr></thead>'
+         '<tbody><tr><td>John</td><td>Doe</td></tr><tr><td>Mary</td><td>Moe'
+         '</td></tr></tbody></table>', [
+            '┌─', '─────────', '─┬─', '────────', '─┐\n',
+            '│ ', ('table_head', 'Firstname'), ' │ ',
+            ('table_head', 'Lastname'), ' │\n',
+            '├─', '─────────', '─┼─', '────────', '─┤\n',
+            '│ ', (None, 'John     '), ' │ ', (None, 'Doe     '), ' │\n',
+            '│ ', (None, 'Mary     '), ' │ ', (None, 'Moe     '), ' │\n',
+            '└─', '─────────', '─┴─', '────────', '─┘',
+         ]),
+        ('<table><thead><tr><th align="left">Name</th><th align="right">Id'
+         '</th></tr></thead><tbody><tr><td align="left">Robert</td>'
+         '<td align="right">1</td></tr><tr><td align="left">Mary</td>'
+         '<td align="right">100</td></tr></tbody></table>', [
+            '┌─', '──────', '─┬─', '───', '─┐\n',
+            '│ ', ('table_head', 'Name  '), ' │ ', ('table_head', ' Id'),
+            ' │\n',
+            '├─', '──────', '─┼─', '───', '─┤\n',
+            '│ ', (None, 'Robert'), ' │ ', (None,  '  1'), ' │\n',
+            '│ ', (None, 'Mary  '), ' │ ', (None, '100'), ' │\n',
+            '└─', '──────', '─┴─', '───', '─┘',
+         ]),
+        ('<table><thead><tr><th align="center">Name</th><th align="right">Id'
+         '</th></tr></thead><tbody><tr><td align="center">Robert</td>'
+         '<td align="right">1</td></tr><tr><td align="center">Mary</td>'
+         '<td align="right">100</td></tr></tbody></table>', [
+            '┌─', '──────', '─┬─', '───', '─┐\n',
+            '│ ', ('table_head', ' Name '), ' │ ', ('table_head', ' Id'),
+            ' │\n',
+            '├─', '──────', '─┼─', '───', '─┤\n',
+            '│ ', (None, 'Robert'), ' │ ', (None,  '  1'), ' │\n',
+            '│ ', (None, ' Mary '), ' │ ', (None, '100'), ' │\n',
+            '└─', '──────', '─┴─', '───', '─┘',
+         ]),
+        ('<table><thead><tr><th>Name</th></tr></thead><tbody><tr><td>Foo</td>'
+         '</tr><tr><td>Bar</td></tr><tr><td>Baz</td></tr></tbody></table>', [
+            '┌─', '────', '─┐\n',
+            '│ ', ('table_head', 'Name'), ' │\n',
+            '├─', '────', '─┤\n',
+            '│ ', (None, 'Foo '), ' │\n',
+            '│ ', (None, 'Bar '), ' │\n',
+            '│ ', (None, 'Baz '), ' │\n',
+            '└─', '────', '─┘',
+         ]),
+        ('<table><thead><tr><th>Column1</th></tr></thead><tbody><tr><td></td>'
+         '</tr></tbody></table>', [
+            '┌─', '───────', '─┐\n',
+            '│ ', ('table_head', 'Column1'), ' │\n',
+            '├─', '───────', '─┤\n',
+            '│ ', (None,  '       '), ' │\n',
+            '└─', '───────', '─┘',
+         ]),
         ('<span class="katex-display">some-math</span>', ['some-math']),
         ('<span class="katex">some-math</span>', ['some-math']),
         ('<ul><li>text</li></ul>', ['', '  * ', '', 'text']),
@@ -1417,7 +1469,10 @@ class TestMessageBox:
         'link_sametext', 'link_sameimage', 'link_differenttext',
         'link_userupload', 'link_api', 'link_serverrelative_same',
         'listitem', 'listitems',
-        'br', 'br2', 'hr', 'hr2', 'img', 'img2', 'table', 'math', 'math2',
+        'br', 'br2', 'hr', 'hr2', 'img', 'img2', 'table_default',
+        'table_with_left_and_right_alignments',
+        'table_with_center_and_right_alignments', 'table_with_single_column',
+        'table_with_the_bare_minimum', 'math', 'math2',
         'ul', 'strikethrough_del', 'inline_image', 'inline_ref',
         'emoji', 'preview-twitter', 'zulip_extra_emoji', 'custom_emoji'
     ])
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -10,7 +10,7 @@ from re import ASCII, match
 from threading import Thread
 from typing import (
     Any, Callable, DefaultDict, Dict, FrozenSet, Iterable, List, Optional, Set,
-    Tuple, Union,
+    Tuple, Union, cast,
 )
 
 import lxml.html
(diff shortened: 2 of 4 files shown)


kaustubh-nair:
@preetmishra This is a great feature to have :tada: 
I've tested this locally a little bit, and it seems to work well.

It would be useful to be able to support the syntax used by the web-app
If I use the example given [here](https://zulipchat.com/help/format-your-message-using-markdown#tables), I run into an `IndexError` exception.

preetmishra:
@kaustubh-nair Thanks for the review. :+1:  I have sorted out the bug that you found.

preetmishra:
Another thing I have noticed that in-table styling doesn't work in ZT yet (for instance, alignment). I was thinking that we might want to support that in v2 but if it doesn't seem like a good idea, I would move ahead to support that in v1 itself.

preetmishra:
@neiljp Thanks for the review. I have made the changes that you suggested for v1.

I have:
* Removed all the horizontal lines (except the top, middle and bottom border).
* Used Unicode characters for drawing the table which makes it cleaner overall.
* Added support for alignments.
* Refactored the code quite a bit (including splitting the previous render function into two - `parse_html_table` and `render_table`).


I have left comments in the code for explaining a few points, let me know if it is too much or if we need more. :)

tests/ui/test_ui_tools.py
```diff
@@ -1397,7 +1397,15 @@ def test_private_message_to_self(self, mocker):
         ('<hr/>', ['[RULER NOT RENDERED]']),
         ('<img>', ['[IMAGE NOT RENDERED]']),
         ('<img/>', ['[IMAGE NOT RENDERED]']),
-        ('<table>stuff</table>', ['[TABLE NOT RENDERED]']),
+        ('<table><thead><tr><th>Firstname</th><th>Lastname</th></tr></thead>'
+         '<tbody><tr><td>John</td><td>Doe</td></tr><tr><td>Mary</td><td>Moe'
+         '</td></tr></tbody></table>', [
+            '┌───────────┬──────────┐\n',
+            ('bold', '│ Firstname │ Lastname │\n'),
```
neiljp: I'm unsure about the best render, but it would seem better to me to have just the cell contents in `bold` or some other head-style, rather than the box parts too? This is particularly evident in the blue and light themes, where I first tested this.

tests/ui/test_ui_tools.py
```diff
@@ -1397,7 +1397,15 @@ def test_private_message_to_self(self, mocker):
...
+            '├───────────┼──────────┤\n',
+            '│ John      │ Doe      │\n',
+            '│ Mary      │ Moe      │\n',
+            '└───────────┴──────────┘']),
```
neiljp: You only have one test here, but support at least different alignments, of which there is no indication here?

zulipterminal/helper.py
```diff
@@ -493,3 +493,95 @@ def notify(title: str, html_text: str) -> None:
...
                              stderr=subprocess.STDOUT)
+
+
+def parse_html_table(table_element: Any) -> Tuple[List[str], List[List[str]]]:
+    """
+    Parses an HTML table to extract cell items and column alignments.
+    """
+    headers = table_element.thead.tr.find_all('th')
+    rows = table_element.tbody.find_all('tr')
+    alignments = []
```
neiljp: This is on a column-basis, right? So maybe `column_alignments`?

zulipterminal/helper.py
```diff
@@ -493,3 +493,95 @@ def notify(title: str, html_text: str) -> None:
...
+    alignments, cells = parse_html_table(table_element)
+
+    # Calculate the width required for each column.
+    column_widths = [
+        len(max(column, key=lambda string: len(string)))
+        for column in zip(*cells)
+    ]
+
+    top_border = unicode_border(u'┌', u'─', u'┬', u'┐', column_widths)
+    middle_border = unicode_border(u'├', u'─', u'┼', u'┤', column_widths)
```
neiljp: I like the way you've extracted the box-drawing elements here :+1: 

You might like to do the same for the vertical bars?

zulipterminal/helper.py
```diff
@@ -493,3 +493,95 @@ def notify(title: str, html_text: str) -> None:
...
+
+    # Construct the table, row-by-row.
+    table = []  # type: List[Union[str, Tuple[str, str]]]
+    for row_num, row in enumerate(cells):
+        styled_row = style_row(alignments, column_widths, row)
+        if row_num == 0:
+            table.append(top_border)
+            # Make the header bold.
+            table.append(('bold', styled_row))
+            table.append(middle_border)
```
neiljp: Would this be clearer as two sections, one generating the header and the other the rest (the latter being the loop)?

zulipterminal/helper.py
```diff
@@ -493,3 +493,95 @@ def notify(title: str, html_text: str) -> None:
...
+        cells[0].append(header.text)
+
+    # Fill up the `cells` with body rows.
+    for index, row in enumerate(rows, start=1):
+        for tdata in row.find_all('td'):
+            cells[index].append(tdata.text)
+    return (alignments, cells)
+
+
+def unicode_border(lcorner: str, line: str, connector: str, rcorner: str,
```
neiljp: Is this more accurately a `unicode_row`? Or `row_with_only_borders`?

This would make it clearer when you interleave it with a `row_with_styled_content`, perhaps.

zulipterminal/helper.py
```diff
@@ -493,3 +493,95 @@ def notify(title: str, html_text: str) -> None:
...
+                   widths: List[int], newline: bool=True) -> str:
+    """
+    Given left corner, line, connecter and right corner unicode character,
+    constructs border for markup table.
+    """
+    border = lcorner
+    for width in widths:
+        # Add +2 to the width to compensate for the extra space added around
+        # each string while styling the row by `style_row`.
```
neiljp: It would be simpler to centralize this padding convention, rather than have it over two methods?

zulipterminal/helper.py
```diff
@@ -493,3 +493,95 @@ def notify(title: str, html_text: str) -> None:
...
+    """
+    headers = table_element.thead.tr.find_all('th')
+    rows = table_element.tbody.find_all('tr')
+    alignments = []
+
+    # The table cells are stored row-wise:
+    # [[row0, row0, row0],
+    # [row1, row1, row1],
+    # [row2, row2, row2]].
+    # Add +1 to count the header row as well.
```
neiljp: This looks like it's part of the comment above? I assume it is not?

Perhaps the cell description should go into the docstring?

zulipterminal/helper.py
preetmishra: Yes, `column_alignments` is a better choice.

zulipterminal/helper.py
preetmishra: No, they are different. I'll move the upper part into the docstring.

zulipterminal/helper.py
preetmishra: Yes, it on a column-basis, `column_alignments` will be perfect.

preetmishra: Yes, it is on a column-basis, column_alignments will be perfect.

preetmishra: Yes, it is on a column-basis, `column_alignments` will be perfect.

zulipterminal/helper.py
preetmishra: I agree, these will be more accurate.

zulipterminal/helper.py
preetmishra: I have added a `vertical_bar` to store the Unicode character to make that cleaner. I believe the way we have it right now is okay, doing it this way will make things complicated.

preetmishra: Thanks. :+1: 

I have added a `vertical_bar` to store the Unicode character to make that cleaner. I believe the way we have it right now is okay, doing it this way will make things complicated.

zulipterminal/helper.py
```diff
@@ -493,3 +493,141 @@ def notify(title: str, html_text: str) -> None:
...
+        connector_char,
+        cell_content,
+        connector_char,
+        rightmost_char,
+    ]
+    """
+    fill = fill_char * fill_width
+
+    # Pad the leftmost box-drawing character.
+    row_strip[0] = row_strip[0] + fill  # type: ignore
```
preetmishra, line 551: `mypy` complains about this concatenation.

preetmishra, line 551: I am not sure why `mypy` complains about these concatenations.

preetmishra:
@neiljp Thanks. I have added extra test cases (to support different alignments), fixed the styling issue and incorporated the other changes that you suggested.

I have restructured the row styling functions to return a list of strings and tuples (style, string). This not only helps in centralizing the padding routine but also helps in setting styles individually. In addition to these, this also makes the work required for adding inline-table-styles in v2 rather straightforward. 

Once we reach a consensus about the implementation, I would add the tests for the helper functions.

preetmishra:
@neiljp Thanks. I have added extra test cases (to support different alignments), fixed the styling issue and incorporated the other changes that you suggested.

I have restructured the row styling functions to return a list of strings and tuples (style, string). This not only helps in centralizing the padding routine but also helps in setting styles individually. In addition to these, this also makes the work required for adding inline-table-styles in v2 rather straightforward. 

I would add the tests for the helper functions once we reach a consensus about the implementation.

preetmishra:
Update: Extended the docstring for `pad_row_strip`.

zulipterminal/helper.py
```diff
@@ -493,3 +493,142 @@ def notify(title: str, html_text: str) -> None:
...
+        for tdata in row.find_all('td'):
+            cells[index].append(tdata.text)
+    return (column_alignments, cells)
+
+
+TABLE_ROW = List[Union[str, Tuple[Optional[str], str]]]
+
+
+def pad_row_strip(row_strip: TABLE_ROW, fill_char: str=' ',
+                  fill_width: int=1) -> TABLE_ROW:
```
neiljp, line 530: I like this new method :+1:

zulipterminal/helper.py
```diff
@@ -493,3 +493,142 @@ def notify(title: str, html_text: str) -> None:
...
+                                         column_widths,
+                                         row_style='table_head'))
+    table.extend(middle_border)
+
+    # Add the body rows to the table followed by the bottom-most border in the
+    # end.
+    for row in cells:
+        table.extend(row_with_styled_content(row, column_alignments,
+                                             column_widths))
+    table.extend(bottom_border)
```
neiljp, line 632: This layout for constructing the table content looks cleaner to me :+1:

sumanthvrao:
@neiljp This is the TinglingGit Output for All the PR's which clash with the file ui_tools/boxes.py
and their diff
```
PR#397(+2-0)
PR#530(+4-0)
PR#278(+16-8)
PR#539(+1-1)
PR#540(+63-22)
PR#542(+26-18)
PR#287(+51-1)
PR#424(+8-1)
PR#572(+38-38)
PR#573(+11-10)
PR#335(+18-10)
PR#354(+2-0)
PR#550(+4-3)
PR#359(+12-1)
PR#371(+21-15)
PR#501(+6-2)
```

zulipterminal/helper.py
```diff
@@ -493,3 +493,142 @@ def notify(title: str, html_text: str) -> None:
...
+        for column in zip(*cells)
+    ]
+
+    top_border = row_with_only_border(u'┌', u'─', u'┬', u'┐', column_widths)
+    middle_border = row_with_only_border(u'├', u'─', u'┼', u'┤', column_widths)
+    bottom_border = row_with_only_border(u'└', u'─', u'┴', u'┘', column_widths,
+                                         newline=False)
+
+    # Construct the table, row-by-row.
+    table = []  # type: List[Union[str, Tuple[Optional[str], str]]]
```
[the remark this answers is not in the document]
preetmishra: Yes, this seems reasonable. I think `StyleTableData` will be good.

tests/ui/test_ui_tools.py
```diff
@@ -1397,7 +1397,41 @@ def test_private_message_to_self(self, mocker):
...
+         '<td align="right">1</td></tr><tr><td align="center">Mary</td>'
+         '<td align="right">100</td></tr></tbody></table>', [
+            '┌─', '──────', '─┬─', '───', '─┐\n',
+            '│ ', ('table_head', ' Name '), ' │ ', ('table_head', ' Id'),
+            ' │\n',
+            '├─', '──────', '─┼─', '───', '─┤\n',
+            '│ ', (None, 'Robert'), ' │ ', (None,  '  1'), ' │\n',
+            '│ ', (None, ' Mary '), ' │ ', (None, '100'), ' │\n',
+            '└─', '──────', '─┴─', '───', '─┘',
+         ]),
```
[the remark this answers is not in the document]
preetmishra: The tables will have at least two rows (the header and an empty body row) if nothing is defined in the body.

Though it is possible for a table to have one column. I'll add a few more test cases.

zulipterminal/helper.py
preetmishra: I tried using `type: str` for `fill` but that didn't seem to solve the issue. I'll add comments regarding the ignores.

preetmishra:
@neiljp Thanks. I have added a comment for the first commit (it lets `row_with_styled_content` add `None` display attribute) and added a few more test cases for the table.

I have replaced `TABLE_ROW` with `StyledTableData` as it seemed more reasonable. On a related note, `urwid_Size` type alias uses underscore with CamelCase, was this intentional or should we refactor?

The `vertical_bar` is now a function argument for `row_with_styled_data`. This would make replacing box-drawing characters easy, if we ever feel like to.

If it is not much of an overhead, we can move the MessageBox in a separate file along with the new changes and perhaps separate tests for the table from `soup2mark` tests.

preetmishra:
@neiljp Thanks. I have:

* Added a comment for the first commit (it lets `row_with_styled_content` add `None` display attribute).
* Added a few more test cases for the table.
* Replaced `TABLE_ROW` with `StyledTableData` as it seemed more reasonable. On a related note, `urwid_Size` type alias uses underscore with CamelCase, was this intentional or should we refactor?
* The `vertical_bar` is now a function argument for `row_with_styled_data`. This would make replacing box-drawing characters easy if we ever feel like to.
* Added comments for the `type: ignore` issues. I tried using a `type: str` for `fill` but it didn't seem to work.

If it is not much of an overhead, we can move the MessageBox in a separate file along with the new changes and perhaps separate the table test from `soup2mark` tests (thoughts?).

preetmishra:
@neiljp Thanks. I understand the idea for `urwid_Size` now.

I was able to fix all the `mypy` issues using `cast`, which assists `mypy` with the type, and what you suggested in https://github.com/zulip/zulip-terminal/pull/550#discussion_r403626995 :+1:.

neiljp:
@preetmishra As I mentioned on the stream, this is now merged with a few minor tweaks :tada: This is really great :) I'm sure you can find it in the list of commits by now!

preetmishra:
@neiljp Thanks so much, it was fun working on this. :)

To address the things that we would like to see in the future iterations, I'll file an issue.

Issue #544, opened 2020-03-17
Inventory styles and ensure they are appropriately used & named


neiljp:
I'm pretty sure that we currently re-use some style names for elements which are un-related (eg. unread counts). This is mostly harmless, though confusing to read in the code, and now that we use more styles it would be useful to have a clearer idea of which are used and for what purposes.

The end result might involve having descriptive comments next to each entry in `required_styles` (if the names are insufficiently clear, in particular), and possibly changing the names of some styles, perhaps with prefixes.

Looking through the UI code will help find these instances, so this would be good to learn the way the UI works. To ensure that these cases are found, we *may* be able to use tests, but this may require manual checks - though automating this and having a tool for future use would be beneficial.

zulipbot:
Hello @neiljp, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

PR #553, opened 2020-03-21
light theme: Improve selections, quotes, stars & times.

neiljp:
I normally only use the default theme, but in brighter light conditions have been trying the light theme, leading to these suggested changes so far:
* Use a consistent black-text/green-background for selections
* Change the color background for blockquote to be legible
* Adjust background of date/time and star-status to match light theme


kaustubh-nair:
@neiljp Could you add a screenshot with this too? So we can be sure we're looking at the same thing.
Here's what mine looks like 
DeepinScreenshot_x-terminal-emulator_20200323115816

I'm wondering if it would look better if the light colored text ( unread counts, user ids) is changed to something darker, so there's no need for a grey background for them?

neiljp:
@kaustubh-nair My coloring is a little darker, but it depends on the default terminal colors I believe.

screenshot-20200323-10:27:50

As I said, in addition to green in the middle column, it's green in the side columns too (something which we could do for the default theme too, maybe others).

The time and star colors just match the unselected background now, rather than being darker, and something similar for the blockquote.

I hadn't explored the usernames, so just pushed an update with a different green name; better?

Regarding the counts and the user list, this is a limitation of the current styling rules (see #544), so I chose to leave this for now.

sumanthvrao:
@neiljp LGTM!

PR #554, opened 2020-03-21
boxes: Merge date/time and optional prev-year fields to improve styling.

neiljp:
While not visible in the default theme, in the light theme the two parts of the
date are separated by a gap, due to how the optional year feature was
added as an extra column. By merging the columns and explicitly adding
spaces this now displays more cleanly and appears as one block of text.

Tests amended.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1629,9 +1629,9 @@ class TestMessageBox:
     @pytest.mark.parametrize('starred_msg', ['this', 'last', 'neither'],
                              ids=['this_starred', 'last_starred', 'no_stars'])
     @pytest.mark.parametrize('expected_header, to_vary_in_last_message', [
-        (['alice', ' ', ' ', 'DAYDATETIME'], {'sender_full_name': 'bob'}),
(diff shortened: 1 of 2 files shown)


kaustubh-nair:
This might be unrelated to the commit, but how do you feel about adding some space to surround the text? The text feels a little cut-off here:
DeepinScreenshot_select-area_20200323124924

Adding a blank space would make it look better maybe like we do it for topics:
DeepinScreenshot_select-area_20200323124953

neiljp:
@kaustubh-nair Thanks for the review :+1:

Regarding the blank space, this could be something to look into more generally and separately. There's no harm in mentioning these things here, but it will reach more people if discussed on the stream :) I do think the spacing may look better, but we need to be careful as while these seem like small changes and only extra characters here and there, it also takes up marginally more space, which can make a difference on narrow screens.

sumanthvrao:
@neiljp The changes look nice!

Also, nice observation skills for spotting the extra gap ;)!

neiljp:
@sumanthvrao It's a little minor, but it does look a bit messy - I'll merge this now :tada:

PR #555, opened 2020-03-23
run: themes: transparent background is now optional (#513)

### LINKED ISSUE #513: Re-enable configurable background transparency support
In #378 we allowed the terminal to set the background color, to support transparent terminals.

However, due to the range of terminals and customization available, it is likely better if this is an option rather than assumed by default. For example, some users report that with the default scheme then the background color looks strange - this is likely the default background color from the terminal, and we likely want 'black' instead, unless the user specifically requests otherwise (for transparency)

We would want to enable setting this option in the config file and on the command line, as with other options.

**UPDATE**: Transparency support was lost in a recent theme data restructuring. This feature therefore needs adding back, through a configurable option (command-line, config) which likely should default to off.

mregdos:
Default background is not transparent anymore.
Add --transparent flag and analogous option in config file which
may be used to set background's color based on terminal.

Fixes #513.

### DIFF
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -61,7 +61,8 @@ def test_main_help(capsys, options):
         '--config-file CONFIG_FILE, -c CONFIG_FILE',
         '--autohide',
         '--no-autohide',
-        '-v, --version'
+        '-v, --version',
+        '--transparent'
     }
     optional_argument_lines = {line[2:] for line in lines
                                if len(line) > 2 and line[2] == '-'}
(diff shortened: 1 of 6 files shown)


neiljp:
@mregdos Thanks for working on this; we've restructured themes since this, so we'll need to take a different approach for the theme-specific side of things at least. I plan to close this and refer others to this as a reference in #513, unless you want to take this up again?

neiljp:
As per previous comment, closing in favor of #1415.

PR #556, opened 2020-03-24
refactor: helper: Separate setting counts for model & UI.

sumanthvrao:
I got this Idea after looking at #548 since our `set_count` function is particularly complex (not so well structured code). 

The function is now separated into out into set_model and set_UI. Though these two are not completely uncoupled, this should be a good start along those lines.

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -93,19 +93,19 @@ def _set_count_in_model(id_list: List[int], new_count: int,
         if it's passed in, but is not tied to it).
     """
     for id in id_list:
-        msg = messages[id]
+        message = messages[id]
 
-        if msg['type'] == 'stream':
-            key = (messages[id]['stream_id'], msg['subject'])
+        if message['type'] == 'stream':
+            key = (message['stream_id'], message['subject'])
             unreads = unread_counts['unread_topics']
         # self-pm has only one display_recipient
         # 1-1 pms have 2 display_recipient
-        elif len(msg['display_recipient']) <= 2:
-            key = messages[id]['sender_id']
+        elif len(message['display_recipient']) <= 2:
+            key = message['sender_id']
             unreads = unread_counts['unread_pms']  # type: ignore
         else:  # If it's a group pm
             key = frozenset(  # type: ignore
-                recipient['id'] for recipient in msg['display_recipient']
+                recipient['id'] for recipient in message['display_recipient']
             )
             unreads = unread_counts['unread_huddles']  # type: ignore
 
@@ -137,17 +137,18 @@ def _set_count_in_view(id_list: List[int], controller: Any, new_count: int,
     all_msg = controller.view.home_button
     all_pm = controller.view.pm_button
     for id in id_list:
-        user_id = messages[id]['sender_id']
+        message = messages[id]
+        user_id = message['sender_id']
 
         # If we sent this message, don't increase the count
         if user_id == controller.model.user_id:
             continue
 
-        msg_type = messages[id]['type']
+        msg_type = message['type']
         add_to_counts = True
         if msg_type == 'stream':
-            stream_id = messages[id]['stream_id']
-            msg_topic = messages[id]['subject']
+            stream_id = message['stream_id']
+            msg_topic = message['subject']
             if controller.model.is_muted_stream(stream_id):
                 add_to_counts = False  # if muted, don't add to eg. all_msg
             else:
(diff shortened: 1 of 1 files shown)


kaustubh-nair:
This is a good one, `set_counts` is a little cleaner now :+1:

zulipterminal/helper.py
```diff
@@ -84,14 +84,13 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
-    messages = controller.model.index['messages']
-    unread_counts = controller.model.unread_counts  # type: UnreadCounts
-
+def _set_count_in_model(id_list: List[int], controller: Any, new_count: int,
+                        messages: Dict[int, Message],
+                        unread_counts: UnreadCounts) -> None:
+    """
+        This function contains the logic for setting the count in the model.
+    """
+    print(messages, flush=True)
```
neiljp: Nice debug :)

zulipterminal/helper.py
```diff
@@ -84,14 +84,13 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
 
-def set_count(id_list: List[int], controller: Any, new_count: int) -> None:
-    # This method applies new_count for 'new message' (1) or 'read' (-1)
-    # (we could ensure this in a different way by a different type)
-    assert new_count == 1 or new_count == -1
-
-    messages = controller.model.index['messages']
-    unread_counts = controller.model.unread_counts  # type: UnreadCounts
-
+def _set_count_in_model(id_list: List[int], controller: Any, new_count: int,
```
neiljp: We don't use `controller` here?

zulipterminal/helper.py
```diff
@@ -84,14 +84,13 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+                        messages: Dict[int, Message],
+                        unread_counts: UnreadCounts) -> None:
+    """
+        This function contains the logic for setting the count in the model.
```
neiljp: It is probably worth mentioning that this doesn't explicitly set the count in the model, but rather just updates `unread_counts` (which can update the model if it's passed in, but is not tied to it). Mutating values passed in is something we should comment on in any case.

zulipterminal/helper.py
```diff
@@ -84,14 +84,13 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+    """
+    print(messages, flush=True)
     for id in id_list:
         msg = messages[id]
```
neiljp: Seeing this now, all we need is the messages that correspond to `id_list`, so we could simplify these function parameters further, potentially?

The same applies to the other function too.

zulipterminal/helper.py
sumanthvrao: Oof. A classic example of what happens when you rush things 😱

zulipterminal/helper.py
```diff
@@ -187,14 +184,14 @@ def set_count(id_list: List[int], controller: Any, new_count: int) -> None:
     assert new_count == 1 or new_count == -1
     messages = controller.model.index['messages']
     unread_counts = controller.model.unread_counts  # type: UnreadCounts
-
-    _set_count_in_model(id_list, new_count, messages, unread_counts)
+    message_list = [messages[id] for id in messages]
```
neiljp: `...for id in id_list` ?

Could `message_list` be better named?

zulipterminal/helper.py
```diff
@@ -84,28 +84,25 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
     """
         This function doesn't explicitly set counts in model,
         but updates `unread_counts` (which can update the model
         if it's passed in, but is not tied to it).
     """
-    for id in id_list:
-        msg = messages[id]
-
-        if msg['type'] == 'stream':
-            key = (messages[id]['stream_id'], msg['subject'])
```
neiljp: This commit both changes the variable name to `message` and simplifies the function call, so maybe we could be clearer by unifying `messages[id]` and `msg` to `message`, then making the other change in another commit?

sumanthvrao: True. I wanted to change the `msg` name for a while and though this would be a good time :). Will keep the 2 commits separate!

zulipterminal/helper.py
sumanthvrao: Thanks for catching this!

neiljp:
@sumanthvrao Thanks for this work :tada: This refactor looks like a good split which was just waiting to be done :)

To finalize this work we need to really split up the tasks cleanly between these two new methods, so it's clearer. As I mentioned, we could then rename them, but doing that before the tasks are split would result in inappropriate names.

PR #562, opened 2020-03-26
Themes: Separate unread count and idle styles

neiljp:
This resolves one aspect of #544, by separating out a style for unread counts from that used for idle.

This in turn allows an improvement in the styling of unread count text, which has been a blocker to improving the user-list styling in the light and blue themes.

Given the small code changes, for review purposes I'm more interested in how this looks compared to master for people on different platforms.

I have a followup to improve the blue theme further, both along the same lines as #553 and now including lightening the user list (if we want this).

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -165,7 +165,7 @@ THEMES = {
         ('footer',       'white',           'dark red',   'bold'),
         ('starred',      'light red, bold', 'white'),
         ('category',     'dark gray, bold', 'light gray'),
-        ('unread_count', 'yellow',          'dark gray'),
+        ('unread_count', 'dark blue, bold', 'white'),
     ],
     'blue': [
         (None,           'black',           'light blue'),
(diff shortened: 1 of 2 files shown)


neiljp:
@preetmishra You looked at blue too, right?

preetmishra:
@neiljp Yes, I checked both the light and blue theme.

kaustubh-nair:
This looks a lot better :+1:
I would expect the background to be same as highlighted bar (green) when the screen is highlighted, but I'm not sure if it's relevent to this PR

DeepinScreenshot_x-terminal-emulator_20200327153205
DeepinScreenshot_x-terminal-emulator_20200327153250

preetmishra:
Adding screenshots for reference.
The screenshots were taken on Ubuntu's default terminal with the default theme.

#### Light
Screenshot from 2020-03-27 17-34-55

#### Blue
Screenshot from 2020-03-27 17-35-57

kaustubh-nair:
Which background color is being used for the light theme? Is it light gray?

neiljp:
@kaustubh-nair Are you referring to the difference between the selected-message and selections in the side bars being different? Or something else?

kaustubh-nair:
When the stream name is highlighted green, the background color of the unread count (even the #) stays white. It would be cool if that background changes to green too. But if that happens the text color of the unread count would have to change too, since it probably won't be visible.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

neiljp:
@kaustubh-nair I agree with what you mean regarding this, but that seems a separate improvement that we could look into.

I'll merge this now.

PR #565, opened 2020-03-28
Improve consistency of selection styles

neiljp:
I've noticed this before, but my feeling is that it's confusing when moving between the side panels and the message list, that the selection is different.

I explored matching them in the light theme in 2c211bb, and I think that's an improvement.

This PR updates the blue and default themes, not yet touching the gruvbox theme. I also adjusted the unread style, as that's important to show a clear distinction between unread/active/read.

Any thoughts on this approach in general, and also if you think it's good, of how to approach it in gruvbox, would be great to hear.

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -54,7 +54,7 @@ THEMES = {
     'default': [
         (None,           'white',           ''),
         ('selected',     'white',           'dark blue'),
-        ('msg_selected', 'light green',     ''),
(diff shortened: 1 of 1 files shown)


neiljp:
I just pushed a version with `black` on `white` for gruvbox too.

The only thought I have is whether there is good enough contrast with the unread style in gruvbox now?

neiljp:
I've just pushed a version with no changes to the unread styling, which should work for now; we can discuss unread styling further on ZT next while still having the benefits of the selection consistency :+1:

PR #568, opened 2020-03-29
Minor theme improvements

neiljp:
This is the last work I have on themes so far:
* Blue theme: adjust background of date/time and star-status to match blue theme.

These are not hugely important, but are a little extra polish for the blue theme.

Any other minor theme polish suggestions are welcome here, so that we have a good set of working themes in the repo to move on with.


neiljp:
@preetmishra Thanks for that suggestion - I've added that :+1:

neiljp:
@preetmishra I had made some changes last night including putting the unread-contrast changes into this PR, but hadn't pushed, so good catch on the text :+1:

preetmishra:
@neiljp Oh, that explains it. Thanks. :)

neiljp:
I've pushed a few more changes based on feedback on #**zulip-terminal**.

I think this is good, but as I mentioned in the stream, I'm tempted to keep the links with a background for now, to highlight them; thoughts?

@kaustubh-nair Thanks for the feedback for this on your system.

@preetmishra Does this still look good on your system?

neiljp:
Thanks to those on #**zulip-termainal** for the extra testing - I'm merging this now as it sets a good baseline which appears to work well.

We can work on simplifying the palettes (maybe extract monochrome palettes using #563 or similar) and adding further color support after other discussion.

PR #569, opened 2020-03-30
refactor: tests: Eliminate hard-coded size variables from keypress tests

preetmishra:
The PR has been updated to use a factory fixture instead. This would help in compacting our tests and also correcting our keypress sizes (see asserts in the code).

Every `Use/Migrate ...` commit adds extra assert statements for sanity-check and every `Remove ...` commit removes those statements along with `size`. These are ordered alternatively, so we could either squash them in pair or let them be as-is.

The prior discussion for this can be found here: https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Compacting.20tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -875,3 +875,20 @@ def navigation_key_expected_key_pair(request):
     The expected key is the one which is passed to the super `keypress` calls.
     """
     return request.param
+
+
+@pytest.fixture
+def widget_size():
+    """
+    Returns widget size for any widget.
+    """
+    def _widget_size(widget):
+        widget_type, *_ = widget.sizing()
+        if widget_type == 'box':
+            return (200, 20)
+        elif widget_type == 'flow':
+            return (20, )
+        else:
+            None
+
+    return _widget_size
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -220,8 +220,6 @@ class TestView:
         view.body = mocker.Mock()
         view.user_search = mocker.Mock()
         size = widget_size(view)
-        assert size != (20,), 'Incorrect size used for a Box widget'
-        assert size == (200, 20)
 
         super_keypress = mocker.patch("zulipterminal.ui.urwid.WidgetWrap"
                                       ".keypress")
@@ -238,8 +236,6 @@ class TestView:
         view.body.focus_col = None
         view.controller.is_in_editor_mode = lambda: False
         size = widget_size(view)
-        assert size != (20,), 'Incorrect size used for a Box widget'
-        assert size == (200, 20)
         view.model.controller.show_all_mentions = mocker.Mock()
 
         view.keypress(size, key)
@@ -260,8 +256,6 @@ class TestView:
         view.left_panel = mocker.Mock()
         view.right_panel = mocker.Mock()
         size = widget_size(view)
-        assert size != (20,), 'Incorrect size used for a Box widget'
-        assert size == (200, 20)
 
         super_view = mocker.patch("zulipterminal.ui.urwid.WidgetWrap.keypress")
         view.controller.is_in_editor_mode = lambda: False
@@ -291,8 +285,6 @@ class TestView:
         view.left_panel.is_in_topic_view = False
         view.right_panel = mocker.Mock()
         size = widget_size(view)
-        assert size != (20,), 'Incorrect size used for a Box widget'
-        assert size == (200, 20)
 
         super_view = mocker.patch("zulipterminal.ui.urwid.WidgetWrap.keypress")
         view.controller.is_in_editor_mode = lambda: False
@@ -318,8 +310,6 @@ class TestView:
 
         view.controller.is_in_editor_mode = lambda: True
         size = widget_size(view)
-        assert size != (130, 28), 'Different box size value'
-        assert size == (200, 20)
 
         view.keypress(size, key)
 
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -16,6 +16,9 @@ class TestView:
     @pytest.fixture
     def view(self, mocker):
         main_window = mocker.patch('zulipterminal.ui.View.main_window')
+        # View is an urwid.Frame instance, a Box widget.
+        mocker.patch('zulipterminal.ui.View.sizing',
+                     return_value=frozenset({'box'}))
         return View(self.controller)
 
     def test_init(self, mocker):
@@ -210,13 +213,15 @@ class TestView:
         else:
             view.body.options.assert_not_called()
 
-    def test_keypress_normal_mode_navigation(self, view, mocker,
+    def test_keypress_normal_mode_navigation(self, view, mocker, widget_size,
                                              navigation_key_expected_key_pair):
         key, expected_key = navigation_key_expected_key_pair
         view.users_view = mocker.Mock()
         view.body = mocker.Mock()
         view.user_search = mocker.Mock()
-        size = (20,)
+        size = widget_size(view)
+        assert size != (20,), 'Incorrect size used for a Box widget'
+        assert size == (200, 20)
 
         super_keypress = mocker.patch("zulipterminal.ui.urwid.WidgetWrap"
                                       ".keypress")
@@ -228,11 +233,13 @@ class TestView:
         super_keypress.assert_called_once_with(size, expected_key)
 
     @pytest.mark.parametrize('key', keys_for_command('ALL_MENTIONS'))
-    def test_keypress_ALL_MENTIONS(self, view, mocker, key):
+    def test_keypress_ALL_MENTIONS(self, view, mocker, key, widget_size):
         view.body = mocker.Mock()
         view.body.focus_col = None
         view.controller.is_in_editor_mode = lambda: False
-        size = (20,)
+        size = widget_size(view)
+        assert size != (20,), 'Incorrect size used for a Box widget'
+        assert size == (200, 20)
         view.model.controller.show_all_mentions = mocker.Mock()
 
         view.keypress(size, key)
@@ -243,7 +250,8 @@ class TestView:
     @pytest.mark.parametrize('key', keys_for_command('SEARCH_PEOPLE'))
     @pytest.mark.parametrize('autohide', [True, False], ids=[
         'autohide', 'no_autohide'])
-    def test_keypress_autohide_users(self, view, mocker, autohide, key):
+    def test_keypress_autohide_users(self, view, mocker, autohide, key,
+                                     widget_size):
         view.users_view = mocker.Mock()
         view.body = mocker.Mock()
         view.controller.autohide = autohide
@@ -251,7 +259,9 @@ class TestView:
         view.user_search = mocker.Mock()
         view.left_panel = mocker.Mock()
         view.right_panel = mocker.Mock()
-        size = (20,)
+        size = widget_size(view)
+        assert size != (20,), 'Incorrect size used for a Box widget'
+        assert size == (200, 20)
 
         super_view = mocker.patch("zulipterminal.ui.urwid.WidgetWrap.keypress")
         view.controller.is_in_editor_mode = lambda: False
@@ -269,7 +279,8 @@ class TestView:
     @pytest.mark.parametrize('key', keys_for_command('SEARCH_STREAMS'))
     @pytest.mark.parametrize('autohide', [True, False], ids=[
         'autohide', 'no_autohide'])
-    def test_keypress_autohide_streams(self, view, mocker, autohide, key):
+    def test_keypress_autohide_streams(self, view, mocker, autohide, key,
+                                       widget_size):
         view.stream_w = mocker.Mock()
         view.left_col_w = mocker.Mock()
         view.stream_w.stream_search_box = mocker.Mock()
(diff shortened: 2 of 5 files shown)


neiljp:
@preetmishra I know we discussed this on the stream, but one solution may be to use the `.sizing()` function.

tests/conftest.py
```diff
@@ -781,3 +781,14 @@ def navigation_key_expected_key_pair(request):
...
+
+@pytest.fixture
+def widget_size(request):
+    """
+    Returns widget size for any widget.
+    """
+    widget = request.getfixturevalue(request.param)
+    widget_type, *_ = widget.sizing()
+    return (20, ) if widget_type == 'flow' else \
+           (200, 20) if widget_type == 'box' else None
```
neiljp, line 794: I've not seen the `indirect` approach; it looks a little unclear to someone who hasn't seen it before, though I can guess what is happening. Could this be translated to returning a function to be applied to a widget? That might be simpler? Potentially we could try caching the values, if that had a speed factor, though that might reduce test isolation.

tests/ui/test_ui_tools.py
```diff
@@ -1106,6 +1106,7 @@ def pop_up_view(self, mocker):
         self.super_keypress = mocker.patch(VIEWS + '.urwid.ListBox.keypress')
         self.pop_up_view = PopUpView(self.controller, self.widgets,
                                      self.command)
+        return self.pop_up_view
```
neiljp, line 1134: Are all these changes to support the `.sizing()` calls in the next commit? If so, we might want to structure it differently, or at least identify this in the commit text.

preetmishra: Yes. I'll amend the commit text.

tests/conftest.py
preetmishra: On second thought, this does look a bit unclear. I'll see if this can be translated into a simpler form.

preetmishra: @neiljp I wonder if using a function would be less complicated?

```python
def widget_size(widget):
    """
    Returns widget size for any widget.
    """
    sizes = {
        'box': (200, 20),
        'flow': (20, ),
    }
    widget_type, *_ = widget.sizing()
    return sizes.get(widget_type, None)
```

In tests,  we could do:

```python
view.keypress(widget_size(widget), key)
```

However, this would require an import (unlike fixtures).

neiljp: We could resolve this by simply using a factory fixture:
https://docs.pytest.org/en/stable/fixture.html#factories-as-fixtures

preetmishra:
@neiljp Thanks for the pointer. :+1:

I have updated the PR and its [description](https://github.com/zulip/zulip-terminal/pull/569#issue-395715437) to list the changes.

neiljp:
@preetmishra I think you have some rebase issue - 24 commits seems unlikely :)

preetmishra:
@neiljp Yes, I have split the commits hoping it would make it easier to review.

neiljp (commented):
@preetmishra Other than the first commit being a little hybrid (as if two commits) and not following the same style, the rest of  the commit structure is fairly clear. The net result is good and mergeable, however there are a lot of commits with very similar changes in; below are some thoughts on how we might achieve a similar migration but with some extra clarity - I'm interested to hear what you think :+1:

I do like the idea of the "switch over, with asserts" -> tests pass -> "remove asserts and old code" approach, though there's a lot of duplicate code that's just added and then removed straight away in the next commit. To avoid some of the duplication I think we can simply treat the `size` variables as dynamically assigned now and add temporary asserts of what they are. We can indicate where the sizes have changed in the commit text, instead of adding and then removing the assert in adjacent commits. The resulting `assert size == ...` lines should then be sufficiently identifiable to remove in a single commit. What do you think?

In terms of grouping the commits by different classes, this could be useful to see similar sizes being maintained or corrected. My other thought is grouping per those changes (no-change vs wrong-size vs multiple-valid-size-types), as that makes it clearer where the test code has meaningfully changed or not.

preetmishra:
@neiljp Thanks for the review. As per our discussion in PM, this has been updated to separate the fixture-introduction commit, group changes by three/four types (the last type is a little different from the wrong-size change but could be squashed it onto the incorrect size commit) and remove asserts altogether independently.

tests/conftest.py
```diff
@@ -875,3 +875,20 @@ def navigation_key_expected_key_pair(request):
     The expected key is the one which is passed to the super `keypress` calls.
     """
     return request.param
+
+
+@pytest.fixture
+def widget_size():
```
neiljp, line 881: Adding this isn't a refactor.

tests/ui/test_ui.py
```diff
@@ -16,6 +16,9 @@ def mock_external_classes(self, mocker):
     @pytest.fixture
     def view(self, mocker):
         main_window = mocker.patch('zulipterminal.ui.View.main_window')
+        # View is an urwid.Frame instance, a Box widget.
+        mocker.patch.object(main_window, 'sizing',
+                            return_value=frozenset(['box']))
```
neiljp, line 21: This appears equivalent to adding a parameter to the first patch, settings sizing equal to a lambda?
Normally it seems cleaner to combine the mock setup in one line if possible.

tests/ui/test_ui.py
```diff
@@ -210,13 +213,15 @@ def test_show_right_panel(self, mocker, view,
...
 
-    def test_keypress_normal_mode_navigation(self, view, mocker,
+    def test_keypress_normal_mode_navigation(self, view, mocker, widget_size,
                                              navigation_key_expected_key_pair):
         key, expected_key = navigation_key_expected_key_pair
         view.users_view = mocker.Mock()
         view.body = mocker.Mock()
         view.user_search = mocker.Mock()
-        size = (20,)
+        size = widget_size(view.main_window)
```
neiljp, line 222: Why does this size depend on `main_window` when we're calling the test on `view`?
(similarly elsewhere)

tests/ui/test_ui.py
```diff
@@ -16,6 +16,9 @@ def mock_external_classes(self, mocker):
     @pytest.fixture
```
neiljp, line 16: I'd consider the fixing of the incorrect keypress sizes to be a bugfix rather than (pure) refactor?

tests/ui/test_ui.py
```diff
@@ -309,17 +309,19 @@ def test_keypress_autohide_streams(self, view, mocker, autohide, key,
...
+    def test_keypress_edit_mode(self, view, mocker, key, widget_size):
         view.users_view = mocker.Mock()
         view.body = mocker.Mock()
         view.user_search = mocker.Mock()
 
         super_view = mocker.patch("zulipterminal.ui.urwid.WidgetWrap.keypress")
 
         view.controller.is_in_editor_mode = lambda: True
-        size = (130, 28)
+        size = widget_size(view.main_window)
```
neiljp, line 312: This commit references `view.main_window` but tests `view`, too.

tests/ui/test_ui_tools.py
```diff
@@ -1513,10 +1515,12 @@ def edit_mode_view(self, mocker):
         (2, 'change_all'),
     ])
     @pytest.mark.parametrize('key', keys_for_command('ENTER'))
-    def test_select_edit_mode(self, mocker, edit_mode_view,
+    def test_select_edit_mode(self, mocker, edit_mode_view, widget_size,
                               index_in_widgets, mode, key):
         radio_button = edit_mode_view.widgets[index_in_widgets]
-        size = (20, 20)
+        size = widget_size(edit_mode_view)
```
neiljp, line 1457: `edit_mode_view` or `radio_button`?

tests/ui/test_ui_tools.py
```diff
@@ -1556,22 +1560,26 @@ def test_keypress_navigation(self, mocker, widget_size,
         super_keypress.assert_called_once_with(size, expected_key)
 
     @pytest.mark.parametrize('key', (*keys_for_command('ENTER'), ' '))
-    def test_checkbox_toggle_mute_stream(self, mocker, key):
+    def test_checkbox_toggle_mute_stream(self, mocker, key, widget_size):
         mute_checkbox = self.stream_info_view.widgets[3]
         toggle_mute_status = self.controller.model.toggle_stream_muted_status
         stream_id = self.stream_info_view.stream_id
-        size = (20, 20)
+        size = widget_size(self.stream_info_view)
```
neiljp, line 1499: `mute_checkbox`?

tests/ui/test_ui_tools.py
```diff
@@ -1556,22 +1560,26 @@ def test_keypress_navigation(self, mocker, widget_size,
...
         toggle_mute_status.assert_called_once_with(stream_id)
 
     @pytest.mark.parametrize('key', (*keys_for_command('ENTER'), ' '))
-    def test_checkbox_toggle_pin_stream(self, mocker, key):
+    def test_checkbox_toggle_pin_stream(self, mocker, key, widget_size):
         pin_checkbox = self.stream_info_view.widgets[4]
         toggle_pin_status = self.controller.model.toggle_stream_pinned_status
         stream_id = self.stream_info_view.stream_id
-        size = (20, 20)
+        size = widget_size(self.stream_info_view)
```
neiljp, line 1510: `pin_checkbox`?

neiljp (commented):
@preetmishra Thanks for your patience with this :+1:

I've made more comments inline than I expected - I think this is simply because the different structure makes it much easier to read and review!

Most points are minor in terms of commit text or style, but there are a number of cases in the later commits where you seem to use a different widget to calculate the size than would at least be naively assumed - ie. `size` based on `A` but then testing `B`. That may be related to how the `size` gets passed around, but it's certainly confusing compared to how you're applying the fixture in the other cases, so I'd like to understand where this comes from, ie. why you're using a different widget - and whether we should be doing that elsewhere too.

tests/ui/test_ui.py
preetmishra: The `View` class is a `WidgetWrap` subclass, which means its keypress size depends on the main widget passed in its `init` call.

As the `widget_size` uses the widget instance to deduce the widget type and ultimately the `size` tuple, the actual widget, which happens to be the one that we return from the `main_window`, is being passed.

```python3
class View(urwid.WidgetWrap):
        ...
        super().__init__(self.main_window())
```
```python3
    def main_window(self) -> Any:
        ...
        w = urwid.Frame(self.body, title_bar, focus_part='body',
                        footer=self.footer_view())
        return w
```

tests/ui/test_ui.py
preetmishra: Yeah, that would be appropriate. Thanks!

tests/ui/test_ui.py
preetmishra: (Posted an explanation in the other comment.)

tests/ui/test_ui_tools.py
preetmishra: This should have been `radio_button`. Thanks! Fixed for the other comments as well.

preetmishra:
@neiljp Thanks for the review and pointing out the inaccuracies. I have incorporated the suggested changes and left an in-line comment regarding the latter point.

tests/ui/test_ui.py
neiljp: From what I can tell, a `WidgetWrap` subclass acquires at least the minimal widget interface of the passed-in widget by default, including sizing? I assume we patch `main_window` to avoid constructing the UI; can we simply then patch `View.sizing` to provide what `super().__init__` would do for us, or something like that? We should add comments as to why we're doing that, similar to what you have now, but that keeps the tests themselves clearer?

It's still not ideal in that we're explicitly fixing the sizing of the widget in a mock, but I'm not sure we have an alternative given the current code layout? If we later changed the implementation however, the sizing should still come from the view - eg. if we were to inherit from `Frame` instead of `WidgetWrap` then the sizing would come from the view and the mock could be removed - but the tests would work the same.

```python3
>>> class X(urwid.WidgetWrap):
...     def __init__(self):
...             super().__init__(urwid.Text("hi"))
... 
>>> x = X()
>>> x
<X flow widget>
>>> dir(x)
['_DelegateToWidgetMixin__super', '_WidgetWrap__super', '_Widget__super', '_X__super', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_command_map', '_emit', '_invalidate', '_not_a_container', '_raise_old_name_error', '_repr_attrs', '_repr_words', '_selectable', '_set_w', '_sizing', '_w', '_wrapped_widget', 'base_widget', 'focus', 'focus_position', 'get_cursor_coords', 'get_pref_col', 'keypress', 'mouse_event', 'move_cursor_to_coords', 'pack', 'render', 'rows', 'selectable', 'sizing', 'w']
>>> x.sizing()
frozenset({'flow'})

preetmishra: Thanks for the exposition; patching `View.sizing` seems like a better approach given the current implementation and how it makes the tests clearer and easier to change later.

neiljp:
@preetmishra Thanks for all your work on this - I know this has been a long time coming! I'm going to merge this now :tada:

PR #571, opened 2020-04-01
docker: Add dockerfile for building images from latest code (upstream).

sumanthvrao:
The current docker files in docker/Dockerfile.* build images by pulling
our releases from pypi. This commit introduces the 2 docker files at
docker/dev.Dockerfile.* which build images by using the latest code from
master, merged upstream.

README updated to reflect the new change.

### DIFF
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -8,8 +8,8 @@ RUN addgroup -S zulipgr \
 USER zulip
 WORKDIR /home/zulip
 
-RUN set -ex; python3 -m venv zulip-terminal-venv \
-    && source zulip-terminal-venv/bin/activate \
+RUN set -ex; python3 -m venv zt_venv \
+    && source zt_venv/bin/activate \
     && pip3 --disable-pip-version-check install zulip-term \
     && rm -rf /home/zulip/.cache
 
@@ -23,5 +23,5 @@ RUN addgroup -S zulipgr \
 COPY --from=builder --chown=zulip:zulipgr /home/zulip /home/zulip
 USER zulip
 WORKDIR /home/zulip
-ENTRYPOINT ["/home/zulip/zulip-terminal-venv/bin/zulip-term"]
+ENTRYPOINT ["/home/zulip/zt_venv/bin/zulip-term"]
 CMD ["--config-file", "/.zulip/zuliprc"]
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -1,23 +1,24 @@
 FROM python:3.7-alpine AS builder
 
 RUN addgroup -S zulipgr \
-&& adduser -S zulip -G zulipgr \
-&& apk add --no-cache ca-certificates gcc musl-dev libffi-dev openssl-dev libxml2-dev libxslt-dev
-
+    && adduser -S zulip -G zulipgr \
+    && apk add --no-cache ca-certificates gcc musl-dev \
+                          libffi-dev openssl-dev libxml2-dev \
+                          libxslt-dev
 USER zulip
 WORKDIR /home/zulip
 
 RUN set -ex; python3 -m venv zulip-terminal-venv \
-&& source zulip-terminal-venv/bin/activate \
-&& pip3 --disable-pip-version-check install zulip-term \
-&& rm -rf /home/zulip/.cache
+    && source zulip-terminal-venv/bin/activate \
+    && pip3 --disable-pip-version-check install zulip-term \
+    && rm -rf /home/zulip/.cache
 
 FROM python:3.7-alpine
 
 RUN addgroup -S zulipgr \
-&& adduser -S zulip -G zulipgr \
-&& apk add --no-cache ca-certificates libffi openssl libxml2 libxslt \
-&& rm -rf /var/cache/apk/*
+    && adduser -S zulip -G zulipgr \
+    && apk add --no-cache ca-certificates libffi openssl libxml2 libxslt \
+    && rm -rf /var/cache/apk/*
 
 COPY --from=builder --chown=zulip:zulipgr /home/zulip /home/zulip
 USER zulip
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -2,15 +2,23 @@ FROM python:3.7-alpine AS builder
 
 RUN addgroup -S zulipgr \
     && adduser -S zulip -G zulipgr \
-    && apk add --no-cache ca-certificates gcc musl-dev \
+    && apk add --no-cache git ca-certificates gcc musl-dev \
                           libffi-dev openssl-dev libxml2-dev \
                           libxslt-dev
 USER zulip
 WORKDIR /home/zulip
 
+ARG SOURCE=pypi
+ARG GIT_URL=https://github.com/zulip/zulip-terminal.git@master
+
 RUN set -ex; python3 -m venv zt_venv \
     && source zt_venv/bin/activate \
-    && pip3 --disable-pip-version-check install zulip-term \
+    && if [ "$SOURCE" = "pypi" ] ; \
+        then pip3 --disable-pip-version-check install zulip-term ; \
+       fi\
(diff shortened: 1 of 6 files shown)


kaustubh-nair:
@sumanthvrao This seems to work well, but I noticed that you're pulling the dependencies from github's master? This means, if I make modifications `requirements.txt` the packages won't get installed/updated?

docker/README.md
```diff
@@ -16,10 +16,16 @@ cd zulip-terminal/docker
 # Build the docker image with your preferred Dockerfile, here based on Alpine
 docker build -t zulip-terminal:latest -f Dockerfile.alpine .
 
+# Similarly, to build docker image with your preferred dev.Dockerfile, here based on Buster
+docker build -t zulip-terminal-dev:latest -f dev.Dockerfile.buster .
```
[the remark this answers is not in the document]
sumanthvrao: Changed 👍

docker/README.md
```diff
@@ -16,10 +16,16 @@ cd zulip-terminal/docker
...
+
 # Run zulip-terminal image and setup your zulip for the first time
 # (next time the $HOME/.zulip/zuliprc file will be used)
 mkdir $HOME/.zulip
```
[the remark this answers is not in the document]
sumanthvrao: Added another commit for this. Thanks for pointing it out.

sumanthvrao:
@preetmishra @kaustubh-nair Thanks for testing the PR and I'm glad it works. 

I have added another commit which clarifies the README and the use-case of the files.

@kaustubh-nair The requirement here was for users to use the latest code straight out of the box (On docker of course); Hence if they do developmental changes (which are unmerged) the docker image doesn't build from source. Note that if the changes are merged to upstream then the requirements.txt is run again as the cache is invalidated.

docker/dev.Dockerfile.alpine
```diff
@@ -0,0 +1,29 @@
+FROM python:3.7-alpine AS builder
+
+RUN addgroup -S zulipgr \
+    && adduser -S zulip -G zulipgr \
+    && apk add --no-cache git ca-certificates gcc musl-dev \
```
neiljp, line 5: These indents seem clearer; if we don't merge files, we could tidy the other files too.

sumanthvrao:
I pushed a bunch of commits here addressing the review @neiljp gave. (Apologies on leaving this PR go stale!)

* The first commit does a urwid upgrade to fix the locale issues (as mentioned by @fredrikekre in #800).
* The next two does minor reformatting
* The last adds an option to pass `--build-args` to decide on whether the image should be built using latest develpment code merged upstream or the latest pypi release.

fredrikekre:
I think it would be useful to be able to customize the branch and even the remote repo. Would make it very easy to test out PRs and also local changes.

sumanthvrao:
There have also been some discussions of having a third image which allows some sort of a docker-based development setup. I think that might fit the use case you suggested better.

neiljp (commented):
@sumanthvrao Thanks for reworking and updating this - this looks a lot simpler and clearer now :+1:

docker/Dockerfile.alpine
```diff
@@ -8,8 +8,8 @@ RUN addgroup -S zulipgr \
 USER zulip
 WORKDIR /home/zulip
 
-RUN set -ex; python3 -m venv zulip-terminal-venv \
-    && source zulip-terminal-venv/bin/activate \
+RUN set -ex; python3 -m venv zt_venv \
+    && source zt_venv/bin/activate \
```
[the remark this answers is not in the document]
sumanthvrao: This doesn't achieve anything as you pointed out, but my goal was to have a similar venv naming scheme in both places - just for consistency. Feel free to drop this commit if you feel it might not be something significant - hence I added it in a separate commit as well.

docker/Dockerfile.alpine
```diff
@@ -2,15 +2,21 @@ FROM python:3.7-alpine AS builder
...
 RUN addgroup -S zulipgr \
     && adduser -S zulip -G zulipgr \
-    && apk add --no-cache ca-certificates gcc musl-dev \
+    && apk add --no-cache git ca-certificates gcc musl-dev \
                           libffi-dev openssl-dev libxml2-dev \
                           libxslt-dev
 USER zulip
 WORKDIR /home/zulip
 
+ARG ENV
```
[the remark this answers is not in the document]
sumanthvrao: Sounds good! `source` seems like a descriptive and short name :) 

Yup, I felt the general `else` could be made more specific. I'm proposing we move to a two `if` statement option with `SOURCE=pypi` and `SOURCE=git`

sumanthvrao:
@neiljp Thanks for the review :) . I pushed an update addressing the points you brought up

1) Indented the if-then cases
2) Replaced the variable to `SOURCE`
3) Replaced the branch with `$ZT_BRANCH` (which is hardcoded right now)
4) Replaced `dev` option to `git`
5) Feel free to take a call on the zt_venv name as mentioned above

neiljp (commented):
@sumanthvrao This looks great to me :+1:

@fredrikekre Will the ZT_BRANCH option satisfy your request?

fredrikekre:
Yea looks good. My only comment is that `ZT_BRANCH` might not be the best name since it sets the repo *and* branch.

docker/Dockerfile.buster
```diff
@@ -4,16 +4,24 @@ RUN useradd --user-group --create-home zulip
 USER zulip
 WORKDIR /home/zulip
 
-RUN set -ex; python3 -m venv zulip-terminal-venv \
-&& . zulip-terminal-venv/bin/activate \
-&& pip3 --disable-pip-version-check install zulip-term \
-&& rm -rf /home/zulip/.cache
+ARG SOURCE
```
fredrikekre, line 7: Perhaps it is useful if this is the default?
```suggestion
ARG SOURCE=pypi
```

neiljp: I did wonder about this, but thought it might be better to leave it needing to be explicit? I suppose we could also change `pypi` to `latest_release` or similar, in case users aren't familiar with what pypi is.

Re the `ZT_BRANCH` issue, we could have `GIT_URL`?
```
ARG GIT_URL=https://github.com/zulip/zulip-terminal.git@master
```
and later
```
    then pip3 --disable-pip-version-check install git+$GIT_URL ;

Comment (fredrikekre) [reply docker/Dockerfile.buster:7]: > I did wonder about this, but thought it might be better to leave it needing to be explicit?

I think the opposite, I think it would be a better "default experience" if
```
docker build -t zulip-terminal:latest -f Dockerfile.alpine .
```
worked. Messing with `--build-arg`s seems unnecessary if you just want to build the image and get started IMO.

neiljp: @fredrikekre To clarify:
* I think it would be reasonable to take the approach you suggest - perhaps describing this in a "quick start" section with no build args needing to be specified, and then listing the git or pypi options?
* Was the :+1: to the `GIT_URL` approach?

@sumanthvrao What do you think?

fredrikekre: > I think it would be reasonable to take the approach you suggest - perhaps describing this in a "quick start" section with no build args needing to be specified, and then listing the git or pypi options?

Okay, I misunderstood then. Sounds good.

> Was the +1 to the GIT_URL approach?

:+1:

sumanthvrao:
Appreciate the feedback on this! I pushed an update which has both the items we were talking about, above.

* Renaming the `ZT_BRANCH` to `GIT_URL`
* Adding the default source to `pypi`
* Splitting out the readme into `Quick start` and `Advanced setup` sections.

Additionally in the README, I added a link to our `pypi` page, in case folks don't know what that is and can click the link to get a fair idea.

neiljp:
@sumanthvrao This looks like it resolves all the points :+1:
@fredrikekre Agree?

Thanks for everyone's input into this - this should make running ZT even easier :tada:

neiljp:
Merging now :tada: Thanks all, and please don't hesitate to post followup issues or PRs if things could be improved for you :+1:

PR #572, opened 2020-04-01
Refactor autocomplete methods

kaustubh-nair:
As per the new changes, adding the footer text for typeaheads will now cause changes only in the `generic_autocomplete` method. This should decrease a lot of redundancy faced in #540

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -23,86 +23,73 @@ class TestWriteBox:
         assert write_box.view == self.view
         assert write_box.msg_edit_id is None
 
-    @pytest.mark.parametrize('text, prefix_string, state', [
-        ('@Boo', '@', 0),
-        ('@Boo', '@', 1),
-        ('@_Boo', '@_', 0),
-        ('@_Boo', '@_', 1),
-        ('Plain Text', '', 0),
-        ('Plain Text', '', 1),
+    @pytest.mark.parametrize('text, state', [
+        ('Plain Text', 0),
+        ('Plain Text', 1),
     ])
-    def test_generic_autocomplete(self, mocker, write_box, text,
-                                  prefix_string, state):
-        autocomplete_mentions = mocker.patch(WRITEBOX+'.autocomplete_mentions')
+    def test_generic_autocomplete_no_prefix(self, mocker, write_box, text,
+                                            state):
         return_val = write_box.generic_autocomplete(text, state)
-        if prefix_string == '':
-            assert return_val == text
-        else:
-            autocomplete_mentions.assert_called_once_with(text, state,
-                                                          prefix_string)
+        assert return_val == text
 
-    @pytest.mark.parametrize('text, state, prefix_string,\
-                              required_typeahead', [
-        ('@Human', 0, '@', '@**Human Myself**'),
-        ('@Human', 1, '@', '@**Human 1**'),
-        ('@Human', 2, '@', '@**Human 2**'),
-        ('@Human', -1, '@', '@**Human 2**'),
-        ('@Human', -2, '@', '@**Human 1**'),
-        ('@Human', -3, '@', '@**Human Myself**'),
-        ('@Human', -4, '@', None),
-        ('@_Human', 0, '@_', '@_**Human Myself**'),
-        ('@_Human', 1, '@_', '@_**Human 1**'),
-        ('@_Human', 2, '@_', '@_**Human 2**'),
-        ('@H', 1, '@', '@**Human 1**'),
-        ('@Hu', 1, '@', '@**Human 1**'),
-        ('@Hum', 1, '@', '@**Human 1**'),
-        ('@Huma', 1, '@', '@**Human 1**'),
-        ('@Human', 1, '@', '@**Human 1**'),
-        ('@Human 1', 0, '@', '@**Human 1**'),
-        ('@_H', 1, '@_', '@_**Human 1**'),
-        ('@_Hu', 1, '@_', '@_**Human 1**'),
-        ('@_Hum', 1, '@_', '@_**Human 1**'),
-        ('@_Huma', 1, '@_', '@_**Human 1**'),
-        ('@_Human', 1, '@_', '@_**Human 1**'),
-        ('@_Human 1', 0, '@_', '@_**Human 1**'),
-        ('@Group', 0, '@', '@*Group 1*'),
-        ('@Group', 1, '@', '@*Group 2*'),
-        ('@G', 0, '@', '@*Group 1*'),
-        ('@Gr', 0, '@', '@*Group 1*'),
-        ('@Gro', 0, '@', '@*Group 1*'),
-        ('@Grou', 0, '@', '@*Group 1*'),
-        ('@G', 1, '@', '@*Group 2*'),
-        ('@Gr', 1, '@', '@*Group 2*'),
-        ('@Gro', 1, '@', '@*Group 2*'),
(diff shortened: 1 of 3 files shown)


zulipterminal/helper.py
```diff
@@ -477,6 +477,14 @@ def canonicalize_color(color: str) -> str:
         raise ValueError('Unknown format for color "{}"'.format(color))
 
 
+def styled_typeahead(typeaheads: List[str],
+                     wrapping_text: str) -> List[str]:
```
neiljp, line 481: I assume this method is a precursor to make styled (with bugdown/markdown) versions separate from the versions in the bottom bar? If so, this feels like it belongs with that work.

We could include it now, but it's clearer seeing it 'in action' - demonstrating why it's much better - though arguably it does that to some extent already.

zulipterminal/helper.py
```diff
@@ -442,7 +442,7 @@ def match_stream(stream: Any, text: str) -> bool:
     return False
 
 
-def match_groups(group_name: str, text: str) -> bool:
+def match_group(group_name: str, text: str) -> bool:
```
neiljp, line 445: I can see why this minimal rename could be useful, but you can explain why in the commit message. This helps ensure that reviewers and those who look at commits later understand your motivation - did you just change the name to add a commit? ;)

zulipterminal/ui_tools/boxes.py
```diff
@@ -132,18 +130,17 @@ def autocomplete_mentions(self, text: str, prefix_string: str
 
     def autocomplete_streams(self, text: str) -> List[str]:
         streams_list = self.view.pinned_streams + self.view.unpinned_streams
-        stream_names = [stream[0]
-                        for stream in streams_list
-                        if match_stream(stream, text[1:])]
+
+        streams = filtered_typeahead_names(streams_list, text, match_stream, 1)
+        stream_names = [stream[0] for stream in streams]
```
neiljp, line 135: This change has two list comprehensions, one inside a function, compared to just one list comprehension. That doesn't make it necessarily unwise, but if this can be seen more clearly worthwhile in the context of the other PR, let's put it there?

tests/ui_tools/test_boxes.py
```diff
@@ -17,6 +17,24 @@ def write_box(self, mocker):
...
+    @pytest.fixture()
+    def write_box_users(self, mocker, users_fixture, user_groups_fixture,
+                        write_box):
+        write_box.view.users = users_fixture
+        write_box.model.user_group_names = [
+            groups['name'] for groups in user_groups_fixture]
+        return write_box
+
+    @pytest.fixture()
+    def write_box_streams(self, mocker, streams_fixture, write_box):
```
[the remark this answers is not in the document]
kaustubh-nair: Right, I'll put this over with the footer PR. These fixtures get used multiple times over there.

tests/ui_tools/test_boxes.py
```diff
@@ -166,6 +167,6 @@ def test_autocomplete_streams(self, write_box_streams, text,
         ])
     def test_autocomplete_emojis(self, write_box, emojis_fixture,
                                  text, state, required_typeahead):
-        typeahead_string = write_box.autocomplete_emojis(text,
-                                                         state, emojis_fixture)
+        zulipterminal.emoji_names.EMOJI_NAMES = emojis_fixture
```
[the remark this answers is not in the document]
kaustubh-nair: Yes, I wanted to do that too. But is it possible to mock local variables? I couldn't find any reasonable way to do it. Please let me know if you have any ideas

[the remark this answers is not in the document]
kaustubh-nair: I don't think we would need an emoji version of `write_box`; `write_box` doesn't contain any emoji related attributes. Unless you're suggesting setting `emoji_names` as a `write_box` attribute?

zulipterminal/helper.py
kaustubh-nair: Right, this is mainly to decrease the clutter in the footer PR. It'll be better if I put it there :+1:

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Done :+1:

zulipterminal/helper.py
kaustubh-nair: It was really bothering me that all other `match_` methods were singular, but `match_groups` had an `s` at the end. :stuck_out_tongue_closed_eyes:

kaustubh-nair:
Thanks for the review @neiljp 
I have removed the footer PR related commits , so this one becomes much simpler now. 
Commits 2 and 3 are the most important ones - they should really help avoid duplication while adding new features. I'll add the commits I removed from here to the footer PR once these two commits are merged.

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
neiljp: The mock should be possible; I think this is more straightforward since `emoji_names` can be fully mocked.

tests/ui_tools/test_boxes.py
```diff
@@ -41,82 +41,70 @@ def test_init(self, write_box):
...
-        ('Plain Text', '', 0),
-        ('Plain Text', '', 1),
+    @pytest.mark.parametrize('text, state', [
+        ('Plain Text', 0),
+        ('Plain Text', 1),
     ])
-    def test_generic_autocomplete(self, mocker, write_box, text,
-                                  prefix_string, state):
-        autocomplete_mentions = mocker.patch(WRITEBOX+'.autocomplete_mentions')
+    def test_generic_autocomplete_no_prefix(self, mocker, write_box, text,
```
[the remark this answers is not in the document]
neiljp: You can also note that this test set is split into it's own test.

zulipterminal/ui_tools/boxes.py
```diff
@@ -96,18 +96,23 @@ def stream_box_view(self, caption: str='', title: str='') -> None:
...
+            typeahead = self.autocomplete_mentions(text, '@_')
         elif text.startswith('@'):
-            return self.autocomplete_mentions(text, state, '@')
+            typeahead = self.autocomplete_mentions(text, '@')
         elif text.startswith('#'):
-            return self.autocomplete_streams(text, state)
+            typeahead = self.autocomplete_streams(text)
         elif text.startswith(':'):
-            return self.autocomplete_emojis(text, state)
+            typeahead = self.autocomplete_emojis(text)
```
neiljp, line 105: We may be able to simplify this block further - it seems rather repetitive - though we don't have to do it in this PR necessarily.

kaustubh-nair: I'll probably look at this in the footer PR, some of the changes made there could affect the refactor here.

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
kaustubh-nair: Done :+1:

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
kaustubh-nair: Good idea :+1: done

kaustubh-nair:
@neiljp Thanks for the review! I've updated the PR according to your suggestions.

neiljp:
@kaustubh-nair I made the few small changes remaining and pushed this just now - thanks :tada:

There were a few minor typos in the function names in the commit titles/descriptions.

Other differences were just with regard to how the mocking can be simplified (no need to import the module to mock, and mock attributes can be set on initialization).

PR #584, opened 2020-04-08
Message reactions: Add spacing and refactor

neiljp:
I was diving into this code to look into the `:crossed_fingers:` vs `:fingers_crossed` situation and noticed this code could be cleaner.

This PR is mainly refactoring to make this clearer to read and more pythonic.

However, partly as a result of the refactoring (or vice versa) this PR adds "unstyled spaces" between reactions - this is particularly noticeable for messages such as in **#announce** with lots of reactions (and of different types), and clearly associates the number with the reaction itself (particularly when the message is selected). The other side effect is that there is also no trailing "styled" space.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -376,18 +376,23 @@ class MessageBox(urwid.Pile):
                     std_reaction_stats[reaction['emoji_code']] += 1
                 else:
                     # Includes realm_emoji and zulip_extra_emoji
-                    custom_reaction_stats[":"+reaction['emoji_name']+":"] += 1
-            dis = [
-                '\\U{:0>8} {} '.format(reaction, count)
+                    custom_reaction_stats[reaction['emoji_name']] += 1
+
+            std_reactions = [
+                '{} {} '.format(
+                    '\\U{:0>8}'.format(reaction)
+                    .encode().decode('unicode-escape'),
+                    count)
                 for reaction, count in std_reaction_stats.items()
             ]
-            emojis = ''.join(e.encode().decode('unicode-escape') for e in dis)
-            custom_emojis = ''.join(
-                ['{} {} '.format(reaction, count)
-                 for reaction, count in custom_reaction_stats.items()])
+            custom_reactions = [
+                ':{}: {} '.format(reaction, count)
+                for reaction, count in custom_reaction_stats.items()
+            ]
+            all_emoji_text = ''.join(std_reactions + custom_reactions)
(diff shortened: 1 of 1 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -368,25 +369,35 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
         if not reactions:
             return ''
         try:
-            reacts = defaultdict(int)  # type: Dict[str, int]
-            custom_reacts = defaultdict(int)  # type: Dict[str, int]
+            std_stats = Counter()  # type: typing.Counter[str]
+            custom_stats = Counter()  # type: typing.Counter[str]
```
preetmishra, line 373: Could we use `std_reaction_stats` and `custom_reaction_stats` or something similar instead to increase verbosity?

zulipterminal/ui_tools/boxes.py
```diff
@@ -368,25 +369,35 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
                 else:
                     # Includes realm_emoji and zulip_extra_emoji
-                    custom_reacts[":"+reaction['emoji_name']+":"] += 1
-            dis = [
-                '\\U' + '0'*(8-len(emoji)) + emoji + ' ' + str(reacts[emoji]) +
-                ' ' for emoji in reacts]
-            emojis = ''.join(e.encode().decode('unicode-escape') for e in dis)
-            custom_emojis = ''.join(
-                ['{} {} '.format(r, custom_reacts[r]) for r in custom_reacts])
+                    custom_stats[":"+reaction['emoji_name']+":"] += 1
```
preetmishra, line 379: [suggestion] We could `.format` here to make this clearer; `':{}:'.format(reaction['emoji_name'])`.

neiljp: This is a corner case where small string addition could be clearer/faster. When we are only working with python3.6 this may even be less clearer with f-strings.

However, I've pushed a revised version which avoids this completely, since we don't need this in the stats part, only really in the series of strings.

neiljp:
@preetmishra I updated the commits as above - thanks for the review :+1:

sumanthvrao:
Refactored changes look neat and clean! Thanks @neiljp

zulipterminal/ui_tools/boxes.py
preetmishra: Thanks for the explanation. :+1:

PR #589, opened 2020-04-11
Update CHANGELOG to add changes since 0.4.0.

neiljp:
Feedback welcome on suggested changes, or things I've missed!

I'm hoping we can do a release soonish before GSoC 2020, before which we will update for any last-minute changes and adjust the text slightly to represent that the next release is... 0.4.1?

I'm working on a simple script to automate the release separately, which will replace the PR Aman worked on some time ago which involved adding release capability into `setup.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,47 @@
 # Zulip-terminal Changelog
 
+## Next Release
+
+Expected to be a point release, this contains much of the remaining work from last year's GSoC students, and various other contributors - thanks to you all! (180+ commits)
+
+If you've been waiting for autocomplete, popup notifications, mentions, better themes & search...this release is for you!
+
+### HIGHLIGHTS
+- Autocomplete users/groups/streams/emoji when composing messages, given a prefix (<kbd>ctrl</kbd>+<kbd>f</kbd> / <kbd>ctrl</kbd>+<kbd>r</kbd>)
+- Popup notifications for private messages & mentions (NOTE: disabled by default; tools may be required on WSL/Linux)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,46 @@
...
+- Add support to render tables in messages
+- Improve rendering of links in messages
+- Categorize and further reorganize help menu
+- Show the year of a message, if it is not the current year
+- Show server name & description in the title bar
+
+### Important bugfixes
+- On changing to a narrow which contains the focused message, retain focus on that message
+- Message search results reset between searches
+- Improved unread count handling
```
sumanthvrao, line 35: One more bug which existed in the previous release was (https://github.com/zulip/zulip-terminal/pull/462); I happened to install v0.4.0 for testing and ran into this within 5 minutes. I think folks would be glad to hear that the UI is no longer distorted during some startups.

neiljp: Right, good catch! I've marked this as #456, as that was the bug itself?

neiljp:
@sumanthvrao I've not focused on a specific release number at this point, other than just going for 0.4.1 unofficially. Given we're not at 1.x.y yet, then we have less information to include, and as long as pypi gets updated with a new more-stable and full-featured version that people upgrade to, I'm not sure the version is too important right now? Happy to discuss on czo of course.

neiljp:
I just pushed an update to include that we now support WSL, including WSL2 @amanagr :)

neiljp:
Merged with updates for number of commits and to mention platform support as a reason to be interested :)

PR #591, opened 2020-04-12
Fix issue with quotes and distribution on WSL.

amanagr:
WSL2 uses 'microsoft' and WSL1 uses 'Microsoft' in its release name.
Remove unnecessary quotes in WSL command.


neiljp:
@amanagr As discussed via PM, just merged with more text as f445bca7b :+1:

PR #592, opened 2020-04-14
setup: Allow running with Python 3.8.x

punchagan:
With the previous check `<=3.8`, `3.8.2` didn't work.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,7 @@ setup(
         'Programming Language :: Python :: Implementation :: CPython',
         'Programming Language :: Python :: Implementation :: PyPy',
     ],
-    python_requires='>=3.5, <=3.8',
+    python_requires='>=3.5, <3.9',
     keywords='',
     packages=find_packages(exclude=['tests', 'tests.*']),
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -75,7 +75,7 @@ def long_description():
         'Programming Language :: Python :: Implementation :: CPython',
         'Programming Language :: Python :: Implementation :: PyPy',
     ],
-    python_requires='>=3.5, <=3.8',
+    python_requires='>=3.5, <3.9',
```
neiljp, line 78: If you look, this was the form we had in a previous iteration. I accepted feedback that it should be equivalent, except it doesn't take minor versions into account :-1:

neiljp:
@punchagan Merged with amended text as e3f9014f6.

neiljp:
@kaustubh-nair I'm concerned about the travis issue too; my thought would be that for some reason the travis version is identifying as exactly 3.8 rather than 3.8.x, but you're welcome to dig into it.

Supporting 3.8.x seems reasonable (for all x), so <3.9 should be good.

PR #596, opened 2020-04-15
notify: Use powershell style escapes to make notify work on WSL.

amanagr:
@neiljp I found today that notifications were not working for some quoted messages. After some digging, I found that escaping quotes working differently for powershell, It uses <kbd>`</kbd>  instead of <kbd>\\</kbd> to escape characters.


neiljp:
@amanagr Commit merged manually after adjusting comments slightly as 1693fb701 - thanks! :tada:

PR #608, opened 2020-04-20
boxes: reactions: Stop using emoji package to obtain emoji names.

### LINKED ISSUE #599: Reactions can show differently in multiple places
Some reactions appear to show differently below messages and in the message-info box - and also possibly compared to in the webapp (on hover, for the name).

This looks confusing and like a major bug, even though it typically shows at least a similar reaction, so we should probably figure this out - it just looks bad.

This may be related to data which we don't directly have? @kaustubh-nair

kaustubh-nair:
Fixes #599 
Here is a topic where I've already tested it a bit https://chat.zulip.org/#narrow/stream/7-test-here/topic/test.20reactions/near/855591

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -6,7 +6,6 @@ from time import ctime, time
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 from urllib.parse import urljoin, urlparse
 
-import emoji
 import urwid
 from bs4 import BeautifulSoup
 from bs4.element import NavigableString, Tag
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -366,34 +366,23 @@ class MessageBox(urwid.Pile):
         if not reactions:
             return ''
         try:
-            std_reaction_stats = Counter()  # type: typing.Counter[str]
-            custom_reaction_stats = Counter()  # type: typing.Counter[str]
+            reaction_stats = Counter()  # type: typing.Counter[str]
             for reaction in reactions:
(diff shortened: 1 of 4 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -366,30 +365,19 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
+                emoji_stats[reaction['emoji_name']] += 1
+
+            spaced_emoji_stats = [
                 ':{}: {}'.format(reaction, count)
-                for reaction, count in custom_reaction_stats.items()
+                for reaction, count in emoji_stats.items()
             ]
-            all_emoji = std_reactions + custom_reactions
+
             spaced_emoji = [
```
neiljp, line 377: I know you've not adjusted the name of this variable, but I think it'd be useful to have a consistent name here - `spaced_reaction_texts`?

zulipterminal/ui_tools/boxes.py
```diff
@@ -366,30 +365,19 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
         if not reactions:
             return ''
         try:
```
neiljp, line 367: Perhaps as a follow-up commit we could narrow down what this `try` block is intended to do? Not everything needs to be in it, and wrapping everything makes it less clear.

kaustubh-nair:
@neiljp Thanks for the review, I've pushed the changes.
I put all the code outside the `try` block but I still haven't run into any exceptions. Any ideas on how to investigate this?

neiljp:
@kaustubh-nair Thanks, just merged with moving the `import` removal between the commits - great to have this fixed! :tada: 

Regarding the exception aspect, the main unknown data is the reactions data - everything else should work?

PR #610, opened 2020-04-22
themes: Add 256-color support to default theme.

### LINKED ISSUE #283: Augment 'default' theme with 256-color version

neiljp:
This is a more cautious updated version of #310, without the deeper black and attempting to match the more 'subtle' colors I see in my terminal with the default theme normally.

The colors match those observed using Tango theme in gnome-terminal.

For those with a 256-color terminal (ie. most) this may alleviate some of the concerns discussed recently, comparing preferences for the 'gruvbox' theme over 'default'.

Fixes #283.

NOTE: This PR **removes support** for transparency by default, with the intention that a different approach such as #555 / #513 applies to all themes in a sensible way. The 'default' theme is not intended as the only transparent theme, and if we end up switching to a different theme being the default, then transparency will currently disappear in any case.


neiljp:
Thanks to all those who tested this and posted positive feedback on #**zulip-terminal** :+1:  I'm merging this now :tada:

PR #612, opened 2020-04-22
themes/run: Rename themes & add compatibility aliases.

neiljp:
This implements the renaming of themes as discussed in the 'Rename default theme' topic.

Also included is an 'alias' table, to allow the old names to continue working, at least for now.

### DIFF
--- a/tests/config/test_themes.py
+++ b/tests/config/test_themes.py
@@ -7,7 +7,7 @@ from zulipterminal.config.themes import (
 
 
 expected_complete_themes = {
-    'default', 'gruvbox', 'light', 'blue'
+    'zt_dark', 'gruvbox_dark', 'zt_light', 'zt_blue',
 }
 
 
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -94,7 +94,7 @@ def test_valid_zuliprc_but_no_connection(capsys, mocker, minimal_zuliprc,
     lines = captured.out.strip().split("\n")
     expected_lines = [
         "Loading with:",
-        "   theme 'default' specified with no config.",
+        "   theme 'zt_dark' specified with no config.",
         "   autohide setting 'no_autohide' specified with no config.",
         "\x1b[91m",
         ("Error connecting to Zulip server: {}.\x1b[0m".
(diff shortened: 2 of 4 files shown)


tests/config/test_themes.py
```diff
@@ -7,7 +7,7 @@
 
 
 expected_complete_themes = {
-    'default', 'gruvbox', 'light', 'blue'
+    'zt_dark', 'gruvbox_dark', 'zt_light', 'zt_blue',
```
amanagr, line 10: Can we leave the name of `gruvbox` as it is? This will give error for users who use `gruvbox` via zuliprc.

neiljp: You should find that you can still use `gruvbox` via the theme aliases.

neiljp:
The only thing missing here is that I could add some extra test code.

preetmishra:
@neiljp We could add a deprecation warning if we're planning to drop `THEME_ALIASES` in the future.

neiljp:
@preetmishra Re deprecation warnings, I did include that in my initial version, but I think we'll be good to keep the theme aliases for a while? At the very least we are pointing out that the aliases exist already (on loading), which is important - did I actually label `zt_dark` as the theme in my `zuliprc`? ;) We could consider deprecating them if we get to a later stage - though there is maybe an argument that now is a better stage to even remove them straight away!

preetmishra:
@neiljp I agree. We shouldn't remove them straight away but rather indicate that we plan to remove them at some stage. Moreover, if we decide to keep the aliases, would we list them in our themes list or elsewhere?

neiljp:
@preetmishra Are you suggesting a change beyond this PR? My intention at this point is that existing theme names being used (anywhere) 'just work' but get converted to the new names via the aliases - but this is indicated during loading via the updated messages (eg. 'default' in config being treated as 'zt_dark'). However, we wouldn't advertise the old names explicitly otherwise - eg. something like `--list-themes` (#613) would only show the new names.

We could open an issue to track when we want to make this breaking change, assuming we don't do it now.

preetmishra:
@neiljp Yes. The plan you've proposed looks good! :+1:

neiljp:
Rebased to check new linting rules, and now merging :tada:

Issue #606, opened 2020-04-20
Wrong rendering of reaction removal if multiple already present


neiljp:
* Find a message with multiple thumbs-up
* Toggle thumbs-up (+)
* Toggle again

One of those last two, depending on your previous thumbs-up reaction to that message, currently completely removes that reaction from being rendered - and would probably do the same for other reactions when we add that feature in future.

kaustubh-nair:
@neiljp I can reproduce this but not all the time.
Sometimes it leads to different behaviour
This is what happens:
*  Find a message with multiple thumbs-up
*  Toggle thumbs-up (+)
* Toggle again to remove thumbs-up
* Previous thumbs-ups remain, but now I cannot add my thumbs up again.

preetmishra:
I suspect that the reason is the following code block in `model.py`. There are two key issues here: 
* it removes all the reaction (we could warrant a `break`),
* it removes whichever reaction it finds first (the comment has an explanation).

https://github.com/zulip/zulip-terminal/blob/7bc9da6c7d762f93554b26abb434f6a65ce68f79/zulipterminal/model.py#L840-L844

Ezio-Sarthak:
@preetmishra Are you planning to push the changes in [your local commit](https://github.com/preetmishra/zulip-terminal/commit/121827541d93f7498db013a0851db7e128aae9a0) ?

I guess it would be good to merge this before #913. so as to avoid provoking the removal issue in the EmojiPicker?

neiljp:
@Ezio-Sarthak I have a WIP branch for this which improves some tests as well. I'll work on digging it out.

PR #614, opened 2020-04-22
Show reactions you contributed to in a new style

### LINKED ISSUE #479: Toggle styling of reactions to indicate whether user has contributed
This is visible in the webapp by a slight change to the style when toggling a reaction, and allows awareness of participation in a reaction when returning to older messages.

The idea of including something similar in ZT came from my comment on #476, ie. that we could achieve something similar by toggling the boldness.

This would avoid needing to go into the `i` menu to determine whether you have already reacted to a message, or toggling it to see if the count goes up or down.

This would probably be a follow-up to #476.

**see also an accidental duplicate in #512, which mentions an alternative styling**

neiljp:
* Added test for `MessageBox.reactions_view`; this is an 'internal' method, but is a good small piece to test

* Main feature:
  - Added `reactions_mine` style (inverted form of regular reactions)
  - Amended test added in first commit
  - Amended `reactions_view` method to use styling appropriately

I've not refactored the added parametrization into a fixture at this point, but added a FIXME to indicate we could do this in future. I have another active branch which may parametrize over message reactions, so that could be where to work on this.

Fixes #479.


neiljp:
I just updated the first commit to actually fix a bug, since our existing reactions weren't reliably ordered (demonstrated by the results from python 3.5). I've gone for alphabetical in the first instance, and we can always update the ordering later - but with a defined ordering in there, we have a known output at least :)

kaustubh-nair:
@neiljp This looks great in `default` theme :) 
I found a bug where if there is an already existing reaction, and I add and remove my reaction - the style remains inverted.


I'm also getting startup errors in other themes,
This is what I get on `gruvbox``
```
urwid.display_common.AttrSpecError: Unrecognised color specification in background ('h132, bold')

```

neiljp:
@kaustubh-nair Glad you like it - I enjoy it too :+1:

Re existing reaction, is this #606?

I just pushed an update which fixes the issues in the other themes, as well as an additional commit which should report some of these theme-related errors more cleanly  in future.

zulipterminal/ui_tools/boxes.py
```diff
@@ -365,17 +365,27 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
         if not reactions:
             return ''
         try:
-            reaction_stats = Counter()  # type: typing.Counter[str]
+            reaction_stats = defaultdict(set)
             for reaction in reactions:
-                reaction_stats[reaction['emoji_name']] += 1
+                user_id = int(reaction['user'].get('id', -1))
+                if user_id == -1:
+                    user_id = int(reaction['user']['user_id'])
```
kaustubh-nair, line 372: Why not something like -
```
  user_id = reaction['user']['id'] 
           if 'id' in reaction['user'] 
           else reaction['user']['user_id']
```

neiljp: Regarding this version of the code, I follow advice I was given previously and agree with, which is that if these conditional expressions go beyond one line, then it's probably unclear - or at least equivalent to just using something like I did, or equivalently a version of what you did:
```python3
if 'id' in reaction['user']:
    user_id = reaction['user']['id']
else:
    user_id = reaction['user']['user_id']
```
I would claim that is clearer than the conditional (if an extra line), and we could indeed use it...

... but that said, the real solution here is to do the normalization of the `user` dict in the first place, which will make this kind of check go away completely!

kaustubh-nair: Right, fixing the mismatch in the events should render this unnecessary. :+1:

neiljp:
OK, this sounds good, so merging this now :tada:

PR #619, opened 2020-04-23
tests: Add command keys for exit popup keypress tests.

### LINKED ISSUE #561: Add additional keypresses for `GO_BACK` keypress tests for popups.
Currently, the `GO_BACK` keypress tests for popups test whether a popup closes when the `GO_BACK` key is pressed. However, this behavior can also be triggered with the respective trigger command keys (`i` for info views and `?` for help) for popups.

We should add these extra command keys in tests along with `GO_BACK` and rename the tests as well (since it would not be testing just the `GO_BACK` key anymore).

preetmishra:
This brings minor changes to exit popup keypress tests (previously 'GO_BACK' tests).
* The first and the second commit are renaming refactors to improve consistency (more on this is in the respective commit description).
* The third commit adds command keys to exit popup keypress tests (see #561).
* The fourth commit adds exit popup keypress test for `StreamInfoView`.

Fixes #561.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1170,8 +1170,9 @@ class TestHelpMenu:
         self.help_view.keypress(size, key)
         assert not self.controller.exit_popup.called
 
-    @pytest.mark.parametrize('key', keys_for_command("GO_BACK"))
-    def test_keypress_GO_BACK(self, key):
+    @pytest.mark.parametrize('key', {*keys_for_command('GO_BACK'),
+                                     *keys_for_command('HELP')})
+    def test_keypress_exit_popup(self, key):
         size = (200, 20)
         self.help_view.keypress(size, key)
         assert self.controller.exit_popup.called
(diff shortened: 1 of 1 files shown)


preetmishra:
@neiljp Thanks for the review and the comments.  :+1:  I'll keep these things in mind.

My intention was to make reviewing easier for the reviewer; though I should have mentioned whether I would like them to be squashed or left untouched before merging.

I have squashed the second ('exit_popup' name change) and the third commit (command keys addition) into one as they seem pretty straightforward and related. I have left the test addition commit as it is; I am not sure whether this should be squashed too?

Re `TestHelpMenu`, I understand what you're suggesting. I am holding of this particular commit; I'll wait to accumulate more similar changes. :)

neiljp:
@preetmishra As I said, commits can always be squashed, but this new layout is much clearer and ready to merge :+1: If commits stand alone cleanly in what they do, I don't see harm in small commits, as long as there aren't other good changes to go with them.

PR #621, opened 2020-04-24
Migrate from pycodestyle to flake8 in CI & locally

neiljp:
Flake8 covers a lot more coding errors than pycodestyle, including finding unused imports - just as one example. This PR translates the dev and CI environments to use flake8. No documentation should need to be updated, as long as people are just using `tools/lint-all` rather than raw pycodestyle.

The 'barebones' flake8 takes ~6s to run locally for me right now, compared to the ~3s for pycodestyle, but combined with mypy and isort, this isn't as big a proportional change.

This PR removes default installation of pycodestyle (it is installed via flake8, though). We could keep it around as an intentional developer tool, but then we'd need to keep the settings synchronized.

flake8 also has various plugins, which we could investigate further if we go ahead with this.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,11 +23,27 @@ atomic=True
 # This ensures compatibility with pytest-pep8
 lines_after_imports=2
 
-# This is a first step from using the old pep8 plugin to pytest
-# We could use something else in future, like a combination linter
-# (flake8, pylint, zulint)
-[pycodestyle]
-ignore=E121,E123,E126,E226,E24,E704,W503,W504,E117,E252
+# flake8 takes a little more time than pycodestyle, but finds more things
+# (we could also look at pylint, zulint, etc.)
+[flake8]
+ignore=
+  # E121 - continuation line under-indented for hanging indent
+  E121,
+  # E123 - closing bracket does not match indentation of opening bracket's line
+  E123,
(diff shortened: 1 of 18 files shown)


setup.cfg
```diff
@@ -23,11 +23,27 @@ atomic=True
...
+# (we could also look at pylint, zulint, etc.)
+[flake8]
+ignore=
+  # E121 - continuation line under-indented for hanging indent
+  E121,
+  # E123 - closing bracket does not match indentation of opening bracket's line
+  E123,
+  # E126 - continuation line over-indented for hanging indent
+  E126,
+  # E226 - Missing whitespace around arithmetic - FIXME possibly change this
```
sumanthvrao, line 36: Is there a reason we ignore this?

sumanthvrao:
@neiljp Liking this stricter linter check especially with unused imports! PR LGTM. Left a small comment.

setup.cfg
neiljp: There are quite a few cases of this, and I'm hoping we can implement a general fixer tool with the possibility of pre- or post-commit hook which will fix this for us - but I wasn't sure if we actually wanted this, particularly given our line length, though we can always reformat code accordingly of course.

PR #625, opened 2020-04-25
Use right-triangle symbol to separate stream & topic.

neiljp:
I was looking at our use of '>' recently and wondered if we could do better. This is the best character I've found so far.

This moves the separator from using '>' to using '▶'.

Tests updated.


sumanthvrao:
I was wondering if a space beside the "right triangle" would make things a little more aesthetic?

image

neiljp:
@sumanthvrao I do like the extra space, though I'm often the person who is over-cautious about this, for those with narrow screens!

Interested to hear others' views.

neiljp:
I just pushed a commit for the extra space, which we can squash into the previous commit. If you wish to test with and without, you can simple checkout HEAD~ from this PR and then back again to compare :)

amanagr:
@neiljp This looks cool! I support the extra space.

kaustubh-nair:
This looks nice! And I think extra space fits better too.
I'm wondering if we should add the triangle on the search bar as well - for the sake of completeness?

neiljp:
I've just squashed the commits together and slightly rephrased the text, and plan to merge this tomorrow after the positive feedback unless there's anything to the contrary before then :+1:

@kaustubh-nair This is an interesting point, but let's consider this separately? Also, this is really a 'current narrow' area, which sometimes has search in it - is that what you meant and understand it to be? We could consider this as part of what I started talking about more specifically in the 'Topic narrow text' topic in #**zulip-terminal**

neiljp:
Thanks for the feedback everyone - I'm merging this now, as promised :tada:

PR #626, opened 2020-04-25
cleanup: Results of pyupgrade with --py3-only --keep-percent-format.

### LINKED ISSUE #360: Drop support for py3.4 & migrate to py3.5+ code and any new features

neiljp:
This tool automatically upgrades some common early python.

Changes made include:
* removal of u prefix from unicode strings
* remove some excessive uses of set() where {...} is sufficient
* use 'yield from' for spinning_cursor
* remove extra parentheses

Fixes #360, at least to a reasonable automated level at this point.

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -264,7 +264,7 @@ def all_themes() -> List[str]:
 
 def complete_and_incomplete_themes() -> Tuple[List[str], List[str]]:
     complete = {name for name, styles in THEMES.items()
-                if set(s[0] for s in styles).issuperset(required_styles)}
+                if {s[0] for s in styles}.issuperset(required_styles)}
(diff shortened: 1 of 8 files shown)


sumanthvrao:
It's cool that such a tool exists. LGTM! 

With this, we can finally bid 'au revoir' to Py3.4 :wave:

PR #627, opened 2020-04-25
README: Update installation, running & configuration notes.

neiljp:
NOTE: Some of this was inspired by trying to simplify and focus the README better, but also include some of the tips we've found recently, such as installing a development version straight from git using pip.

This is mainly a restructuring of text, with a few additions.

* Continue to recommend installing in a virtualenv, but place those
  details lower and integrate them with the 'different sessions' note
* Emphasize the main installation command is from PyPI
* Add a way to install current development version from git using pip
* Add that we include some docker build scripts
* Remove the simple 'running' part and 'what it shows when running'
* Turn the zuliprc file text into a 'Running for the first time' section
* Convert the 'example zuliprc file' text into a 'Configuration' section
* Link to 'Configuration' in the top links, not 'Installation & Running'

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 An interactive terminal interface for [Zulip](https://zulipchat.com).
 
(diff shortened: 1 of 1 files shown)


sumanthvrao:
LGTM! 👍

PR #630, opened 2020-04-27
Refactoring focusing on tidying/migrating msg_list & msg_view

neiljp:
The motivation for this PR started as an exploratory migration of the `msg_list` and `msg_view` attributes of the `Model` into a more appropriate location - the `View`. These variables have stuck out each time I look at the model due to their UI focus, and that they're initialized to meaningful values outside of the `Model`. One is used in some model methods, but similarly to `set_count`, these are cases where there are model and view changes, which can be split - and so the UI ones can hopefully be migrated into the `View` after this change.

Some less directly connected refactoring commits are included: the first three are not strictly related, so could be merged separately or first, but should be small or separate to understand and review.

The last 4 commits are the thrust of the PR:
* remove use of `msg_view`, since it is a shorthand for `msg_list.log`
* do the migration of `msg_list` into `View`
* rename the existing `message_view` method in `View`
* rename `msg_list` to `message_view` to improve the accuracy/clarity

This is a large change, but I believe worth the effort!

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -74,7 +74,6 @@ class TestController:
                               stream_button, index_stream) -> None:
         controller.model.narrow = []
         controller.model.index = index_stream
-        controller.model.msg_view = mocker.patch('urwid.SimpleFocusListWalker')
         controller.model.msg_list = mocker.patch('urwid.ListBox')
         controller.model.stream_dict = {
             205: {
@@ -87,9 +86,10 @@ class TestController:
         assert controller.model.stream_id == stream_button.stream_id
         assert controller.model.narrow == [['stream',
                                             stream_button.stream_name]]
-        controller.model.msg_view.clear.assert_called_once_with()
+        controller.model.msg_list.log.clear.assert_called_once_with()
 
-        widget = controller.model.msg_view.extend.call_args_list[0][0][0][0]
+        widget = (controller.model.msg_list.log
+                  .extend.call_args_list[0][0][0][0])
         stream_id = stream_button.stream_id
         id_list = index_stream['stream_msg_ids_by_stream_id'][stream_id]
         assert {widget.original_widget.message['id']} == id_list
@@ -100,7 +100,6 @@ class TestController:
                            ['topic', msg_box.topic_name]]
         controller.model.narrow = []
         controller.model.index = index_topic
-        controller.model.msg_view = mocker.patch('urwid.SimpleFocusListWalker')
         controller.model.msg_list = mocker.patch('urwid.ListBox')
         controller.model.stream_dict = {
             205: {
@@ -112,9 +111,10 @@ class TestController:
         controller.narrow_to_topic(msg_box)
         assert controller.model.stream_id == msg_box.stream_id
         assert controller.model.narrow == expected_narrow
-        controller.model.msg_view.clear.assert_called_once_with()
+        controller.model.msg_list.log.clear.assert_called_once_with()
 
-        widget = controller.model.msg_view.extend.call_args_list[0][0][0][0]
+        widget = (controller.model.msg_list.log
+                  .extend.call_args_list[0][0][0][0])
         stream_id, topic_name = msg_box.stream_id, msg_box.topic_name
         id_list = index_topic['topic_msg_ids'][stream_id][topic_name]
         assert {widget.original_widget.message['id']} == id_list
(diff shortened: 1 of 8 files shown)


preetmishra:
@neiljp Thanks for working on this. I have only examined the first three commits for this pass. The changes and the commit (with description) look good. They'll improve the readability quite a bit. :+1: 

We could merge the first three commits for now and rebase the changes [to resolve conflicts] for the next pass.

neiljp:
@preetmishra Thanks for the review - I just pushed the first three commits to master :tada:

I've pushed rebased versions of the others here, which are ready for general review.

There may be some remaining clean-up of excess test material which we no longer need (likely some mocks).

tests/core/test_core.py
```diff
@@ -72,23 +74,26 @@ def test_editor_mode_exits_after_entering(self, mocker, controller):
 
     def test_narrow_to_stream(self, mocker, controller,
                               stream_button, index_stream) -> None:
+
```
preetmishra, line 77: Blank line?

tests/core/test_core.py
```diff
@@ -30,8 +30,10 @@ def controller(self, mocker) -> None:
         self.theme = 'default'
```
preetmishra, line 30: The area section in the commit title is unsorted.

zulipterminal/ui_tools/views.py
```diff
@@ -467,15 +467,15 @@ def mouse_event(self, size: urwid_Size, event: str, button: int, col: int,
 class MiddleColumnView(urwid.Frame):
     def __init__(self, view: Any, model: Any,
                  write_box: Any, search_box: Any) -> None:
-        msg_list = MessageView(model, view)
+        message_view = MessageView(model, view)
```
preetmishra, line 470: :+1:

tests/core/test_core.py
neiljp: It's odd how these sneak in!

tests/core/test_core.py
neiljp: We're not really consistent with this, but in terms of the other commits I agree this makes it clearer.

tests/ui/test_ui.py
```diff
@@ -135,17 +139,17 @@ def test_main_window(self, mocker):
...
         left.assert_called_once_with()
-        center.assert_called_once_with()
+        # NOTE: Don't check center here, as we're monkeypatching it
         right.assert_called_once_with()
 
         expected_column_calls = [
             mocker.call([
                 (View.LEFT_WIDTH, left()),
-                ('weight', 10, center()),
+                ('weight', 10, mocker.ANY),  # ANY is a center
```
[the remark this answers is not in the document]
neiljp: * `mocker.ANY` matches all values passed in. In some cases you can get away with `.called`, not caring about the values, and in others you care about each parameter; if you can't test or don't care about one, but want to match the others, you can use this for the `anything` parameter.
* The challenge in this test is that I added the `assert` you referred to and queried the location of in another comment, which relies on `message_view` setting `msg_list`. if we want to check `message_view` is called then we need the equivalent of a mock, but we need to set the view attribute before the `__init__` is called (to avoid the `assert` failing), which depends on the existence of the object to set - or at least that's my interpretation. With `monkeypatch` we replace the method itself, but then can't check the mock. Other ideas welcome!

neiljp:
@preetmishra I've pushed an updated version with a few changes which should address the points you raised - let me know :)

tests/ui/test_ui.py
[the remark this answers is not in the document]
preetmishra: Ah, I understand it now. Thanks for the clarification and the comments.

tests/ui/test_ui.py
```diff
@@ -114,7 +114,7 @@ def test_main_window(self, mocker, monkeypatch):
 
         # NOTE: Use monkeypatch not patch, as view doesn't exist until later
         def just_set_msg_list(self):
```
preetmishra, line 116: Shouldn't the method name change here to `just_set_message_view` too?

neiljp:
@sumanthvrao I wanted to do this some time ago, and hopefully this will make understanding some things easier :)

Regarding the lengthy commits, this was mainly moving/renaming so hopefully was not too onerous to review - I made sure it passed linting and pytest after each commit too. The scale of the changes is simply related to how interconnected this variable is. We could have gone with encapsulation first, but the inconsistent naming felt like a feasible step to help understanding in the meantime.

@preetmishra Thanks for that test change - I've updated that test name :+1: 

After letting CI complete, I'm going to merge the minor update shortly.

PR #634, opened 2020-04-28
Use bullet character in list & remove newlines.

neiljp:
This change:
* makes it clearer when markdown text is recognized as bullets
* saves space by removing extra lines that are in the html

Tests amended.

New tests added for changed code, to ensure newlines removed:
* between ul and li elements
* between li and p elements
---
To see a good example of the reduction in newlines this makes, have a look at:
* simple bulleted elements in eg. #**test here**
* the bash++ topic in #**learning** (there were extra newlines *after* a bullet, before text)

A side by side comparison makes the difference very obvious.

The bullet characters make it clearer when a bullet is used, since `* foo` previously appeared like a bullet, but wasn't always recognized by the markdown parser eg. depending on indent, so wasn't always obviously rendered differently.

Note that the bullet symbol doesn't cause rendering issues for me, and is already used in the user list, so should be good to use.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1451,8 +1451,12 @@ class TestMessageBox:
          .format(SERVER_URL),
             [('msg_link', '{}/some/relative_url'.format(SERVER_URL))]),
         ('<li>Something', ['  \N{BULLET} ', '', 'Something']),
+        ('<li>\n<p>Something', ['  \N{BULLET} ', '', '', '', 'Something']),
         ('<li>Something<li>else',  # NOTE Real items are newline-separated?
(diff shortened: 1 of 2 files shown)


neiljp:
@preetmishra I do think this does look better visually, with the bullets (vs stars) and less whitespace - but this is also about the presence of what looks like 'wrong' newlines, between a bullet and its text. Either way, a good thing :)

What rebase work needs doing? What gives that impression?

preetmishra:
@neiljp Indeed.

My bad. I ran `./tools/lint-all` on `./tools/fetch-pull-request` (instead of `./tools/fetch-rebase-pull-request`) accidentally. Nevermind, everything seems to be okay. 👍

neiljp:
@preetmishra This should be less of an issue when we're more stable on linters etc, though we could of course adjust further.

PR #635, opened 2020-04-29
Don't corrupt screen if notifying fails; indicate to user once instead

neiljp:
Previously if notifications were enabled, but we cannot find the notification executable, then a traceback appears over the UI - I only noticed this since I copied a zuliprc file to another machine where the notification commands weren't available.

This small commit fixes this by wrapping the subprocess.run in a try/except block and logging FileNotFoundError exceptions, currently to `debug.log`.

**NOTE** This fixes the issue as it stands, but could spam the debug file and not be very obvious. I've noted some alternatives in the comments; we could merge this now to avoid the messed-up UI and work on the method in another issue/PR, or decide how best to approach this.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -11,6 +11,7 @@ from urllib.parse import urlparse
 import zulip
 from mypy_extensions import TypedDict
 
+from zulipterminal.config.keys import keys_for_command
 from zulipterminal.helper import (
     Message, asynch, canonicalize_color, classify_unread_counts,
     index_messages, initial_index, notify, set_count,
@@ -79,6 +80,8 @@ class Model:
         self.server_url = '{uri.scheme}://{uri.netloc}/'.format(
                           uri=urlparse(self.client.base_url))
         self.server_name = ""
+
+        self._notified_user_of_notification_failure = False
 
         self.event_actions = OrderedDict([
             ('message', self._handle_message_event),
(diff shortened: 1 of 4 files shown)


neiljp:
I just pushed a minimal improved version that pops up a notice a maximum of once, while still continuing to send notifications subsequently.

@aero31aero @preetmishra @sumanthvrao

zulipterminal/model.py
```diff
@@ -701,7 +707,19 @@ def _handle_message_event(self, event: Event) -> None:
                 self.controller.update_screen()
 
         # We can notify user regardless of whether UI is rendered or not.
```
sumanthvrao, line 709: Now that we need the UI to have rendered to show notifications, should we remove this comment? Should we do some bounded waiting here?

neiljp: I've updated this comment accordingly, and changed the code to only render to the UI if it is ready. We could really do with an analysis of the code to understand if we can simplify this, as this kind of check exists elsewhere. I suspect it will be easier if the loading screen is in `urwid`, as the UI loads first then in any case.

neiljp:
@preetmishra This was reference on #**zulip-terminal** - we could add a test-notifications option to make this easier, and potentially add that to the message we show in this popup.

neiljp:
The latest version is a little different, but should have roughly the same behavior but but look better, and is now ready for a proper review ;)

Tip: I rename my fetched PRs so that when I grab the next one I can do a diff between iterations :)

neiljp:
@preetmishra Regarding the 'keep doing it', I'm open to changing this in the future, but the main priority here was to a) remove the traceback b) still inform the user c) not inform the user on every failure! This should allow ZT to stay running, and the problem to be 'fixed' in the background, and notifications just start working. 

If anyone else would like to weigh in on this, please do :)

zulipterminal/model.py
sumanthvrao: Thanks for the quick update. Yes, we should simplify this.

PR #637, opened 2020-04-30
linting: Switch to check line break before binary operator (W504) & for presence of continuation characters

neiljp:
This commit slightly adjusts the layout of many files, with other minor
adjustments to:
* combine quoted strings on adjacent lines
* simplify some conditionals using .get()
* return directly the result of boolean expressions
* generally adjust indentation/ordering for readability & linting

An example of the change in acceptable behavior is from (accepting W503, ignoring W504):

    (foo and
     bar)

to (accepting W504, ignoring W503):

    (foo
     and bar)

While both are acceptable according to PEP8, the latter is now recommended:
https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,8 +33,8 @@ ignore=
   E123,
   # E126 - continuation line over-indented for hanging indent
   E126,
-  # W504 - line break after binary operator - we have this currently, vs W503
-  W504,
+  # W503 - line break before binary operator - new PEP8 best practice, cf W504
+  W503,
   # E241 - Multiple spaces after ',' - breaks layouts like themes.py
   E241,
   # E252 - Spaces in parameters, a style deliberately avoided
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -254,10 +254,8 @@ class Model:
         existing_reactions = [
             reaction['emoji_code']
             for reaction in message['reactions']
-            if ('user_id' in reaction['user'] and
-                reaction['user']['user_id'] == self.user_id) or
-               ('id' in reaction['user'] and
-                reaction['user']['id'] == self.user_id)
+            if (reaction['user'].get('user_id', None) == self.user_id or
+                reaction['user'].get('id', None) == self.user_id)
         ]
         if reaction_to_toggle_spec['emoji_code'] in existing_reactions:
             response = self.client.remove_reaction(reaction_to_toggle_spec)
@@ -610,8 +608,7 @@ class Model:
         Handle changes in subscription (eg. muting/unmuting streams)
         """
         if hasattr(self.controller, 'view'):
-            if ('property' in event and
-                    event['property'] == 'in_home_view'):
+            if event.get('property', None) == 'in_home_view':
                 stream_id = event['stream_id']
 
                 # FIXME: Does this always contain the stream_id?
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -499,8 +499,8 @@ class Model:
                             if aggregate_status != 'active':
                                 aggregate_status = status
                         if status == 'offline':
-                            if aggregate_status != 'active' and\
-                                    aggregate_status != 'idle':
+                            if (aggregate_status != 'active'
+                                    and aggregate_status != 'idle'):
                                 aggregate_status = status
 
                 status = aggregate_status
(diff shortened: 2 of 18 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -747,8 +747,10 @@ def topics_view(self, stream_button: Any) -> Any:
         return w
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
-        if (is_command_key('SEARCH_STREAMS', key) or
-                is_command_key('SEARCH_TOPICS', key)):
+        if (
+            is_command_key('SEARCH_STREAMS', key)
+            or is_command_key('SEARCH_TOPICS', key)
+           ):
```
kaustubh-nair, line 753: Is the bracket change supposed to happen?

zulipterminal/ui_tools/boxes.py
```diff
@@ -724,8 +727,10 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         elif is_command_key('TOGGLE_NARROW', key):
             self.model.unset_search_narrow()
             if self.message['type'] == 'private':
-                if (len(self.model.narrow) == 1 and
-                        self.model.narrow[0][0] == 'pm_with'):
+                if (
+                    len(self.model.narrow) == 1
+                    and self.model.narrow[0][0] == 'pm_with'
+                   ):
```
preetmishra, line 733: A compact alternative could be:

```python
if len(self.model.narrow) == 1 \
   and self.model.narrow[0][0] == 'pm_with':
```

zulipterminal/ui_tools/views.py
preetmishra: Same as the other 'alternative' in-line comment.

setup.cfg
```diff
@@ -35,8 +35,8 @@ ignore=
   E126,
   # E226 - Missing whitespace around arithmetic - FIXME possibly change this
   E226,
-  # W504 - line break after binary operator - we have this currently, vs W503
-  W504,
+  # W503 - line break after binary operator - we have this currently, vs W504
```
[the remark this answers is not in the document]
neiljp: Good catch - late night commenting!

zulipterminal/ui_tools/boxes.py
neiljp: Right - I'm trying to avoid continuation characters.

I'm not fond of the single-line parentheses, but I do think they're cleaner.

Also see https://www.python.org/dev/peps/pep-0008/#maximum-line-length

zulipterminal/ui_tools/views.py
neiljp: This is the same feedback as from @preetmishra ?

neiljp:
Regarding the continuation character/parentheses comments, I remember I was wondering if allowing is_command_key to take an Iterator (set, list, etc) in addition (or instead), to allow syntax like `is_command_key({'SEARCH_STREAMS', 'SEARCH_TOPICS'}, key)`, would be worthwhile.

I actually just saw ui.py - line 152 - which due to the continuation character, flake8 missed the binary operator, presumably!

zulipterminal/ui_tools/views.py
neiljp: See my comment below re continuation character vs parentheses.

Also possible alternative style with iterable passed to `is_command_key`... and how flake8 missed some things due to the presence of continuation characters!

preetmishra:
Re `is_command_key`, I believe we have multiple use-cases for the 'iterator' support. It would be worthwhile to add that syntax support.

Right, that's a bit awkward, this means flake8 would allow the 'unwanted' style if we use continuation characters.

neiljp:
@preetmishra I've found a lot more of cases breaking this change, due to continuations. Just looking at them now.

sumanthvrao:
Like this new Linting rule 👌   ! PR LGTM

neiljp:
Please re-review - it should be fairly simple, but needs attention to detail :) 

Latest push includes a few separate refactoring commits as general improvements, but also to make the linting changes cleaner.

In addition, since the original flake8 change didn't find the changes that were hidden by continuation characters, I've added `flake8-continuation` as another plugin to enforce that we don't use them, after adjusting the code to not use them in the prior commit.

zulipterminal/model.py
```diff
@@ -778,9 +780,9 @@ def _handle_update_message_event(self, event: Event) -> None:
             # 'subject' is not present in update event if
             # the event didn't have a 'subject' update.
             if 'subject' in event:
+                new_subject = event['subject']
```
preetmishra, line 783: I like this change to reduce dictionary lookups. :+1:

neiljp: This might make a marginal difference if there are multiple message ids in the event; this is mainly for readability and also to some degree having it fit on one line easily ;)

Issue #528, opened 2020-03-08
Integrate Zulint into Zulip Terminal.


amanagr:
Zulip Terminal uses linters like `isort`, `mypy`, `pep8` and pytest for running tests. Integrate all into Zulint and add it to Zulip Terminal.

preetmishra:
@amanagr Great! I was going to bring this up in #zulip-terminal, my idea was to add a section in the `README` where we would explicitly encourage contributors to run `isort`, `mypy`, `pycodestyle` along with `pytest` locally before pushing their changes but Zulint seems more plausible. This will greatly help new contributors to make better commits.

neiljp:
@amanagr I favour bringing zulint into the project, but at least initially my preference would be to use it for additional linting.

We could then discuss integrating it with the others later. I enjoy seeing different linting pass/fail in CI, and if zulint supports splitting linting elements up I feel that integration could be useful, though I'm not sure if including pytest fits.
 
This is related to #508 (meta-linters)

@preetmishra There is some text there regarding this, but if you have ideas of how to improve it, please discuss or propose a change :)

preetmishra:
The text in the README is pretty descriptive already though I think it might be useful to add a redirect link to the README (or maybe add a section) in the message that @zulipbot comments whenever someone claims any issue for the first time.

zulipbot:
Hello @amanagr, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

neiljp:
This is being most recently worked on in #980.

Issue #508, opened 2020-02-22
Other linters/fixers or tools to improve code style/consistency


neiljp:
We currently have a reasonable linting setup to run locally and in CI, so this is not highly important,  but worth tracking - and discussing on #zulip-terminal. See the `Linting/formatting` topic for some past discussion.

Note that there are linters and (re/auto)-formatters, and not all tools do both, or cover the same feature-set.

Tools to consider so far:
* **autopep8** (reformats based on pycodestyle rules)
* **zulint** - @amanagr helped to extract the linting tool from the server repository into an external project (https://github.com/zulip/zulint) last Sumemr; this can incorporate external linters including pycodestyle
* **flake8** (and plugins, eg. bugbear)
* **black** (auto formatter) - many projects are using this, but it notoriously lacks configuration!
* **pylint** - this is very configurable, but can be slow?

neiljp:
`autopep8`, `autoflake` and `flake8` are now dev dependencies, although only the latter is in CI as the others are code fixers.

We also use flake8-continuation and flake8-quotes.

There has also been talk of using `codespell` in the zulip/zulip repo, and I've applied it on occasion, eg. the updated merged commit in #881, but also notes in the former that it may be not so simple to add to CI.

timabbott:
Now that the server project has moved to black, it's probably a good idea to migrate this too.

@andersk may be able to help with tooling for making it easy to rebase PRs.

neiljp:
We now use a lot of standard linting and fixing tools that are well integrated, including most of the above tools or faster versions, so I'm closing this issue in favor of any issues for specific improvements.

PR #640, opened 2020-05-01
tools/lister: Remove six dependency

andersk:


neiljp:
@andersk I can see this is a legacy import that we would never need, except for potentially keeping sync with the originally imported version, but did you use a tool to identify this change or just happen to notice?

andersk:
It was found by [pyupgrade](https://github.com/asottile/pyupgrade/). I wouldn’t have bothered to PR it, except we don’t declare a dependency on `six`, so it could fix an `ImportError`.

timabbott:
@amanagr why do we still have a copy of `lister` in this project?  We should rework this project to just use `zulint` (or `lister` from there, if needed).

neiljp:
@timabbott `lister` is used by the current version of `run-mypy` - which zulip now imports from `zulint`. We haven't fully discussed next stages to integrate zulint, but see #508 & #528.

I was looking at the dependencies, and while some dev dependencies require it - so it shouldn't be an issue for an install, currently, assuming those don't change - apparently `zulip` depends on it still too? While ZT doesn't depend on the latest version, even that still has dependencies upon `six`, according to the package:
https://pypi.org/project/zulip/

So, see #zulip/python-zulip-api#597 for a remedy for the `zulip` issue.

neiljp:
I was leaving this tiny PR open for responses, but will merge it now, thanks @andersk :+1: Feel free to comment on the zulint issues/pulls or discuss elsewhere.

PR #643, opened 2020-05-03
bugfix: TopicsView: Focus Frame header (search box) on search.

neiljp:
Previously the focus was moved into the search box part of the Frame
header, but the header itself was not focused. The search worked, but
entries remained selected in the topic list (body).

Test added.


neiljp:
I'm merging this now, since it's a small independent fix which improves the UX to what is expected.

For reference, this issue was mentioned by @preetmishra at the end of #641.

PR #644, opened 2020-05-03
Render ordered lists

neiljp:
This is a followup to #634 - not only does it render numbered lists, but since it handles them explicitly then we can remove extra newlines in that case, as we did in the other PR.

Numbering starts with the first number specified in the list, as per the Zulip revised scheme:
https://github.com/zulip/zulip/blob/master/docs/subsystems/markdown.md#lists

soup2markup now has a kw-args parameter 'state', in which a 'list_index'
entry is set by the ordered-lists code to have an index, incremented
through each list item, starting at the value specified in the 'start'
attribute of the html ordered list (ol).

Tests added.


neiljp:
@sumanthvrao Thanks for looking - I'm glad to have this in :)

Issue #552, opened 2020-03-21
Show search-key & prefix in search boxes if search is enabled


neiljp:
We conveniently show the keys to jump to search boxes, and when typing text it should be clear that this is filtering the results, or maybe just after pressing 'enter', also hopefully obviously.

However, once the search is enabled, ie. the list of items are filtered, we should probably show the key at the end of the box again, though the actual search text of course will not include this key text. This makes it clear to users that they can go back to the search.

It would also be good to show that a search is 'in progress'/'enabled', by prefixing some text. This is similar to the approach taken in the message list when searching/filtering messages.

kaustubh-nair:
@zulipbot claim

kaustubh-nair:
@zulipbot abandon
Will take it up after finishing current prs

PR #645, opened 2020-05-04
Indicate panel searches in progress

neiljp:
This applies the styling in the message-view which indicates when a search is in progress, to the streams, topics and users views - making it visually clearer when the values shown in the view (being listed) are filtered, ie. a search result. This goes partially toward fixing #552.

GIven there were no tests for the `PanelSearchBox`, modified for this feature, this PR also includes a commit adding tests for this class first.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -1,9 +1,10 @@
 import pytest
 
-from zulipterminal.ui_tools.boxes import WriteBox
+from zulipterminal.config.keys import keys_for_command
+from zulipterminal.ui_tools.boxes import PanelSearchBox, WriteBox
 
 
-WRITEBOX = "zulipterminal.ui_tools.boxes.WriteBox"
+BOXES = "zulipterminal.ui_tools.boxes"
 
 
 class TestWriteBox:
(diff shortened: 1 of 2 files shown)


tests/ui_tools/test_boxes.py
```diff
@@ -146,3 +147,74 @@ def test_generic_autocomplete_emojis(self, write_box, emojis_fixture, text,
...
+
+        # Leave editor mode
+        assert panel_search_box.panel_view.view.controller.editor_mode is False
+
+        # Switch focus to body; focus should return to previous in body
+        panel_search_box.panel_view.set_focus.assert_called_once_with("body")
+
+        # pass keypress back  # FIXME Should just do a callback?
+        (panel_search_box.panel_view.keypress
+         .assert_called_once_with(size, 'esc'))
```
preetmishra, line 229: This seems okay (reply to the FIXME comment).

tests/ui_tools/test_boxes.py
```diff
@@ -1,9 +1,10 @@
 import pytest
 
-from zulipterminal.ui_tools.boxes import WriteBox
+from zulipterminal.config.keys import keys_for_command
+from zulipterminal.ui_tools.boxes import PanelSearchBox, WriteBox
 
 
-WRITEBOX = "zulipterminal.ui_tools.boxes.WriteBox"
+BOXES = "zulipterminal.ui_tools.boxes"
```
preetmishra, line 7: :+1:

tests/ui_tools/test_boxes.py
```diff
@@ -202,11 +209,13 @@ def test_keypress_ENTER(self, panel_search_box,
     def test_keypress_GO_BACK(self, panel_search_box, back_key):
         size = (20,)
         panel_search_box.panel_view.view.controller.editor_mode = True
+        panel_search_box.set_caption("Search Results")
```
preetmishra, line 212: I wonder if we should use the 'right-triangle symbol' here as well?

neiljp: I think my concern with this is that this is associated with a stream-topic separation. I'm looking at safe unicode and wondering about ⛏ instead.

tests/ui_tools/test_boxes.py
neiljp: I should have made this clearer:
* 'esc' could be generalized
* maybe we should trigger the search-panel update a different way than faking a keypress.

Is that clearer? I can update the FIXME.

neiljp:
I've just pushed an updated version with a more explanatory FIXME comment and with a 'PICK' emoji (hopefully signifying digging for results!) added - also for the message search text too.

tests/ui_tools/test_boxes.py
preetmishra: This Unicode character (emoji) blends in with the background and is hard to distinguish.

image

image

tests/ui_tools/test_boxes.py
preetmishra: With the recent push, it is clearer.

As far as `PanelSearchBox` is concerned, it is only supposed to trigger the `panel_view`'s keypress method with `esc` (or `GO_BACK`) key. I think this callback is sufficient.

neiljp: To clarify, I'm not saying this doesn't work, just that it's a bit fragile and might be clearer expressed a different way.

tests/ui_tools/test_boxes.py
neiljp: I only see the non-color (non-emoji) version of this, which doesn't have the same problem - it's just a white pick on green-ish background; see:
http://www.unicode.org/emoji/charts/emoji-variants.html#26cf

I've explored whether this can be expressed using a variant selector / presentation sequence, but it doesn't seem to change the output for me, and I can't find clear examples of how this is used in python.

I'm going to remove the pick for now, and we can explore this separately if we want to, possibly after discussion on the stream.

neiljp:
Updated with the pick removed due to the above issue (we can handle this later).

tests/ui_tools/test_boxes.py
```diff
@@ -179,11 +184,15 @@ def test_keypress_ENTER(self, panel_search_box,
...
         panel_search_box.panel_view.view.controller.editor_mode = True
         panel_search_box.panel_view.log = log
+        panel_search_box.set_caption("")
         panel_search_box.edit_text = "key words"
 
         panel_search_box.keypress(size, enter_key)
 
         # Update this display
+        # We can't test for the styled version?
+        # expected_caption = [('filter_results', 'Search Results'), ' ']
```
sumanthvrao, line 194: Just to clarify, did you miss this (or) is this on purpose?

neiljp: I left this in as a note - perhaps I should change to FIXME - but couldn't find an obvious way to test for the applied style of the caption.

neiljp: Adjusted slightly to indicate this.

neiljp:
I just pushed with a slight adjustment to the comment that @sumanthvrao pointed out, but otherwise plan to merge this later today.

PR #649, opened 2020-05-05
Render nested lists

neiljp:
This builds upon #644 (the first commit is that PR) and renders nested lists with appropriate indents, and different bullets for different indents.

This was pending a fix to the bugdown rendering in the server, as fixed by @aero31aero recently.

We could include many other tests for lists (mixed order & unordered), but for now this seems a good start - though if you want specific cases to be included, please mention it or feel free to generate some :)


preetmishra:
@neiljp Thanks for the working on this. :+1: I had a quick look. The two things that strike out to me are:

* For some tests (e.g. https://chat.zulip.org/#narrow/stream/7-test-here/topic/%28no.20topic%29/near/868288), there are unexpected newlines between the list. Is this an expected behaviour?
image

* Shouldn't the bullet characters for 'd'/'dd' and 'e' be in reverse order?
image

neiljp:
@preetmishra You'll have to indicate the bugdown source for this, as I don't know precisely which cases you're referring to.

For the first case, my understanding is that this is a bug in the rendering (or how markdown is defined, even) that just happens to appear right on the server, for these "mixed" lists, ie. where sub-lists are of different types. I'm guessing your source is:
```
* This
  1. a
    * here
    * there
  2. b
    2. that
* That
```
On the contrary, the following does not render right on either:
```
* This
  1. a
    3. here
    3. there
  2. b
     * that
* That
```
 Maybe @aero31aero could weigh in here.

For the rendering in the second case, note that it's cycling through 3 symbols. Your screenshot indicates these are very similar, but for me they are different. On that basis we should probably choose other symbols - what do you think of `HYPHEN` or `EN DASH` as alternatives to `BULLET OPERATOR`? The first one looks better to me; try substituting it in `boxes.py` and see what you think.

preetmishra:
@neiljp Indeed, the first looks like an issue at the server end.

Re symbols, `HYPHEN` looks better in comparison with `EN DASH` (it is too long?).

neiljp:
@preetmishra Thanks for taking another look at this - it really does make looking at lists so much nicer, I think :)

I'm going to merge this now, given that it resolves the bug we discussed in #**zulip-terminal** as a side-effect.

PR #653, opened 2020-05-08
Fetch and indicate server version and feature level 

### LINKED ISSUE #616: Fetch & indicate server version (& 'feature level')
It would be useful to show the server version - whether in the title or in some 'about' type box?

The side effect of us fetching this data, is that it should also make available the `zulip_feature_level`, as added in this commit: https://github.com/zulip/zulip/commit/e3b90a5ec86e426f4fa27e372e47655428927ca5

That in turn should make handling newer features, and migration of API versions, easier over time.

preetmishra:
This introduces `AboutView` to show server version and feature level within the app and makes other related amendments.

Fixes #616.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,10 @@ source ~/.zshenv
 
 **Note:** You can use `arrows`, `home`, `end`, `Page up` and `Page down` keys to move around in Zulip-Terminal.
 
+## Supported Server Versions
+
+The minimum server version that Zulip Terminal supports is [`2.1.0`](https://zulip.readthedocs.io/en/latest/overview/changelog.html#id7). Though, it may still work with earlier versions.
+
 ## Contributor Guidelines
 
 Zulip Terminal is being built by the awesome [Zulip](https://zulip.com/team) community.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,7 @@ from zulipterminal.config.keys import keys_for_command
 from zulipterminal.helper import initial_index as helper_initial_index
 from zulipterminal.ui_tools.boxes import MessageBox
 from zulipterminal.ui_tools.buttons import StreamButton, UserButton
+from zulipterminal.version import MINIMUM_SUPPORTED_SERVER_VERSION
 
 
 @pytest.fixture(autouse=True)
@@ -467,6 +468,10 @@ def initial_data(logged_on_user, users_fixture, streams_fixture):
         'last_event_id': -1,
         'muted_topics': [],
         'realm_user_groups': [],
+        # Deliberately use hard-coded zulip version and feature level to avoid
+        # adding extra tests unnecessarily.
+        'zulip_version': MINIMUM_SUPPORTED_SERVER_VERSION[0],
+        'zulip_feature_level': MINIMUM_SUPPORTED_SERVER_VERSION[1],
     }
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,7 +8,9 @@ from zulipterminal.config.keys import keys_for_command
 from zulipterminal.helper import initial_index as helper_initial_index
 from zulipterminal.ui_tools.boxes import MessageBox
 from zulipterminal.ui_tools.buttons import StreamButton, UserButton
-from zulipterminal.version import MINIMUM_SUPPORTED_SERVER_VERSION
+from zulipterminal.version import (
+    MINIMUM_SUPPORTED_SERVER_VERSION, SUPPORTED_SERVER_VERSIONS,
+)
 
 
 @pytest.fixture(autouse=True)
@@ -311,6 +313,19 @@ def messages_successful_response() -> Dict[str, Any]:
         'result': 'success',
         'msg': '',
     })
+
+
+@pytest.fixture(
+    params=SUPPORTED_SERVER_VERSIONS,
+    ids=(lambda param:
+         'server_version:{}-server_feature_level:{}'.format(*param)),
+)
+def zulip_version(request):
+    """
+    Fixture to test different components based on the server version and the
+    feature level.
+    """
+    return request.param
 
 
 @pytest.fixture
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -53,6 +53,10 @@ class TestModel:
                                                    num_after=10,
                                                    anchor=None)
         assert model.initial_data == initial_data
+        assert model.server_version == initial_data['zulip_version']
+        assert model.server_feature_level == (
+            initial_data.get('zulip_feature_level')
+        )
         assert model.user_id == user_profile['user_id']
         assert model.user_full_name == user_profile['full_name']
         assert model.user_email == user_profile['email']
@@ -136,6 +140,7 @@ class TestModel:
             'muted_topics',
             'realm_user',
             'realm_user_groups',
+            'zulip_version',
         ]
         model.client.register.assert_called_once_with(
                 event_types=event_types,
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -17,6 +17,7 @@ from zulipterminal.ui_tools.views import (
     ModListWalker, MsgInfoView, PopUpConfirmationView, PopUpView,
     RightColumnView, StreamInfoView, StreamsView, TopicsView, UsersView,
 )
+from zulipterminal.version import MINIMUM_SUPPORTED_SERVER_VERSION, ZT_VERSION
 
 
 VIEWS = "zulipterminal.ui_tools.views"
@@ -1246,8 +1247,11 @@ class TestAboutView:
         mocker.patch.object(self.controller, 'maximum_popup_dimensions',
                             return_value=(64, 64))
         mocker.patch(VIEWS + '.urwid.SimpleFocusListWalker', return_value=[])
+        server_version, server_feature_level = MINIMUM_SUPPORTED_SERVER_VERSION
         self.about_view = AboutView(self.controller, 'About',
-                                    zt_version='0.5.1+git')
+                                    zt_version=ZT_VERSION,
+                                    server_version=server_version,
+                                    server_feature_level=server_feature_level)
 
     @pytest.mark.parametrize('key', {*keys_for_command('GO_BACK'),
                                      *keys_for_command('ABOUT')})
@@ -1269,6 +1273,21 @@ class TestAboutView:
         super_keypress = mocker.patch(VIEWS + '.urwid.ListBox.keypress')
         self.about_view.keypress(size, key)
         super_keypress.assert_called_once_with(size, expected_key)
+
+    def test_feature_level_content(self, mocker, zulip_version):
+        self.controller = mocker.Mock()
+        mocker.patch.object(self.controller, 'maximum_popup_dimensions',
+                            return_value=(64, 64))
+        mocker.patch(VIEWS + '.urwid.SimpleFocusListWalker', return_value=[])
+        server_version, server_feature_level = zulip_version
+
+        about_view = AboutView(self.controller, 'About', zt_version=ZT_VERSION,
+                               server_version=server_version,
+                               server_feature_level=server_feature_level)
+
+        assert len(about_view.feature_level_content) == (
+            1 if server_feature_level else 0
+        )
 
 
 class TestPopUpConfirmationView:
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -98,6 +98,11 @@ class Model:
         # lose any updates while messages are being fetched.
         self._update_initial_data()
 
+        self.server_version = self.initial_data['zulip_version']
+        self.server_feature_level = (
+            self.initial_data.get('zulip_feature_level')
+        )
+
         self.users = self.get_all_users()
 
         subscriptions = self.initial_data['subscriptions']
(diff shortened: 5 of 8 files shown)


zulipterminal/model.py
```diff
@@ -119,6 +119,17 @@ def __init__(self, controller: Any) -> None:
         self.new_user_input = True
         self._start_presence_updates()
 
+    def get_versions(self) -> Dict[str, str]:
```
neiljp, line 122: We shouldn't need to explicitly fetch this information. You can check the new https://chat.zulip.org/api/changelog, or `git checkout` various versions of the source, but this can be requested to be (and later will be included by default) in the initial data sent by the server.

zulipterminal/model.py
```diff
@@ -119,6 +119,17 @@ def __init__(self, controller: Any) -> None:
...
+        """
+        Returns feature level and server version.
+        """
+        settings = self.client.get_server_settings()
+        versions = dict()
+        if settings['result'] == 'success':
+            versions['feature_level'] = str(settings['zulip_feature_level'])
```
neiljp, line 129: In the original issue, I was not originally intending that we should show the feature level, as it's just an integer and so not necessarily useful, but *if* it's present (see my note on 'supported server versions` in #**zulip-terminal**) we could include it.

zulipterminal/ui_tools/views.py
```diff
@@ -806,6 +806,30 @@ def __init__(self, controller: Any,
         super().__init__(controller, widgets, 'GO_BACK')
 
 
+class AboutView(PopUpView):
+    def __init__(self, controller: Any, versions: Dict[str, str]) -> None:
+        content = [
+            ('Feature level', versions['feature_level']),
+            ('Server version', versions['server_version']),
+        ]
```
neiljp, line 814: This box seems like a great place to include the running version number of ZT? That might make it easier to split this commit into pieces - adding an about box, then adding various features to it?

sumanthvrao: Yeah This is a nice idea! 👌

sumanthvrao:
One point I noticed was that focus by default was on the last row of the popup; However the up/down arrows didn't work. This makes me wonder if we don't to remove marking anything as 'focus' in this popup, since we may not show enough data for it to be scrollable at any point.

preetmishra:
@sumanthvrao The keys don't work because there's nothing to scroll in the about popup; they will work fine if there's more data. Were you referring to `popup_contrast` (markup) when you said:
> One point I noticed was that focus by default was on the last row of the popup

README.md
```diff
@@ -111,6 +111,7 @@ source ~/.zshenv
 | Command                                               | Key Combination                               |
 | ----------------------------------------------------- | --------------------------------------------- |
 | Show/hide help menu                                   | <kbd>?</kbd>                                  |
+| Show/hide about menu                                  | <kbd>a</kbd>                                  |
```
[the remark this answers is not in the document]
preetmishra: I agree. How does `meta ?` sound?

zulipterminal/ui_tools/views.py
preetmishra: That sounds plausible, thanks! :+1:

preetmishra:
@sumanthvrao @neiljp Thank you for the reviews and the suggestions. I have incorporated the suggestions and split the commits into two parts:
* the first introduces an about menu in the app, using `AboutView`, to show `ZT_VERSION`
* the second adds server version and feature level in the about menu.

zulipterminal/ui_tools/views.py
```diff
@@ -806,6 +806,29 @@ def __init__(self, controller: Any,
...
+        self.width = sum(max_widths) + 2  # Add +2 to recoup `dividechars=2`.
+
+        widgets = []
+        for index, (label, data) in enumerate(content):
+            widgets.append(urwid.AttrWrap(
+                urwid.Columns([(max_widths[0], urwid.Text(label)),
+                               (max_widths[1], urwid.Text(data))],
+                              dividechars=2),
+                None if index % 2 == 0 else 'popup_contrast'
+            ))
```
neiljp, line 830: In this commit we don't have more than one line, so including this alternating color code in this first commit is a bit confusing? Also the other multi-line capabilities. This may be less relevant if we ended up merging the other PR first.

zulipterminal/ui_tools/views.py
```diff
@@ -807,8 +807,12 @@ def __init__(self, controller: Any,
...
 class AboutView(PopUpView):
-    def __init__(self, controller: Any, ZT_VERSION: str) -> None:
-        content = [
+    def __init__(self, controller: Any, ZT_VERSION: str, server_version: str,
+                 server_feature_level: str) -> None:
+        content = ([('Server feature level', server_feature_level)]
+                   if server_feature_level else [])
+        content += [
+            ('Server version', server_version),
             ('Zulip Terminal version', ZT_VERSION),
```
neiljp, line 816: Would it be clearer to put the ZT version first?

If we move to categories based on the other PR, we could group them instead?

tests/conftest.py
```diff
@@ -467,6 +467,8 @@ def initial_data(logged_on_user, users_fixture, streams_fixture):
         'last_event_id': -1,
         'muted_topics': [],
         'realm_user_groups': [],
+        'zulip_version': '2.2-dev',
+        'zulip_feature_level': 1,
```
neiljp, line 471: This could be a good place to start exploring how we might use parametrization to encode variations per feature level and/or version. That is, our test data (and expectations) varies if the server covers a range of versions. This is already the case here, as the code is testing for the presence of `zulip_feature_level`, which we're not testing being present/absent.

tests/ui/test_ui_tools.py
```diff
@@ -1227,7 +1227,8 @@ class TestAboutView:
     def mock_external_classes(self, mocker):
         self.controller = mocker.Mock()
         mocker.patch(VIEWS + '.urwid.SimpleFocusListWalker', return_value=[])
-        self.about_view = AboutView(self.controller, '0.5.0+git')
+        self.about_view = AboutView(self.controller, '0.5.0+git', '2.2-dev',
+                                    '3')
```
neiljp, line 1231: Maybe test this with variations of server values?

zulipterminal/model.py
```diff
@@ -930,7 +934,8 @@ def _register_desired_events(self, *, fetch_data: bool=False) -> str:
             'update_message_flags',
             'muted_topics',
             'realm_user',  # Enables cross_realm_bots
-            'realm_user_groups'
+            'realm_user_groups',
+            'zulip_version'
```
neiljp, line 938: Also worth adding a comment that this is present by default from one particular `feature_level`?

preetmishra:
@neiljp Thanks for the review and the in-line suggestions. I have split the commits to tidy up the PR and incorporated the suggestion that you left. 

I have added a refactor commit before the commit that fetches zulip version and feature level. It was necessary to support the `zulip_version` fixture in `initial_data`  (a detailed explanation is in the commit description).

The `zulip_version` fixture allows testing how `model.py` behaves with different zulip versions.

zulipterminal/ui_tools/views.py
```diff
@@ -879,10 +879,14 @@ def __init__(self, controller: Any,
...
-        contents = [
-            ('Version', [('Zulip Terminal', zt_version)]),
+    def __init__(self, controller: Any, title: str, *, zt_version: str,
+                 server_version: str, server_feature_level: str) -> None:
+        contents = ([('', [('Server feature level', server_feature_level)])]
+                    if server_feature_level else [])
+
+        contents += [
+            ('Version', [('Zulip Terminal', zt_version),
+                         ('Server', server_version)]),
```
neiljp, line 889: I think it'd be better to group the server properties together; when I merged the first two commits I set the `Zulip Terminal` category to be in `Application`, and so I'd suggest we have a `Server` category with the extra details.

tests/model/test_model.py
```diff
@@ -1108,10 +1108,32 @@ def test__handle_reaction_event(self, mocker, model, response, index):
...
+          'emoji_name': 'thumbs_up',
+          'type': 'reaction',
+          'op': 'add'
+          }, {
+            'messages': {
+                1: {},
+            }
+          }),
+    ])
+    def test__handle_reaction_event_empty_index(self, mocker, model, response,
```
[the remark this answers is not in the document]
preetmishra: https://github.com/zulip/zulip-terminal/blob/49d3fa5244eab398fa2597156e7309f5bfa6c24c/tests/model/test_model.py#L1112

The line mentioned above mutates the index parameter (see commit description). This causes the test to fail (it raises a KeyError exception for 'reactions') with fixtures that make the test run more than once.

I have pushed a branch on my fork to demonstrate how the test fails. Link: https://github.com/preetmishra/zulip-terminal/commit/15a756a477fabb658b92ac26560d3f41afe208ec.

tests/conftest.py
```diff
@@ -312,6 +312,18 @@ def messages_successful_response() -> Dict[str, Any]:
     })
 
 
+@pytest.fixture(params=['', '1'])
+def zulip_feature_level(request):
```
preetmishra, line 316: While `zulip_feature_level` isn't used anywhere in this particular commit, this appeared to be a good place to introduce the fixture as this is where we start to fetch the server version and the feature level.

tests/ui/test_ui_tools.py
```diff
@@ -1270,6 +1272,18 @@ def test_keypress_navigation(self, mocker,
...
+        mocker.patch.object(self.controller, 'maximum_popup_dimensions',
+                            return_value=(64, 64))
+        mocker.patch(VIEWS + '.urwid.SimpleFocusListWalker', return_value=[])
+        about_view = AboutView(self.controller, 'About', zt_version='0.5.1',
+                               server_version='2.2-dev',
+                               server_feature_level=zulip_feature_level)
+        assert len(about_view.feature_level_content) == (
+            1 if zulip_feature_level else 0
+        )
+
```
preetmishra, line 1286: I have added this is an extra test to avoid running all the tests in `TestAboutView` for different feature levels with `zulip_feature_level` fixture (relating to the point about unnecessary extra tests that you mentioned in the review).

tests/ui/test_ui_tools.py
```diff
@@ -1247,7 +1247,9 @@ def mock_external_classes(self, mocker):
                             return_value=(64, 64))
         mocker.patch(VIEWS + '.urwid.SimpleFocusListWalker', return_value=[])
         self.about_view = AboutView(self.controller, 'About',
-                                    zt_version='0.5.1+git')
+                                    zt_version='0.5.1+git',
+                                    server_version='2.2-dev',
+                                    server_feature_level='1')
```
preetmishra, line 1252: I have used hard-coded values here as `self.about_view` is used in the other tests that don't 'care' about the version or the feature_level (see also the other in-line comment in this commit).

preetmishra:
@neiljp Thanks for the review and the merge. :tada: 

To avoid adding extra tests, I have hard-coded the value in `initial_data` (which means that the refactor commit won't be necessary anymore). In addition to that, I have added a new fixture, `zulip_feature_level`, which would make it easier to test against different feature levels. I believe it would be reasonable to only use this fixture for implementations that vary according to the feature level.

Moreover, I have moved the server version and the feature level under 'Server' category.

Furthermore, since the refactor commit isn't necessary anymore, I have moved it to the end. We could shelve it if you'd like to address it after reactions test refactor work.

preetmishra:
@neiljp Thanks for the review and the comments on CZO. I have updated the PR with `SUPPORTED_SERVER_VERSIONS` and `MINIMUM_SUPPORTED_SERVER_VERSION`. Additionally, I have amended the fixture to spit out version tuples and the code (with tests) in `AboutView`.

zulipterminal/version.py
```diff
@@ -1 +1,8 @@
 ZT_VERSION = '0.5.1+git'
+
+SUPPORTED_SERVER_VERSIONS = [
+    ('2.1', None),
+    ('2.2', 1),
```
preetmishra, line 5: Would 1, as a feature level, be sufficient for now or should I extend it till 14 (the latest feature level)?

neiljp: Any finite value seems ok for now, but we should add a comment given that 2.2 is not actually yet released!

preetmishra:
Updated with the comment about `2.2` and fixed commit order (GitHub).

zulipterminal/version.py
```diff
@@ -1 +1,8 @@
 ZT_VERSION = '0.5.1+git'
+
+SUPPORTED_SERVER_VERSIONS = [
+    ('2.1', None),
```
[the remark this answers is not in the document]
preetmishra: I have added a note in the README (README felt more appropriate than FAQ for a declaration).

Re notifying users, should I file an issue?

preetmishra:
@neiljp Thanks for the review. I have made the amendments accordingly.

kaustubh-nair:
This looks good for the use case that I need it for :tada:

preetmishra:
I have updated the references to `2.2` with `3.0` as per the recent discussion in [general > Next release 3.0](https://chat.zulip.org/#narrow/stream/2-general/topic/Next.20release.203.2E0/near/913436).

zulipterminal/version.py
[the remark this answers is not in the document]
neiljp: I think we could have an issue for this, but it's definitely a needs-discussion style issue, as we hadn't decided how to make it reasonable?

neiljp:
@preetmishra Thanks for working on this! I just merged this manually with a minor change to a commit title and moving the README change to a more prominent position - thanks for persevering, since I know this was more than we originally intended for this PR.

Anyhow, merged as last 3 commits ending at d9dc81c63868b8cfa45cf60163199cd4e97f0d20 :tada:

PR #655, opened 2020-05-11
Shrink space to left of message by using unicode to mark edited messages.

neiljp:
By moving from 'EDITED' to '✎……' we avoid having a distracting label to
the left of the message, and can shrink wasted space to the left of the
message.

This commit reduces the space to the left by 3.

Based on a suggestion by Tim Abbott.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1889,8 +1889,8 @@ class TestMessageBox:
         assert len(view_components) == 1
         assert isinstance(view_components[0], Padding)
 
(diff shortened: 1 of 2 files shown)


preetmishra (commented):
@neiljp This does help in shrinking the space and allowing more screen real estate for messages.  :+1: 

Another alternative worth considering could be `\N{WRITING HAND}`.

neiljp:
This was discussed in #**zulip-terminal>EDITED marking #T655** on chat.zulip.org

PR #656, opened 2020-05-11
Improve monochrome rendering of narrows & streams.

neiljp:
I noticed this while testing the thin-colorer-barbranch.

Prior to this commit:
* narrow text (excluding topics) was plain, with topic names in
  standout; this commit broadly inverts this, setting non-topics in
  standout and topics in bold, to produce the expected visual ordering
* stream '#'/'P' marks and unreads were visually undistinguishable from
  the stream name; this commit sets these all as bold


neiljp:
@preetmishra Thanks for taking a look - it's pretty small and makes it more usable, and I was just going to merge, so good to hear you feel the same way :+1:

Issue #658, opened 2020-05-14
Stream names don't get rendered for streams that don't have any description


preetmishra:
To reproduce:
* Navigate to any stream that doesn't have any description (e.g. 'Django')
* Press `i`

preetmishra:
@zulipbot "bug"

preetmishra:
@zulipbot add "bug"

PR #659, opened 2020-05-14
views: Refactor popups to expedite the popup creation process.

### LINKED ISSUE #654: Tidy up white space in StreamInfoView
The mentioned popup/view show a blank line after the content which isn't as pleasing to look at. We could either remove the blank line or add a blank line at the top for consistency.

For comparison, look at the help menu (it doesn't have any unnecessary blank lines).

preetmishra:
The intention is to extract and generalize duplicate code to expedite the popup creation process.

Along with that, it tidies up popups, that derive from PopUpView, by calculating their required width and height based on the content instead of using hard-coded values.

Tests amended.

As a side effect, fixes #654 and #658.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1154,10 +1154,13 @@ class TestPopUpView:
     @pytest.fixture(autouse=True)
     def pop_up_view(self, mocker):
         self.controller = mocker.Mock()
+        mocker.patch.object(self.controller, 'maximum_popup_dimensions',
+                            return_value=(64, 64))
         self.command = 'COMMAND'
         self.title = 'Generic title'
         self.width = 16
         self.widget = mocker.Mock()
+        mocker.patch.object(self.widget, 'rows', return_value=1)
         self.widgets = [self.widget, ]
         self.list_walker = mocker.patch(VIEWS + '.urwid.SimpleFocusListWalker',
                                         return_value=[])
@@ -1208,6 +1211,8 @@ class TestHelpMenu:
     @pytest.fixture(autouse=True)
     def mock_external_classes(self, mocker, monkeypatch):
         self.controller = mocker.Mock()
+        mocker.patch.object(self.controller, 'maximum_popup_dimensions',
+                            return_value=(64, 64))
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
         self.help_view = HelpView(self.controller, 'Help Menu')
 
@@ -1279,6 +1284,8 @@ class TestStreamInfoView:
     @pytest.fixture(autouse=True)
     def mock_external_classes(self, mocker, monkeypatch):
         self.controller = mocker.Mock()
+        mocker.patch.object(self.controller, 'maximum_popup_dimensions',
+                            return_value=(64, 64))
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
         self.stream_info_view = StreamInfoView(self.controller, color='',
                                                desc='', title='# stream-name')
@@ -1303,6 +1310,8 @@ class TestMsgInfoView:
     @pytest.fixture(autouse=True)
     def mock_external_classes(self, mocker, monkeypatch, message_fixture):
         self.controller = mocker.Mock()
+        mocker.patch.object(self.controller, 'maximum_popup_dimensions',
+                            return_value=(64, 64))
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
         self.msg_info_view = MsgInfoView(self.controller, message_fixture,
                                          'Message Information')
@@ -1321,7 +1330,7 @@ class TestMsgInfoView:
         assert self.controller.exit_popup.called
 
     def test_height_noreactions(self):
-        expected_height = 5
+        expected_height = 4
         assert self.msg_info_view.height == expected_height
 
     # FIXME This is the same parametrize as MessageBox:test_reactions_view
@@ -1368,7 +1377,8 @@ class TestMsgInfoView:
         varied_message = dict(message_fixture, **to_vary_in_each_message)
         self.msg_info_view = MsgInfoView(self.controller, varied_message,
                                          'Message Information')
-        expected_height = 8
+        # 7 = 3 labels + 4 reactions.
+        expected_height = 7
         assert self.msg_info_view.height == expected_height
 
     def test_keypress_navigation(self, mocker,
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1155,6 +1155,7 @@ class TestPopUpView:
     def pop_up_view(self, mocker):
         self.controller = mocker.Mock()
         self.command = 'COMMAND'
+        self.title = 'Generic title'
         self.width = 16
         self.widget = mocker.Mock()
         self.widgets = [self.widget, ]
@@ -1163,11 +1164,12 @@ class TestPopUpView:
         self.super_init = mocker.patch(VIEWS + '.urwid.ListBox.__init__')
         self.super_keypress = mocker.patch(VIEWS + '.urwid.ListBox.keypress')
         self.pop_up_view = PopUpView(self.controller, self.widgets,
-                                     self.command, self.width)
+                                     self.command, self.width, self.title)
 
     def test_init(self):
         assert self.pop_up_view.controller == self.controller
         assert self.pop_up_view.command == self.command
+        assert self.pop_up_view.title == self.title
         assert self.pop_up_view.width == self.width
         self.list_walker.assert_called_once_with(self.widgets)
         self.super_init.assert_called_once_with(self.pop_up_view.log)
@@ -1207,7 +1209,7 @@ class TestHelpMenu:
     def mock_external_classes(self, mocker, monkeypatch):
         self.controller = mocker.Mock()
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
-        self.help_view = HelpView(self.controller)
+        self.help_view = HelpView(self.controller, 'Help Menu')
 
     def test_keypress_any_key(self):
         key = "a"
@@ -1278,7 +1280,8 @@ class TestStreamInfoView:
     def mock_external_classes(self, mocker, monkeypatch):
         self.controller = mocker.Mock()
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
-        self.stream_info_view = StreamInfoView(self.controller, '', '', '')
+        self.stream_info_view = StreamInfoView(self.controller, color='',
+                                               desc='', title='# stream-name')
 
     @pytest.mark.parametrize('key', {*keys_for_command('GO_BACK'),
                                      *keys_for_command('STREAM_DESC')})
@@ -1301,7 +1304,8 @@ class TestMsgInfoView:
     def mock_external_classes(self, mocker, monkeypatch, message_fixture):
         self.controller = mocker.Mock()
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
-        self.msg_info_view = MsgInfoView(self.controller, message_fixture)
+        self.msg_info_view = MsgInfoView(self.controller, message_fixture,
+                                         'Message Information')
 
     def test_keypress_any_key(self):
         key = "a"
@@ -1316,8 +1320,7 @@ class TestMsgInfoView:
         self.msg_info_view.keypress(size, key)
         assert self.controller.exit_popup.called
 
-    def test_height_noreactions(self, message_fixture):
-        self.msg_info_view = MsgInfoView(self.controller, message_fixture)
+    def test_height_noreactions(self):
         expected_height = 5
         assert self.msg_info_view.height == expected_height
 
@@ -1363,7 +1366,8 @@ class TestMsgInfoView:
         ])
     def test_height_reactions(self, message_fixture, to_vary_in_each_message):
         varied_message = dict(message_fixture, **to_vary_in_each_message)
-        self.msg_info_view = MsgInfoView(self.controller, varied_message)
+        self.msg_info_view = MsgInfoView(self.controller, varied_message,
+                                         'Message Information')
         expected_height = 8
         assert self.msg_info_view.height == expected_height
 
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1330,7 +1330,7 @@ class TestMsgInfoView:
         assert self.controller.exit_popup.called
 
     def test_height_noreactions(self):
-        expected_height = 4
+        expected_height = 3
         assert self.msg_info_view.height == expected_height
 
     # FIXME This is the same parametrize as MessageBox:test_reactions_view
@@ -1377,8 +1377,8 @@ class TestMsgInfoView:
         varied_message = dict(message_fixture, **to_vary_in_each_message)
         self.msg_info_view = MsgInfoView(self.controller, varied_message,
                                          'Message Information')
-        # 7 = 3 labels + 4 reactions.
-        expected_height = 7
+        # 9 = 3 labels + 1 blank line + 1 'Reactions' (category) + 4 reactions.
+        expected_height = 9
         assert self.msg_info_view.height == expected_height
 
     def test_keypress_navigation(self, mocker,
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1155,6 +1155,7 @@ class TestPopUpView:
     def pop_up_view(self, mocker):
         self.controller = mocker.Mock()
         self.command = 'COMMAND'
+        self.width = 16
         self.widget = mocker.Mock()
         self.widgets = [self.widget, ]
         self.list_walker = mocker.patch(VIEWS + '.urwid.SimpleFocusListWalker',
@@ -1162,11 +1163,12 @@ class TestPopUpView:
         self.super_init = mocker.patch(VIEWS + '.urwid.ListBox.__init__')
         self.super_keypress = mocker.patch(VIEWS + '.urwid.ListBox.keypress')
         self.pop_up_view = PopUpView(self.controller, self.widgets,
-                                     self.command)
+                                     self.command, self.width)
 
     def test_init(self):
         assert self.pop_up_view.controller == self.controller
         assert self.pop_up_view.command == self.command
+        assert self.pop_up_view.width == self.width
         self.list_walker.assert_called_once_with(self.widgets)
         self.super_init.assert_called_once_with(self.pop_up_view.log)
 
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -4,7 +4,7 @@ import sys
 import time
 from functools import partial
 from platform import platform
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Tuple
 
 import urwid
 import zulip
@@ -88,11 +88,18 @@ class Controller:
     def draw_screen(self, *args: Any, **kwargs: Any) -> None:
         self.loop.draw_screen()
 
+    def maximum_popup_dimensions(self) -> Tuple[int, int]:
+        """
+        Returns 3/4th of the screen estate's columns and rows.
+        """
+        max_cols, max_rows = map(lambda num: 3 * num // 4,
+                                 self.loop.screen.get_cols_rows())
+        return max_cols, max_rows
+
     def show_pop_up(self, to_show: Any) -> None:
         double_lines = dict(tlcorner='╔', tline='═', trcorner='╗',
                             rline='║', lline='║',
                             blcorner='╚', bline='═', brcorner='╝')
-        cols, rows = self.loop.screen.get_cols_rows()
         self.loop.widget = urwid.Overlay(
             urwid.LineBox(to_show,
                           to_show.title,
@@ -102,7 +109,7 @@ class Controller:
             valign='middle',
             # +2 to both of the following, due to LineBox
             width=to_show.width + 2,
-            height=min(3 * rows // 4, to_show.height) + 2
+            height=to_show.height + 2,
         )
 
     def exit_popup(self) -> None:
@@ -122,8 +129,8 @@ class Controller:
                                           "# {}".format(name))
         self.show_pop_up(show_stream_view)
 
-    def popup_with_message(self, text: str, width: int, height: int) -> None:
-        self.show_pop_up(NoticeView(self, text, width, height, "NOTICE"))
+    def popup_with_message(self, text: str, width: int) -> None:
+        self.show_pop_up(NoticeView(self, text, width, "NOTICE"))
 
     def search_messages(self, text: str) -> None:
         # Search for a text in messages
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -88,14 +88,14 @@ class Controller:
     def draw_screen(self, *args: Any, **kwargs: Any) -> None:
         self.loop.draw_screen()
 
-    def show_pop_up(self, to_show: Any, title: str) -> None:
+    def show_pop_up(self, to_show: Any) -> None:
         double_lines = dict(tlcorner='╔', tline='═', trcorner='╗',
                             rline='║', lline='║',
                             blcorner='╚', bline='═', brcorner='╝')
         cols, rows = self.loop.screen.get_cols_rows()
         self.loop.widget = urwid.Overlay(
             urwid.LineBox(to_show,
-                          title,
+                          to_show.title,
                           **double_lines),
             self.view,
             align='center',
@@ -109,20 +109,21 @@ class Controller:
         self.loop.widget = self.view
 
     def show_help(self) -> None:
-        help_view = HelpView(self)
-        self.show_pop_up(help_view, "Help Menu (up/down scrolls)")
+        help_view = HelpView(self, "Help Menu (up/down scrolls)")
+        self.show_pop_up(help_view)
 
     def show_msg_info(self, msg: Message) -> None:
-        msg_info_view = MsgInfoView(self, msg)
-        self.show_pop_up(msg_info_view,
-                         "Message Information (up/down scrolls)")
+        msg_info_view = MsgInfoView(self, msg,
+                                    "Message Information (up/down scrolls)")
+        self.show_pop_up(msg_info_view)
 
     def show_stream_info(self, color: str, name: str, desc: str) -> None:
-        show_stream_view = StreamInfoView(self, color, name, desc)
-        self.show_pop_up(show_stream_view, "# {}".format(name))
+        show_stream_view = StreamInfoView(self, color, desc,
+                                          "# {}".format(name))
+        self.show_pop_up(show_stream_view)
 
     def popup_with_message(self, text: str, width: int, height: int) -> None:
-        self.show_pop_up(NoticeView(self, text, width, height), "NOTICE")
+        self.show_pop_up(NoticeView(self, text, width, height, "NOTICE"))
 
     def search_messages(self, text: str) -> None:
         # Search for a text in messages
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -777,8 +777,23 @@ class PopUpView(urwid.ListBox):
         self.command = command
         self.title = title
         self.log = urwid.SimpleFocusListWalker(widgets)
-        self.width = requested_width
+
+        max_cols, max_rows = controller.maximum_popup_dimensions()
+
+        self.width = min(max_cols, requested_width)
+
+        height = self.calculate_popup_height(widgets, self.width)
+        self.height = min(max_rows, height)
+
         super().__init__(self.log)
+
+    @staticmethod
+    def calculate_popup_height(widgets: List[Any], popup_width: int) -> int:
+        """
+        Returns popup height. The popup height is calculated using urwid's
+        .rows method on every widget.
+        """
+        return sum(widget.rows((popup_width, )) for widget in widgets)
 
     @staticmethod
     def calculate_table_widths(contents: PopUpViewTableContent,
@@ -850,9 +865,7 @@ class NoticeView(PopUpView):
     def __init__(self, controller: Any,
                  notice_text: str,
                  width: int,
-                 height: int,
                  title: str) -> None:
-        self.height = height
         widgets = [
             urwid.Divider(),
             urwid.Padding(urwid.Text(notice_text), left=1, right=1),
@@ -877,7 +890,6 @@ class HelpView(PopUpView):
             help_menu_content, len(title))
         widgets = self.make_table_with_categories(help_menu_content,
                                                   column_widths)
-        self.height = len(widgets)
 
         super().__init__(controller, widgets, 'HELP', popup_width, title)
 
@@ -921,12 +933,10 @@ class PopUpConfirmationView(urwid.Overlay):
 class StreamInfoView(PopUpView):
     def __init__(self, controller: Any, color: str,
                  desc: str, title: str) -> None:
-        # TODO: Height handling could be improved
         # Add 4 (for 2 Unicode characters on either side) to the popup title
         # length to make sure that the title gets displayed even when the
         # content is shorter than the title length (+4 Unicode characters).
         width = max(len(desc) + 2, len(title) + 4)
-        self.height = 2
         stream_info_content = [urwid.Text(desc, align='center')]
         super().__init__(controller, stream_info_content, 'STREAM_DESC', width,
                          title)
@@ -959,7 +969,6 @@ class MsgInfoView(PopUpView):
                   for field, data in msg_info.items()]
         max_widths = [max(width) for width in zip(*widths)]
         width = sum(max_widths)
-        self.height = len(msg_info['Reactions']) + 4 if msg['reactions'] else 5
 
         msg_info_content = [urwid.AttrWrap(
                 urwid.Columns([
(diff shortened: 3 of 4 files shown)


tests/ui/test_ui_tools.py
```diff
@@ -1308,8 +1308,7 @@ def test_keypress_exit_popup(self, key):
         assert self.controller.exit_popup.called
 
     def test_height_noreactions(self, message_fixture):
-        self.msg_info_view = MsgInfoView(self.controller, message_fixture)
```
preetmishra, line 1311: This was redundant.

preetmishra:
I'll proceed to add the tests, for the new functions, if the implementation and the approach look good to everyone.

zulipterminal/ui_tools/views.py
```diff
@@ -775,6 +777,55 @@ def __init__(self, controller: Any, widgets: List[Any],
...
+        col_max_widths = [max(row_width) for row_width in zip(*row_widths)]
+        self.width = max(heading_width, sum(col_max_widths) + dividechars)
+        return col_max_widths
+
+    def build_widgets(self, contents: PopUpViewContent, title_len: int,
+                      dividechars: int=2) -> List[Any]:
+        """
+        Return a list of widgets to be rendered and sets height accordingly.
+        """
+        col_max_widths = self.max_widths(contents, title_len, dividechars)
```
neiljp, line 804: You're passing these values in (contents, title_len, dividechars), to use them in an interior method and only contents otherwise. How might this look if you didn't nest the calls, but called them sequentially?

zulipterminal/ui_tools/views.py
```diff
@@ -807,37 +858,15 @@ def __init__(self, controller: Any,
...
-                    )
-                )
-
-        self.height = len(help_menu_content)
-
+            help_menu_content.append((HELP_CATEGORIES[category], [
+                (binding['help_text'], ', '.join(binding['keys']))
+                for binding in [binding for binding in KEY_BINDINGS.values()
+                                if binding['key_category'] == category]
+            ]))
```
neiljp, line 868: You've compacted the generation of this data-structure, and while it's great that the data is more localized than it was when mixed with the UI elements, it's not quite as clear as it was before - it has a lot of nested elements.

zulipterminal/ui_tools/views.py
```diff
@@ -775,6 +777,55 @@ def __init__(self, controller: Any, widgets: List[Any],
...
+            for row in content:
+                if isinstance(row, str):
+                    heading_width = max(len(row), heading_width)
+                elif isinstance(row, tuple):
+                    row_widths.append(list(map(len, row)))
+        col_max_widths = [max(row_width) for row_width in zip(*row_widths)]
+        self.width = max(heading_width, sum(col_max_widths) + dividechars)
+        return col_max_widths
+
+    def build_widgets(self, contents: PopUpViewContent, title_len: int,
```
neiljp, line 799: I know that `title_len` is what is really passed in, but this value need not be the length of the title - what does it mean to pass in a certain value, for the purposes of these functions? Does that give you an idea for a better name?

zulipterminal/ui_tools/views.py
```diff
@@ -878,50 +907,25 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
 
 class MsgInfoView(PopUpView):
-    def __init__(self, controller: Any, msg: Message) -> None:
-        self.msg = msg
-
+    def __init__(self, controller: Any, msg: Message, title_len: int) -> None:
+        msg_info = [
+            ('', [('Date & Time', time.ctime(msg['timestamp'])[:-5])]),
+            ('', [('Sender', msg['sender_full_name'])]),
+            ('', [('Sender\'s Email ID', msg['sender_email'])]),
```
neiljp: Is this intended as 3 separate categories with no name?

zulipterminal/core.py
```diff
@@ -109,16 +109,17 @@ def exit_popup(self) -> None:
...
-        self.show_pop_up(msg_info_view,
-                         "Message Information (up/down scrolls)")
+        title = 'Message Information (up/down scrolls)'
+        msg_info_view = MsgInfoView(self, msg, len(title))
+        self.show_pop_up(msg_info_view, title)
 
     def show_stream_info(self, color: str, name: str, desc: str) -> None:
-        show_stream_view = StreamInfoView(self, color, name, desc)
+        show_stream_view = StreamInfoView(self, color, desc, len('# ' + name))
         self.show_pop_up(show_stream_view, "# {}".format(name))
```
neiljp, line 123: Is there a reason you didn't tidy this title duplication in addition to the other title changes?

zulipterminal/ui_tools/views.py
preetmishra: Yes.

zulipterminal/core.py
preetmishra: Thanks for pointing it out. I missed it.

zulipterminal/ui_tools/views.py
preetmishra: There are corner cases wherein the calculated width for the 'content' is smaller than the popup title. I use `title_len` to make sure that the calculated width is wide enough to display the popup title every time.

I will rename this to something better.

preetmishra:
@neiljp Thanks for the review. I have restructured the commit history to separate fixes and new changes from refactors, and also to make the refactors incremental.

A few comments about the commits:
* The first two commits are small fixes that could be squashed but I have left them separate for review.
* The next two commits extract `make_chart` (suggestions are welcome for the name) and `charts_width` from `HelpView` and `StreamInfoView`. These static methods are solely useful for popups that want to have or already have a similar chart-like structure.
* The subsequent commit introduces a new static method `get_popup_height` to dynamically calculate popup height that should be useful for any popup.
* The second last commit uses these new methods to simplify `MsgInfoView`.
* The last commit should be self-explanatory.

Re `__init__` vs `make_chart` (previously `build_widgets`), `make_chart` would only be useful for popups that want to have a similar chart-like structure which is why I have kept it separate.

zulipterminal/ui_tools/views.py
```diff
@@ -775,6 +779,34 @@ def __init__(self, controller: Any, widgets: List[Any],
         self.log = urwid.SimpleFocusListWalker(widgets)
         super().__init__(self.log)
 
+    @staticmethod
+    def make_chart(contents: PopUpViewChartContent, column_widths: List[int],
+                   dividechars: int=2) -> List[Any]:
+        """
+        Returns a list of widgets to render a chart.
+        """
+        strip_no = 0  # To track strips (do not increase for categories).
```
neiljp, line 824: Should this be in the loop? I can't test now due to the czo reactions issue, but does the help menu look the same as before?
I think that was the intention of the `enumerate`.

zulipterminal/ui_tools/views.py
```diff
@@ -814,28 +846,17 @@ def __init__(self, controller: Any) -> None:
...
-                        None if help_item_number % 2 else 'popup_contrast'
-                    )
-                )
+            key_bindings = []
+            for binding in keys_in_category:
+                key_bindings.append((binding['help_text'],
+                                     ', '.join(binding['keys'])))
+            help_menu_content.append((HELP_CATEGORIES[category], key_bindings))
 
+        help_menu_content = PopUpView.make_chart(help_menu_content, max_widths)
```
neiljp: Although this is static, you can still use `self.make_chart` to access the method. For the same reason as using `super()`, that avoids coupling in case the name changes.

zulipterminal/ui_tools/views.py
```diff
@@ -775,6 +779,34 @@ def __init__(self, controller: Any, widgets: List[Any],
...
+                   dividechars: int=2) -> List[Any]:
+        """
+        Returns a list of widgets to render a chart.
+        """
+        strip_no = 0  # To track strips (do not increase for categories).
+        widgets = []
+        for category, content in contents:
+            if category:
+                if strip_no != 0:
+                    widgets.append(urwid.Text(''))  # Add a blank line.
```
neiljp, line 829: It's worth noting when/why; this was in the original code?

zulipterminal/ui_tools/views.py
```diff
@@ -880,8 +880,11 @@ def keypress(self, size: urwid_Size, key: str) -> str:
 class StreamInfoView(PopUpView):
     def __init__(self, controller: Any, color: str,
                  name: str, desc: str) -> None:
-        # TODO: Width & Height handling could be improved
-        self.width = max(len(desc), len("# {}".format(name))) + 2
+        # TODO: Height handling could be improved
+        # Add 4 (for 2 Unicode characters on either side) to the popup title
+        # length to make sure that the title gets displayed even when the
+        # content is shorter than the title length (+4 Unicode characters).
+        self.width = max(len(desc), len("# {}".format(name)) + 4)
```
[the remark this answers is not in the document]
preetmishra: Yes, it is how the bug gets fixed.

zulipterminal/ui_tools/views.py
preetmishra: Earlier, in the help menu, the style started alternating from every 'category'. For instance,
```diff
Category 1
+
-
+

Category 2
+
-
+
```

However, now the style starts from the first strip and keeps alternating for every strip. For instance,
```diff
Category 1
+
-
+

Category 2
-
+
-
```

In addition to that, the new alternate styling is needed for the change. Otherwise, consecutive stripes that don't have categories would end up looking like:
```diff
+
+
+

Category 1
+
-
+
```
Instead of:
```diff
+
-
+

Category 1
-
+
-
```

`+`: `popup_constrast`
`-`: `None`

preetmishra:
Yes, I left `MsgInfoView` because the refactors would end up 'changing' its behaviour and structure. However, `StreamInfoView` was straightforward to include.

zulipterminal/ui_tools/views.py
```diff
@@ -880,13 +880,13 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
                  name: str, desc: str) -> None:
-        # TODO: Height handling could be improved
+        description = urwid.Text(desc, align='center')
         # Add 4 (for 2 Unicode characters on either side) to the popup title
         # length to make sure that the title gets displayed even when the
         # content is shorter than the title length (+4 Unicode characters).
         self.width = max(len(desc), len("# {}".format(name)) + 4)
-        self.height = 2
-        stream_info_content = [urwid.Text(desc, align='center')]
+        self.height = description.rows((self.width, ))
```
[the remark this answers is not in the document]
preetmishra: Thanks! I'll fix it.

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
preetmishra: Yes, that is how the bug gets fixed.

zulipterminal/ui_tools/views.py
neiljp: My preference is for the existing consistent line style after the category title; are you changing the behavior since you prefer the alternating style, or because the code is currently simpler with the un-categorized style, and results in that change as a side-effect?

Can we retain the original behavior if we group the '+++' lines into one category instead?

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
neiljp: Well, this has two effects, as I understand it:
* it fixes the bug with an empty description, where the title isn't listed (?)
* it removes a border of one space either side where there are descriptions

I understand that this is a bugfix for the former, but the latter isn't obvious.

To clarify, we don't always have left/right padding in popups, but I'm not sure if we want to change this behavior for now?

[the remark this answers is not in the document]
preetmishra: I understand the concern now. I will edit the commit to only do the former i.e. the bugfix. However, since the help menu doesn't add extra spaces, the two extra spaces will eventually get removed in the refactoring commit (`chart_widths`). Would that be okay?

zulipterminal/ui_tools/views.py
preetmishra: I only changed the original behaviour to avoid having '+++' lines for stripes that do not have categories. 

Yes, grouping them into one category would work. Thanks!

preetmishra:
@neiljp Thanks for the feedback. I have updated the PR to address the suggestions that you left. 

The PR only involves `HelpView` in the refactoring commits and addresses other popups in separate commits as the extracted methods change their behaviours to some extent (which are mentioned in the commit message).

zulipterminal/ui_tools/views.py
```diff
@@ -945,40 +944,26 @@ def __init__(self, controller: Any, color: str, desc: str,
...
-    def __init__(self, controller: Any, msg: Message) -> None:
+    def __init__(self, controller: Any, msg: Message,
+                 popup_title: str) -> None:
         self.msg = msg
 
+        msg_info = [
+            ('', [('Date & Time', time.ctime(msg['timestamp'])[:-5]),
+                  ('Sender', msg['sender_full_name']),
+                  ('Sender\'s Email ID', msg['sender_email'])]),
+        ]
```
preetmishra, line 919: I have grouped these into a category as per the discussion in the other comment.

zulipterminal/ui_tools/views.py
```diff
@@ -779,6 +779,35 @@ def __init__(self, controller: Any, widgets: List[Any],
...
+        title_width = popup_title_len + 4
+
+        category_width = 0
+        text_width = 0
+        strip_widths = []
+        for category, content in contents:
+            category_width = max(category_width, len(category))
+            for row in content:
+                if isinstance(row, str):
+                    text_width = max(text_width, len(row))
```
neiljp, line 809: Same point here as the previous commit; HelpView doesn't use this?

zulipterminal/ui_tools/views.py
```diff
@@ -838,13 +867,7 @@ def __init__(self, controller: Any,
...
 
 class HelpView(PopUpView):
-    def __init__(self, controller: Any) -> None:
-        widths = [(len(binding['help_text']) + 4,
-                   len(", ".join(binding['keys'])))
-                  for binding in KEY_BINDINGS.values()]
-        max_widths = [max(width) for width in zip(*widths)]
-        self.width = sum(max_widths)
-
+    def __init__(self, controller: Any, popup_title: str) -> None:
```
neiljp, line 877: Do we actually need to pass in the title? Didn't you settle on the length, previously? This is like a `minimum_width` parameter. That said, I see the appeal of just expecting the title, as then we can manage the extra spacing internally (eg. the +4)

zulipterminal/ui_tools/views.py
```diff
@@ -779,6 +779,35 @@ def __init__(self, controller: Any, widgets: List[Any],
...
+            category_width = max(category_width, len(category))
+            for row in content:
+                if isinstance(row, str):
+                    text_width = max(text_width, len(row))
+                if isinstance(row, tuple):
+                    strip_widths.append(list(map(len, row)))
+        column_widths = [max(width) for width in zip(*strip_widths)]
+
+        popup_width = max(title_width, category_width, text_width,
+                          sum(column_widths) + dividechars)
```
neiljp, line 815: This is actually different behavior - arguably better and more general, but again not just a refactor!
(some of that comes from the other extra feature, ie. the `text_width`, of course)

zulipterminal/ui_tools/views.py
```diff
@@ -814,28 +845,18 @@ def __init__(self, controller: Any) -> None:
...
-                    )
-                )
-
+            key_bindings = []
+            for binding in keys_in_category:
+                key_bindings.append((binding['help_text'],
+                                     ', '.join(binding['keys'])))
+            help_menu_content.append((HELP_CATEGORIES[category], key_bindings))
+
+        help_menu_content = self.make_table_with_categories(help_menu_content,
```
neiljp, line 891: This reassigning is a bit confusing; it's perfectly valid python, and strictly `Sequence[...]` is compatible with `List[Any]` (can you see where I got those from?), but just using a different name would probably make this clearer :)

zulipterminal/ui_tools/views.py
```diff
@@ -883,8 +891,7 @@ def __init__(self, controller: Any, popup_title: str) -> None:
         self.width = popup_width
         help_menu_content = self.make_table_with_categories(help_menu_content,
                                                             column_widths)
-        self.height = len(help_menu_content)
-
+        self.height = self.get_popup_height(help_menu_content, popup_width)
```
neiljp, line 893: As per the other comment re `help_menu_content` being re-used, using a different name would emphasize why the order is important here.

zulipterminal/ui_tools/views.py
```diff
@@ -945,40 +944,26 @@ def __init__(self, controller: Any, color: str, desc: str,
...
-                urwid.Columns([
-                    urwid.Text(field),
-                    (max_widths[1], urwid.Text(data))
-                ], dividechars=2),
-                None if index % 2 else 'popup_contrast')
-             for index, (field, data) in enumerate(msg_info.items())]
+                (reaction['emoji_name'], reaction['user']['full_name'])
+                for reaction in msg['reactions']
+            )
+            msg_info.append(('Reactions', reactions))
```
neiljp, line 925: I like the reactions in a different category - that seems clearer for sure - but what do you think of grouping responses by alternating color line? Currently (in 0.5.1) it's all one color, which is the other extreme. I see we could do this by reaction, or *maybe* even by user? Grouping would also mean we could avoid having 'tada' listed however many times, for example!

If this fits within the current feature of the revised popup code, we could make that change in this commit, but otherwise we could delay this until later.

zulipterminal/ui_tools/views.py
```diff
@@ -932,15 +932,15 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
-        # Add 4 (for 2 Unicode characters on either side) to the popup title
-        # length to make sure that the title gets displayed even when the
-        # content is shorter than the title length (+4 Unicode characters).
-        self.width = max(len(desc), len("# {}".format(name)) + 4) + 2
-        description = urwid.Text(desc, align='center')
-        self.height = description.rows((self.width, ))
-        stream_info_content = [description]
+    def __init__(self, controller: Any, color: str, desc: str,
+                 popup_title: str) -> None:
+        stream_info_content = [('', [desc])]
```
neiljp: Is this the commit where we'd introduce the one-string form of content? (ie. one column)

zulipterminal/ui_tools/views.py
```diff
@@ -932,15 +932,15 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        popup_width, column_widths = self.table_widths(stream_info_content,
+                                                       len(popup_title))
+        self.width = popup_width
+        stream_info_content = self.make_table_with_categories(
+            stream_info_content, column_widths)
+        self.height = self.get_popup_height(stream_info_content, popup_width)
```
neiljp: Do we actually need to use this here? I suppose we will need it if we enhance this popup.

zulipterminal/ui_tools/views.py
```diff
@@ -932,15 +932,15 @@ def keypress(self, size: urwid_Size, key: str) -> str:
 
 
 class StreamInfoView(PopUpView):
-    def __init__(self, controller: Any, color: str,
-                 name: str, desc: str) -> None:
-        # Add 4 (for 2 Unicode characters on either side) to the popup title
-        # length to make sure that the title gets displayed even when the
-        # content is shorter than the title length (+4 Unicode characters).
-        self.width = max(len(desc), len("# {}".format(name)) + 4) + 2
```
neiljp: Does this change match the refactoring? Another reason that moving the one-column code  into this commit might make sense is that the `len(desc)` code is replaced by the `max(textwidth, ...)` code? So what we use in the first commit should translate to the changes here.

zulipterminal/ui_tools/views.py
```diff
@@ -880,8 +880,11 @@ def keypress(self, size: urwid_Size, key: str) -> str:
 class StreamInfoView(PopUpView):
     def __init__(self, controller: Any, color: str,
                  name: str, desc: str) -> None:
-        # TODO: Width & Height handling could be improved
-        self.width = max(len(desc), len("# {}".format(name))) + 2
+        # TODO: Height handling could be improved
+        # Add 4 (for 2 Unicode characters on either side) to the popup title
+        # length to make sure that the title gets displayed even when the
+        # content is shorter than the title length (+4 Unicode characters).
+        self.width = max(len(desc), len("# {}".format(name)) + 4) + 2
```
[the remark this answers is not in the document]
preetmishra: Yes, the +4 adds to the title length (to compensate for those 4 Unicode box drawing characters). I initially removed the +2 but that changed the original behaviour i.e. removed the two extra spaces.

Re 'desc+2, name+4', I think the +2, in the current iteration, is clearer, it indicates that we are explicitly adding 2 extra spaces to the minimum required width returned by the `max` function.

zulipterminal/ui_tools/views.py
preetmishra: Yes, this seems appropriate.

zulipterminal/ui_tools/views.py
preetmishra: The reason I didn't go with `minimum_width` is that even when we pass `minimum_width=len(popup_title)`, we need the +4 in the `table_widths` method which doesn't justify the 'minimum width' name?

An alternative could be passing `len(popup_title) + 4` but that wouldn't be clear either (why we're adding +4 to the length?). To address that, we could add a comment but that'd require adding a comment before every call.

The approach where we handle the +4 internally allows adding a single comment.

I could use a suggestion about how to approach this.

preetmishra:
@neiljp Thanks for the review and the suggestions, splitting and arranging the commits in an incremental order has definitely made it easier to comprehend the changes and spot any new changes.

I have moved code and updated the commits to introduce changes only when they're required and also removed 'extra' code from the refactor commits.

I have added a separate commit for `StreamInfoView` that essentially incorporates single column/text logic in the existing methods. We could either hold it off or merge it but I think it'd be better to include it while we're at it.

Re grouping reactions together, I second that. I would like to address it in a comparatively small follow-up PR.

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
neiljp: The issue is that we don't want to unconditionally add 2, as this could end up with len(title)+6.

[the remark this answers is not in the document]
preetmishra: Ah, right. `max(len(desc) + 2, len("# {}".format(name)) + 4)` would be better. Thanks!

preetmishra:
Updated with `max(len(desc) + 2, len("# {}".format(name)) + 4)`.

zulipterminal/ui_tools/views.py
```diff
@@ -778,6 +778,23 @@ def __init__(self, controller: Any, widgets: List[Any],
         self.log = urwid.SimpleFocusListWalker(widgets)
         super().__init__(self.log)
 
+    @staticmethod
+    def table_widths(contents: PopUpViewTableContent,
+                     dividechars: int=2,
+                     extra_width: int=0) -> Tuple[int, List[int]]:
```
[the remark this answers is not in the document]
preetmishra: It isn't used anywhere for now.

I added for making it easier to add extra width if we ever want to. Should we remove it and add it only when it is needed?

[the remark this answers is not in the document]
neiljp: Let's leave it for now, unless we use it - actually, could it be used for padding the message info?

zulipterminal/ui_tools/views.py
```diff
@@ -925,16 +933,16 @@ def keypress(self, size: urwid_Size, key: str) -> str:
 
 
 class StreamInfoView(PopUpView):
-    def __init__(self, controller: Any, color: str,
-                 name: str, desc: str) -> None:
-        # Add 4 (for 2 Unicode characters on either side) to the popup title
-        # length to make sure that the title gets displayed even when the
-        # content is shorter than the title length (+4 Unicode characters).
-        self.width = max(len(desc) + 2, len("# {}".format(name)) + 4)
-        description = urwid.Text(desc, align='center')
```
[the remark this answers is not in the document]
preetmishra: While centered and padded content will look good as a standalone item, it will a bit odd in popups which also have two columns (where the left column is left and the right column is right aligned).

Re 'snug', I agree. I see that you've suggested 'mimicking the existing code for now' (which I assume means using the left column for text and leaving the right column blank). 

On another note, the `StreamInfoView` looks good without the changes that are being introduced in this commit. We could also hold off this commit as "we don't know what we'll need in the future".

Thoughts?

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
preetmishra: It can only be used to increase the width from the inside - the content in the left column will be still left-aligned and the content in the right column will be still right-aligned. I will leave it for now.

zulipterminal/ui_tools/views.py
preetmishra: @neiljp With the current (PR) code, one way to group could be using the reaction name for the first reactor and then leaving the subsequent left column blank for other reactors (with alternating `popup_contrast` style).

For instance,
```
+1       User 1 (popup_contrast)
         User 2
         User 3 (popup_contrast)
tada     User 1 
         User 2 (popup_contrast)
         User 3 

```
```python
[('+1', 'User 1'), ('', 'User 2'), ('', 'User 3'), ('tada', 'User 1'), ('', 'User 2'), ('', 'User 3')]
```


Does the proposed styling look good or should I make amendments to the `make_table_with_categories` code to incorporate the other behaviour (grouping responses by alternating colour line) that you suggested?

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
neiljp: I was meaning to use padded and centered style on the single-column version, ie. the specialized version introduced in that commit, to retain the existing style. So two columns => help menu style, text => StreamInfo style (as it is now).

We could leave this commit for a later merge in this PR or elsewhere. Once things are tidied from the other comments I could merge some commits and we could either work on the remaining commit(s) here or move them to another branch/PR.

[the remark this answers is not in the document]
preetmishra: I understand what you're suggesting. Though, I am concerned about the scenario where we would want to have both in the same popup i.e. the help menu style and the stream info style. It would look a bit off to have some content centre-aligned and some left/right-aligned. That said, if we decide to stick with either one of the styles in any popup (not mix single-column and two-column), that would work.

preetmishra:
@neiljp Thanks for the review. I have incorporated the in-line suggestions (leaving the ones that are related to `StreamInfoView` for now) and split the MsgInfoView commit into three parts.

Re grouping the reactions, I have left a proposal in https://github.com/zulip/zulip-terminal/pull/659#discussion_r432652509. I'll make the amendments as per your suggestion.

zulipterminal/ui_tools/views.py
neiljp: This is one option. I think originally I was considering you could have the users joined with newlines, so the above would be more like:
```
+1    User 1 (popup_contrast)
      User 2 (popup_contrast)
      User 3 (popup_contrast)
tada  User 1
      User 2
      User 3
smile User 4 (popup_contrast)
```
I thought this might use the contrast to group the reactions better, perhaps.

zulipterminal/ui_tools/views.py
```diff
@@ -880,13 +880,13 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
                  name: str, desc: str) -> None:
-        # TODO: Height handling could be improved
         # Add 4 (for 2 Unicode characters on either side) to the popup title
         # length to make sure that the title gets displayed even when the
         # content is shorter than the title length (+4 Unicode characters).
         self.width = max(len(desc) + 2, len("# {}".format(name)) + 4)
-        self.height = 2
-        stream_info_content = [urwid.Text(desc, align='center')]
+        description = urwid.Text(desc, align='center')
+        self.height = description.rows((self.width, ))
```
neiljp: I was just going to merge this, but realized we have a subtle nuance here, as `self.width` is always requested to be big enough to hold the description, so we end up wit ha height of 1. The popup can be scrolled, but that's probably not what we want!

Given that the height is only known when showing the popup, we likely want to use a different approach for cases where we want some kind of minimum height visible, or use the column width we know we get in the popup to calculate it. The controller popup code needs improving, whether a special case for a 'minimum height visible' case like this, or a general change where the height is calculated later, maybe all from rows?

As part of this, we may wish to limit the maximum popup width, like we do with popup height?

preetmishra:
@neiljp I have updated the PR with the new height calculation approach (that we discussed in the PM) and with grouped reactions. 

Re `MsgInfoView`, since the original commit has been split into three, you would come across a few strange behaviours if you review it commit by commit. I have explained these [here](https://chat.zulip.org/#narrow/pm-with/2406,10457,14317-group/near/891242). We could squash these three commits post review.

preetmishra:
Updated to fix the commit order on GitHub.

zulipterminal/ui_tools/views.py
```diff
@@ -772,10 +772,11 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
 
 class PopUpView(urwid.ListBox):
     def __init__(self, controller: Any, widgets: List[Any],
-                 command: str) -> None:
+                 command: str, width: int) -> None:
```
neiljp: I like having this first commit early, but maybe also have the last commit just afterwards, which applies the same logic to the title?
Then you can promptly follow it with the 'Consider category and title width...' commit, which means at least one of the latter commits doesn't break, in terms of not allowing space for the window title?

zulipterminal/ui_tools/views.py
```diff
@@ -942,27 +941,14 @@ def __init__(self, controller: Any, msg: Message) -> None:
...
-                None if index % 2 else 'popup_contrast')
-             for index, (field, data) in enumerate(msg_info.items())]
-
-        super().__init__(controller, msg_info_content, 'MSG_INFO', width)
+        msg_info = [
+            ('', [('Date & Time', time.ctime(msg['timestamp'])[:-5]),
+                  ('Sender', msg['sender_full_name']),
+                  ('Sender\'s Email ID', msg['sender_email']),
+                  ('Reactions', reactions if msg['reactions']
+                   else '---None---')]),
```
neiljp: While I know that this code is replaced when you use a category instead, I'm a little concerned that the layout code for columns doesn't work in general, in case we end up with a wide column again? Or is something else happening here?

preetmishra: In this particular case, `reactions`, the data structure, that is being passed isn't what `make_table_with_categories` expects. 

However, since the old code uses a hardcoded width (`column_widths[1]`), this is likely to break for wide columns.

A fix could be to fall back `('weight', 1, urwid.Text(data))` whenever the maximum allowed width is smaller than `column_widths[1]`. Though, it would require `max_cols` to be passed into the method.

I couldn't come up with a way to pass maximum allowed width into `calculate_table_witdhs` without tampering the 'flow' of the commits in this PR.

I was thinking of passing a `maximum_width` argument, where `maximum_width=controller.maximum_popup_dimensions()[0]` and then making further amendments to `calculate_popup_width` and `make_table_with_categories` accordingly.

preetmishra:
@neiljp Thank you for the review. I have made the proposed amendments and restructured the PR as per your in-line comments.

However, I couldn't address the column breakage bit in the `MsgInfoView`. I have left my findings in https://github.com/zulip/zulip-terminal/pull/659#discussion_r434257471. It would be really helpful if you could share a few pointers about the same.

zulipterminal/ui_tools/views.py
```diff
@@ -813,7 +813,10 @@ def calculate_table_widths(contents: PopUpViewTableContent,
         for category, content in contents:
             category_width = max(category_width, len(category))
             for row in content:
-                strip_widths.append(list(map(len, row)))
+                # Replace text with the longest string if the text is seperated
+                # by newlines.
+                row_max = map(lambda text: max(text.split('\n'), key=len), row)
```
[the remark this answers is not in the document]
preetmishra: I don't have any preference for `map` or `lambda` syntax as such. It's better to use them in small doses, I agree.

I have combined the proposed implementation with the other `len` in the latest update. Thanks.

zulipterminal/ui_tools/views.py
```diff
@@ -1,6 +1,6 @@
 import threading
 import time
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
```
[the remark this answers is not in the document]
preetmishra: This change amends the +2 extra width and the alignment that we don't have any test for. The commit doesn't change any other functionality.

preetmishra:
@neiljp Thanks for the review and the in-line suggestion.

I have updated that commit with `comprehension` syntax and improved the commit messages. 

Furthermore, I have moved the https://github.com/zulip/zulip-terminal/pull/659#discussion_r438341690 change into that other commit (where we had an issue with reactions) and used `''.join()` to make `reactions` compatible with `make_table_with_categories`. This should resolve the issue (partially).

Yes, I gather it would be better to hold the last commit off for now. I have removed it from the PR. I will rebase and push a WIP PR after we have merged these changes.

zulipterminal/ui_tools/views.py
```diff
@@ -814,7 +813,13 @@ def calculate_table_widths(contents: PopUpViewTableContent,
...
             category_width = max(category_width, len(category))
             for row in content:
-                strip_widths.append(list(map(len, row)))
+                # Measure the longest line if the text is seperated by
+                # newline(s).
+                max_row_lengths = [
+                    len(max(text.split('\n'), key=len))
+                    for text in row
+                ]
+                strip_widths.append(max_row_lengths)
```
neiljp, line 822: This looks clearer in this commit than the version with lambdas before! This does seem to fix the issue we saw earlier with this commit, unless the window/width gets super-narrow, which is probably going to be an issue to use ZT in any case.

zulipterminal/ui_tools/views.py
neiljp: This seems to be mostly fixed by the updated commit in any case :+1:

PR #667, opened 2020-05-20
Show loading screen in urwid.

kaustubh-nair:
These commits are not perfectly isolated, because there are a lot of things to be done here:
* Show `LoadingView` first and run `Model` and `View` initialization asynchronously.
* Move settings from `run.py` to `LoadingView`
* Move spinning cursor from `Controller` to `LoadingView`

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -16,8 +16,8 @@ from zulipterminal.model import Model
 from zulipterminal.ui import Screen, View
 from zulipterminal.ui_tools.utils import create_msg_box_list
 from zulipterminal.ui_tools.views import (
-    AboutView, HelpView, MsgInfoView, NoticeView, PopUpConfirmationView,
-    StreamInfoView,
+    AboutView, HelpView, LoadingView, MsgInfoView, NoticeView,
+    PopUpConfirmationView, StreamInfoView,
 )
 from zulipterminal.version import ZT_VERSION
 
@@ -39,12 +39,14 @@ class Controller:
 
         self._editor = None  # type: Optional[Any]
 
-        self.show_loading()
+        self.capture_stdout()
         self.client = zulip.Client(config_file=config_file,
                                    client='ZulipTerminal/{} {}'.
                                           format(ZT_VERSION, platform()))
+        self.loading_view = LoadingView(self)
         self.init_model_view()
 
+    @asynch
     def init_model_view(self) -> None:
         self.model = Model(self)
         self.view = View(self)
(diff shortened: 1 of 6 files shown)


kaustubh-nair:
Making a new PR for this. 
I'll keep the current tutorial PR #583 only for tutorial related enhancements.

Adding tests.

sumanthvrao:
@kaustubh-nair I tried this out manually and it looks like great! 👌

neiljp:
@kaustubh-nair I replied on the stream about this - functionally this does what we want, but the structure is a little confusing and smaller commits may help explain which changes you're making and why. It's great that you're picking up others' older PRs and moving them forward, but it can be challenging precisely since they're not your code

Some aspects to consider:
* what (current) aspects of the model and view can be started in parallel or not?
* we have various cases in the code where we check whether the view is set up (`hasattr`); if we can ensure the ordering in the initial setup then can we remove these?
* while `Controller()` and its `.main(...)` are run consecutively, if we change the order we should understand/explain why
* when commits simply "move" code around, make sure that the code hasn't been adjusted by other commits between this - this is particularly true with rebasing older PRs!

zulipterminal/core.py
```diff
@@ -273,16 +273,28 @@ def show_all_mentions(self, button: Any) -> None:
...
         sys.exit(0)
 
     @asynch
     def set_main_view(self) -> None:
+        def spinning_cursor() -> Any:
+            while True:
+                for cursor in '|/-\\':
+                    yield cursor
+
+        self.capture_stdout()
```
sumanthvrao, line 292: Do we need to capture stdout here again? We already do this now in the previous commit in the `__init__` stage itself.

zulipterminal/ui_tools/views.py
```diff
@@ -1127,4 +1134,4 @@ def keypress(self, size: Tuple[int, int], key: str) -> str:
         return key
 
     def set_spinner(self, spinner: str) -> None:
-        self.base_widget.set_text(self.text + [spinner])
+        self.body.base_widget.set_text(self.text + [spinner])
```
sumanthvrao, line 1137: Is this something necessary for this commit? Couldn't we move this to the previous commit where you introduce the `set_spinner` class?

sumanthvrao (commented):
@kaustubh-nair Thanks for rebasing 👍 . 

Additionally, I think this may avoid the UI distort race issue which we have seen in the past. I left a couple of commits but looks good overall.

zulipterminal/core.py
```diff
@@ -273,16 +273,28 @@ def show_all_mentions(self, button: Any) -> None:
                         mentioned=True)
 
     def deregister_client(self) -> None:
-        queue_id = self.model.queue_id
-        self.client.deregister(queue_id, 1.0)
+        # Deregister before model has loaded.
```
Ezio-Sarthak, line 276: Is this necessary in this commit?

Issue #642, opened 2020-05-02
Negative number of new messages in a topic


osa1:
See the screenshot (`Salsa q` line). Observed in http://rust-lang.zulipchat.com/.

![Screenshot_2020-05-02_14-39-28](https://user-images.githubusercontent.com/448274/80863144-dd05bf80-8c82-11ea-8eed-650043a82e42.png)

neiljp:
@osa1 Thanks for the report! Could you confirm which version you are using? (`zulip-term -v`)

We're aware of this bug, which has been present previously but would then present as the stream suddenly appearing muted instead of having a negative total.

osa1:
```
~ $ zulip-term -v
Zulip Terminal 0.5.0
```

neiljp:
@osa1 This has improved somewhat with a recent change in master, but is not fully fixed.

Rohitth007:
I was able to successfully reproduce a particular case of negative count this would be fixed after #977 is merged.
Steps to reproduce:

    Search for something.
    Wait for a message sent to any other stream apart from those in the results.
    Go to that message and mark to as read
    Count becomes negative if initially count was zero.

This is applicable for topic count and user count aswell

Issue #574, opened 2020-04-02
Handle setting unread count for muted topics


kaustubh-nair:
Marking muted topics as read decreases count in its respective stream. This could lead to negative unread counts in the stream. 

Steps to reproduce:
1. Mute a topic.
2. Receive messages in the topic.
3. Mark the messages as read.
4. Unread count in the stream should decrease.

kaustubh-nair:
@zulipbot add "bug"

Issue #566, opened 2020-03-28
Don't update unread counts for PMs whose view is open


kaustubh-nair:
Ran into this while working on https://github.com/zulip/zulip-terminal/pull/548
We don't have code to handle PMs with open views, like we do for streams. This leads to invalid unread counts. Like in this screenshot, the unread count is -1 
![DeepinScreenshot_select-area_20200328213735](https://user-images.githubusercontent.com/30197132/77827623-cbd60a00-713c-11ea-8445-36bffd733210.png)

Steps to reproduce this:
1. Make sure the PM view with another user is open.
2. Receive a PM from the other user. (Note that because the view is open, the unread count won't increase)
3. Select the new message to mark it as read - unread count goes to -1

kaustubh-nair:
@zulipbot add "bug"

neiljp:
If I'm following this correctly, I can't reproduce this. This is with me chatting with my own bot.

However, I can reproduce this issue if the PM view is **not** open.

Could you describe more what you think the problem relates to?

My suspicion is that it is something which would be resolved with correct handling of the `unread_message` data, as in your open WIP PR.

kaustubh-nair:
Hmm, I'm not able to entirely pinpoint the exact issue here, maybe @sumanthvrao would have a better idea?

sumanthvrao:
I tried reproducing the issue with my bot, but am not able to. Neither when PM view is open nor when it's not open. 😕

preetmishra:
I can't reproduce it, either with PM opened or closed, using the dev server (on latest `master`).

PR #670, opened 2020-05-24
Improve tests for Model._handle_message_event

kaustubh-nair:
This should hopefully be a fix for #642 #574 and #566

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -697,7 +697,8 @@ class TestModel:
          assert_called_once_with(model, [message_fixture['id']],
                                  last_message=expected_last_msg))
 
-    def test__handle_message_event_with_flags(self, mocker,
+    @pytest.mark.parametrize('flags', [['read', 'mentioned'], [], ])
+    def test__handle_message_event_with_flags(self, mocker, flags,
                                               model_message_event,
                                               message_fixture):
         model = model_message_event
@@ -705,21 +706,15 @@ class TestModel:
                      return_value=["msg_w"])
         model.msg_list.log = [mocker.Mock()]
         set_count = mocker.patch('zulipterminal.model.set_count')
+        event = {'message': message_fixture, 'flags': flags}
 
-        # Test event with flags
-        event = {'message': message_fixture, 'flags': ['read', 'mentioned']}
         model._handle_message_event(event)
-        # set count not called since 'read' flag present.
-        set_count.assert_not_called()
 
-        # Test event without flags
-        model.notify_user.assert_called_once_with(event['message'])
-        model.msg_list.log = [mocker.Mock()]
-        event = {'message': message_fixture, 'flags': []}
-        model._handle_message_event(event)
-        # set count called since the message is unread.
-        set_count.assert_called_once_with([event['message']['id']],
-                                          self.controller, 1)
+        if('read' in flags):
+            set_count.assert_not_called()
+        else:
+            set_count.assert_called_once_with([event['message']['id']],
+                                              self.controller, 1)
 
     @pytest.mark.parametrize('response, narrow, recipients, log', [
         ({'type': 'stream', 'stream_id': 1, 'subject': 'FOO',
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -660,6 +660,18 @@ class TestModel:
         model.found_newest = True
         return model
 
+    def test__handle_message_event_notify_users(self, mocker,
+                                                model_message_event,
+                                                message_fixture):
+        model = model_message_event
+        create_msg_box_list = mocker.patch('zulipterminal.model.'
+                                           'create_msg_box_list',
+                                           return_value=["msg_w"])
+        event = {'message': message_fixture}
+
+        model._handle_message_event(event)
+        model.notify_user.assert_called_once_with(event['message'])
+
     def test__handle_message_event_with_Falsey_log(self, mocker,
                                                    model_message_event,
                                                    message_fixture):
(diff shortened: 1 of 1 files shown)


tests/model/test_model.py
```diff
@@ -783,6 +783,36 @@ def test__handle_message_event(self, mocker, user_profile, response,
...
+        (True),
+        pytest.param(False,
+                     marks=pytest.mark.xfail(reason="Unread count bug")),
+        ])
+    @pytest.mark.parametrize('event', [
+        ({'message': {'type': 'stream', 'stream_id': 1, 'subject': 'HELLO',
+          'id': 1}, 'flags': []}),
+        ])
+    def test__handle_message_event_set_count_all_narrows(self, mocker, event,
+                                                         model, found_newest,
```
neiljp, line 795: This uses `model`, which has an initial default narrow of 'all messages', so this doesn't replicate the 'in a different narrow' test case we started with. This didn't turn out to be the issue, so I'm not sure if we want to explore behavior over different narrows.

zulipterminal/model.py
```diff
@@ -727,8 +727,10 @@ def _handle_message_event(self, event: Event) -> None:
             self.controller.update_screen()
             self._notified_user_of_notification_failure = True
 
+        self.index = index_messages([message], self, self.index)
+        if 'read' not in message['flags']:
+            set_count([message['id']], self.controller, 1)
```
neiljp, line 732: minor: follow this with a newline, and possibly add a comment to this block re updating counts and how the message needs to be in the index to be identified by `set_count`.

From other conversations, I wonder if we might update `set_count` to just accept the message(s) themselves to avoid an index dependence, but that's another discussion. That would avoid the coupling between these two actions (add to index, set count), as the message need not necessarily be in the index.

tests/model/test_model.py
```diff
@@ -785,8 +785,7 @@ def test__handle_message_event(self, mocker, user_profile, response,
 
     @pytest.mark.parametrize('found_newest', [
         (True),
-        pytest.param(False,
-                     marks=pytest.mark.xfail(reason="Unread count bug")),
+        (False),
```
neiljp, line 788: I understand this is a simple adjustment from the previous commit and so a small diff, but I wonder if this should shrink onto one line now.

tests/model/test_model.py
```diff
@@ -783,6 +783,36 @@ def test__handle_message_event(self, mocker, user_profile, response,
         # LOG REMAINS THE SAME IF UPDATE IS FALSE
         assert model.msg_list.log == log
 
+    @pytest.mark.parametrize('found_newest', [
+        (True),
```
neiljp, line 787: For a one-parameter entry, this is just `True`, ie. the parentheses are unnecessary.

neiljp (commented):
@kaustubh-nair Thanks for filing this and working on the extra test :+1: 

Some of my points will disappear if we squash the two commits.

Other points are discussion points that we could use to explore other improvements.

tests/model/test_model.py
kaustubh-nair: Being in a different narrow won't make a difference since it's only `found_newest` which affects behaviour?

tests/model/test_model.py
```diff
@@ -783,6 +783,36 @@ def test__handle_message_event(self, mocker, user_profile, response,
...
+                                 return_value=initial_index)
+        model.found_newest = found_newest
+        model.index = initial_index
+
+        model._handle_message_event(event)
+
+        index_msg.assert_called_once_with([event['message']], model,
+                                          model.index)
+        set_count.assert_called_once_with([event['message']['id']],
+                                          self.controller, 1)
```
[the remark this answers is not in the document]
kaustubh-nair: @neiljp As in, change in state of unread counts and index?

tests/model/test_model.py
kaustubh-nair: Renamed the test to `test__handle_message_event_set_count_found_newest`, since we are not directly considering different narrows.

PS. Now that I look at this, it feels a little too much to write such a big test for such a small change.
Wouldn't a comment saying "Update counts irrespective of found_newest" have been enough?

kaustubh-nair:
@neiljp Thanks for the review, I've pushed the updated commits.

sumanthvrao:
@kaustubh-nair Thanks for working on this 👍 . I tried some manual testing and the results are promising.

One follow-up point worth mentioning here (in case you or someone else wants to pick it) is to try and bridge the 'holes' in index forming which this might trigger. For instance, the new logic  will solve the negative unread_count issues we have been seeing, but suppose we index a non-current message which is received and then the user scrolls down (keypress down) causing new message to be fetched, how will the index handle this? Will the re-index cause any problem since the message may potentially exist already in the index.

This don't need to be solved right away, but it's worth keeping in mind of the impact this might have on the indexing changes.

neiljp:
@kaustubh-nair I just pushed the code part of this PR, since this is an important fix that I don't want to leave unmerged for too long - both because it improves the overall experience, and since it may expose other bugs.

Let's discuss whether you want to leave this PR open to improve the tests for this method, or whether to add this as another issue so we don't forget it.

kaustubh-nair:
Pushing some commits for refactoring tests now.

PS. If you think the current commits don't require any more changes, you can merge them. I'll add commits on top.

tests/model/test_model.py
```diff
@@ -645,17 +645,23 @@ def test__stream_info_from_subscriptions(self, initial_data, streams,
...
+                     return_value={})
         model.msg_list = mocker.Mock(log=[])
+        model.notify_user = mocker.Mock()
+        model.found_newest = True
+        return model
+
+    def test__handle_message_event_with_Falsey_log(self, mocker,
+                                                   model_message_event,
+                                                   message_fixture):
+        model = model_message_event
```
kaustubh-nair, line 673: Doing this because `model_message_event` is a really long variable name.

kaustubh-nair:
The main commit here was merged, so renamed the PR.
Does need a rebase though.

PR #672, opened 2020-05-25
Ensure recipient headers are shown in message list

### LINKED ISSUE #671: Long message recipients are not shown in narrow windows
This is most notable with many PM recipients or with long stream/topic names. The recipients are still shown in the 'current messages recipients' box, but not in the list of messages. This is troublesome in stream, all-pm and presumably all-message narrows, where these headings are relied upon to understand which messages are in which narrow, within a wider/broader narrow (eg. messages in a given topic, in the stream).

neiljp:
This fixes #671, which is reproducible by resizing the window to be very narrow and noting the 'blank' headers that appear where there is a change in recipients.

An additional commit is included, which improves the spacing between the header text and the new lines/bars, both fixing an existing bug (for PMs) and ensuring a space is shown in the narrow-window situation.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1716,7 +1716,8 @@ class TestMessageBox:
         assert isinstance(view_components[0], Columns)
 
         assert isinstance(view_components[0][0], Text)
-        assert isinstance(view_components[0][1], Divider)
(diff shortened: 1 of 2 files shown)


neiljp:
@preetmishra Thanks for the review :+1: Merging now.

PR #673, opened 2020-06-01
Encapsulate editor/editor_mode

neiljp:
This was previously partially encapsulated in WriteBox for local use,
but this wraps all functionality with access via the Controller using:
* is_in_editor_mode()
* enter_edit_mode_with(editor)
* exit_editor_mode()
* current_editor()

This then allows further tests for this and refactoring, as well as being more descriptive.

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -43,6 +43,23 @@ class TestController:
         self.model.poll_for_events.assert_called_once_with()
         assert controller.theme == self.theme
 
+    def test_initial_editor_mode(self, controller):
+        assert not controller.is_in_editor_mode()
+
+    def test_editor_mode_entered_from_initial(self, mocker, controller):
+        editor = mocker.Mock()
+
+        controller.enter_editor_mode_with(editor)
+
+        assert controller.is_in_editor_mode()
+        assert controller.current_editor() == editor
+
+    def test_editor_mode_exits_after_entering(self, mocker, controller):
+        controller.enter_editor_mode_with(mocker.Mock())
+        controller.exit_editor_mode()
+
+        assert not controller.is_in_editor_mode()
+
     def test_narrow_to_stream(self, mocker, controller,
                               stream_button, index_stream) -> None:
         controller.model.narrow = []
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -46,6 +46,10 @@ class TestController:
     def test_initial_editor_mode(self, controller):
         assert not controller.is_in_editor_mode()
 
+    def test_current_editor_error_if_no_editor(self, controller):
+        with pytest.raises(AssertionError):
+            controller.current_editor()
+
     def test_editor_mode_entered_from_initial(self, mocker, controller):
         editor = mocker.Mock()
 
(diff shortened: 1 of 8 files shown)


tests/ui_tools/test_boxes.py
```diff
@@ -208,7 +208,9 @@ def test_reset_search_text(self, panel_search_box):
     def test_keypress_ENTER(self, panel_search_box,
                             enter_key, log, expect_body_focus_set):
         size = (20,)
-        panel_search_box.panel_view.view.controller.editor_mode = True
+        panel_search_box.panel_view.view.controller.is_in_editor_mode = (
```
sumanthvrao, line 211: The previous diff looked like it was setting the editor mode to True, which according to our helper functions should have called `enter_edit_mode_with()`. Did you use `is_in_editor_mode`  to makes things simpler? If that is the case, maybe adding a comment here would help.

tests/ui/test_ui.py
```diff
@@ -249,29 +253,31 @@ def test_keypress_autohide_streams(self, view, mocker, autohide, key):
...
-        assert view.controller.editor == view.stream_w.stream_search_box
+        (view.controller.enter_editor_mode_with
+         .assert_called_once_with(view.stream_w.stream_search_box))
 
     @pytest.mark.parametrize('key', keys_for_command('SEARCH_PEOPLE'))
     def test_keypress_edit_mode(self, view, mocker, key):
         view.users_view = mocker.Mock()
         view.body = mocker.Mock()
         view.user_search = mocker.Mock()
-        size = (20,)
```
sumanthvrao, line 269: Move this dead code to a separate commit?

tests/core/test_core.py
```diff
@@ -43,6 +43,23 @@ def test_initialize_controller(self, controller, mocker) -> None:
         self.model.poll_for_events.assert_called_once_with()
         assert controller.theme == self.theme
 
+    def test_initial_editor_mode(self, controller):
```
sumanthvrao, line 46: Maybe extract this change along with the core.py changes into a previous commit?

zulipterminal/core.py
```diff
@@ -34,8 +34,7 @@ def __init__(self, config_file: str, theme: ThemeSpec,
         self.autohide = autohide
         self.notify_enabled = notify
 
-        self._editor_mode = False
-        self._editor = None  # type: Any
+        self._editor = None  # type: Optional[Any]
```
sumanthvrao, line 37: If I understand this right, we are doing away with `self._editor_mode` altogether. Maybe, add a small note in commit description about this?

tests/ui_tools/test_boxes.py
neiljp: What do you mean by 'previous diff'? Do you mean what this commit replaces? I agree that setting it to `True` as it stands (without encapsulation) is literally the same as setting editor_mode to True via doing `enter_edit_mode_with`. However, that assumes we have all of the state in the controller available. Instead of mocking the controller state (as previously), these lambda methods are essentially mocking the controller behavior (on the assumption that it doesn't matter that `is_in_editor_mode` *always* returns `True` in this test).

tests/core/test_core.py
neiljp: These tests are for the new functions, which replace the previous direct setting of the attributes. Do you mean to add the new methods and tests in one commit, and then migrate the code to use them in another? I suppose I could make the attributes private in the second step.

tests/ui/test_ui.py
neiljp: I think there's just this one line and a comment? That seems rather minor to have separately?

tests/ui_tools/test_boxes.py
sumanthvrao: Re `previous diff`, I meant what this commit replaces. Yup. Thanks for clarifying. Re state: That was what I suspected as well. :)

tests/core/test_core.py
sumanthvrao: Yes, exactly.

neiljp:
@sumanthvrao Thanks for the review :+1:

PR #681, opened 2020-06-08
Update zulipchat.com links and emails to zulip.com

andersk:


andersk:
@neiljp Subdomains like `realm.zulipchat.com` and `your-org.zulipchat.com` are not being moved, at least as of now.

PR #683, opened 2020-06-11
README: Update pip install command.

kaustubh-nair:
Update the command to install development requirements using pip.
Zsh does not match expressions unless surrounded with quotes.
Thus, executing the command without quotes causes it to fail.


neiljp:
@kaustubh-nair Thanks for reporting, debugging and filing the PR! Merging now :+1:

Issue #617, opened 2020-04-23
Maintain position in topic list when toggling topic list


neiljp:
When we enter/leave the topic list, we currently reset the focus in the topic list to the top - even if previously it was some distance down the list.

I'd suggest we want to maintain focus instead, perhaps even when moving between different streams and back again? (at least, within one ZT session)

For implementation, this will involve storing the state when last in a stream (if there is any), on changing between stream and topic views in the left panel. Note that topic names and ordering may change between such times, so we should also fall back to an almost match, including if a topic is resolved.

Subhasish-Behera:
@neiljp i would like to work on this issue as my first issue.can u point me to the directory where changes are needed to be made?

neiljp:
@Subhasish-Behera This will likely be a change in one of the ui_tools/ files, maybe `views.py`.

I'd suggest generally understanding the logic involved in following from a 'key' (eg. `?` for the help menu) to the action, and then which state it updates.

If you need support with getting started, please drop in on chat.zulip.org in #zulip-terminal :)

Subhasish-Behera:
@israelsgalaxy Thanks for showing interest in ZT.Thanks for showing interest in ZT. I've made substantial progress and am almost finished. Given the progress I've made, I think it would be best for me to wrap this up, and you could work on a different issue.
Sorry if there was some confusion as I don't generally `claim` any issue .

neiljp:
Fixed by #1277.

PR #695, opened 2020-06-18
Toggle stream settings

kaustubh-nair:
This PR adds checkboxes to `StreamInfoView` for toggling muted and pinned settings for streams.

There are two parts to this PR and a bunch of minor refactors:
**1. Add support for muting** - We already handle mute events, so all we have to do here is add a checkbox that will toggle mute status.
**2. Add support for pinning** - First support for pinning events are added, then a checkbox is added to `StreamInfoView`.

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -190,13 +190,13 @@ def test_powerset(iterable, map_func, expected_powerset):
 
 @pytest.mark.parametrize('muted_streams, muted_topics, vary_in_unreads', [
     ([99], [['Some general stream', 'Some general unread topic']], {
-        'all_msg': 9,
+        'all_msg': 8,
         'streams': {99: 1},
         'unread_topics': {(99, 'Some private unread topic'): 1},
         'all_mentions': 0,
     }),
     ([1000], [['Secret stream', 'Some private unread topic']], {
-        'all_msg': 11,
+        'all_msg': 8,
         'streams': {1000: 3},
         'unread_topics': {(1000, 'Some general unread topic'): 3},
         'all_mentions': 0,
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1399,6 +1399,7 @@ class TestStreamInfoView:
         mocker.patch.object(self.controller, 'maximum_popup_dimensions',
                             return_value=(64, 64))
         self.controller.model.is_muted_stream.return_value = False
+        self.controller.model.is_pinned_stream.return_value = False
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
         stream_id = 10
         self.controller.model.stream_dict = {stream_id: {'name': 'books',
@@ -1430,6 +1431,17 @@ class TestStreamInfoView:
         mute_checkbox.keypress(size, key)
 
         toggle_mute_status.assert_called_once_with(stream_id)
+
+    @pytest.mark.parametrize('key', (*keys_for_command('ENTER'), ' '))
+    def test_checkbox_toggle_pin_stream(self, mocker, key):
+        pin_checkbox = self.stream_info_view.widgets[4]
+        toggle_pin_status = self.controller.model.toggle_stream_pinned_status
+        stream_id = self.stream_info_view.stream_id
+        size = (20, 20)
+
+        pin_checkbox.keypress(size, key)
+
+        toggle_pin_status.assert_called_once_with(stream_id)
 
 
 class TestMsgInfoView:
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -2443,9 +2443,8 @@ class TestStreamButton:
     def test_text_content(self, mocker,
                           is_private, expected_prefix,
                           width, count, short_text, caption='caption'):
-        mocker.patch(STREAMBUTTON + ".mark_muted")
         controller = mocker.Mock()
-        controller.model.muted_streams = {}
+        controller.model.is_muted_stream.return_value = False
         properties = [
             caption, 5, '#ffffff', is_private, 'Some Stream Description'
         ]
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1399,8 +1399,10 @@ class TestStreamInfoView:
         mocker.patch.object(self.controller, 'maximum_popup_dimensions',
                             return_value=(64, 64))
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
-        self.stream_info_view = StreamInfoView(self.controller, color='',
-                                               desc='', title='# stream-name')
+        stream_id = 10
+        self.controller.model.stream_dict = {stream_id: {'name': 'books',
+                                                         'description': 'hey'}}
+        self.stream_info_view = StreamInfoView(self.controller, stream_id)
 
     @pytest.mark.parametrize('key', {*keys_for_command('GO_BACK'),
                                      *keys_for_command('STREAM_DESC')})
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1398,6 +1398,7 @@ class TestStreamInfoView:
         self.controller = mocker.Mock()
         mocker.patch.object(self.controller, 'maximum_popup_dimensions',
                             return_value=(64, 64))
+        self.controller.model.is_muted_stream.return_value = False
         mocker.patch(VIEWS + ".urwid.SimpleFocusListWalker", return_value=[])
         stream_id = 10
         self.controller.model.stream_dict = {stream_id: {'name': 'books',
@@ -1418,6 +1419,17 @@ class TestStreamInfoView:
         super_keypress = mocker.patch(VIEWS + '.urwid.ListBox.keypress')
         self.stream_info_view.keypress(size, key)
         super_keypress.assert_called_once_with(size, expected_key)
+
+    @pytest.mark.parametrize('key', (*keys_for_command('ENTER'), ' '))
+    def test_checkbox_toggle_mute_stream(self, mocker, key):
+        mute_checkbox = self.stream_info_view.widgets[3]
+        toggle_mute_status = self.controller.model.toggle_stream_muted_status
+        stream_id = self.stream_info_view.stream_id
+        size = (20, 20)
+
+        mute_checkbox.keypress(size, key)
+
+        toggle_mute_status.assert_called_once_with(stream_id)
 
 
 class TestMsgInfoView:
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -451,7 +451,8 @@ def classify_unread_counts(model: Any) -> UnreadCounts:
             unread_counts['streams'][stream_id] = count
         else:
             unread_counts['streams'][stream_id] += count
-        unread_counts['all_msg'] += count
+        if stream_id not in model.muted_streams:
+            unread_counts['all_msg'] += count
 
     # store unread count of group pms in `unread_huddles`
     for group_pm in unread_msg_counts['huddles']:
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -623,8 +623,16 @@ class Model:
 
     def _handle_subscription_event(self, event: Event) -> None:
         """
-        Handle changes in subscription (eg. muting/unmuting streams)
+        Handle changes in subscription (eg. muting/unmuting,
+                                        pinning/unpinning streams)
         """
+        def get_stream_by_id(streams: List[List[Any]], stream_id: int
+                             ) -> List[Any]:
+            for stream in streams:
+                if stream[1] == stream_id:
+                    return stream
+            return []
+
         if hasattr(self.controller, 'view'):
             if event.get('property', None) == 'in_home_view':
                 stream_id = event['stream_id']
@@ -640,7 +648,28 @@ class Model:
                 else:  # Muting streams
                     self.muted_streams.add(stream_id)
                     stream_button.mark_muted()
+                self.controller.update_screen()
+            elif event.get('property', None) == 'pin_to_top':
+                stream_id = event['stream_id']
 
+                # FIXME: Does this always contain the stream_id?
+                stream_button = (
+                    self.controller.view.stream_id_to_button[stream_id]
+                )
+
+                if event['value']:
+                    stream = get_stream_by_id(self.unpinned_streams, stream_id)
+                    if stream:
+                        self.unpinned_streams.remove(stream)
+                        self.pinned_streams.append(stream)
+                else:
+                    stream = get_stream_by_id(self.pinned_streams, stream_id)
+                    if stream:
+                        self.pinned_streams.remove(stream)
+                        self.unpinned_streams.append(stream)
+                sort_streams(self.unpinned_streams)
+                sort_streams(self.pinned_streams)
+                self.controller.view.left_panel.update_structure()
                 self.controller.update_screen()
 
     def _handle_typing_event(self, event: Event) -> None:
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -621,6 +621,18 @@ class Model:
         response = self.client.update_subscription_settings(request)
         return response['result'] == 'success'
 
+    def is_pinned_stream(self, stream_id: int) -> bool:
+        return stream_id in list(stream[1] for stream in self.pinned_streams)
+
+    def toggle_stream_pinned_status(self, stream_id: int) -> bool:
+        request = [{
+            'stream_id': stream_id,
+            'property': 'pin_to_top',
+            'value': not self.is_pinned_stream(stream_id)
+        }]
+        response = self.client.update_subscription_settings(request)
+        return response['result'] == 'success'
+
     def _handle_subscription_event(self, event: Event) -> None:
         """
         Handle changes in subscription (eg. muting/unmuting,
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -57,6 +57,13 @@ class ServerConnectionFailure(Exception):
     pass
 
 
+def sort_streams(streams: List[List[str]]) -> None:
+    """
+    Used for sorting model.pinned_streams and model.unpinned_streams.
+    """
+    streams.sort(key=lambda s: s[0].lower())
+
+
 class Model:
     """
     A class responsible for storing the data to be displayed.
@@ -570,6 +577,13 @@ class Model:
         for subscription in subscriptions:
             subscription['color'] = canonicalize_color(subscription['color'])
 
+        pinned_streams = [[stream[key] for key in stream_keys]
+                          for stream in subscriptions if stream['pin_to_top']]
+        unpinned_streams = [[stream[key] for key in stream_keys]
+                            for stream in subscriptions
+                            if not stream['pin_to_top']]
+        sort_streams(pinned_streams)
+        sort_streams(unpinned_streams)
         # Mapping of stream-id to all available stream info
         # Stream IDs for muted streams
         # Limited stream info sorted by name (used in display)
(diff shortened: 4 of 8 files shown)


neiljp:
@kaustubh-nair For splitting the second commit, I'd suggest we add the UI change last. Just supporting the subscription events would be a good start - eg. with changes via the webapp.

The muting event change looks pretty good; I'll have another more detailed look later.

zulipterminal/ui_tools/views.py
```diff
@@ -950,14 +950,29 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
         # content is shorter than the title length (+4 Unicode characters).
-        width = max(len(desc) + 2, len(title) + 4)
-        stream_info_content = [urwid.Text(desc, align='center')]
-        super().__init__(controller, stream_info_content, 'STREAM_DESC', width,
-                         title)
+        # 20 is the length required to show stream settings labels.
+        width = max(len(desc) + 2, len(title) + 4, 20)
+        self.controller = controller
+        self.stream_id = stream_id
+        muted_setting = urwid.CheckBox(label="Mute stream",
```
neiljp, line 969: I know this is the label used in the webapp, but it does read a bit strangely to me for some reason. Consistency with the webapp may be important, but this seems like an 'action' rather than a status. Maybe something to think about for a v2, or discuss with frontend.

One idea would be 'muted', which leaves out 'Stream', but perhaps we want a 'Title' above the checkboxes in any case, ie. something like 'Stream settings'?

zulipterminal/ui_tools/views.py
```diff
@@ -950,14 +950,29 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+                 desc: str, stream_id: int, title: str) -> None:
         # Add 4 (for 2 Unicode characters on either side) to the popup title
         # length to make sure that the title gets displayed even when the
         # content is shorter than the title length (+4 Unicode characters).
-        width = max(len(desc) + 2, len(title) + 4)
-        stream_info_content = [urwid.Text(desc, align='center')]
-        super().__init__(controller, stream_info_content, 'STREAM_DESC', width,
-                         title)
+        # 20 is the length required to show stream settings labels.
+        width = max(len(desc) + 2, len(title) + 4, 20)
```
neiljp, line 966: I'm not sure how you plan to develop this further, but it may be useful to simply extract the label text used later and place a `len` expression in here, so the code will express what is happening rather than needing another comment.

zulipterminal/ui_tools/views.py
```diff
@@ -950,14 +950,29 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        self.controller = controller
+        self.stream_id = stream_id
+        muted_setting = urwid.CheckBox(label="Mute stream",
+                                       state=controller.model.is_muted_stream(
+                                           stream_id),
+                                       checked_symbol='✓')
```
neiljp, line 972: I like this, though this may be something we want to standardize on throughout rather than hard-coding. You might like to include that change in a commit with things like extracting the stream 'triangle' and a few other things we've discussed? Maybe in a file in `config`?

zulipterminal/ui_tools/views.py
```diff
@@ -950,14 +950,29 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        urwid.connect_signal(muted_setting, 'change', self.toggle_mute_status)
+
+        # Include blank line between description and stream settings
+        self.stream_info_content = [urwid.Text(desc, align='center'),
```
neiljp, line 983: Do we need this `self`?

zulipterminal/ui_tools/views.py
```diff
@@ -950,14 +950,29 @@ def keypress(self, size: urwid_Size, key: str) -> str:
 
 class StreamInfoView(PopUpView):
     def __init__(self, controller: Any, color: str,
-                 desc: str, title: str) -> None:
+                 desc: str, stream_id: int, title: str) -> None:
```
neiljp, line 961: Possibly for a separate refactor later (or before this), we could consider just passing in the `stream_id`, since we're already calling into the `model` to activate the muting, so could call in to fetch whatever details are required. We don't actually use the color right now.

zulipterminal/ui_tools/views.py
```diff
@@ -950,14 +950,29 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        urwid.connect_signal(muted_setting, 'change', self.toggle_mute_status)
+
+        # Include blank line between description and stream settings
+        self.stream_info_content = [urwid.Text(desc, align='center'),
+                                    urwid.Text('', align='center'),
+                                    muted_setting]
+        super().__init__(controller, self.stream_info_content, 'STREAM_DESC',
+                         width, title)
+
+    def toggle_mute_status(self, button: Any, new_state: bool) -> None:
```
neiljp, line 989: `new_state` is unused; did this code come from somewhere else? Should we name it accordingly or elide it, if it's unused?

tests/ui/test_ui_tools.py
```diff
@@ -1338,6 +1340,15 @@ def test_keypress_navigation(self, mocker,
         self.stream_info_view.keypress(size, key)
         super_keypress.assert_called_once_with(size, expected_key)
 
+    @pytest.mark.parametrize('key', (*keys_for_command('ENTER'), ' '))
+    def test_checkbox_toggle_mute_stream(self, mocker, key):
+        mute_checkbox = self.stream_info_view.stream_info_content[2]
+        toggle_mute_status = self.controller.model.toggle_stream_muted_status
+        stream_id = self.stream_info_view.stream_id
+        size = (10, )
+        mute_checkbox.keypress(size, key)
```
neiljp, line 1350: Blank lines around this line, ideally.

zulipterminal/ui_tools/views.py
kaustubh-nair: Yes, a status message would be more appropriate in case of a checkbox.

zulipterminal/ui_tools/views.py
kaustubh-nair: It's urwids default function signature for callback functions

zulipterminal/ui_tools/views.py
kaustubh-nair: Hmm, I still need a comment to explain why I'm adding 4

zulipterminal/ui_tools/views.py
kaustubh-nair: No, removed.

kaustubh-nair: I need it while testing

zulipterminal/ui_tools/views.py
kaustubh-nair: Added stream header indicator for now, I'll make a new issue for moving the rest of them

kaustubh-nair:
@neiljp Thanks for the review! I updated the PR, and added some refactors as you suggested.
I split the pin/unpin commit - though there are some weird bugs that I can't seem to fix. I'll probably have a look tomorrow.

kaustubh-nair:
Fixed the bugs, this is ready for review!

zulipterminal/helper.py
```diff
@@ -421,7 +421,8 @@ def classify_unread_counts(model: Any) -> UnreadCounts:
             unread_counts['streams'][stream_id] = count
         else:
             unread_counts['streams'][stream_id] += count
-        unread_counts['all_msg'] += count
+        if stream_id not in model.muted_streams:
+            unread_counts['all_msg'] += count
```
kaustubh-nair, line 425: I didn't use `model.is_muted_stream()` because I can't figure out a way to mock it during testing, without rewriting the whole test.

tests/ui/test_ui_tools.py
```diff
@@ -2274,9 +2274,8 @@ class TestStreamButton:
     def test_text_content(self, mocker,
                           is_private, expected_prefix,
                           width, count, short_text, caption='caption'):
-        mocker.patch(STREAMBUTTON + ".mark_muted")
         controller = mocker.Mock()
-        controller.model.muted_streams = {}
+        controller.model.is_muted_stream.return_value = False
```
kaustubh-nair, line 2288: I'm not sure why this works, but `mocker.patch('controller', 'model.is_muted', return_value=False) doesn't.

zulipterminal/config/unicode_characters.py
```diff
@@ -0,0 +1 @@
+BLACK_RIGHT_POINTING_TRIANGLE = '▶'
```
preetmishra, line 1: We may want to use names that are more descriptive so that it is easier for our end-users to depict what modifies which UI element. Also, you could include other Unicode characters that are present in the app.

See also [#zulip-terminal > Unicode characters](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Unicode.20characters/near/862659).

zulipterminal/ui_tools/views.py
preetmishra: @kaustubh-nair Which 4?

kaustubh-nair: Ah, sorry this is outdated
This is what I'm referring to
```python3
        # Add 4 to stream settings label for accomodating checkbox.
        width = max(len(desc) + 2, len(title) + 4,
                    len(muted_setting.label) + 4)
```

zulipterminal/config/unicode_characters.py
kaustubh-nair: @preetmishra We could do that but I thought it would be better to stick to standard names from wikipedia?
The problem with names related to UI elements is that the same characters may be reused in a lot of places, for e.g. lineboxes.

neiljp: @kaustubh-nair There are quite a few cases not included in commit - I thought you were going to put this into a separate PR/issue?

I agree with @preetmishra that we probably want semantic names rather than direct ones, ie. here the equivalent of 'stream-topic-separator' rather than the currently-used version. That would mean we could change in one place and all the code+tests should update, and ZT run with different characters.

tests/helper/test_helper.py
```diff
@@ -190,13 +190,13 @@ def test_powerset(iterable, map_func, expected_powerset):
...
     ([99], [['Some general stream', 'Some general unread topic']], {
-        'all_msg': 9,
+        'all_msg': 8,
         'streams': {99: 1},
         'unread_topics': {(99, 'Some private unread topic'): 1},
         'all_mentions': 0,
     }),
     ([1000], [['Secret stream', 'Some private unread topic']], {
-        'all_msg': 11,
+        'all_msg': 8,
```
neiljp, line 199: If these two values vary, is this really a bugfix rather than refactor?

zulipterminal/helper.py
neiljp: As per another comment, is this a bugfix for muted streams?

zulipterminal/ui_tools/buttons.py
```diff
@@ -164,7 +164,7 @@ def __init__(self, properties: List[Any],
 
         # Mark muted streams 'M' during button creation.
         if self.model.is_muted_stream(self.stream_id):
-            self.mark_muted()
+            self.update_widget('M')
```
neiljp, line 167: Did the model changes expose something different here?

Looking at the unread counts handling being here, I'd love to move that into a model method to mark a stream/topic muted; it's rather hacky being in UI code.

zulipterminal/config/unicode_characters.py
kaustubh-nair: There are a lot of unicode characters in lineboxes so it will be a little time-consuming, hence a new PR would be better.

kaustubh-nair: @neiljp How would you specify semantic names for lines in lineboxes that uses a different style in each corner?
Also the same lines are used as pane seperators, hence I thought a general name would be better.

zulipterminal/ui_tools/views.py
kaustubh-nair: To clarify, I meant while writing the tests.
I need a way to obtain the checkbox object while testing it

Refer to tests above in the same commit
```python3
mute_checkbox = self.stream_info_view.stream_info_content[4]
```

tests/helper/test_helper.py
kaustubh-nair: This is testing `classify_unread_counts`, which as per old behavior, did not update `all_msgs` for muted streams. This was done instead in the UI while generating stream buttons. 
This commit moved that part into `classify_unread_counts`, hence there's a change in the values returned.
The overall external behavior did not change, so I'm not sure if it's correct to call it a bugfix

zulipterminal/ui_tools/buttons.py
kaustubh-nair: Yeah, that seems like a useful refactor too.

zulipterminal/ui_tools/views.py
preetmishra: The +4 is `len([UNICODE]space)`.

Also, there is a typo in the comment.

zulipterminal/ui_tools/views.py
```diff
@@ -949,11 +949,13 @@ def keypress(self, size: urwid_Size, key: str) -> str:
 
 
 class StreamInfoView(PopUpView):
-    def __init__(self, controller: Any, color: str,
-                 desc: str, title: str) -> None:
+    def __init__(self, controller: Any, stream_id: int) -> None:
         # Add 4 (for 2 Unicode characters on either side) to the popup title
         # length to make sure that the title gets displayed even when the
         # content is shorter than the title length (+4 Unicode characters).
+        self.controller = controller
```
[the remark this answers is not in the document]
kaustubh-nair: Used in the callback, moved to next commit.

zulipterminal/ui_tools/views.py
```diff
@@ -951,12 +962,16 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        stream_info_content = [('', [desc])]
+        _, column_widths = self.calculate_table_widths(
+            stream_info_content, len(title))
         # Add 4 (for 2 Unicode characters on either side) to the popup title
         # length to make sure that the title gets displayed even when the
         # content is shorter than the title length (+4 Unicode characters).
-        width = max(len(desc) + 2, len(title) + 4)
-        stream_info_content = [urwid.Text(desc, align='center')]
-        super().__init__(controller, stream_info_content, 'STREAM_DESC', width,
+        popup_width = max(len(desc) + 2, len(title) + 4)
```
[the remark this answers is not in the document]
preetmishra: Ah, my bad. I am fixing that in my WIP PR.

zulipterminal/config/unicode_characters.py
kaustubh-nair: I intend to remove this commit, opened #717

zulipterminal/ui_tools/buttons.py
kaustubh-nair: Unrelated, so opened #722

zulipterminal/model.py
```diff
@@ -606,6 +606,18 @@ def toggle_stream_muted_status(self, stream_id: int) -> bool:
         response = self.client.update_subscription_settings(request)
         return response['result'] == 'success'
 
+    def is_pinned_stream(self, stream_id: int) -> bool:
+        return stream_id in list(stream[1] for stream in self.pinned_streams)
+
+    def toggle_stream_pinned_status(self, stream_id: int) -> bool:
```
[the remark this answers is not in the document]
kaustubh-nair: When I add all of them, I usually exceed the line limit so I just add the file with the major changes.

zulipterminal/ui_tools/views.py
```diff
@@ -982,12 +993,16 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
```
preetmishra: @kaustubh-nair Note that the +2 is extra width which we had was for the center alignment. We might want to remove it as we're not using the alignment with `Text`.

zulipterminal/ui_tools/views.py
```diff
@@ -982,12 +993,16 @@ def keypress(self, size: urwid_Size, key: str) -> str:
 class StreamInfoView(PopUpView):
     def __init__(self, controller: Any, color: str,
                  desc: str, title: str) -> None:
+        stream_info_content = [('', [desc])]
+        _, column_widths = self.calculate_table_widths(
```
preetmishra, line 1009: Could we accept the `popup_width` from here and other widths to it later in the subsequent commits? That would avoid the calculation that you're doing in https://github.com/zulip/zulip-terminal/pull/695/commits/89e53d41bc190524e256aeac06fffed1446a4842#diff-5b93264509b3ec67b9370b2bcbaa7798R1002.

zulipterminal/ui_tools/views.py
preetmishra: (See also https://github.com/zulip/zulip-terminal/pull/695#discussion_r456789654.)

zulipterminal/ui_tools/views.py
kaustubh-nair: Yeah, that sounds like a good idea, updated :+1:

zulipterminal/ui_tools/views.py
```diff
@@ -782,6 +782,16 @@ def topics_view(self, stream_button: Any) -> Any:
             )
         return w
 
+    def update_structure(self) -> None:
+        if self.is_in_topic_view:
+            return
```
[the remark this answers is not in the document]
kaustubh-nair: Fixed

zulipterminal/model.py
```diff
@@ -57,6 +57,13 @@ class ServerConnectionFailure(Exception):
     pass
 
 
+def sort_streams(streams: List[List[str]]) -> None:
+    """
+    Used for sorting model.pinned_streams and model.unpinned_streams.
+    """
+    streams.sort(key=lambda s: s[0].lower())
+
+
```
kaustubh-nair, line 66: I hope it is okay to put it here because I can't make it a class method. I don't see a reason to make this a static method, so this was the only choice.
Also, I don't think I understand why `_stream_info_from_subscriptions()` is a static method? Why would you need static methods at all when there is only one `Model` instance?

tests/ui/test_ui_tools.py
```diff
@@ -1417,6 +1421,28 @@ def test_keypress_navigation(self, mocker,
...
+        mute_checkbox.keypress(size, key)
+
+        toggle_mute_status.assert_called_once_with(stream_id)
+
+    @pytest.mark.parametrize('key', (*keys_for_command('ENTER'), ' '))
+    def test_checkbox_toggle_pin_stream(self, mocker, key):
+        pin_checkbox = self.stream_info_view.widgets[4]
+        toggle_pin_status = self.controller.model.toggle_stream_pinned_status
+        stream_id = self.stream_info_view.stream_id
+        size = (200, 20)
```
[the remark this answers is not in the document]
kaustubh-nair: Yes, these should be the same - checkbox is a box widget here. Updated.

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
neiljp: Could we retain focus on that stream too?

zulipterminal/model.py
neiljp: This is fine here for now. I'm expecting there may be more refactoring of stream data in the near future, so just having it local to the model is sufficient.

A static method is just in a different namespace. I think the only difference we could make would be to have it with a leading underscore, but we're not really consistent otherwise right now.

I think `_stream_info_from_subscriptions` might be inlined in future. The advantage is that the individual functions can be tested, and static methods are guaranteed not to modify any data in an object, except potentially that passed in - which can be caught with a type checker, ideally.

neiljp:
@kaustubh-nair I'm merging this now - thanks for working on this and keeping in with the updates! :tada: This definitely addresses the initial design, and I'd love to have this in for the release - including that first commit which we've debated in chat!

The outstanding points I raised are likely:
* coupling model state to the popup checkbox state (whether via toggles or setting absolute state - see above)
* improving stream focus position retention, which already likely exists as #617 

Other than the subscriber count which I mentioned, another extension might be the per-stream notifications? We added support for that recently, so it'd be great to make that configurable, though that could require examining which settings take precedence.

PR #697, opened 2020-06-19
Fix unmuting unread counts bug

kaustubh-nair:


zulipterminal/helper.py
```diff
@@ -92,29 +92,30 @@ def _set_count_in_model(new_count: int, changed_messages: List[Message],
         but updates `unread_counts` (which can update the model
         if it's passed in, but is not tied to it).
     """
+    # broader unread counts (for all_* and streams) are updated
+    # later conditionally.
```
neiljp, line 96: In the previous refactoring we missed the point that this now occurs 'later' - in another method! So it would be useful to improve the comment here.

zulipterminal/helper.py
```diff
@@ -92,29 +92,30 @@ def _set_count_in_model(new_count: int, changed_messages: List[Message],
...
+            if unreads[key] == 0:
+                unreads.pop(key)
+        elif new_count == 1:
+            unreads[key] = new_count
+
     for message in changed_messages:
         if message['type'] == 'stream':
             key = (message['stream_id'], message['subject'])
-            unreads = unread_counts['unread_topics']
+            update_unreads(unread_counts['unread_topics'], key)
```
neiljp, line 109: Perhaps it would be clearer here to also put the `key`s inline, so we don't then need the key variables, including the extra variable you add in the second commit.

zulipterminal/helper.py
```diff
@@ -92,29 +92,30 @@ def _set_count_in_model(new_count: int, changed_messages: List[Message],
         but updates `unread_counts` (which can update the model
         if it's passed in, but is not tied to it).
     """
+    # broader unread counts (for all_* and streams) are updated
+    # later conditionally.
+    def update_unreads(unreads: Any, key: Any) -> None:
```
neiljp, line 97: This may be typed using `TypeVar`s?

neiljp:
@kaustubh-nair This PR either exposes another bug with unreads, or breaks them slightly. I've not figured out the exact issue, but possibly something like the stream count seems bigger, maybe a topic is higher in the topic list (ie. has new messages?) and when I read it now, the *topic* goes negative (in the topic list).

When this happens I definitely see sum(topic_unreads) != stream_unreads for a given stream, in the topic list.

zulipterminal/helper.py
```diff
@@ -21,7 +21,29 @@
 LINUX = platform.system() == "Linux"
 WSL = 'microsoft' in platform.release().lower()
 
-Message = Dict[str, Any]
+Message = TypedDict('Message', {
```
neiljp, line 24: As I mentioned via PM, this would be a good change to make separately. If you know which parameters require the `total=False` then we can also note that, as with python 3.6 we should be able to inherit `TypedDict`s, and remove the need for every parameter to be non-total. This could also be used effectively for API versioning.

zulipterminal/helper.py
```diff
@@ -53,7 +75,7 @@
     edited_messages=set(),
     topics=defaultdict(list),
     search=set(),
-    messages=defaultdict(dict),
+    messages=defaultdict(lambda: Message()),
```
neiljp, line 78: Good catch of the `mypy` bug! It would be useful to put a comment here referring to why it is necessary to use the `lambda`, referring to eg. the `mypy` issue.

neiljp:
@kaustubh-nair Great to have this bug fixed and the extra typing is going to be useful :+1: Merged with minor adjustment to first commit title :tada:

Issue #551, opened 2020-03-21
Going through results of a search which is "empty" looks confusing


neiljp:
Currently we appear to treat an empty search string the same as regular ones, which is confusing since on 'enter' it shows an empty search box and jumps to eg. the full selection of streams, leaving the search box blank.

Solutions may be to treat 'enter' with an empty string:
* the same as 'esc', ie. that an empty string search should cancel the search (and return somewhere?)
* the same as 'esc', but moving explicitly into the stream list at some point?
* as doing nothing: checking if the box contains a valid value before allowing enter to do anything, ie. locking the user into the box until they press esc explicitly

This is marked as a good first issue since I expect it to be fairly straightforward, but we may need to discuss different approaches in the #**zulip-terminal** stream on chat.zulip.org first.

meelanp:
@zulipbot claim

zulipbot:
Hello @meelanp, it looks like we've already sent you a collaboration invite at https://github.com/zulip/zulip-terminal/invitations, but you haven't accepted it yet!

Please accept the invite and try to claim this issue again afterwards. We look forward to your contributions!

meelanp:
@zulipbot claim

zulipbot:
Hello @meelanp, it looks like you've currently claimed 1 issue in this repository. We encourage new contributors to focus their efforts on at most 1 issue at a time, so please complete your work on your other claimed issues before trying to claim this issue again.

We look forward to your valuable contributions!

meelanp:
@zulipbot claim

PR #699, opened 2020-06-19
[WIP] panel_search_box: Enter on empty string now cancels the search.

meelanp:
An empty search string shows an empty search box and
jumps to eg. the full selection of streams,
leaving the search box blank. This caused confusion because
it was an empty search.

This fix causes an empty string search,
with enter, to now act the same as 'esc';ie. it
cancels the search and returns to the full list of streams/users.
A test for this behavior was also added.

Fixes: #551.


meelanp:
@neiljp 

Thanks for reviewing! Regarding message search, I see a couple inconsistencies. If a search is performed and has a non-empty set of results, and then esc is pressed, the results are reset to all messages and the focus is set to that. However, if a search is performed with an empty string or a non-empty string that results in a empty set of results, and then esc is pressed, the search text is removed but the results are not reset to all messages. Furthermore, if there is a search that results in a empty set of results, the green "Search Results" tag is not added to the search bar. 

I am not sure if all this is a bug or intended behavior. Though, i think the inconsistent behavior with esc and an empty set of results should be resolved, possibly as a separate issue, so that if i implement a fix to cause empty string searches to reset the results, it will be consistent with the behavior of esc. 

This is probably a bit confusing to read but if you try it yourself I think you will see what I mean. Let me know what you think.

neiljp:
@meelanp Message search does have other issues, it's true, but I was trying to address the 'empty search' point in the first instance.

The issue with `esc` being pressed in the central column, including in the message search, is that this is used to reset to the all-messages 'narrow', so it is ambiguous what key should be used to exit search.  However, that doesn't mean that `enter` with an 'empty' search has to trigger the `esc` key in the same way as for `PanelSearchBox`.

Empty narrows are a general issue which we should return to fixing, likely via #278; I'm not sure if @sumanthvrao will have time to rebase that in the short term?

@meelanp If you want to focus on the PanelSearchBox for now, based on my review, we can merge that commit separately in any case.

meelanp:
@neiljp When you say first instance, do you mean when the user's first/initial message search is with an 'empty' string? As in, if the first/initial search is with an 'empty' string, act as `esc`, ie. return focus to the messages.

neiljp:
@meelanp In my first paragraph, I meant a solution for hitting enter when there is an 'empty' search (initially `''`; I'm not sure if the same extension applies to messages as to streams etc), in which case we should likely just cancel the search, like you wrote code for in the other case. The difference here is that we don't want it to *literally* trigger an `esc` - but maybe return to a sensible 'previous focus'. We do that for the other searches too, though not as cleverly as we could do, perhaps.

preetmishra:
> I also just found another bug related to our recent search changes, where a space gives zero results

@neiljp Thanks for the report. :+1: 

The existing `PanelSearchBox`'s update functions don't handle spaces. We didn't tweak these in our recent search improvements.

The PR that is linked above has a fix.

meelanp:
@neiljp  I added a check for space-only strings and 'empty' message searches now return to the body. Let me know what you think. I'm going to briefly look into generalizing the tests now. Also, as a heads up and if it is possible, getting this pull request merged before Thursday, June 25th 11am EST. will get me some extra credit on my school assignment! If not, no worries! I will be actively working on this today and tomorrow.

neiljp:
@meelanp Are your commits a series of iterations towards a working version to consider as one change, or separate commits? It perhaps seems like the former, whereas we tend to merge the latter so it's clearer what the intentions of each commit are. You could have separate commits for the panelsearchbox and searchbox if that improves the clarity of your changes. I'm not sure how much git experience you have but there is a Zulip git guide in the main project, and specifically here an interactive git rebase may help adjust earlier commits. If you focus each commit on one change, it might be easier to merge them individually.

In addition, could you rebase against current master from github? You appear to have a change in one commit that looks like some recent changes we merged.

@preetmishra @meelanp It seems that #700 resolves the separate issue of filtering as-we-type in the panelbox with 'empty' characters, vs in this PR what we do if we try and search with just 'empty' characters? These both seem important to fix, but it would be useful to synchronize them afterwards to ensure that both pieces of code focuses on avoiding the same set of characters.

neiljp:
To clarify:
* @preetmishra #700 seems to fix a specific bug in the current implementation :+1:
* I'm leaning towards leaving the `space` case in searching for now, as perhaps we explicitly want to be able to search for it - ie. results with a space in. Our new match-stream resolution is allowing multiple words, but maybe at the cost of not being able to search for divider characters. This is not a priority right now, but also I think we shouldn't focus on a solution until we know what we a space should mean!
* @meelanp For this issue, let's just leave it for the empty string, as that will simplify things, in line with my point above.

I hope that's clear and hasn't confused the issue - we can certainly discuss further here or on chat.zulip.org.

meelanp:
@neiljp Okay, I will take the space case out of my checks. I don't have too much experience with git, but I'll look into adjusting the commits to a commit for PanelSearchBox, SearchBox, and test_boxes.

neiljp:
@meelanp One commit will be fine for what you have; if it's one change though, it would be good to squash the commits you've made together, which again you can do with an interactive rebase - it's a good git skill to learn!

meelanp:
@neiljp I definitely just messed something up... Not sure what to do here. Is there a way to just start over fresh?

preetmishra:
@meelanp You could run `$ git reset --hard origin/master` to reset your branch. That would help you start afresh. 

Note that before resetting you might want to `cherry-pick` or reference the new commit (that you worked on) in some other way.

meelanp:
@preetmishra @neiljp Thanks! I reset to upstream/master, implemented my changes again, and created a new pull request at #701 . I created a feature branch this time.

neiljp:
@meelanp My personal workflow is to `git fetch upstream && git rebase upstream/master` when I have `master` checked out, before I branch something new from recent work (such as with `git checkout -b new-branch-name`). That same rebase command should tidy your existing branch, but not knowing quite what you did it's a little difficult to identify. Also since you worked on your local master that can get slightly more confusing too.

PR #700, opened 2020-06-23
bugfix: Left strip search text to avoid empty search results

preetmishra:
This amends `PanelSearchBox`'s update functions to avoid empty search results for spaces.

@neiljp Thanks for reporting the bug in https://github.com/zulip/zulip-terminal/pull/699#pullrequestreview-434369722. :+1:

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -317,6 +317,7 @@ class StreamsView(urwid.Frame):
 
             self.log.clear()
             self.log.extend(streams_display)
+            self.stream_search_box.reset_active_search_focus()
             self.view.controller.update_screen()
 
     def mouse_event(self, size: urwid_Size, event: str, button: int, col: int,
@@ -386,6 +387,7 @@ class TopicsView(urwid.Frame):
             ]
             self.log.clear()
             self.log.extend(topics_to_display)
+            self.topic_search_box.reset_active_search_focus()
             self.view.controller.update_screen()
 
     def update_topics_list(self, stream_id: int, topic_name: str,
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -283,6 +283,8 @@ class StreamsView(urwid.Frame):
 
     @asynch
     def update_streams(self, search_box: Any, new_text: str) -> None:
+        new_text = new_text.lstrip()
+
         if not self.view.controller.is_in_editor_mode():
             return
         # wait for any previously started search to finish to avoid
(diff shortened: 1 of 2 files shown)


neiljp:
@preetmishra I'd like to get #699 fixed in a similar way, as I commented there, but this looks reasonable as a v0 in terms of the specific issue I mentioned there, from a quick test.

However, I just gave this a try and for at least the stream list we seem to scroll it even with a space. Is there a different way to fix this that would resolve that too, do you think? Or is that a different bug?

zulipterminal/ui_tools/views.py
```diff
@@ -331,7 +331,7 @@ def __init__(self, topics_btn_list: List[Any], view: Any,
 
     @asynch
     def update_topics(self, search_box: Any, new_text: str) -> None:
-        if not self.view.controller.is_in_editor_mode():
+        if not self.view.controller.is_in_editor_mode() or not new_text:
```
preetmishra, line 336: The existing code in `update_user_list` already handles empty search which is why this commit doesn't have any change for user search.

preetmishra:

Re scrolling issue, I gather it is a different bug. I have added a fix, in a separate commit, to resolve this.

preetmishra:
Note: The stream/topic list doesn't get updated when we completely clear out/<kbd>backspace</kbd> an active search as well. This might feel a bit unintuitive but this behaviour already exists in the user search.

Would we want to eradicate this behaviour from all of the `PanelSearchBox`'s update functions or leave them as they are?

neiljp:
@preetmishra The behavior of the user search has been what I consider buggy for some time - eg. `p` `backspace` doesn't return to an empty case. That this PR seems to have a regression like the user list for the other (panel) searches, makes this worse. I would hope that deleting back to nothing resets to show all the options, as otherwise I don't know what the 'global' list might look like. Admittedly if you have so many entries that you can't see them at all, this doesn't help, but I can't see much of a reason to *not* reset the display if the search is empty?

Another aspect that this appears to fail on is that while the left-stripped search strings do find results, `<space>` `<enter>` doesn't cancel the search, ie. the search string is treated as meaningful, like before @meelanp 's commit. That is, `<enter>` exits the search, but `<space><enter>` does not, which seems strange if `<space>` doesn't mean anything for searches, at least at the left.

preetmishra:
@neiljp Thanks for the review and the exposition. In the latest update, I have resolved the user search issue, the cancel issue and reworked the approach to fix the scrolling issue.

However, the scrolling issue [fix](https://github.com/zulip/zulip-terminal/pull/700/commits/cb285de0469eabb49707d0d6ee3cafc1ec9852cf), the last commit, has a regression where the stream/topic list is shifted when `esc` is pressed post a search (note that the `focus_index_before_search` logic still works). I can't seem to figure out what is causing this particular behaviour. I would greatly appreciate any pointers. 

Besides, the first three commits work as expected and should be good to go.

neiljp:
@preetmishra Thanks for working on this and the associated earlier commits - which were merged already. As per agreement elsewhere, we'll likely move forward with an approach like in #784, though we can always return to this if we change our mind :+1:

PR #701, opened 2020-06-24
boxes: Enter on a empty string cancels the stream/user/message search.

### LINKED ISSUE #551: Going through results of a search which is "empty" looks confusing

meelanp:
An empty search string shows an empty search box and
jumps to eg. the full selection of streams,
leaving the search box blank. This caused confusion because
it was an empty search.

This fix causes an empty string search for a stream/user,
with enter, to now act the same as 'esc';ie. it
cancels the search and returns to the full list of streams/users.
For messages, it returns to the body.
A test for this behavior in PanelSearchBox was also added.

Fixes: #551.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -200,28 +200,34 @@ class TestPanelSearchBox:
         assert panel_search_box.caption == ""
         assert panel_search_box.edit_text == panel_search_box.search_text
 
-    @pytest.mark.parametrize("log, expect_body_focus_set", [
-        ([], False),
-        (["SOMETHING"], True)
+    @pytest.mark.parametrize("log, expect_body_focus_set, edit_text", [
+        ([], False, ""),
+        ([], False, "some search with no results"),
+        (["SOMETHING"], True, "key words")
     ])
     @pytest.mark.parametrize("enter_key", keys_for_command("ENTER"))
     def test_keypress_ENTER(self, panel_search_box,
-                            enter_key, log, expect_body_focus_set):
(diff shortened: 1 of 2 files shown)


neiljp:
@meelanp To get the fix in, I've tidied up your changes to `boxes.py` and merged them manually with a few minor adjustments as f4dc8198044876232ca960b9aaeadf2843cbba27, since we can work on the tests separately as you develop your git skills further :tada:

I see you've got some idea of how the parametrize-ation works, but these asserts look like they're closer to those in a different test, which matches how you've coded it too. The names of existing tests can be changed, if necessary, in case that made you want to put the change in this test function instead?

We can close this and you can work on tests in another PR, or just rebase your changes and resolve the (likely minor) conflicts due to my minor changes and work on it here.

neiljp:
This looks like it could be closed with the adjustment/addition of tests to cover what was the new/fixed behavior.

PR #703, opened 2020-06-26
MessageBox: Extract in-line message links from the message content and show them as footlinks

### LINKED ISSUE #618: List links at end of message (and/or in popup?)
There is already an existing older PR for listing the links in a message in a popup and handling them (and would benefit from being broken into smaller commits). Discussion in #**zulip-terminal** in the 'links' topic suggested the approach below as an alternative or additional route to take, and in any case extracting the links may be useful for both cases (check that discussion first in any case)

Currently links are shown pretty much as in raw markdown, which is workable but can be very verbose and disrupt the flow of a message, for such simple cases as the linkifiers, eg. `This might be solved with #T1, but maybe not` in chat.zulip.org expands to `This might be solved with [#T1](https://github.com/zulip/zulip-terminal/pull/1), but maybe not`, which disrupts the flow in the middle of the sentence.

One idea to improve the rendering has been to use a more 'reference list' or 'footnote' based approach, so the above sentence might render as:
```
This might be solved with #T1 [1], but maybe not
1 : https://github.com/zulip/zulip-terminal/pull/1
```
where the `#T1`, `[1]` and second line has different styling somehow to indicate they fit together (and without requiring a blank line)

preetmishra:
The initial commits extract in-line message links from the message content and add them below it to improve the reading experience. 

The commits are small and incremental in order to make reviewing manageable. Though, they could be squashed before merging.

The last commit adds in-line message links to `MsgInfoView` (needs discussion; see in-line comment).

Fixes #618 and #452 partially.


zulipterminal/ui_tools/boxes.py
```diff
@@ -676,6 +673,7 @@ def transform_content(self) -> Tuple[None, Any]:
             padding_char = '░'
             self.indent_quoted_content(soup, padding_char)
 
+        self.link_index = 1
```
preetmishra, line 677: For maintaining a link counter, I couldn't make a recursive approach work and eventually had to go with an attribute. I could use pointers on how to make it clearer.

zulipterminal/config/keys.py
```diff
@@ -99,6 +99,11 @@
         'help_text': 'New message to a person or group of people',
         'key_category': 'msg_actions',
     }),
+    ('MSG_LINKS', {
+        'keys': {'m'},
```
preetmishra, line 103: Which key would we want to warrant for `MSG_LINKS`?

zulipterminal/ui_tools/boxes.py
```diff
@@ -476,14 +476,11 @@ def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
...
-                    # If the link and text are same
-                    # usually the case when user just pastes
-                    # a link then just display the link
-                    markup.append(('msg_link', text))
-                else:
-                    markup.append(
-                        ('msg_link', '[' + text + ']' + '(' + link + ')'))
+                text = text if text else link
+                markup.append(
+                    ('msg_link', '{} [{}]'.format(text, self.link_index))
```
preetmishra, line 482: Currently, I am showing link as text whenever text is not present. This, however, could be noisy when the link is comparatively long.

zulipterminal/ui_tools/views.py
```diff
@@ -901,9 +901,12 @@ def __init__(self, controller: Any, title: str, *, zt_version: str,
 
 class MsgLinksView(PopUpView):
     def __init__(self, controller: Any, msg: Message, title: str) -> None:
-        width = len(title) + 4
+        width = max(len(title) + 4, 64)
         widgets = []
-
+        for index, (text, link) in enumerate(msg['links'].items(), start=1):
+            widgets.append(
+                urwid.Text('[{}]: {} - {}'.format(index, text, link))
```
preetmishra, line 908: As link can be extremely long, I am inclining towards only showing text with mark - `[1]: Text` in `MsgLinksView`. Thoughts?

preetmishra: Also, I haven't used `PopUpView` here as I am planning to use buttons from Shreya Malviya's old PR later (for handling internal/external links).

zulipterminal/ui_tools/views.py
```diff
@@ -899,6 +899,14 @@ def __init__(self, controller: Any, title: str, *, zt_version: str,
         super().__init__(controller, widgets, 'ABOUT', popup_width, title)
 
 
+class MsgLinksView(PopUpView):
```
preetmishra, line 902: The related issue, #618, and the discussion [here](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/links) mentioned that we would want links to be listed at the end (with a different style) and combined with the `MsgInfoView`. 

On the contrary, this doesn't add links in `MsgInfoView` considering how 'Reactions' can potentially take a significant amount of space in the view and how it would be easier to work with a popup [than trailing links in `MessageBox`] later (e.g. handling internal/external links). Thoughts?

zulipterminal/ui_tools/boxes.py
neiljp: Could we use `state` for this?

zulipterminal/ui_tools/boxes.py
neiljp: Having duplicate content seems confusing and noisy, yes.

zulipterminal/ui_tools/boxes.py
```diff
@@ -200,11 +200,11 @@ def __init__(self, message: Message, model: Any,
                  last_message: Any) -> None:
         self.model = model
         self.message = message
+        self.message['links'] = OrderedDict()
```
neiljp, line 203: Currently we treat the message data as essentially the data raw from the API. I'm a little concerned about adding extra data there since we don't know if the server may add a `links` data structure for messages, which this will then override.

zulipterminal/ui_tools/views.py
neiljp: I'd suggest for a v1 we can integrate links as the end of a message, and also as a category above reactions in the message-info-view. I don't expect too many links in a single message?

zulipterminal/ui_tools/views.py
neiljp: I'm not sure precisely what you mean here, but perhaps this warrants discussion in light of checkboxes being used with @kaustubh-nair 's stream settings work, to work towards a general solution?

kaustubh-nair: Since both PRs are being worked on simultaneously, I think it would be better if we merge them both first and then do a refactor - this would save time as well as make it easier to ensure that the refactor works for both the PRs

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Yeah, I'd prefer only having one when there's no text.

zulipterminal/ui_tools/boxes.py
preetmishra: (Resolved in PM.)

zulipterminal/ui_tools/boxes.py
preetmishra: As per the recent discussion, the inclination is towards only using the link in either case.

zulipterminal/ui_tools/views.py
preetmishra: In the latest update, I have used the existing code from `PopUpView` but, yes, the use-case will likely evolve if we decide to incorporate new changes over the course of time. I believe it would be reasonable to refactor later.

zulipterminal/ui_tools/views.py
```diff
@@ -968,14 +968,20 @@ def __init__(self, controller: Any, color: str,
...
         self.msg = msg
 
         msg_info = [
             ('', [('Date & Time', time.ctime(msg['timestamp'])[:-5]),
                   ('Sender', msg['sender_full_name']),
                   ('Sender\'s Email ID', msg['sender_email'])]),
         ]
+        if message_links:
+            msg_info.append(('Message Links (see footlinks)', [
+                (str(index), text)
```
preetmishra, line 982: I attempted adding `text` and `link` but they disrupt `MsgInfoView` and trigger the wide column issue that we decided to tackle later in https://github.com/zulip/zulip-terminal/pull/659#discussion_r434257471. That said, even if we fix the issue, we'd end up with very wide `MsgInfoView` mostly.

I have added `index` and `text`, for now, but they don't seem to add much value.

Would it be desirable to introduce a message links popup inside `MsgInfoView` (like the edit history popup)?

zulipterminal/ui_tools/boxes.py
```diff
@@ -476,14 +476,11 @@ def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
...
 
-                if link == text:
-                    # If the link and text are same
-                    # usually the case when user just pastes
-                    # a link then just display the link
-                    markup.append(('msg_link', text))
-                else:
-                    markup.append(
-                        ('msg_link', '[' + text + ']' + '(' + link + ')'))
+                text = text if text else link
```
preetmishra, line 497: I was wondering if it would be better to trim the links that do not have any text to improve the flow of a message? We'd show the complete link in the footnote anyway.

preetmishra:
@neiljp Thanks for the feedback. I have incorporated your suggestions in the latest update.

neiljp:
Another potentially straightforward change would be to simplify/compact any duplicate links, so that we don't end up with
```
1: google.com
2: duckduckgo.com
3: google.com
```
but rather just the first two and a re-use of [1] in the text in place of [3].

zulipterminal/ui_tools/boxes.py
```diff
@@ -403,6 +403,23 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
+            '{}: {}\n'.format(index, link)
+            for index, (text, link) in enumerate(message_links, start=1)
+        ]
+        indexed_links[-1] = indexed_links[-1][:-1]  # Remove the last newline.
+
+        footlinks = [
+            ('msg_footlink', indexed_link)
+            for indexed_link in indexed_links
+        ]
+        return urwid.Padding(urwid.Text(footlinks), align='left', left=10,
```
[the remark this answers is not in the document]
preetmishra: **Before**:

Screenshot from 2020-07-06 23-11-05

**After**:

Screenshot from 2020-07-06 23-17-20

I prefer the _after_ as it adds a bit of distinction, between the content and the footlinks, and gives it more width. Additionally, I don't see much interference. Thoughts?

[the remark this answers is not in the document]
neiljp: I'm not sure I have a strong preference at this stage, but I think this change will be easier to adjust later, and may change if we have either of
* link text & names (eg. link & google.com) styled as currently, but the numbers styled the same with the new style
* the ellipsis to avoid long links wrapping

zulipterminal/ui_tools/boxes.py
preetmishra: Yes, I meant the last segment. I have applied it in the latest update.

zulipterminal/ui_tools/boxes.py
```diff
@@ -477,6 +478,7 @@ def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
                     link = urljoin(self.model.server_url, link)
 
                 text = text if text else link
+                self.message_links.append((text, link))
```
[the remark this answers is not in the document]
preetmishra: I have amended and moved this in the commit where this is used and tested. The methods which use this change have the respective tests.

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
preetmishra: I agree. I have pushed the new update with all these three changes - namely _after_ alignment, new styling only for link indexes and ellipsis for now.

zulipterminal/ui_tools/views.py
```diff
@@ -812,13 +813,14 @@ def calculate_table_widths(contents: PopUpViewTableContent,
...
-                ]
-                strip_widths.append(max_row_lengths)
+                if isinstance(row, tuple):
+                    # Measure the longest line if the text is seperated by
+                    # newline(s).
+                    max_row_lengths = [
+                        len(max(text.split('\n'), key=len))
+                        for text in row
+                    ]
+                    strip_widths.append(max_row_lengths)
```
preetmishra, line 824: I haven't considered other widgets' width because they disrupt `MsgInfoView`.

zulipterminal/ui_tools/boxes.py
```diff
@@ -476,14 +501,24 @@ def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
...
+                text = text if text else link
+
+                # Detect duplicate links to save screen real estate.
+                if link not in self.message_links:
+                    self.message_links[link] = (text, self.link_index)
+                    self.link_index += 1
                 else:
-                    markup.append(
-                        ('msg_link', '[' + text + ']' + '(' + link + ')'))
+                    saved_text, saved_link_index = self.message_links[link]
```
preetmishra, line 514: The else bit is to retain `text` if the link is already present in `message_links`.

preetmishra:
@neiljp Thanks for the comprehensive and thorough review.

I have reworked the commit structure. The first commit should only have the migration that is required from the previous view. Next, the subsequent commits should only have higher-level functionalities incrementally. The 4th and the 5th commit could be squashed together (I have split them for review).

Further, I have:
- Wrapped links in the footlinks with ellipses to save screen state.
- Handled duplicate links - https://github.com/zulip/zulip-terminal/pull/703#issuecomment-653816325.
- Styled link indexes differently than `link`/`text`.
- Trimmed link `text` to only show the last segment of the link if the text is redundant.
- Modified/hacked `PopUpView` to show message links in `MsgInfoView` properly.
- Used the suggested style in https://github.com/zulip/zulip-terminal/pull/703#pullrequestreview-442628731 for `MsgInfoView`.

zulipterminal/ui_tools/boxes.py
```diff
@@ -402,6 +405,28 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
         except Exception:
             return ''
 
+    # Use quotes as a workaround for OrderedDict typing issue.
+    # See https://github.com/python/mypy/issues/6904.
```
neiljp, line 409: Thanks for the comment :+1:

zulipterminal/ui_tools/boxes.py
```diff
@@ -402,6 +405,28 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
+    def footlinks_view(self,
+                       message_links: 'OrderedDict[str, Tuple[str, int]]',
+                       ) -> Any:
```
neiljp, line 412: While we don't currently type return type of urwid widgets, I think it might be clearer if we return `None` instead of the empty string, so that we are effectively returning `Optional[Any]`. That's not worth annotating now, but if we do move to a stage of urwid widget types, `Union[str, <widget>]` will be less clear than `Union[None, <widget>]`, ie. `Optional[<widget>]`.

zulipterminal/ui_tools/boxes.py
```diff
@@ -476,14 +501,24 @@ def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
...
+                    self.message_links[link] = (text, self.link_index)
+                    self.link_index += 1
                 else:
-                    markup.append(
-                        ('msg_link', '[' + text + ']' + '(' + link + ')'))
+                    saved_text, saved_link_index = self.message_links[link]
+                    if saved_text != text:
+                        self.message_links[link] = (
+                            '{}, {}'.format(saved_text, text), saved_link_index
+                        )
```
neiljp, line 518: You don't have a test for this?

zulipterminal/config/themes.py
```diff
@@ -29,6 +29,7 @@
     'reaction_mine': 'standout',
     'msg_mention': 'bold',
     'msg_link': '',
+    'msg_link_index': 'bold',
```
neiljp, line 32: While the previous commit is fine first, if you have this change first, you don't need to change all the uses of it to be the new names.

tests/ui/test_ui_tools.py
```diff
@@ -1526,21 +1526,18 @@ def test_private_message_to_self(self, mocker):
           ('msg_link', 'Foo'), ' ', ('msg_link_index', '[1]'),
           ('msg_link', 'Bar'), ' ', ('msg_link_index', '[2]')]),
         ('<a href="http://foo">http://foo</a>',
-         [('msg_link', 'http://foo'), ' ', ('msg_link_index', '[1]')]),
+         [('msg_link', 'foo'), ' ', ('msg_link_index', '[1]')]),
```
neiljp, line 1529: This seems unexpected, but it might just be because this 'domain name' is so short.

This is something of a tricky problem, as I've seen links which are eg. query strings, so the text becomes `some_file?this_thing`, which looks unexpected (and real cases even more so). I'm reconsidering how we should handle them, though I don't see an easy solution straight away.

For now this looks OK.

zulipterminal/ui_tools/boxes.py
```diff
@@ -502,6 +502,9 @@ def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
                     link = urljoin(self.model.server_url, link)
 
                 text = text if text else link
+                # Only use the last segment if the text is a link itself.
+                if text == link:
+                    text = text.rstrip('/').split('/')[-1]
```
neiljp, line 507: We could consider parsing the URL using the standard library, as we do in other places.

neiljp:
@preetmishra I'm using this now, and I actually wonder if we should avoid showing links in the footlinks if they're just eg. `example.com` with no text - we don't gain anything? The user doesn't gain any information other than perhaps continuity of numbering - seeing example.com twice is actually more text than we had before! It also exposes how we're implementing the auto link/text association, after a message author just wrote what looks like a link with no description. They could still be shown in other contexts (popups).

tests/ui/test_ui_tools.py
```diff
@@ -1654,6 +1669,7 @@ def test_soup2markup(self, content, markup):
         self.model.server_url = SERVER_URL
         # NOTE Absence of previous (last) message should not affect markup
         msg_box = MessageBox(message, self.model, None)
+        msg_box.link_index = 1
```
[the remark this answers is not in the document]
preetmishra: It was necessary as I was initializing the variable in `soup2markup`'s caller method. That said, I gather it better belongs to the `__init__`.

Re `len(message_links)`, I am not sure if this would work here.

zulipterminal/ui_tools/boxes.py
preetmishra: Resolved in PM. (We're moving ahead without the library call for now due to few inconsistencies. Though, we may want to switch to it later.)

zulipterminal/config/themes.py
preetmishra: True. Notwithstanding, the changes in this particular commit express where the new style is used explicitly [in GitHub's diff].

zulipterminal/ui_tools/boxes.py
preetmishra: The 'saved text' is not used anywhere currently and is for the popup (where we'll show `1: Text \n Link`). However, the code is still useful to have so that we don't miss the 'saved text' even though we're not using it.

I have added extra test cases that are remotely related.

zulipterminal/ui_tools/boxes.py
```diff
@@ -468,22 +494,33 @@ def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
                 markup.append(('msg_mention', element.text))
             elif element.name == 'a':
                 # LINKS
-                link = element.attrs['href']
+                link = element.attrs['href'].rstrip('/')
                 text = element.img['src'] if element.img else element.text
+                text = text.rstrip('/')
```
preetmishra, line 503: The rstrips are useful to have to avoid anomalies and edge cases like https://google.com vs https://google.com/.

neiljp: This would be a useful comment :)

preetmishra:
@neiljp Thanks for the review. I have incorporated your suggestions in the latest update and replied to your other in-line comments. Moreover, I have introduced a new commit to address https://github.com/zulip/zulip-terminal/pull/703#issuecomment-656915338.

Additionally, I have dropped the popup commits for now. I will push a new PR after #695 is settled.

preetmishra:
Updated with the comment in https://github.com/zulip/zulip-terminal/pull/703#discussion_r453336979.

preetmishra:
Strange! I pushed the commits successfully when GitHub was down, but they were not showing up. I had to `--amend --no-edit` my commit and force push again.

zulipterminal/config/themes.py
neiljp: I agree that the ordering you have is reasonable, it just ends up with a larger diff.

zulipterminal/ui_tools/boxes.py
```diff
@@ -402,6 +405,28 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
+        for link, (text, index) in message_links.items():
+            footlinks.extend([
+                '{}:'.format(index),
+                ' ',
+                ('msg_link', link),
+                '\n',
+            ])
+        footlinks[-1] = footlinks[-1][:-1]  # Remove the last newline.
+
+        return urwid.Padding(urwid.Text(footlinks, wrap='ellipsis'),
```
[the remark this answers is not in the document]
neiljp: As we discussed elsewhere, when we add the ellipsis, we should add a FAQ entry regarding this.

tests/ui/test_ui_tools.py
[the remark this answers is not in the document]
neiljp: By my reading, `link_index` is really 'next link index', so equivalent to `len(message_links)+1`? So we could avoid needing to synchronize two variables.

zulipterminal/ui_tools/boxes.py
```diff
@@ -413,6 +413,10 @@ def footlinks_view(self,
                        ) -> Any:
         footlinks = []
         for link, (text, index) in message_links.items():
+            # Skip the footlink if its text only has the network location.
+            if text == urlparse(link).netloc:
+                continue
```
neiljp, line 418: I'm not sure if it's just this code/commit or the combination with previous commits - this works well for simple cases that I tested manually, but I think we may wish to be careful with handling the link scheme...or even the entire link?

Given a message of `example.com http://example.com https://example.com`, the webapp shows 3 links expressed as they are in the message (with 'text' exactly as written in the message), the first two pointing to the same address. We point the first two to the same address (based on the HTML, presumably), but we remove the scheme, so they *look* identical but are confusingly not, ie. we generate `example.com [1] example.com [1] example.com [2]`. Given we don't show the footlinks for these now after this commit, this is rather confusing! So, at a minimum, I think we should leave the scheme as it is, and probably use the assumed href in the anchor where no scheme is present?

After some further manual testing, note that even a long link like eg `example.com/blah/foooo/agikhasdglkashdglkashdg/asdfgklhasg/asdjkgalskjdhalskjdhlkasdjhsalkdghsasiougypoadisgypsdgypodsgypoadgysapdgy` gets shown twice identically (in message and footlinks). I don't think we gain from this being in the footlinks, much like with the simple `example.com` case? So, if there is no text, this makes me think we should remove it in all cases. Thoughts?

tests/ui/test_ui_tools.py
```diff
@@ -1515,17 +1516,35 @@ def test_private_message_to_self(self, mocker):
         ('<blockquote>stuff', [('msg_quote', ['', 'stuff'])]),
         ('<div class="message_embed">',
             ['[EMBEDDED CONTENT NOT RENDERED]']),  # FIXME Unsupported
-        ('<a href="http://foo">http://foo</a>', [('msg_link', 'http://foo')]),
+        ('<a href="http://foo">Foo</a><a href="https://bar.org">Bar</a>',
+         [('msg_link', 'Foo'), ' ', '[1]',
+          ('msg_link', 'Bar'), ' ', '[2]']),
+        ('<a href="http://foo">Foo</a><a href="http://foo">Another foo</a>',
+         [('msg_link', 'Foo'), ' ', '[1]',
+          ('msg_link', 'Another foo'), ' ', '[1]']),
```
neiljp: minor: A variant of this showing the stripping of trailing slashes (maybe even with identical text?) would be useful, as I don't think we have a test of that?

neiljp:
Regarding the hiding of all full-text links, I should clarify that if there is a 'file name' then this works well as the 'text' substitute, but otherwise I'm not sure that another part of the URL makes sense on its own.

tests/ui/test_ui_tools.py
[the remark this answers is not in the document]
preetmishra: Thanks for the tip! I have implemented it in the latest update.

zulipterminal/config/themes.py
preetmishra: Yeah, that's there. However, it only fits with the previous commit which is pretty large already.

zulipterminal/ui_tools/boxes.py
preetmishra: Thanks for the report. It turns out that the server translates any scheme-less link to `HTTP` (e.g. `example.com` becomes `<a href="http://example.com">example.com</a>`). The core problem was with the `text == link` check that I had previously. I have replaced it with a 'scheme-less' check to figure out whether the 'text' part is identical.

Moreover, I have amended the last segment bit to only work when the text is not too simple (e.g. `example.com`) and to patch texts like these with their link itself. Consequently, `example.com http://example.com https://example.com` come out as `http://example.com [1] http://example.com [1] https://example.com [2]` and long links work more reliably.

Furthermore, I have not redacted the last segment bit for long links as I believe they improve the reading flow drastically. We could surely improve how we detect the 'significant' segment later but I think they work well for now.

preetmishra: (See also https://github.com/zulip/zulip-terminal/pull/703/commits/a782d546470df674c78e7d52268fd89765b405f0#diff-3e518af2a88ffb14e1771a621641edd5R1552-R1558.)

preetmishra:
@neiljp Thanks for the review and discovering those examples. They helped me greatly to improve the last segment and the skip footlink bit. :+1: 

I have removed the `link_index` attribute and moved to only using the `message_links`, and added tests for the trailing slash bit. Moreover, I have reworked the last segment detection bit and the related skip footlink. Hopefully, they should be more reliable now.

Thanks for helping me with the commit structure. They do seem to tell a 'story' with the incremental changes and look coherent when compared to what we started with. :+1: The duplicate links excerpt is necessary to have with the initial commit otherwise the `OrderedDict` keeps getting overridden and we end up with rather funny link indexes.

Re hiding of all full-text links, see https://github.com/zulip/zulip-terminal/pull/703#discussion_r454576011.

tests/ui/test_ui_tools.py
```diff
@@ -1536,19 +1536,26 @@ def test_private_message_to_self(self, mocker):
...
-         [('msg_link', '{}/some/relative_url'.format(SERVER_URL)), ' ',
-          ('msg_link_index', '[1]')]),
+         [('msg_link', 'relative_url'), ' ', ('msg_link_index', '[1]')]),
+        ('<a href="http://foo.com/bar">foo.com/bar</a>',
+         [('msg_link', 'bar'), ' ', ('msg_link_index', '[1]')]),
+        ('<a href="http://foo.com">foo.com</a>'
+         '<a href="http://foo.com">http://foo.com</a>'
+         '<a href="https://foo.com">https://foo.com</a>'
+         '<a href="http://foo.com">Text</a>',
+         [('msg_link', 'http://foo.com'), ' ', ('msg_link_index', '[1]'),
```
neiljp, line 1555: This is my only remaining query. It's minor really, but if you look at #**commits** then you'll see some on a `chat.zulip.org` branch - it gets shown as `http://chat.zulip.org`, which of course isn't the branch name :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -412,6 +412,11 @@ def footlinks_view(self,
                        ) -> Any:
         footlinks = []
         for link, (text, index) in message_links.items():
+            # Skip the footlink if it is identical to its text (see soup2mark
```
neiljp, line 415: `soup2markup`?

preetmishra:
@neiljp Thanks for the review! In the latest update, I have:
* Fixed the link text behaviour that you mentioned here and we talked about on CZO. Now, `example.com http://example.com https://example.com` come out as `example.com [1] http://example.com [1] https://example.com [2]` (without any footlink) and the last segment is only used when the link text has something else than just the 'domain name'.
* Improved the tests and added the mentioned TODO comment.
* Restructured the tests to amend them incrementally as the refinements are added.

neiljp:
@preetmishra Thanks for the update and all your work on this :+1: I've just merged this! :tada:

The only outstanding thing I think we could tidy up is the cases where we show the last entry (or not), but that's definitely fine tuning now, and we have a lot of test cases that we can re-examine to determine if we want different behavior. On that note, it might be worth investigating if we can incorporate the ids inline (`pytest.param`?), since we have so many parametrizations in `soup2markup` now!

The popup and link-opening work will be great to build off this. Perhaps we can use the same parametrizations as a fixture?

preetmishra:
@neiljp Thanks for the merge! I'll proceed with the follow-ups now. :+1:

PR #705, opened 2020-06-29
Visually separate pinned and unpinned streams while searching

### LINKED ISSUE #577: Separate pinned and unpinned streams while searching.
We already use a divider to separate pinned and unpinned streams in the streams view. However, we don't add a divider while showing the search results at the moment.

We would want to have a divider at both times to address the distinction.

preetmishra:
The first commit is a refactor that extracts `streams_view_divider()` from `streams_view()`. 

The `streams_view_divider()` is used in the second commit, in `update_streams`, for inserting a divider between search results.

Fixes #577.

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -287,6 +287,25 @@ class StreamsView(urwid.Frame):
             ]
             streams_display = match_stream(stream_buttons, new_text,
                                            self.view.pinned_streams)[0]
+
+            # Add a divider to separate pinned streams from the rest.
+            pinned_stream_names = [
+                stream[0]
+                for stream in self.view.pinned_streams
+            ]
+            first_unpinned_index = len(streams_display)
+            for index, stream in enumerate(streams_display):
+                if stream.stream_name not in pinned_stream_names:
+                    first_unpinned_index = index
+                    break
+            if first_unpinned_index not in [0, len(streams_display)]:
+                # Do not add a divider when it is already present. This can
+                # happen when new_text=''.
+                if not isinstance(streams_display[first_unpinned_index],
+                                  StreamsViewDivider):
+                    streams_display.insert(first_unpinned_index,
+                                           StreamsViewDivider())
+
             self.log.clear()
             self.log.extend(streams_display)
             self.view.controller.update_screen()
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -245,6 +245,18 @@ class MessageView(urwid.ListBox):
         self.model.mark_message_ids_as_read(read_msg_ids)
 
 
+class StreamsViewDivider(urwid.Divider):
+    """
+    A custom urwid.Divider to visually separate pinned and unpinned streams.
+    """
+    def __init__(self) -> None:
+        # FIXME: Necessary since the divider is treated as a StreamButton.
+        # NOTE: This is specifically for stream search to work correctly.
+        self.stream_id = -1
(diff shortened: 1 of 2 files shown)


kaustubh-nair:
@preetmishra This looks like a useful visual improvement :+1: 
One bug I noticed is two lines are displayed if I search for something then clear the text with backspace.

preetmishra:
@kaustubh-nair Thanks for the bug report. :+1:

I couldn't notice it initially it as I only have a few streams pinned and the stream list would update past the pinned streams whenever I cleared the search with <kbd>backspace</kbd> (#700 will fix this). However, with a bunch of streams pinned, it was pretty straightforward to reproduce.

I have pushed an update that will resolve the bug.

zulipterminal/helper.py
```diff
@@ -529,6 +530,21 @@ def format_string(names: List[str], wrapping_text: str) -> List[str]:
     return [wrapping_text.format(name) for name in names]
 
 
+def streams_view_divider() -> Any:
```
neiljp, line 533: This is now basically a factory function which initializes a divider in another way - so maybe let's just make this a simple subclass?

zulipterminal/ui_tools/views.py
```diff
@@ -245,6 +245,19 @@ def read_message(self, index: int=-1) -> None:
         self.model.mark_message_ids_as_read(read_msg_ids)
 
 
+class StreamsViewDivider(urwid.Divider):
+    """
+    A custom urwid.Divider to visually separate pinned streams from the
+    unpinned streams.
+    """
+    def __init__(self, div_char: str='-', top: int=0, bottom: int=0) -> None:
```
preetmishra, line 253: Unlike the previous implementation, I have set the `div_char='-'` from the `__init__` instead. This shall help to maintain consistency (between `StreamsView` and `LeftColumnView` stream view dividers) if we ever wish to update the `div_char`.

preetmishra:
@neiljp Thanks for the review and the pointers.

I have extracted the divider as a subclass now and moved it into in `views.py`. The subclass does improve readability and make it cleaner and the code does fit better in `views.py`.

zulipterminal/ui_tools/views.py
```diff
@@ -288,6 +288,25 @@ def update_streams(self, search_box: Any, new_text: str) -> None:
...
+                stream[0]
+                for stream in self.view.pinned_streams
+            ]
+            first_unpinned_index = len(streams_display)
+            for index, stream in enumerate(streams_display):
+                if stream.stream_name not in pinned_stream_names:
+                    first_unpinned_index = index
+                    break
+            if (first_unpinned_index != 0
+                    and first_unpinned_index != len(streams_display)):
```
neiljp, line 303: Can we simplify this with `not in`?

zulipterminal/ui_tools/views.py
```diff
@@ -288,6 +288,25 @@ def update_streams(self, search_box: Any, new_text: str) -> None:
...
+                # Check whether a divider already exists.
+                if not isinstance(streams_display[first_unpinned_index],
+                                  urwid.Divider):
```
neiljp, line 306: `StreamsViewDivider`?

Also, I remember you discussed the need for this section, but could you explain how it can happen in the comment, ie. what situation you're avoiding?

preetmishra:
@neiljp Thanks for the review. I have incorporated all the suggestions in the latest update.

neiljp:
@preetmishra Thanks for working on this! I'm merging this now :tada:

PR #707, opened 2020-07-03
WIP: Add reactions to messages

kaustubh-nair:

PR #708, opened 2020-07-04
Handle internal links

### LINKED ISSUE #352: Cross reference internal links to their respective narrows within ZT
**Internal links** in ZT (those which point to conversation in other narrows in the same realm), are clickable in the sense they are opened in the browser through the web-app which redirect then to the narrow. A similar implementation would be to switch to the respective narrow the link is meant for from within ZT itself (Cross reference). A v2 version of this would be to enable copy/paste the current `link to conversation` in a message later. This when clicked would do the same redirection to the respective target narrow from within ZT. Implementation of this can be split up into phases.

v1 :
-  [ ] Identify the current narrow is an internal link.
-  [ ]   Switch the narrow to the message addressed on clicking the link.

v2 :  
-  [ ]   Enable copy/paste of `link to conversation` - terminal/OS based solution.

preetmishra:
The PR adds support for handling internal links - stream, topic and near narrow.

#### Commits
The commits introduce `MessageLinkButton` in `MsgInfoView` and then incrementally add support for different narrows.

I would greatly appreciate any feedback on the changes, the commit chronology and the structure in general.

Partially fixes #352 and #764.

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -11,6 +11,7 @@ from typing import (
     Any, Callable, DefaultDict, Dict, FrozenSet, Iterable, List, Set, Tuple,
     TypeVar, Union,
 )
+from urllib.parse import unquote
 
 import lxml.html
 from typing_extensions import TypedDict
@@ -629,3 +630,13 @@ def display_error_if_present(response: Dict[str, Any], controller: Any
                              ) -> None:
     if response['result'] == 'error' and hasattr(controller, 'view'):
         controller.view.set_footer_text(response['msg'], 3)
+
+
+def hash_util_decode(string: str) -> str:
+    """
+    Returns a decoded string given a hash_util_encode() [present in
+    zulip/zulip's zerver/lib/url_encoding.py] encoded string.
+    """
+    # Acknowledge custom string replacements in zulip/zulip's
+    # zerver/lib/url_encoding.py before unquote.
+    return unquote(string.replace('.', '%'))
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -266,6 +266,7 @@ ParsedNarrowLink = TypedDict('ParsedNarrowLink', {
     'narrow': str,
     'stream': DecodedStream,
     'topic_name': str,
+    'message_id': Optional[int],
 }, total=False)
 
 
@@ -315,6 +316,16 @@ class MessageLinkButton(urwid.Button):
             stream_name = hash_util_decode(encoded_stream_data)
             return DecodedStream(stream_id=None, stream_name=stream_name)
 
+    @staticmethod
+    def _decode_message_id(message_id: str) -> Optional[int]:
+        """
+        Returns either the compatible near message ID or None.
+        """
+        try:
+            return int(message_id)
+        except ValueError:
+            return None
+
     @classmethod
     def _parse_narrow_link(cls, link: str) -> ParsedNarrowLink:
         """
@@ -344,6 +355,21 @@ class MessageLinkButton(urwid.Button):
             topic_name = hash_util_decode(fragments[4])
             parsed_link = dict(narrow='stream:topic', stream=stream_data,
                                topic_name=topic_name)
+
+        elif (len_fragments == 5 and fragments[1] == 'stream'
+                and fragments[3] == 'near'):
+            stream_data = cls._decode_stream_data(fragments[2])
+            message_id = cls._decode_message_id(fragments[4])
+            parsed_link = dict(narrow='stream:near', stream=stream_data,
+                               message_id=message_id)
+
+        elif (len_fragments == 7 and fragments[1] == 'stream'
+                and fragments[3] == 'topic' and fragments[5] == 'near'):
+            stream_data = cls._decode_stream_data(fragments[2])
+            topic_name = hash_util_decode(fragments[4])
+            message_id = cls._decode_message_id(fragments[6])
+            parsed_link = dict(narrow='stream:topic:near', stream=stream_data,
+                               topic_name=topic_name, message_id=message_id)
 
         return parsed_link
 
@@ -400,6 +426,13 @@ class MessageLinkButton(urwid.Button):
             if topic_name not in self.model.topics_in_stream(stream_id):
                 return 'Invalid topic name'
 
+        # Validate message ID for near.
+        if 'near' in parsed_link['narrow']:
+            message_id = parsed_link.get('message_id')
+
+            if message_id is None:
+                return 'Invalid message ID'
+
         return ''
 
     def _switch_narrow_to(self, parsed_link: ParsedNarrowLink) -> None:
@@ -411,10 +444,21 @@ class MessageLinkButton(urwid.Button):
             self.stream_id = parsed_link['stream']['stream_id']
             self.stream_name = parsed_link['stream']['stream_name']
             self.controller.narrow_to_stream(self)
+        elif 'stream:near' == narrow:
+            self.stream_id = parsed_link['stream']['stream_id']
+            self.stream_name = parsed_link['stream']['stream_name']
+            self.message = dict(id=parsed_link['message_id'])
+            self.controller.narrow_to_stream(self)
         elif 'stream:topic' == narrow:
             self.stream_id = parsed_link['stream']['stream_id']
             self.stream_name = parsed_link['stream']['stream_name']
             self.topic_name = parsed_link['topic_name']
+            self.controller.narrow_to_topic(self)
+        elif 'stream:topic:near' == narrow:
+            self.stream_id = parsed_link['stream']['stream_id']
+            self.stream_name = parsed_link['stream']['stream_name']
+            self.topic_name = parsed_link['topic_name']
+            self.message = dict(id=parsed_link['message_id'])
             self.controller.narrow_to_topic(self)
 
     def handle_narrow_link(self) -> None:
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -265,6 +265,7 @@ DecodedStream = TypedDict('DecodedStream', {
 ParsedNarrowLink = TypedDict('ParsedNarrowLink', {
     'narrow': str,
     'stream': DecodedStream,
+    'topic_name': str,
 }, total=False)
 
 
@@ -337,6 +338,13 @@ class MessageLinkButton(urwid.Button):
             stream_data = cls._decode_stream_data(fragments[2])
             parsed_link = dict(narrow='stream', stream=stream_data)
 
+        elif (len_fragments == 5 and fragments[1] == 'stream'
+                and fragments[3] == 'topic'):
+            stream_data = cls._decode_stream_data(fragments[2])
+            topic_name = hash_util_decode(fragments[4])
+            parsed_link = dict(narrow='stream:topic', stream=stream_data,
+                               topic_name=topic_name)
+
         return parsed_link
 
     def _validate_and_patch_stream_data(self,
@@ -384,6 +392,14 @@ class MessageLinkButton(urwid.Button):
             if error:
                 return error
 
+        # Validate topic name.
+        if 'topic_name' in parsed_link:
+            topic_name = parsed_link['topic_name']
+            stream_id = parsed_link['stream']['stream_id']
+
+            if topic_name not in self.model.topics_in_stream(stream_id):
+                return 'Invalid topic name'
+
         return ''
 
     def _switch_narrow_to(self, parsed_link: ParsedNarrowLink) -> None:
@@ -395,6 +411,11 @@ class MessageLinkButton(urwid.Button):
             self.stream_id = parsed_link['stream']['stream_id']
             self.stream_name = parsed_link['stream']['stream_name']
             self.controller.narrow_to_stream(self)
+        elif 'stream:topic' == narrow:
+            self.stream_id = parsed_link['stream']['stream_id']
+            self.stream_name = parsed_link['stream']['stream_name']
+            self.topic_name = parsed_link['topic_name']
+            self.controller.narrow_to_topic(self)
 
     def handle_narrow_link(self) -> None:
         """
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -1,9 +1,12 @@
-from typing import Any, Callable, Dict, Optional, Tuple, Union
+import re
+from typing import Any, Callable, Dict, Optional, Tuple, Union, cast
+from urllib.parse import urljoin, urlparse
 
 import urwid
+from typing_extensions import TypedDict
 
 from zulipterminal.config.keys import is_command_key, keys_for_command
-from zulipterminal.helper import StreamData
+from zulipterminal.helper import StreamData, hash_util_decode
 from zulipterminal.urwid_types import urwid_Size
 
 
@@ -254,10 +257,23 @@ class UnreadPMButton(urwid.Button):
         self.email = email
 
 
+DecodedStream = TypedDict('DecodedStream', {
+    'stream_id': Optional[int],
+    'stream_name': Optional[str],
+})
+
+ParsedNarrowLink = TypedDict('ParsedNarrowLink', {
+    'narrow': str,
+    'stream': DecodedStream,
+}, total=False)
+
+
 class MessageLinkButton(urwid.Button):
     def __init__(self, controller: Any, caption: str, link: str,
                  display_attr: Optional[str]) -> None:
         self.controller = controller
+        self.model = self.controller.model
+        self.view = self.controller.view
         self.link = link
 
         super().__init__('')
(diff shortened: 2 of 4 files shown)


zulipterminal/ui_tools/buttons.py
```diff
@@ -25,7 +25,6 @@ def __init__(self, controller: Any, caption: str,
             prefix = prefix_character[1]
         else:
             prefix = prefix_character
-        assert len(prefix) in (0, 1)
```
preetmishra, line 28: What's the reasoning behind this particular assert? The buttons don't render properly if the prefix character is >1.

I need numeric (and incremental) `prefix_character` for in-line message links which could be >1 (e.g. 10).

neiljp: This defines how the current button works, ie. it is not guaranteed to work outside those bounds. If you wish it to be able to handle more, then we likely want to refactor the class to work with longer text.

That said, you're not using the extra feature which the prefix allows, which from memory is just a specific style. If you don't need that then you can skip the prefix and just put it all in the text?

preetmishra: Thanks for the pointer. Given the issues with `caption`, `prefix_character` and that the existing methods weren't needed, I ended up implementing a new `urwid.Button` subclass to better tailor my needs for the message links.

preetmishra:
I was meaning to change the `esc` focus for successful internal link narrows but I see the current behaviour aligns with `esc` key behaviour that we already have elsewhere (e.g. web app and zulip-term itself), that is, `esc` transitions to 'All messages' narrow no matter in which narrow you are in at that moment. Thoughts?

zulipterminal/ui_tools/views.py
```diff
@@ -1023,7 +1026,28 @@ def __init__(self, controller: Any, msg: Message, title: str) -> None:
                     grouped_reactions[reaction] = user
             msg_info.append(('Reactions', list(grouped_reactions.items())))
 
+        # Render the category using the existing table methods if links exist.
+        if message_links:
+            msg_info.append(('Message Links', []))
```
preetmishra, line 1031: I am currently showing `message_links` at the bottom since it is straightforward to implement and we were planning to move reactions away from `MsgInfoView` in #707. Though, I am not sure when that would be integrated. Thoughts?

zulipterminal/ui_tools/views.py
```diff
@@ -1023,7 +1026,28 @@ def __init__(self, controller: Any, msg: Message, title: str) -> None:
...
+                text, link_index, _ = message_links[link]
+                caption = '{}: {}\n{}'.format(index, text, link)
+                message_link_width = max(message_link_width, len(caption))
+
+                display_attr = None if index % 2 else 'popup_contrast'
+                message_link_widgets.append(
+                    urwid.AttrMap(urwid.Text(caption), display_attr)
+                )
+
+            widgets.extend(message_link_widgets)
```
preetmishra, line 1050: I am currently showing `message_links` at the bottom since it is straightforward to implement and we were planning to move reactions away from `MsgInfoView` in #707. Though, I am not sure when that would be integrated. Thoughts?

preetmishra: Resolved in PM. (We're moving ahead with `message_links` before 'Reactions'.)

zulipterminal/ui_tools/views.py
```diff
@@ -1002,14 +1003,19 @@ def __init__(self, controller: Any, color: str,
 
 
 class MsgInfoView(PopUpView):
-    def __init__(self, controller: Any, msg: Message, title: str) -> None:
+    def __init__(self, controller: Any, msg: Message, title: str,
+                 message_links: 'OrderedDict[str, Tuple[str, int, bool]]',
```
sumanthvrao, line 1007: If this can be None, then `Optional[MESSAGE_LINKS]`?

zulipterminal/core.py
```diff
@@ -134,9 +135,12 @@ def show_help(self) -> None:
         help_view = HelpView(self, "Help Menu (up/down scrolls)")
         self.show_pop_up(help_view)
 
-    def show_msg_info(self, msg: Message) -> None:
+    def show_msg_info(self, msg: Message,
+                      message_links: 'OrderedDict[str, Tuple[str, int, bool]]',
```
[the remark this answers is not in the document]
preetmishra: Apart from the conversation that was had in PM, `OrderedDict` typing doesn't work as a type alias or var (see [mypy bug](https://github.com/python/mypy/issues/6904)).

zulipterminal/ui_tools/views.py
preetmishra: This can be an empty `OrderedDict` but I believe mypy covers that currently?

preetmishra:
@sumanthvrao Thanks for the review. As discussed in the PM, the first three commits are merged! :tada: 

I have pushed a rebased version for further review.

preetmishra:
Updated to resolve conflicts.

preetmishra:
Updated with tests.

zulipterminal/ui_tools/buttons.py
```diff
@@ -288,11 +288,36 @@ def handle_link(self, *_: Any) -> None:
...
                 # TODO: Open media in the native viewer.
                 pass
         else:
             # TODO: Open external link in the web browser.
             pass
+
+    def handle_narrow(self) -> None:
+        """
+        Handles stream narrow.
+        """
```
neiljp: The commit title (description) for this could be improved - imagine reading just that line after coming back to it - what does this actually do? (without reading the code)

zulipterminal/ui_tools/buttons.py
```diff
@@ -288,11 +288,36 @@ def handle_link(self, *_: Any) -> None:
...
+        # TODO: Handle topic narrow.
+        link_split = self.link.split('/')
+
+        self.stream_id = int(link_split[5].split('-')[0])
```
neiljp: Parsing the URL might be more clear here?

zulipterminal/ui_tools/buttons.py
```diff
@@ -288,11 +288,36 @@ def handle_link(self, *_: Any) -> None:
...
+
+        # Return if the concerned stream is unsubscribed.
+        if self.stream_id not in self.model.stream_dict:
+            text = ' You are not subscribed to the stream'
+            self.view.set_footer_text(text, duration=3)
+            return
+
+        self.stream_name = self.model.stream_dict[self.stream_id]['name']
```
neiljp, line 318: Is this and `self.stream_id` for `narrow_to_stream`?

If so, it may be worth adding a comment with the assert statements.

zulipterminal/ui_tools/buttons.py
```diff
@@ -288,11 +288,36 @@ def handle_link(self, *_: Any) -> None:
...
+
+        if '/stream/' in self.link:
+            assert hasattr(self, 'stream_id')
+            assert hasattr(self, 'stream_name')
+            self.controller.narrow_to_stream(self)
+
+        # Exit popup if the narrow was successful.
+        if hasattr(self.controller.loop.widget, '_Overlay__super'):
```
neiljp, line 333: This seems rather cryptic - is there a better expression for this? What narrow failure are we expecting? Should the narrow methods return a value instead?

tests/ui/test_ui_tools.py
```diff
@@ -2755,3 +2755,45 @@ def test_update_widget(self, mocker, caption, expected_cursor_position,
             caption, cursor_position=expected_cursor_position
         )
         assert isinstance(mocked_button._w, AttrMap)
+
+    @pytest.mark.parametrize('link, handle_narrow_called', [
+        (SERVER_URL + '/#narrow/stream/1-Stream-1', True),
+        (SERVER_URL + '/user_uploads/some/path/image.png', False),
+        ('https://foo.com', False),
+    ])
```
neiljp, line 2763: ids are always good :)

tests/ui/test_ui_tools.py
```diff
@@ -2755,3 +2755,45 @@ def test_update_widget(self, mocker, caption, expected_cursor_position,
             caption, cursor_position=expected_cursor_position
         )
         assert isinstance(mocked_button._w, AttrMap)
+
+    @pytest.mark.parametrize('link, handle_narrow_called', [
+        (SERVER_URL + '/#narrow/stream/1-Stream-1', True),
+        (SERVER_URL + '/user_uploads/some/path/image.png', False),
+        ('https://foo.com', False),
```
neiljp, line 2762: `(SERVER_URL + '/#narrow/stream/asdasdgasdgasdg', ?),` could be in a message typed in manually, I think? What should happen?

zulipterminal/helper.py
```diff
@@ -618,3 +619,9 @@ def display_error_if_present(response: Dict[str, Any], controller: Any
                              ) -> None:
     if response['result'] == 'error' and hasattr(controller, 'view'):
         controller.view.set_footer_text(response['msg'], 3)
+
+
+def hash_util_decode(string: str) -> str:
+    # Acknowledge custom string replacements in zulip/zulip's
+    # zerver/lib/url_encoding.py before unquote.
+    return unquote(string.replace('.', '%').replace('%2E', '.'))
```
neiljp, line 627: I seem to recall a discussion about wanting decoding of server URLs previously on czo - might be worth a check, maybe it's something that could be added to the python API as a utility?

tests/ui/test_ui_tools.py
```diff
@@ -2772,12 +2772,15 @@ def test_handle_link(self, mocker, link, handle_narrow_called):
         assert self.handle_narrow.called == handle_narrow_called
 
     @pytest.mark.parametrize(['link', 'is_subscribed', 'stream_id',
-                              'stream_name'], [
-        (SERVER_URL + '/#narrow/stream/1-Stream-1', True, 1, 'Stream 1'),
-        (SERVER_URL + '/#narrow/stream/4-Stream-4', False, 4, None),
+                              'stream_name', 'near'], [
+        (SERVER_URL + '/#narrow/stream/1-Stream-1', True, 1, 'Stream 1', None),
+        (SERVER_URL + '/#narrow/stream/4-Stream-4', False, 4, None, None),
+        (SERVER_URL + '/#narrow/stream/1-Stream-1/near/2', True, 1, 'Stream 1',
```
neiljp, line 2778: Could you describe what near vs not-near means in the commit, and maybe in the ids?

zulipterminal/ui_tools/buttons.py
preetmishra: Fixed with a simple expression. I was only expecting failure from the code that I previously had here (early return statements), but I have improved the structure now.

Re narrow methods, I'd be inclined towards handle failures, if any, in the narrow methods itself than returning a value as a separate amendment.

zulipterminal/helper.py
preetmishra: (Being discussed in [#integrations > Decode server URLs](https://chat.zulip.org/#narrow/stream/127-integrations/topic/Decode.20server.20URLs).)

preetmishra:
@neiljp Thanks for the review and the suggestions. In the latest update, I have incorporated the in-line suggestions, improved the error handling, improved the tests with `ids` and restructured the code radically.

The PR is comparatively bigger now. Though, hopefully, the changes should be easy to comprehend with docstrings, comments, commit descriptions and tests.

neiljp:
@preetmishra I'm not sure if you intended to only push one commit, but seeing it alone suggests we can minimize it:
* we can simply leave `handle_link` as an empty placeholder for now? (or urlparse it more cleanly)
* we can skip storing the model and view for this first version, and add them as necessary
* the tests could go into a new `tests/ui_tools/test_buttons.py` given it's a new class (we could have started doing this sooner)

We also don't use the controller, but we know that's going to be used :)

zulipterminal/ui_tools/buttons.py
```diff
@@ -288,11 +289,92 @@ def handle_link(self, *_: Any) -> None:
...
+        self.stream_id = self._extract_stream_id(self.fragments)
+        if not self.stream_id:
+            return 'Could not resolve stream ID'
+
+        self.stream_name = self._get_stream_name(self.stream_id,
+                                                 self.model.stream_dict)
+        if not self.stream_name:
+            # FIXME: Narrow to the concerened stream in a 'preview' mode or
+            # report whether the stream id is invalid itself at least.
+            return 'You are not subscribed to the stream'
```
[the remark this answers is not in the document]
preetmishra: Thanks. I have extracted `Model.is_user_subscribed_to_stream` as method in the latest update.

Yes, we need `stream_name`.

preetmishra:
@neiljp Thanks for the review and extended guidelines. :+1: 

I have restructured the PR entirely to handle the links in stages that you suggested, extracted `Model.is_user_subscribed_to_stream`, improved the error handling, added support for deprecated URLs and incorporated other suggestions that left you regarding the first commit.

However, there's a caveat! The stream names that have `-`  don't get decoded appropriately. I have left a query in [development help > Encoded stream names](https://chat.zulip.org/#narrow/stream/49-development-help/topic/Encoded.20stream.20names).

preetmishra:
Updated with tests and rebase onto upstream/master.

zulipterminal/ui_tools/buttons.py
```diff
@@ -347,6 +359,36 @@ def _parse_narrow_link(cls, link: str) -> ParsedNarrowLink:
...
+                and fragments[3] == 'near'):
+            stream_id, stream_name = cls._decode_stream_data(fragments[2])
+            message_id = cls._decode_message_id(fragments[4])
+            parsed_link = {'narrow': 'stream:near'}
+            if stream_name:
+                parsed_link['stream_name'] = stream_name
+            if stream_id:
+                parsed_link['stream_id'] = stream_id
+            if message_id:
+                parsed_link['message_id'] = message_id
```
neiljp, line 372: Rather than conditionally adding each of these, perhaps add them unconditionally (generate the full dict each time) and handle the validation slightly differently?

zulipterminal/ui_tools/buttons.py
```diff
@@ -418,13 +467,30 @@ def _switch_narrow_to(self, parsed_link: ParsedNarrowLink) -> None:
                 cast(Optional[str], parsed_link.get('stream_name')),
             )
             self.controller.narrow_to_stream(self)
+        elif 'stream:near' == narrow:
+            self.stream_id, self.stream_name = self._get_stream_data(
+                cast(Optional[int], parsed_link.get('stream_id')),
+                cast(Optional[str], parsed_link.get('stream_name')),
+            )
+            self.message = dict()
+            self.message['id'] = parsed_link['message_id']
```
neiljp, line 476: We can combine these lines? (and a few similar others)

zulipterminal/ui_tools/buttons.py
```diff
@@ -272,4 +280,136 @@ def handle_link(self, *_: Any) -> None:
...
+            if stream_id and not self.model.is_user_subscribed_to_stream(
+               stream_id):
+                # TODO: Narrow to the concerened stream in a 'preview' mode or
+                # report whether the stream id is invalid instead.
+                return 'The stream seems to be either unknown or unsubscribed'
+
+            # Validate stream name.
+            if stream_name and not self.model.is_valid_stream(stream_name):
+                return 'Invalid stream name'
+
```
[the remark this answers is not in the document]
preetmishra: @neiljp Are you suggesting it to be moved to `_get_steam_data()`? I think it fits better here for two reasons:
* The overall parse -> validate -> narrow flow seems consistent.
* It is clearer to validate topic name later.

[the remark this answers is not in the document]
preetmishra: Nevermind, I see how what you're suggesting is better. Thanks!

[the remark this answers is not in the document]
preetmishra: Ah, I see how what you're suggesting fits better. Thanks!

zulipterminal/ui_tools/buttons.py
```diff
@@ -249,10 +253,23 @@ def __init__(self, user_id: int, email: str) -> None:
...
 
 
+StreamData = TypedDict('StreamData', {
+    'stream_id': Optional[int],
+    'stream_name': Optional[str],
+})
+
+ParsedNarrowLink = TypedDict('ParsedNarrowLink', {
+    'narrow': str,
+    'stream': StreamData,
```
preetmishra, line 263: The following was confusing `mypy`:
```python3
ParsedNarrowLink = TypedDict('ParsedNarrowLink', {
    'narrow': str,
    'stream_id': Optional[int],
    'stream_name': Optional[str],
...
stream_data = cls._decode_stream_data(fragments[2])
parsed_link = dict(narrow='stream', **stream_data)  # type: ParsedNarrowLink
```

So, to assist `mypy` and make things simpler I introduced a `TypedDict` for the stream data and nested it instead.

preetmishra:
@neiljp Thanks for the suggestions. They make the code clearer and concise.

In the latest update, I have incorporated the suggestions that you left in-line and amended the tests accordingly.

neiljp:
@preetmishra Thanks for working on this - this is another feature that will make ZT easier to use for jumping between streams/topics :+1: This took a while to review due to the size, and there are a few refactors we can do to improve this, but I'm going to merge this now so we can get it out for wider view :tada:

PR #710, opened 2020-07-06
Emoji data storage improvements and custom emoji support

kaustubh-nair:
The main points of this PR are:
* Currently emoji data only has emoji names. This PR makes the emoji data more complete, by storing other properties like emoji code, and emoji type. They are necessary for sending emoji data while adding reactions to messages. Additionally, a distinction between default and custom emojis would be nice, in case we'd want this data in the future.
* Adds support for custom emojis by fetching them from the server and storing them in the model. Since custom emojis are stored in the model, we might as well store the default ones too - instead of fetching emojis from two different places.
* A bunch of variable/filename renaming commits to facilitate this change.
* Adding zulip_extra_emoji to list of emojis


neiljp:
@kaustubh-nair Based on comment in #zulip-terminal, ok to remove 'needs review', as other PRs are expected to come before this?

kaustubh-nair:
@neiljp, this one of supposed to be merged before the others.

tests/model/test_model.py
```diff
@@ -1597,3 +1604,47 @@ def test_is_muted_topic(self, topic, is_muted, stream_dict, model):
         ]
         assert model.is_muted_topic(stream_id=topic[0],
                                     topic=topic[1]) == is_muted
+
+    @pytest.mark.parametrize('response', [
+        {
+            'emoji': {
+                '100': {
+                    'author_id': 5,
```
neiljp, line 1612: These appear to be user ids. If so, we could correlate these to ones we use in other fixtures?

We may want to take into account the variation between 2.1 and 3+ here due to https://www.zulipchat.com/api/get-custom-emoji

tests/model/test_model.py
```diff
@@ -1597,3 +1604,47 @@ def test_is_muted_topic(self, topic, is_muted, stream_dict, model):
...
+                    'name': 'dancing',
+                    'source_url': '/user_avatars/3/emoji/images/3.png'
+                },
+                '81': {
+                    'author_id': 5,
+                    'deactivated': True,
+                    'id': '81',
+                    'name': 'green_tick',
+                    'source_url': '/user_avatars/4/emoji/images/4.png'
+                },
```
neiljp, line 1638: It's worth noting that this is deactivated in a comment - that's why this doesn't correlate to the `custom_emojis` fixture, presumably.

tests/conftest.py
```diff
@@ -180,6 +180,15 @@ def streams_fixture():
     return deepcopy(streams)
```
neiljp: Typo in your commit summary.

zulipterminal/model.py
```diff
@@ -344,6 +347,15 @@ def update_stream_message(self, topic: str, msg_id: int,
         display_error_if_present(response, self.controller)
         return response['result'] == 'success'
 
+    def fetch_custom_emojis(self) -> Dict[str, Dict[str, str]]:
```
neiljp, line 350: We could `TypedDict` the emoji types? There's even scope for using `Literal` here given the limited set of emoji-types, though we need to understand the realm vs custom vs zulip-extra.

zulipterminal/model.py
```diff
@@ -344,6 +347,15 @@ def update_stream_message(self, topic: str, msg_id: int,
...
+        response = self.client.get_realm_emoji()
+        custom_emojis = {}
+        for emoji_code, emoji in response['emoji'].items():
+            if not emoji['deactivated']:
+                custom_emojis[emoji['name']] = {'code': emoji_code,
+                                                'type': 'realm_emoji'}
```
neiljp, line 356: This looks like it would be cleaner as a comprehension?

tests/model/test_model.py
```diff
@@ -75,7 +78,11 @@ def test_init(self, model, initial_data, user_profile,
         assert model.unpinned_streams == []
         self.classify_unread_counts.assert_called_once_with(model)
         assert model.unread_counts == []
-        assert model.emoji_data == unicode_emojis
+        assert unicode_emojis.items() <= model.emoji_data.items()
+        assert custom_emojis.items() <= model.emoji_data.items()
+        print(model.emoji_data)
```
neiljp, line 83: :fearful:

tests/model/test_model.py
```diff
@@ -75,7 +78,11 @@ def test_init(self, model, initial_data, user_profile,
         assert model.unpinned_streams == []
         self.classify_unread_counts.assert_called_once_with(model)
         assert model.unread_counts == []
-        assert model.emoji_data == unicode_emojis
+        assert unicode_emojis.items() <= model.emoji_data.items()
+        assert custom_emojis.items() <= model.emoji_data.items()
```
neiljp, line 82: What are you trying to state by this?

zulipterminal/model.py
```diff
@@ -126,7 +126,10 @@ def __init__(self, controller: Any) -> None:
 
         self.unread_counts = classify_unread_counts(self)
 
-        self.emoji_data = emoji_data.EMOJI_DATA
+        custom_emojis = self.fetch_custom_emojis()
+        all_emojis = (list(emoji_data.EMOJI_DATA.items())
+                      + list(custom_emojis.items()))
+        self.emoji_data = OrderedDict(sorted(all_emojis, key=lambda e: e[0]))
```
neiljp, line 132: Is this really `active_emoji_data`, given the qualifier on the custom emojis?

zulipterminal/model.py
neiljp, line 356: Are these all `realm_emoji`? In reactions we have unicode/realm/zulip_extra?

zulipterminal/model.py
```diff
@@ -126,7 +126,10 @@ def __init__(self, controller: Any) -> None:
 
         self.unread_counts = classify_unread_counts(self)
 
-        self.emoji_data = emoji_data.EMOJI_DATA
+        custom_emojis = self.fetch_custom_emojis()
+        all_emojis = (list(emoji_data.EMOJI_DATA.items())
+                      + list(custom_emojis.items()))
```
neiljp, line 131: Can we end up with duplicates here? Is that OK? Custom emoji can override unicode ones, I believe, but I'm not sure how we want to handle that situation.

tools/convert-emoji-data
```diff
@@ -1,28 +1,36 @@
...
     f.write('# Generated automatically by tools/convert-emoji-data\n')
     f.write('# Do not modify.\n\n')
-    f.write('EMOJI_DATA = [\n')
-    for listitem in emoji_list:
-        f.write('    \'%s\',\n' % listitem)
-    f.write(']\n')
+    f.write('EMOJI_DATA = OrderedDict([\n')
+    for emoji_code, emoji_name in ordered_emojis.items():
+        # {'smile': {'code': '263a', 'type': 'unicode_emoji'}}
+        f.write('    (\'%s\', {\'code\': \'%s\',\n'
```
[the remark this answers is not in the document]
kaustubh-nair: Yeah, the line becomes too long for `mypy`.

tests/model/test_model.py
kaustubh-nair: Makes sure that `model.emoji_data` contains both `unicode_emoji` and `custom_emoji`

zulipterminal/model.py
kaustubh-nair: Custom emojis are all `realm_emoji`

zulipterminal/model.py
kaustubh-nair: The active qualifier is only on custom emoji, but this contains `unicode_emoji` too?

kaustubh-nair: Do we consider `unicode_emojis` active by default?

zulipterminal/model.py
kaustubh-nair: I didn't consider this. The dictionary has emoji name as keys so it cannot have duplicates.
Since custom emojis are added later, the unicode ones with the same names are quietly removed. This does seem to achieve what we want - the overriding is taken care of automatically - but is it okay to achieve it this way?

tests/conftest.py
kaustubh-nair: Fixed

tests/model/test_model.py
kaustubh-nair: We're not using this right now, it would be simpler if I just remove the attribute?

tests/model/test_model.py
neiljp: `set` logic seems appropriate for this, perhaps?

zulipterminal/model.py
neiljp: I'm not sure we can deactivate unicode emoji, though I see what you mean. At least partially I was trying to make the variable name obvious in the sense that if a custom emoji was deactivated subsequent to this point, then we might think to check - in the same way that `stream_dict` doesn't really describe `active_stream_subscriptions`.

zulipterminal/model.py
```diff
@@ -348,12 +350,12 @@ def update_stream_message(self, topic: str, msg_id: int,
...
         return response['result'] == 'success'
 
-    def fetch_custom_emojis(self) -> Dict[str, Dict[str, str]]:
+    def fetch_custom_emojis(self) -> NamedEmojiData:
         response = self.client.get_realm_emoji()
         custom_emojis = {emoji['name']: {'code': emoji_code,
                                          'type': 'realm_emoji'}
                          for emoji_code, emoji in response['emoji'].items()
-                         if not emoji['deactivated']}
+                         if not emoji['deactivated']}  # type: NamedEmojiData
```
kaustubh-nair, line 358: `response` is not typed so it cannot determine `custom_emojis` automatically.

zulipterminal/model.py
```diff
@@ -126,7 +126,11 @@ def __init__(self, controller: Any) -> None:
 
         self.unread_counts = classify_unread_counts(self)
 
-        self.active_emoji_data = unicode_emojis.EMOJI_DATA
+        custom_emojis = self.fetch_custom_emojis()
+        all_emojis = (list(unicode_emojis.EMOJI_DATA.items())
+                      + list(custom_emojis.items()))
+        self.active_emoji_data = OrderedDict(sorted(all_emojis,
+                                                    key=lambda e: e[0]))
```
neiljp, line 134: If this handles the 'custom emoji can replace unicode emoji' case, then can we add a test case for it?

zulipterminal/model.py
```diff
@@ -139,9 +140,10 @@ def __init__(self, controller: Any) -> None:
         unicode_emoji_data = unicode_emojis.EMOJI_DATA
         for name, data in unicode_emoji_data.items():
             data['type'] = 'unicode_emoji'
+        typed_unicode_emoji_data = cast(NamedEmojiData, unicode_emoji_data)
```
kaustubh-nair: This can't be done inline right?

zulipterminal/emoji_names.py
```diff
@@ -1230,7 +1230,6 @@
     'sleepy',
     'sleuth',
     'slight_frown',
-    'slight_smile',
```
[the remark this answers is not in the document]
kaustubh-nair: Fixed

tests/conftest.py
```diff
@@ -181,7 +181,7 @@ def streams_fixture():
 
 
 @pytest.fixture
-def emojis_fixture():
```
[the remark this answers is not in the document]
kaustubh-nair: Fixed.

tools/convert-unicode-emoji-data
```diff
@@ -1,29 +1,38 @@
...
     f.write('# Do not modify.\n\n')
-    f.write('EMOJI_DATA = [\n')
-    for listitem in emoji_list:
-        f.write('    \'%s\',\n' % listitem)
-    f.write(']\n')
+    f.write('EMOJI_DATA = OrderedDict([\n')
+    for emoji_code, emoji_name in ordered_emojis.items():
+        # {'smile': {'code': '263a', 'type': 'unicode_emoji'}}
+        f.write("    ('%s', {'code': '%s',\n"
+                "     'type': '%s'}),\n" % (emoji_code, emoji_name, 'unicode_emoji'))
```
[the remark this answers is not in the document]
kaustubh-nair: Fixed.

tests/model/test_model.py
```diff
@@ -71,6 +75,7 @@ def test_init(self, model, initial_data, user_profile):
         assert model.unpinned_streams == []
         self.classify_unread_counts.assert_called_once_with(model)
         assert model.unread_counts == []
+        assert model.emoji_data == unicode_emojis
```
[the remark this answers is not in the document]
kaustubh-nair: Fixed. Missed this between commits

tests/model/test_model.py
```diff
@@ -1597,3 +1601,39 @@ def test_is_muted_topic(self, topic, is_muted, stream_dict, model):
         ]
         assert model.is_muted_topic(stream_id=topic[0],
                                     topic=topic[1]) == is_muted
+
+    @pytest.mark.parametrize('response', [
+        {
```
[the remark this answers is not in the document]
kaustubh-nair: Added.

zulipterminal/model.py
kaustubh-nair: Added.

kaustubh-nair:
@neiljp Thanks for the review! Updated PR according to comments.


> When I run the fetch/convert scripts on the branch, there seems to be a diff in `unicode_emojis.py`.
I can't see a diff on the unicode data? Did you run this recently


> Do you know if there are emoji events that should update the data we initially fetch?
This has been discussed on czo before, the server needs to share emoji data using an endpoint - there aren't any events for updating this yet.

zulipterminal/model.py
neiljp: You've modified the original, so this seems reasonable :+1:

neiljp:
@kaustubh-nair This looks good and should be mergeable after a rebase :) As per my comments on czo, I was referring to using events from the server to ensure that the custom emojis are kept up to date. The only aspect we might consider prior to merging would be handling failing responses from the `client` call to fetch the emoji, which would fit into the same category as for get-messages and register.

The unicode data issue was fixed in one of your commits, from what I can tell :)

neiljp:
@kaustubh-nair Thanks for the update! I just fixed some recent minor conflicts and merged manually :tada:

As I indicated in the stream, as part of the synchronizing of the realm emoji, we should likely move to treating the initial fetch as being from the initial data from the register, which I've filed as #809.

PR #713, opened 2020-07-10
FAQ: Amend hotkeys section to remove the obsolete context

preetmishra:
The change feels a bit scrupulous but should be useful to have in any case.


neiljp:
@preetmishra Agreed that this is minor, but worthwhile to avoid confusion.

PR #714, opened 2020-07-10
model: Notify user if a request fails.

### LINKED ISSUE #427: Notify user if hitting ratelimit
This shouldn't be an issue, if we design ZT correctly, but if we get a certain response from the server then we can perhaps determine it's a rate-limiting issue?

kaustubh-nair:
This commit displays an error message in the footer for server requests
involving user interactions. By this the user can be notified for
different failures like:
* Hitting API rate limits.
* Posting in streams where user doesn't have the required permissions.
* Message sending/editing errors.
* Toggling reaction errors.
* etc.

Error responses from some requests are not displayed, since they are not
triggered directly by a user action. Example of these include event
polling requests, event registration, and updating presence.

Fixes #427.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -19,6 +19,8 @@ class TestModel:
                                    'Controller.client')
         self.client.base_url = 'chat.zulip.zulip'
         mocker.patch('zulipterminal.model.Model._start_presence_updates')
+        self.display_error_if_present = mocker.patch(
+                            'zulipterminal.model.display_error_if_present')
 
     @pytest.fixture
     def model(self, mocker, initial_data, user_profile):
@@ -322,6 +324,9 @@ class TestModel:
         self.client.get_stream_topics.assert_called_once_with(23)
         assert model.index['topics'] == expected_index
         assert result == return_value
+        if response['result'] != 'success':
+            (self.display_error_if_present.
+             assert_called_once_with(response, self.controller))
 
     @pytest.mark.parametrize("user_key", ['user_id', 'id'])
     @pytest.mark.parametrize("msg_id, existing_reactions, expected_method", [
@@ -331,7 +336,7 @@ class TestModel:
         (5, [dict(user='me', emoji_code='1f614')], 'POST'),
         (5, [dict(user='not me', emoji_code='1f614')], 'POST'),
     ])
-    def test_react_to_message_with_thumbs_up(self, model,
+    def test_react_to_message_with_thumbs_up(self, mocker, model,
                                              user_key,
                                              msg_id,
                                              existing_reactions,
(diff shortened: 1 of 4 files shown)


sumanthvrao:
@kaustubh-nair This looks fairly straightforward as you mentioned. 👍 

One suggestion maybe (not sure if we already had a discussion of this) whether we can display these kinds of error messages in a different way than regular help tips ones.

zulipterminal/model.py
```diff
@@ -373,6 +389,8 @@ def get_topics_in_stream(self, stream_list: Iterable[int]) -> str:
                 self.index['topics'][stream_id] = [topic['name'] for
                                                    topic in response['topics']]
             else:
+                if response['result'] == 'error':
+                    self.controller.view.set_footer_text(response['msg'], 3)
```
neiljp, line 393: Given the apparent similarity in all this error handling, please use a new method for this.

Putting this in a function should make testing simple?

kaustubh-nair:
@neiljp thanks for the review.
Re. displaying error for server registration - the problem is that the view would not load when it is called. Even if we wait for it to load, I'm not sure what happens if initial_data is not retrieved....would zt even show anything?

neiljp:
@kaustubh-nair I agree that if initial data does not exist then we cannot easily show an error message in the UI, though we could extend the startup error handling via exceptions for that case. If we do already have initial data then we can notify of connection failure, which could occur in the same method, but possibly more in `poll_for_events`, ie. checking the return value.

kaustubh-nair:
Opened #716

tests/model/test_model.py
```diff
@@ -535,11 +535,10 @@ def test_fail_get_messages(self, mocker, error_response,
     ], ids=['muting_205', 'unmuting_205', 'first_muted_205',
             'last_unmuted_205'])
     def test_toggle_stream_muted_status(self, mocker, model,
-                                        initial_muted_streams, value):
+                                        initial_muted_streams, value,
+                                        response={'result': 'success'}):
```
preetmishra, line 557: Is this to overcome pytest's caching issue or just to reuse the `dict`?

kaustubh-nair: Just for reuse

kaustubh-nair:
@sumanthvrao @neiljp, I haven't tested every single one of these, but some of them can be tested like
* Holding down `*` to get rate limited.
* Holding down `down` to get rate limited.
* Trying to post in `#announce`.
* Trying to send a message to a stream that doesn't exist.

neiljp:
@kaustubh-nair As per PMs, I just merged the last two commits together - thanks for working on this! :tada:

Time to update the release notes again :+1:

PR #718, opened 2020-07-14
Limit rapid keypresses

kaustubh-nair:
This PR solves the rate-limiting issues by due to rapid keypresses by:
* Preventing concurrent up/down when at the top or bottom of the `MessageView`.
* Rate limits pressing of other keys to 15 consecutive presses.

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -2,8 +2,8 @@ import pytest
 
 import zulipterminal.helper
 from zulipterminal.helper import (
-    canonicalize_color, classify_unread_counts, index_messages, notify,
-    powerset,
+    KeypressCounter, canonicalize_color, classify_unread_counts,
+    index_messages, notify, powerset,
 )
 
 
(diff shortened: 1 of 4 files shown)


neiljp:
Gave some thoughts in #**zulip-terminal>Rate limiting issues**.

neiljp (commented):
@kaustubh-nair I understand the approach you're taking here, and thanks for taking this up :+1:  However, I think we might want to take different approaches - here are some thoughts on what's happening in more detail, based on my interpretation of the PR vs the problem:
* The first commit intends to avoid up/down triggers from keypresses when earlier/later messages are loading; this should limit calls to load_old/new based on the duration to perform those operations. However, if the cpu is sufficiently fast then `duration_of(api_call + update_code)` could still be too small, so callable frequently enough to continue to hit the rate limits. For this API call, we can internally rate-limit our calls somehow, but also if we "know" we are already at the extreme end of the messages in the narrow, then we shouldn't need to check at all. So, I agree about using an internal variable, but the "mid-loading" type doesn't seem quite right.
* The python API library contains rate-limiting code already, that we could use instead of the approach in the second commit. It's also worth noting that this commit doesn't trigger between keys making the same API call, or eg. starring and pressing down - I'm not sure that the server makes a distinction between them.

Off the top of my head, one server-related follow-up might be to send initial (and updated?) notices on suggested client rate limits, which clients could adapt to, perhaps even with each response? @timabbott

timabbott:
Hmm, good question; I'm not sure what the right server API would be for indicating the rate-limiting structure; maybe a good discussion for #backend.

I'm not sure I understand the situation here -- is the user behavior with ZT that causes it to potentially hit rate limits something we expect users to want to do?  The issue may be structurally a UI problem.

PR #720, opened 2020-07-16
Ignore mouse click if compose box is open.

### LINKED ISSUE #652: Clicking on a compose-triggering location cancels current compose
While this could be considered resolved if we had a drafts feature where messages were saved in these situations, this could still be as confusing as it feels now:

* Start replying to a message (eg. stream, PM), and type some text
* Click on any message in the message list, or a user name in the right (which also narrows!)
* ...and the message is lost

While clicking on a user name *might* be relatively expected to cause this behavior, if you click on a message **in the same narrow**, it has the same behavior. This wouldn't be to a different recipient, so this is a significant issue.

Strangely (?) this currently doesn't occur when clicking on the top-left narrow buttons, where the compose window is unchanged but the narrow view changes!

Various options for solving it could be:
* requiring `GO_BACK` to exit from composing (not responding to clicks)
* retaining message content, potentially changing recipients if moving around (dangerous?)
* not changing narrow or resetting compose until the message is sent or cancelled (`GO_BACK`?)

kaustubh-nair:
Clicking on messages/user names when the compose box is open causes the
current message to be lost. This commits ignores the mouse click to
prevent the typed data to be lost, though better handling can be done
when the drafts feature is implemented.

Also, stored controller as an attribute for UsersView. Tests amended for
this.

Test added for mouse click event.

Fixes #652.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -31,6 +31,14 @@ BOXES = "zulipterminal.ui_tools.boxes"
 SERVER_URL = "https://chat.zulip.zulip"
 
 
+@pytest.fixture(
+    params=[True, False],
+    ids=['ignore_mouse_click', 'handle_mouse_click']
+)
+def compose_box_is_open(request):
+    return request.param
+
+
 class TestModListWalker:
     @pytest.fixture
     def mod_walker(self):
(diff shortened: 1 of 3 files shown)


kaustubh-nair:
@neiljp making a method in the controller sounds like a good idea.

I needed `is_in_editor_mode_with()` because we might require different behavior depending on which editor is active ( for example search boxes ) but if simply requiring a `GO_BACK` is enough for now then we think about this in follow-ups.

kaustubh-nair:
@neiljp Thanks for the review, PR updated :+1:

kaustubh-nair:
@neiljp, thanks for the review, updated PR!
Re. Opened an issue for dealing with the cursor.

neiljp:
@kaustubh-nair Thanks for the fix! Just manually merged :tada: (you had a different file in your commit title than in the commit)

PR #721, opened 2020-07-17
0.5.2 changelog and release

neiljp:
This is a draft of commits with release notes/changes to prepare for the 0.5.2 release, which we can hopefully release soon.

### DIFF
--- a/zulipterminal/version.py
+++ b/zulipterminal/version.py
@@ -1,4 +1,4 @@
-ZT_VERSION = '0.5.2'
+ZT_VERSION = '0.5.2+git'
 
 SUPPORTED_SERVER_VERSIONS = [
     ('2.1', None),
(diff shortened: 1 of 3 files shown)


zulipterminal/version.py
```diff
@@ -1,4 +1,4 @@
-ZT_VERSION = '0.5.1+git'
+ZT_VERSION = '0.5.2'
 
 SUPPORTED_SERVER_VERSIONS = [
     ('2.1', None),
```
preetmishra, line 4: Let's remove the comment from the next line.

neiljp: Oh yes, I agree - though that's not shown in this review context...

neiljp: That said, there are features of v3.0 that we don't yet support. Perhaps #731 could include a table for version-specific features.

neiljp:
This was used as the basis for the 0.5.2 release notes, so closing this now :tada:

PR #723, opened 2020-07-17
Do not mark as read on search narrows

### LINKED ISSUE #624: Don't mark as read on search narrows
Like the web app, ZT should not mark messages as read on search narrows to deliver a better and intuitive user experience.
This should be a good starting point for implementing any 'not-reading' behaviour in ZT (see #585).

Conversation link: https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Not.20marking.20read

preetmishra:
This doesn't mark messages as read on any search narrow like the web app.

Fixes #624.


neiljp:
@preetmishra Thanks for working on this, it works well, and shows a clear path to implementing the 'explore' mode and possibly other non-reading modes (all messages?). I just manually merged this with two small commit text adjustments :tada:

PR #725, opened 2020-07-17
Multiple footer reset bug fix

### LINKED ISSUE #647: Footer gets updated as many times as the edit key is pressed on any uneditable message
Steps to reproduce:
* Move focus to any message that you can't edit
* Press e multiple times
* Press any other key to trigger the behaviour

kaustubh-nair:
This PR introduces a queue to keep track of the currently active sleep threads, and resets the footer only if the queue is empty.

Fixes #647
(Alternate fix: #748)

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -69,18 +69,22 @@ class TestView:
 
     def test_set_footer_text_with_duration(self, view, mocker,
                                            custom_text="custom", duration=5.3):
-        mocker.patch('zulipterminal.ui.View.get_random_help',
-                     return_value=['some help text'])
-        mock_sleep = mocker.patch('time.sleep')
+        mocker.patch('zulipterminal.ui.View._reset_footer_text')
 
         view.set_footer_text([custom_text], duration)
 
-        view._w.footer.set_text.assert_has_calls([
-            mocker.call([custom_text]),
-            mocker.call(['some help text'])
-        ])
+        view._w.footer.set_text.assert_called_once_with([custom_text])
+        view.controller.update_screen.assert_called_once_with()
+        view._reset_footer_text.assert_called_once_with(duration)
+
+    def test__reset_footer_text(self, view, mocker, duration=46.2):
+        mock_sleep = mocker.patch('time.sleep')
+        mocker.patch('zulipterminal.ui.View.set_footer_text')
+
+        view._reset_footer_text(duration)
+
         mock_sleep.assert_called_once_with(duration)
-        assert view.controller.update_screen.call_count == 2
+        view.set_footer_text.assert_called_once_with()
 
     @pytest.mark.parametrize('suggestions, state, truncated, footer_text', [
         ([], None, False, [' [No matches found]']),
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -1,3 +1,6 @@
+import time
+from concurrent.futures import ThreadPoolExecutor
+
 import pytest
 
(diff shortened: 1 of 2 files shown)


tests/ui/test_ui.py
```diff
@@ -68,18 +68,21 @@ def test_set_footer_text_specific_text(self, view, text='blah'):
...
 
-        view._w.footer.set_text.assert_has_calls([
-            mocker.call([custom_text]),
-            mocker.call(['some help text'])
-        ])
+        view._w.footer.set_text.assert_called_once_with([custom_text])
+        view.controller.update_screen.assert_called_once_with()
+        view._reset_footer_text.assert_called_once_with(duration)
+
+    def _reset_footer_text(self, view, mocker, duration=46.2):
```
sumanthvrao, line 81: pytest won't run this unless it starts with `test__reset_footer_.....`. 

Do you want to remove the two blank lines below (81, 83)? It feels pretty a little odd. And if we can save some space and not decrease readability, we should go for it.

tests/ui/test_ui.py
```diff
@@ -84,6 +86,15 @@ def _reset_footer_text(self, view, mocker, duration=46.2):
         mock_sleep.assert_called_once_with(duration)
         view.set_footer_text.assert_called_once_with()
 
+    def _reset_footer_text_multiple_resets(self, view, mocker, duration=1):
```
sumanthvrao, line 89: Same comment as previous, it should begin with `test__.....`

tests/ui/test_ui.py
kaustubh-nair: Oops, good catch - the tests weren't running.

@neiljp Suggests having blank lines surrounding the method under test :sweat_smile:

tests/ui/test_ui.py
```diff
@@ -68,18 +71,36 @@ def test_set_footer_text_specific_text(self, view, text='blah'):
...
+        mocker.patch('zulipterminal.ui.View.set_footer_text')
+
+        # Multiple simultaneous calls for resetting footer.
+        with ThreadPoolExecutor(max_workers=3) as executor:
+            executor.submit(view._reset_footer_text, duration=0.5)
+            executor.submit(view._reset_footer_text, duration=0.5)
+            executor.submit(view._reset_footer_text, duration=0.5)
+
+        view.set_footer_text.assert_called_once_with()
+        assert time.sleep.call_count == 3
```
kaustubh-nair, line 103: I tried something different here - this is more of an integration test rather than the unit tests we have everywhere.

kaustubh-nair: (Also, pytest cannot call methods asynchronously?)

tests/ui/test_ui.py
preetmishra: The blank lines clarify the arrange, act and assert segments. Personally, I like having them. :wink:

neiljp: While this does add extra spacing, this is a style I've been moving towards for the reasons that @preetmishra mentioned.

That said, if a test needs zero setup, I might be tempted to avoid the one blank line, but consistency would still be good and we don't currently have many of those - perhaps unfortunately.

sumanthvrao: Sure looks like a cleaner style. 👍

neiljp:
@kaustubh-nair I'm closing this in favor of your later #748 (thanks for both during GSoC!). That is a simpler solution, though doesn't avoid threads per call, and I suspect we may need an improvement later to more cleanly handle 'stacking' footer-text-updates.

PR #730, opened 2020-07-22
Order typeahead by narrow/recipients

kaustubh-nair:
This PR orders the mentions typeahead such that participants/subscribers in the current conversation are shown first.
There are two parts to this:

**Part 1: Add ordering logic (Commits 1 & 2)** - This is done by adding a `recipient_user_ids` attribute to `WriteBox`, storing the user ids of the participants. This is supported only for private messages because subscriber data for streams is not available. Note that storing recipient ids in `MessageBox` is necessary because there is no other way to obtain the user ids.
**Part 2: Add subscriber data for streams thus enabling ordering (Commits 4 & 5)** - Set `include_subscribers` to True while fetching initial data to get stream subscriber data. Just store this in `WriteBox.recipient_user_ids` and voila!

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -134,6 +134,7 @@ general_stream = {
     'stream_weekly_traffic': 0,
     'push_notifications': False,
     'email_address': 'general@example.comm',
+    'subscribers': [1001, 11, 12]
 }
 
 # This is a private stream;
@@ -152,6 +153,7 @@ secret_stream = {
     'desktop_notifications': False,
     'stream_weekly_traffic': 0,
     'push_notifications': False,
+    'subscribers': [1001, 11]
 }
 
 
@@ -173,6 +175,7 @@ def streams_fixture():
             'stream_weekly_traffic': 0,
             'push_notifications': False,
             'email_address': 'stream{}@example.com'.format(i),
+            'subscribers': [1001, 11, 12]
         })
     return deepcopy(streams)
 
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1441,12 +1441,14 @@ class TestModel:
     @pytest.mark.parametrize('event, final_muted_streams, ', [
         (
             {'property': 'in_home_view',
+             'op': 'update',
              'stream_id': 19,
              'value': True},
             {15}
         ),
         (
             {'property': 'in_home_view',
+             'op': 'update',
              'stream_id': 30,
              'value': False},
             {15, 19, 30}
@@ -1480,6 +1482,7 @@ class TestModel:
         (
             {
                 'property': 'pin_to_top',
+                'op': 'update',
                 'stream_id': 6,
                 'value': True
             },
@@ -1489,6 +1492,7 @@ class TestModel:
         (
             {
                 'property': 'pin_to_top',
+                'op': 'update',
                 'stream_id': 8,
                 'value': False
             },
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -147,7 +147,8 @@ class TestModel:
                 event_types=event_types,
                 fetch_event_types=fetch_event_types,
                 apply_markdown=True,
-                client_gravatar=True)
+                client_gravatar=True,
+                include_subscribers=True)
 
     @pytest.mark.parametrize('msg_id', [1, 5, set()])
     @pytest.mark.parametrize('narrow', [
@@ -1525,6 +1526,25 @@ class TestModel:
         update_left_panel = model.controller.view.left_panel.update_structure
         update_left_panel.assert_called_once_with()
         model.controller.update_screen.assert_called_once_with()
+
+    @pytest.mark.parametrize('event, expected_subscribers', [
+        ({'type': 'subscription', 'op': 'peer_add',
+          'stream_id': 99, 'user_id': 12}, [1001, 11, 12]),
+        ({'type': 'subscription', 'op': 'peer_remove',
+          'stream_id': 2, 'user_id': 12}, [1001, 11]),
+    ], ids=[
+        'user_subscribed_to_stream',
+        'user_unsubscribed_from_stream',
+    ])
+    def test__handle_subscription_event_subscribers(self, model, mocker,
+                                                    stream_dict, event,
+                                                    expected_subscribers):
+        model.stream_dict = stream_dict
+
+        model._handle_subscription_event(event)
+
+        new_subscribers = model.stream_dict[event['stream_id']]['subscribers']
+        assert new_subscribers == expected_subscribers
 
     @pytest.mark.parametrize('muted_streams, stream_id, is_muted', [
         ({1},   1, True),
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -147,6 +147,23 @@ class TestWriteBox:
         typeahead_string = write_box.generic_autocomplete(text, state)
         assert typeahead_string == required_typeahead
 
+    @pytest.mark.parametrize('text, state, required_typeahead, recipients', [
+        ('@', 0, '@**Human 2**', [12]),
+        ('@', 1, '@**Human Myself**', [12]),
+        ('@', 2, '@**Human 1**', [12]),
+        ('@', -1, '@*Group 4*', [12]),
+        ('@', 0, '@**Human 1**', [11, 12]),
+        ('@', 1, '@**Human 2**', [11, 12]),
+        ('@', 2, '@**Human Myself**', [11, 12]),
+        ('@', -1, '@*Group 4*', [11, 12]),
+    ])
+    def test_generic_autocomplete_mentions_subscribers(self, write_box, text,
+                                                       required_typeahead,
+                                                       state, recipients):
+        write_box.recipient_user_ids = recipients
+        typeahead_string = write_box.generic_autocomplete(text, state)
+        assert typeahead_string == required_typeahead
+
     @pytest.mark.parametrize('text, state, required_typeahead, to_pin', [
         # With no streams pinned.
         ('#Stream', 0, '#**Stream 1**', []),  # 1st-word startswith match.
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -107,13 +107,13 @@ class TestWriteBox:
         ('@Gro', 1, '@*Group 2*'),
         ('@Grou', 1, '@*Group 2*'),
         # Expected sequence of autocompletes from '@'
-        ('@', 0, '@*Group 1*'),
-        ('@', 1, '@*Group 2*'),
-        ('@', 2, '@*Group 3*'),
-        ('@', 3, '@*Group 4*'),
-        ('@', 4, '@**Human Myself**'),
-        ('@', 5, '@**Human 1**'),
-        ('@', 6, '@**Human 2**'),
+        ('@', 0, '@**Human Myself**'),
+        ('@', 1, '@**Human 1**'),
+        ('@', 2, '@**Human 2**'),
+        ('@', 3, '@*Group 1*'),
+        ('@', 4, '@*Group 2*'),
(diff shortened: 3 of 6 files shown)


tests/ui_tools/test_boxes.py
```diff
@@ -107,13 +107,13 @@ def test_generic_autocomplete_set_footer(self, mocker, write_box,
...
-        ('@', 4, '@**Human Myself**'),
-        ('@', 5, '@**Human 1**'),
-        ('@', 6, '@**Human 2**'),
+        ('@', 0, '@**Human Myself**'),
+        ('@', 1, '@**Human 1**'),
+        ('@', 2, '@**Human 2**'),
+        ('@', 3, '@*Group 1*'),
+        ('@', 4, '@*Group 2*'),
+        ('@', 5, '@*Group 3*'),
+        ('@', 6, '@*Group 4*'),
```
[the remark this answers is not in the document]
kaustubh-nair: Yes group typeahead requires more thought, we could either have a separate trigger for groups or process both group and user typeahead together and give groups some 'middle' priority. For example first name match > group match > last name match, though the first name > last name difference doesn't exist yet either.

zulipterminal/ui_tools/boxes.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, view: Any) -> None:
         self.view = view
         self.msg_edit_id = None  # type: Optional[int]
         self.is_in_typeahead_mode = False
+        self.subscriber_ids = []  # type: List[int]
```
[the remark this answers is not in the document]
kaustubh-nair: I'm not sure, when you say 'recipient' would you associate it to large members of a stream?
'subscribers' isn't the right fit for private messages either though, since you can't choose to subscribe.

zulipterminal/model.py
```diff
@@ -565,7 +565,7 @@ def _stream_info_from_subscriptions(
             subscriptions: List[Dict[str, Any]]
     ) -> Tuple[Dict[int, Any], Set[int], List[List[str]], List[List[str]]]:
         stream_keys = ('name', 'stream_id', 'color',
-                       'invite_only', 'description')
+                       'invite_only', 'description', 'subscribers')
```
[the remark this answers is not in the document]
kaustubh-nair: Oh right, I missed this :+1:

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
kaustubh-nair: Opened #732

kaustubh-nair:
@neiljp Thanks for the review, updated the PR

neiljp:
I also just got a crash from `P` `@` `Esc` `@`, which is probably the same issue as before.

zulipterminal/ui_tools/boxes.py
```diff
@@ -36,8 +37,14 @@ def main_view(self, new: bool) -> Any:
...
 
-    def private_box_view(self, button: Any=None, email: str='') -> None:
+    def private_box_view(self, button: Any=None, email: str='',
+                         subscriber_ids: List[int]=[]) -> None:
         self.set_editor_mode()
+        self.subscriber_ids = subscriber_ids
+        print(self.subscriber_ids, flush=True)
+        # Remove current user.
+        self.subscriber_ids.remove(self.model.user_id)
+        print(self.subscriber_ids, flush=True)
```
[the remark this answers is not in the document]
kaustubh-nair: Oh no, I keep forgetting to check for these :facepalm:

zulipterminal/ui_tools/boxes.py
```diff
@@ -132,13 +139,22 @@ def autocomplete_mentions(self, text: str, prefix_string: str
...
-        users = [user['full_name']
-                 for user in users_list
-                 if match_user(user, text[len(prefix_string):])]
-        user_typeahead = format_string(users, prefix_string + '**{}**')
+        matching_users = [user
+                          for user in users_list
+                          if match_user(user, text[len(prefix_string):])]
+        matching_ids = set([user['user_id'] for user in matching_users])
+        matching_subscriber_ids = set(self.subscriber_ids) & set(matching_ids)
+        # Display subscribered users first.
```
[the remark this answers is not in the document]
kaustubh-nair: Yes, no new changes here.

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
kaustubh-nair: Sounds good.

zulipterminal/ui_tools/boxes.py
```diff
@@ -36,8 +37,14 @@ def main_view(self, new: bool) -> Any:
     def set_editor_mode(self) -> None:
         self.view.controller.enter_editor_mode_with(self)
 
-    def private_box_view(self, button: Any=None, email: str='') -> None:
+    def private_box_view(self, button: Any=None, email: str='',
+                         subscriber_ids: List[int]=[]) -> None:
```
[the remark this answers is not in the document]
kaustubh-nair: OK, a shallow copy is causing the crash here rather than the change in recipients. I think it is fine to not have the current user stored in the recipients at all? That would also be consistent with the other recipient data we're storing in `MessageBox`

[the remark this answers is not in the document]
kaustubh-nair: Updating subscribers requires a little more discussion, opened #737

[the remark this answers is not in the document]
kaustubh-nair: On second thought, this wouldn't be too hard I'll fix this in the same PR

tests/model/test_model.py
```diff
@@ -1392,12 +1392,14 @@ def test__handle_typing_event(self, mocker, model,
...
         (
             {'property': 'in_home_view',
+             'op': 'update',
              'stream_id': 19,
              'value': True},
             {15}
         ),
         (
             {'property': 'in_home_view',
+             'op': 'update',
```
neiljp, line 1403: Making this change seems like a good opportunity to also refactor the code itself to prepare for the next commit, ie. to join the if/elif sequence incorporate the `update`?

zulipterminal/ui_tools/boxes.py
```diff
@@ -61,8 +61,11 @@ def private_box_view(self, button: Any=None, email: str='',
         ]
         self.focus_position = 1
 
-    def stream_box_view(self, caption: str='', title: str='') -> None:
+    def stream_box_view(self, stream_id: int, caption: str='', title: str='',
+                        ) -> None:
         self.set_editor_mode()
+        self.recipient_user_ids = self.model.get_subscribers_in_stream(
```
neiljp, line 67: You seem to have lost this function in this iteration of the PR.

Worth adding a test?

tests/conftest.py
```diff
@@ -134,6 +134,7 @@ def logged_on_user():
     'stream_weekly_traffic': 0,
     'push_notifications': False,
     'email_address': 'general@example.comm',
+    'subscribers': [1001, 11, 12, 13]
```
neiljp, line 137: We don't have a user id with 13?

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Writebox doesn't have any tests at all (except for typeahead) so I'll just add a test for this line and open an issue for the rest.

zulipterminal/ui_tools/boxes.py
```diff
@@ -36,8 +37,10 @@ def main_view(self, new: bool) -> Any:
     def set_editor_mode(self) -> None:
         self.view.controller.enter_editor_mode_with(self)
 
-    def private_box_view(self, button: Any=None, email: str='') -> None:
+    def private_box_view(self, button: Any=None, email: str='',
+                         recipient_user_ids: List[int]=[]) -> None:
```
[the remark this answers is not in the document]
kaustubh-nair: Oh no, that's horrible!

tests/model/test_model.py
kaustubh-nair: I'm assuming you mean using `update` for identifying mute events?

zulipterminal/model.py
```diff
@@ -615,23 +615,24 @@ def _handle_subscription_event(self, event: Event) -> None:
...
 
-                if event['value']:  # Unmuting streams
-                    self.muted_streams.remove(stream_id)
-                    stream_button.mark_unmuted()
-                else:  # Muting streams
-                    self.muted_streams.add(stream_id)
-                    stream_button.mark_muted()
-
-                self.controller.update_screen()
+                    self.controller.update_screen()
```
kaustubh-nair, line 641: I don't like it when git gets confused like this :disappointed:

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Opened #739

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
neiljp: Another approach I've seen which we could try here is to use the type checker and only allow passing in a `Sequence`, but the change you've made looks fine.

tests/model/test_model.py
```diff
@@ -1407,8 +1408,9 @@ def test__handle_typing_event(self, mocker, model,
     ], ids=[
         'remove_19', 'add_30'
     ])
-    def test__handle_subscription_event(self, model, mocker, stream_button,
-                                        event, final_muted_streams):
+    def test__handle_subscription_event_muting(self, model, mocker,
+                                               stream_button, event,
+                                               final_muted_streams):
```
neiljp, line 1413: Is this intentional?

zulipterminal/ui_tools/boxes.py
```diff
@@ -62,8 +63,13 @@ def private_box_view(self, button: Any=None, email: str='',
         ]
         self.focus_position = 1
 
-    def stream_box_view(self, caption: str='', title: str='') -> None:
+    def stream_box_view(self, stream_id: int, caption: str='', title: str='',
+                        ) -> None:
         self.set_editor_mode()
+        subscribers = deepcopy(self.model.get_subscribers_in_stream(stream_id))
```
neiljp, line 69: Why not return a new list in the method, and skip the need for the deepcopy? Alternatively, we're only using it here so far, so we could change it to be `get_other_subscribers_in_stream` and incorporate the extra code below?

neiljp:
@kaustubh-nair In addition to the few minor points above, I'd suggest you checking the other `WriteBox` uses since eg. `R` causes a crash for me from a stream - though this seems to be related to `MessageBox` not having `recipient_ids`.

tests/model/test_model.py
kaustubh-nair: Yes, this test deals only with muting, I'll move this to the previous commit though

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Updated to return new list, but this method can be used for displaying subscriber counts

kaustubh-nair: I hope returning a new list is the same as using deepcopy, so I would just be moving the deepcopy to the model?

kaustubh-nair: OK, modified it to `get_other_subscribers_in_stream()` and used the `stream_dict` directly for subscribers counts. Removed the test because I don't see a point in writing an entire test just for a simple `assert_called_once_with()`. This can added when the whole method is tested.

neiljp:
@kaustubh-nair Thanks for getting this work done, it makes the typeahead much more useful and the footer list more informative :+1:
I just merged this manually after updating a few commit texts only, which you should be able to determine using `git range-diff` :tada:

I know you have other typeahead work underway, but this change of footer ordering suggests we could consider other styles to use in the footer for typeahead or temporary notices, and/or maybe separators to mark different sections.

Issue #731, opened 2020-07-22
Add list of caveats for incomplete/absent features 


neiljp:
The client works rather well at this point, but it's important for prospective users to know whether features are present, hidden/undocumented/difficult-to-find, partially implemented, not implemented. These could refer to issues/PRs, and be put in various categories.

This probably belongs in the FAQ, though perhaps is significant enough to sit in a separate document. This may be too large to fit reasonably in the README, where I've seen it in other projects.

This was previously discussed in #**zulip-terminal>launch checklist** on chat.zulip.org.

preetmishra:
I am inclined towards a separate document that we could link to FAQ.

PR #736, opened 2020-07-23
Improve & sync project description

neiljp:
Given the project is marked as official after the following change, this
is a good opportunity to improve and synchronize the title between the
README.md and setup.py files (used on PyPI):
https://github.com/zulip/zulip/commit/0b3d14a408634ce26ba94acab7451bd185a786d9

To avoid duplication of the project title in the 'long description' text
on PyPI, we remove the first line read from README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,4 @@
-# Zulip Terminal (zulip-term)
-
-An interactive terminal interface for [Zulip](https://zulip.com).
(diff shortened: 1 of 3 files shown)


neiljp:
Just added an additional commit to mark the 3.0 server release and set feature level at 25. We state these as 'supported', but some of the features in 3.0 we do not yet support - also earlier ones, hence the importance of #731.

PR #738, opened 2020-07-24
Check validity of recipients on manual modification

kaustubh-nair:
Users can manually alter the recipients/stream name, but there is
currently to check in place to ensure validity of these modifications.
This commit checks the recipients everytime the user presses TAB on the
boxes (stream name and recipients for private messages).

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -654,6 +654,18 @@ class Model:
                 else:
                     raise RuntimeError("Unknown typing event operation")
 
+    def get_invalid_recipient_emails(self, recipient_emails: List[str]
+                                     ) -> List[str]:
+
+        return [email for email in recipient_emails
+                if email not in self.user_dict]
+
+    def is_valid_stream(self, stream_name: str) -> bool:
+        for stream in self.stream_dict.values():
+            if stream['name'] == stream_name:
+                return True
+        return False
+
     def notify_user(self, message: Message) -> str:
         """
         return value signifies if notification failed, if it should occur
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -232,12 +232,30 @@ class WriteBox(urwid.Pile):
                 return key
             header = self.contents[0][0]
             # toggle focus position
-            if self.focus_position == 0 and self.to_write_box is None:
-                if header.focus_col == 0:
-                    header.focus_col = 1
-                    return key
+            if self.focus_position == 0:
+                if self.to_write_box is None:
+                    if header.focus_col == 0:
+                        stream_name = header[0].edit_text
+                        if not self.model.is_valid_stream(stream_name):
+                            self.view.set_footer_text("Invalid stream name", 3)
+                            return key
+
+                        header.focus_col = 1
+                        return key
+                    else:
+                        header.focus_col = 0
                 else:
-                    header.focus_col = 0
+                    recipient_box = header.original_widget
+                    recipient_emails = [email.strip() for email in
+                                        recipient_box.edit_text.split(',')]
+                    invalid_emails = self.model.get_invalid_recipient_emails(
+                                                              recipient_emails)
+                    if invalid_emails:
+                        invalid_emails_error = ('Invalid recipient(s) - '
+                                                + ', '.join(invalid_emails))
+                        self.view.set_footer_text(invalid_emails_error, 3)
+                        return key
+
             self.focus_position = self.focus_position == 0
             header.focus_col = 0
 
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -230,15 +230,16 @@ class WriteBox(urwid.Pile):
         elif is_command_key('TAB', key):
             if len(self.contents) == 0:
                 return key
+            header = self.contents[0][0]
             # toggle focus position
             if self.focus_position == 0 and self.to_write_box is None:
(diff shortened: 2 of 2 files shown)


zulipterminal/model.py
```diff
@@ -654,6 +654,16 @@ def _handle_typing_event(self, event: Event) -> None:
                 else:
```
sumanthvrao, line 654: Typo: I think the commit description should be 

> currently **no** check in place

zulipterminal/model.py
```diff
@@ -654,6 +654,16 @@ def _handle_typing_event(self, event: Event) -> None:
...
                     raise RuntimeError("Unknown typing event operation")
 
+    def are_valid_recipients(self, recipient_emails: List[str]) -> bool:
+        user_emails = {user['email'] for user in self.users}
```
sumanthvrao, line 658: This seems potentially costly. Unless I'm mistaken, we already have this data in `self.user_dict[email]` through subscription.

kaustubh-nair: Good point :+1:

kaustubh-nair:
@sumanthvrao thanks for the review!
 Showing invalid recipient emails sounds like a good idea, I've added them for group PMs.
PR updated.

zulipterminal/model.py
neiljp: No space before bracket currently.

zulipterminal/ui_tools/boxes.py
```diff
@@ -232,12 +232,30 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+                            return key
+
+                        header.focus_col = 1
+                        return key
+                    else:
+                        header.focus_col = 0
                 else:
-                    header.focus_col = 0
+                    recipient_box = header.original_widget
+                    recipient_emails = recipient_box.edit_text.split(', ')
```
neiljp, line 249: This works for some cases, but assumes precise separation in this way - eg it fails for 'email1 , email2', which is unexpected (and the feedback isn't clear?)

That said, we don't currently really validate what gets passed to the API calls, and could likely use ids instead of emails in future based on a general migration to use ids there in the server.

zulipterminal/ui_tools/boxes.py
```diff
@@ -230,15 +230,34 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
                 else:
-                    self.contents[0][0].focus_col = 0
+                    recipient_box = header.original_widget
+                    recipient_emails = recipient_box.edit_text.split(', ')
+                    invalid_emails = self.model.get_invalid_recipient_emails(
+                                                              recipient_emails)
+                    invalid_emails_str = ', '.join(invalid_emails)
+                    if invalid_emails:
+                        footer_text = "Invalid recipients - {}".format(
+                                                            invalid_emails_str)
```
neiljp, line 255: It seems reasonable to inline `invalid_emails_str` (it's only used in the conditional), and perhaps make `footer_text` more descriptive?

zulipterminal/ui_tools/boxes.py
kaustubh-nair: @neiljp  I can fix the issue with the spaces, but what is the expected feedback?

zulipterminal/ui_tools/boxes.py
kaustubh-nair: How about "User(s) not found with email - abc@abc, abc@xyq" ?

kaustubh-nair: Or just "Could not find email(s) - abc@abc, abc@asd"?

zulipterminal/ui_tools/boxes.py
neiljp: The current situation is twofold:
* that it complains about 'email1 , email2', where I would expect it to be fine - I should have indicated that both are 'valid' in my thought-experiment/test-case. That is, I'd expect 'email1,email2', 'email1, email2' and 'email1 , email2' to be valid? That said, I'm not sure what the server accepts in terms of whitespace of this kind
* the message produced is also a problem. eg. 'chat2406@chat.zulip.org ' also fails but shows what looks like a valid address in the error message, similar for the cases above without `, ` between emails.

zulipterminal/ui_tools/boxes.py
neiljp: To clarify, I only meant the *name* of `footer_text` could be more descriptive.

I think the current text itself is fine, though you could add the parentheses to accommodate the single-recipient case.

zulipterminal/ui_tools/boxes.py
kaustubh-nair: The server ignores whitespaces completely so it doesn't matter where it is. I fixed the code to follow the same, so this should not be a problem anymore.

kaustubh-nair:
PR updated :+1:

neiljp:
@kaustubh-nair Thanks for the updates and adding this feature! :+1: I'm going to merge this now, but I think we'll want to ensure that we get connected issues resolved in the near future if we can, or open issues for them.

Things I'm thinking about off the top of my head are trailing commas and other characters (which still cause potentially confusing messages), and that the only way to resolve the 'invalid' case is by entering valid data - which the user may not know (so if they are mid-compose, then they may potentially have to `esc` to exit it and lose their message). With the completion for streams/topics and ideally something similar for PMs this should become less of an issue, though we may well want to improve the error handling further once we understand more how we plan to handle separators.

PR #359, opened 2019-05-19
WIP: Support displaying of images in msg in the default viewer.

amanagr:

PR #740, opened 2020-07-25
[WIP] Handle media links

### LINKED ISSUE #764: TRACKING: Handle message links
- [ ] Handle external links

preetmishra:
Thanks to @amanagr for his initial work in #359. :+1:

The PR extracts a few common elements from the PR but has changes to make the function compatible with `MessageLinkButton`.

#### Commits
* The first commit is cherry-picked from #708.
* The second commit is from #359 with a few amendments that were necessary to make it work with `MessageLinkButton`.
* The third commit integrates the introduced function in `MesssageLinkButton` to handle media links.
* The fourth commit makes `open_media` asynch and shows downloading updates in the footer.
* The fifth commit is a refactor that extracts `PopUpConfirmationView` instantiation for its subsequent commit where it is used in `open_media`.

#### Potential follow-up
* Skip downloading existing media.

I would greatly appreciate any early feedback and bug/caveats report.

Partially fixes #764.

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -329,12 +329,12 @@ def test_process_media(mocker, media_path='/tmp/zt-somerandomtext-image.png',
     # The command, which is platform dependent, does not matter for the test.
     mocker.patch('zulipterminal.helper.LINUX', True)
     controller = mocker.Mock()
-    mocked_open_media = mocker.patch('zulipterminal.helper.open_media')
 
     process_media(controller, media_link)
 
-    mocked_open_media.assert_called_once_with(controller, 'xdg-open',
-                                              media_path)
+    controller.show_media_confirmation_popup.assert_called_once_with(
+        open_media, 'xdg-open', media_path
+    )
 
 
 @pytest.mark.parametrize('returncode, error', [
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -3,8 +3,12 @@ import pytest
 import zulipterminal.helper
 from zulipterminal.helper import (
     canonicalize_color, classify_unread_counts, display_error_if_present,
-    hash_util_decode, index_messages, notify, powerset,
+    hash_util_decode, index_messages, notify, open_media, powerset,
+    process_media,
 )
+
+
+SERVER_URL = 'https://chat.zulip.zulip'
 
 
 def test_index_messages_narrow_all_messages(mocker,
@@ -314,3 +318,52 @@ def test_hash_util_decode(quoted_string, expected_unquoted_string):
     return_value = hash_util_decode(quoted_string)
 
     assert return_value == expected_unquoted_string
+
+
+def test_process_media(mocker, media_path='/tmp/zt-somerandomtext-image.png',
+                       media_link=SERVER_URL + '/user_uploads/path/image.png'):
+    mocker.patch('zulipterminal.helper.requests')
+    mocker.patch('zulipterminal.helper.open')
+    (mocker.patch('zulipterminal.helper.NamedTemporaryFile').return_value
+     .__enter__.return_value.name) = media_path
+    # The command, which is platform dependent, does not matter for the test.
+    mocker.patch('zulipterminal.helper.LINUX', True)
+    controller = mocker.Mock()
+    mocked_open_media = mocker.patch('zulipterminal.helper.open_media')
+
+    process_media(controller, media_link)
+
+    mocked_open_media.assert_called_once_with(controller, 'xdg-open',
+                                              media_path)
+
+
+@pytest.mark.parametrize('returncode, error', [
+    (0, []),
+    (1, [' The command ', ('bold', 'some_command'), ' did not run successfully'
+         '. Exited with ', ('bold', '1')]),
+])
+def test_open_media(mocker, returncode, error, command='some_command',
+                    media_path='/tmp/zt-somerandomtext-image.png'):
+    mocked_run = mocker.patch('zulipterminal.helper.subprocess.run')
+    mocked_run.return_value.returncode = returncode
+    controller = mocker.Mock()
+
+    open_media(controller, command, media_path)
+
+    assert mocked_run.called
+    if error:
+        controller.view.set_footer_text.assert_called_once_with(error,
+                                                                duration=3)
+
+
+def test_open_media_exception(mocker, command='some_command',
+                              media_path='/tmp/zt-somerandomtext-image.png',
+                              error=[' The command ', ('bold', 'some_command'),
+                                     ' could not be found']):
+    mocker.patch('zulipterminal.helper.subprocess.run',
+                 side_effect=FileNotFoundError())
+    controller = mocker.Mock()
+
+    open_media(controller, command, media_path)
+
+    controller.view.set_footer_text.assert_called_once_with(error, duration=3)
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -181,7 +181,8 @@ class Controller:
             ('bold', media_path),
             '. Do you want to view it?'
         ])
-        self.loop.widget = PopUpConfirmationView(self, question, callback)
+        self.loop.widget = PopUpConfirmationView(self, question, callback,
+                                                 location='center')
 
     def search_messages(self, text: str) -> None:
         # Search for a text in messages
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -173,6 +173,16 @@ class Controller:
                             'Edit History (up/down scrolls)')
         )
 
+    def show_media_confirmation_popup(self, func: Any, command: str,
+                                      media_path: str) -> None:
+        callback = partial(func, self, command, media_path)
+        question = urwid.Text([
+            'Your requested media has been downloaded at ',
+            ('bold', media_path),
+            '. Do you want to view it?'
+        ])
+        self.loop.widget = PopUpConfirmationView(self, question, callback)
+
     def search_messages(self, text: str) -> None:
         # Search for a text in messages
         self.model.index['search'].clear()
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -678,7 +678,7 @@ def process_media(controller: Any, media_link: str) -> None:
     elif MACOS:
         command = 'open'
 
-    open_media(controller, command, media_path)
+    controller.show_media_confirmation_popup(open_media, command, media_path)
 
 
 @asynch
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -665,6 +665,11 @@ def process_media(controller: Any, media_link: str) -> None:
             for chunk in r.iter_content(chunk_size=8192):
                 if chunk:  # Filter out keep-alive new chunks.
                     f.write(chunk)
+                    controller.view.set_footer_text([
+                        ' Downloading ', ('bold', media_name),
+                    ])
+        controller.view.set_footer_text([' Downloaded ', ('bold', media_name)],
+                                        duration=3)
 
     if LINUX:
         command = 'xdg-open'
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -6,6 +6,7 @@ from collections import OrderedDict, defaultdict
 from functools import wraps
 from itertools import chain, combinations
 from re import ASCII, match
+from tempfile import NamedTemporaryFile
 from threading import Thread
 from typing import (
     Any, Callable, DefaultDict, Dict, FrozenSet, Iterable, List, Set, Tuple,
@@ -14,6 +15,7 @@ from typing import (
 from urllib.parse import unquote
 
 import lxml.html
+import requests
 from typing_extensions import TypedDict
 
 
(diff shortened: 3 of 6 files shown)


zulipterminal/helper.py
```diff
@@ -643,10 +644,20 @@ def open_media(controller: Any, media_link: str, message_id: int) -> None:
...
+        content_len = r.headers.get('content-length')
+        downloaded = 0
         with open(img_path, 'wb') as f:
             for chunk in r.iter_content(chunk_size=8192):
                 if chunk:  # Filter out keep-alive new chunks.
                     f.write(chunk)
+                    downloaded += len(chunk)
+                    if content_len:
+                        status = " Downloaded {}/{} bytes"
+                        controller.view.set_footer_text(
```
preetmishra, line 657: I couldn't figure out a way to test this. I was trying to use an `assert` with `...call_count == 2`, but this particular statement doesn't execute due to mocking.

I would greatly appreciate any pointers in the review.

tests/helper/test_helper.py
```diff
@@ -301,3 +306,30 @@ def test_display_error_if_present(mocker, response, footer_updated):
...
+    ('win32', 'explorer'),
+    ('darwin', 'open'),
+])
+def test_open_media(mocker, media_link, message_id, expected_path, platform,
+                    expected_command):
+    controller = mocker.Mock()
+    requests = mocker.patch('zulipterminal.helper.requests')
+    subprocess = mocker.patch('zulipterminal.helper.subprocess')
+    sys = mocker.patch('zulipterminal.helper.sys')
+    sys.platform = platform
```
neiljp, line 328: minor: these can be combined.

tests/helper/test_helper.py
```diff
@@ -301,3 +306,30 @@ def test_display_error_if_present(mocker, response, footer_updated):
...
+@pytest.mark.parametrize('media_link, message_id, expected_path', [
+    (SERVER_URL + '/user_uploads/some/path/image.png', 1,
+     gettempdir() + '/zulip-term/1/image.png'),
+    (SERVER_URL + '/user_uploads/some/path/1.png', 1,
+     gettempdir() + '/zulip-term/1/1.png'),
+])
+@pytest.mark.parametrize('platform, expected_command', [
+    ('linux', 'xdg-open'),
+    ('win32', 'explorer'),
+    ('darwin', 'open'),
```
neiljp, line 320: I'm not convinced it's worth multiplying the number of tests 3-fold vs having a separate test.

This may also change if we use the 'standard' platforms we use elsewhere now.

zulipterminal/helper.py
neiljp: I'm not sure if we want to show and update the footer as rapidly as this. Rather we could show updates every fraction of a second or thereabouts. There is a progress bar widget in urwid we could use?

This seems like it could be a separate commit to build upon the basic functionality, where we could just set 'Downloading <filename>' in the footer and clear it afterwards? (or set it to temporarily be 'downloaded'?)

zulipterminal/core.py
```diff
@@ -182,8 +188,7 @@ def stream_muting_confirmation_popup(self, button: Any) -> None:
                               "center")
         mute_this_stream = partial(self.model.toggle_stream_muted_status,
                                    button.stream_id)
-        self.loop.widget = PopUpConfirmationView(self, question,
-                                                 mute_this_stream)
+        self.show_confirmation_overlay(question, mute_this_stream)
```
neiljp, line 191: I'm not sure that we really gain from this refactor?

I think we could probably benefit from combining the confirmation popups into 'show_pop_up', however, though that doesn't seem necessary to include here but is pretty independent.

I know you use it in the next commit, but generally I'd prefer if we kept helper code clean of controller/UI, and have the latter use the helper code instead.

zulipterminal/helper.py
```diff
@@ -676,4 +677,7 @@ def open_media(controller: Any, media_link: str, message_id: int) -> None:
         'darwin': 'open',
     }
     command = image_viewers[sys.platform]
-    subprocess.run([command, img_path])
+    callback = partial(subprocess.run, [command, img_path])
+    question = urwid.Text("Your requested media has been downloaded."
+                          " Do you want to view it?")
+    controller.show_confirmation_overlay(question, callback)
```
neiljp, line 683: I know we have the controller here, but it'd be cleaner if we kept the prompting separate - maybe add a method to core instead?

zulipterminal/helper.py
```diff
@@ -676,4 +677,7 @@ def open_media(controller: Any, media_link: str, message_id: int) -> None:
         'darwin': 'open',
     }
     command = image_viewers[sys.platform]
-    subprocess.run([command, img_path])
+    callback = partial(subprocess.run, [command, img_path])
+    question = urwid.Text("Your requested media has been downloaded."
+                          " Do you want to view it?")
```
neiljp, line 682: It would be useful to indicate where it has been download to?

zulipterminal/helper.py
```diff
@@ -629,3 +632,37 @@ def display_error_if_present(response: Dict[str, Any], controller: Any
...
+    img_name = media_link.split('/')[-1]
+    target_dir = os.path.join(gettempdir(), 'zulip-term', str(message_id))
+    img_path = os.path.join(target_dir, img_name)
+
+    try:
+        os.makedirs(target_dir, exist_ok=True)
+    except FileExistsError:
+        # Raised when the target path exists and is not a directory.
+        # FIXME: This is highly unlikely. Do we need to handle this?
+        return
```
neiljp, line 653: While the zulip-term namespacing may be useful, I think this limits the possibility of multiple users saving files?

`NamedTemporaryFile` seems like a simpler approach, with appropriate prefixes and suffixes?

zulipterminal/helper.py
```diff
@@ -629,3 +633,51 @@ def display_error_if_present(response: Dict[str, Any], controller: Any
...
+
+
+@asynch
+def open_media(controller: Any, media_link: str, message_id: int) -> None:
+    """
+    Downloads media in /tmp/zulip-term/{message_id} directory and uses native
+    viewer for the output.
+    """
+    img_name = media_link.split('/')[-1]
+    target_dir = os.path.join(gettempdir(), 'zulip-term', str(message_id))
```
andersk, line 645: Never write to known paths in the temporary directory. This leads to security vulnerabilities. For example, an attacker with access to a different user account on the system can create a symlink or hardlink at the known path to trick you into overwriting a file of their choice with your permissions. Use `NamedTemporaryFile`, `TemporaryDirectory`, `mkstemp`, or `mkdtemp`, which do the extra work needed to prevent these attacks.

preetmishra:
@andersk, @neiljp Thanks for the review. I have updated the pull request to use the safer `NamedTemporaryFile` and also incorporated the other in-line suggestions.

Re UI tuning, agreed; the pull request only has a minimal downloading setup. We could follow up on this with an extended view separately.

zulipterminal/helper.py
```diff
@@ -647,3 +649,48 @@ def hash_util_decode(string: str) -> str:
...
+        if exit_status != 0:
+            error = [
+                ' The command ', ('bold', command), ' did not run successfully'
+                '. Exited with ', ('bold', str(exit_status)),
+            ]
+    except FileNotFoundError:
+        error = [' The command ', ('bold', command), ' could not be found']
+
+    if error:
+        controller.view.set_footer_text(error, duration=3)
```
neiljp, line 701: Can we keep the UI-centric parts out of `helper.py`?

zulipterminal/helper.py
```diff
@@ -647,3 +649,48 @@ def hash_util_decode(string: str) -> str:
...
+            for chunk in r.iter_content(chunk_size=8192):
+                if chunk:  # Filter out keep-alive new chunks.
+                    f.write(chunk)
+
+    if LINUX:
+        command = 'xdg-open'
+    elif WSL:
+        command = 'explorer'  # Needs testing.
+    elif MACOS:
+        command = 'open'
```
neiljp, line 679: minor: It would be useful if the user was able to know what the file was going to be opened with in the popup (see other comment)

zulipterminal/ui_tools/buttons.py
```diff
@@ -300,6 +300,11 @@ def handle_link(self, *_: Any) -> None:
         server_url = self.model.server_url
         if self.link.startswith(urljoin(server_url, '/#narrow/')):
             self.handle_narrow_link()
+        elif self.link.startswith(urljoin(server_url, '/user_uploads/')):
+            # Exit pop-up promptly, let the media download in the background.
+            if isinstance(self.controller.loop.widget, urwid.Overlay):
+                self.controller.exit_popup()
```
neiljp, line 306: What does this check avoid? I know you've used it elsewhere, but I don't remember what the conclusion was - can we discuss and then maybe add a comment. Does it belong in `exit_popup`?

zulipterminal/helper.py
```diff
@@ -665,6 +665,11 @@ def process_media(controller: Any, media_link: str) -> None:
             for chunk in r.iter_content(chunk_size=8192):
                 if chunk:  # Filter out keep-alive new chunks.
                     f.write(chunk)
+                    controller.view.set_footer_text([
+                        ' Downloading ', ('bold', media_name),
+                    ])
+        controller.view.set_footer_text([' Downloaded ', ('bold', media_name)],
+                                        duration=3)
```
neiljp, line 672: I'm not sure if this will end up needing it's own commit, but perhaps make these callbacks to keep the UI separate in any case?

zulipterminal/core.py
```diff
@@ -173,6 +173,16 @@ def show_edit_history(
                             'Edit History (up/down scrolls)')
         )
 
+    def show_media_confirmation_popup(self, func: Any, command: str,
+                                      media_path: str) -> None:
+        callback = partial(func, self, command, media_path)
+        question = urwid.Text([
+            'Your requested media has been downloaded at ',
+            ('bold', media_path),
+            '. Do you want to view it?'
```
neiljp, line 182: Given the length of the download URLs for some files, this might well benefit from more vertical spacing, eg.
```
Your requested media has been downloaded to:
<filename>

Do you want the application to open it with <tool>?

File: zulipterminal/helper.py [hunk 14]
```diff
@@ -678,7 +678,7 @@ def process_media(controller: Any, media_link: str) -> None:
     elif MACOS:
         command = 'open'
 
-    open_media(controller, command, media_path)
+    controller.show_media_confirmation_popup(open_media, command, media_path)
```
neiljp, line 681: I'm guessing you're trying to synchronize the asynch download terminating?

zulipterminal/ui_tools/views.py
```diff
@@ -1006,9 +1006,13 @@ def __init__(self, controller: Any, title: str) -> None:
         super().__init__(controller, widgets, 'HELP', popup_width, title)
 
 
+PopUpConfirmationViewLocation = Literal['top-left', 'center']
```
neiljp, line 1009: I think this comment should go before the previous one, as there is nowhere near enough space in the top-left for showing the download information :)

neiljp (commented):
@preetmishra The latest conflict is small, and I wanted to get this in soon if we can - having just tested it again it seems to work fine, and it'll be great to have merged :+1:

It would be good if we could keep the UI out of the helper code, which may require a little restructuring of the control flow but should be fairly straightforward.

@amanagr Could you (or a delegated WSL contributor) check the feasibility of `explorer` on WSL, as used in this PR?

@andersk Do you have further thoughts or concerns here?

amanagr:
(Sorry I no longer have access to WSL)

neiljp:
It may be better to use `os.startfile` on WSL? This still needs testing/exploration, since potentially the appropriate WSL version has `xdg` features we can use in any case?

zulipterminal/helper.py
Ezio-Sarthak: Now that we have `report_*` methods, would it be OK to use here?

zulipterminal/helper.py
Ezio-Sarthak: I'm not sure if I fully understood your comment, but I think the callback would be working fine in any case?

zulipterminal/ui_tools/buttons.py
Ezio-Sarthak: @preetmishra could you elaborate this?

zulipterminal/ui_tools/views.py
Ezio-Sarthak: You mean commit?

zulipterminal/ui_tools/buttons.py
preetmishra: @Ezio-Sarthak The pop-ups open in an urwid Overlay. This check is to make sure `exit_popup` is only called when an Overlay exist.

Ezio-Sarthak: Ah I see that now. If I avoid that check, there is some unwanted synchronous behavior in footer and popup, if that makes any sense.

neiljp:
The followup to this was merged with slight adjustments in #1223 :tada:

Thanks @amanagr, @preetmishra and @Ezio-Sarthak 👍

PR #741, opened 2020-07-25
Move unicode characters to new file

### LINKED ISSUE #717: Move unicode characters to new file
ZT uses a lot of unicode characters for displaying things like separators, lines, etc. These are hard-coded right now, but should be moved to their own file.

kaustubh-nair:
Currently covered the following characters:
* Topic-stream separator `▶` (Merged)
* Message content marker `▒` (Merged)
* Quoted text marker `░` (Merged)
* Message header divider `━` (Merged)
* Checkbox tick mark `✓` (Merged)
* Application title bar divider = `═`
* pinned streams divider = `-`
* List title bar divider = `━`

Fixes #717


zulipterminal/ui_tools/boxes.py
```diff
@@ -314,7 +317,7 @@ def _is_private_message_to_self(self) -> bool:
                 and recipient_list[0]['email'] == self.model.user_email)
 
     def stream_header(self) -> Any:
-        stream_topic_separator = '▶'
+        stream_topic_separator = STREAM_TOPIC_SEPARATOR
```
sumanthvrao, line 321: There seems to be only one occurance of `stream_topic_separator` further below, so we could just replace that with `STREAM_TOPIC_SEPARATOR` instead of doing this substitution.

zulipterminal/config/unicode_characters.py
```diff
@@ -1 +1,3 @@
 STREAM_TOPIC_SEPARATOR = '▶'
+# Used as a separator between messages and 'EDITED'
+MESSAGE_LEFT_PADDING = '▒'  # Options are '█', '▓', '▒', '░'
```
neiljp, line 3: Maybe `MESSAGE_CONTENT_MARKER`?

zulipterminal/ui_tools/boxes.py
```diff
@@ -781,7 +781,7 @@ def transform_content(self) -> Tuple[None, Any]:
         soup = BeautifulSoup(self.message['content'], 'lxml')
         body = soup.find(name='body')
         if body and body.find(name='blockquote'):
-            padding_char = '░'
+            padding_char = QUOTED_TEXT_PADDING
             self.indent_quoted_content(soup, padding_char)
```
neiljp, line 788: Maybe combine these two lines now, as with the other case?

I'm aware the name likely came from the variable, but maybe `QUOTED_TEXT_MARKER`?

tests/ui/test_ui_tools.py
```diff
@@ -2137,6 +2137,7 @@ def test_keypress_EDIT_MESSAGE(self, mocker, message_fixture,
     ])
     def test_transform_content(self, mocker, raw_html, expected_content,
                                messages_successful_response):
+        expected_content = expected_content.replace('{}', QUOTED_TEXT_PADDING)
```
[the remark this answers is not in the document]
kaustubh-nair: It is smaller than using a word, and `{}` automatically suggests that there is something to be substituted here, since `.format()` using the same notation. So it will make the parameters easier to understand.

neiljp:
@kaustubh-nair Thanks for extracting these - I just slightly adjusted the commit text to fit better and cherry-picked these first five commits :tada: Adding the others I indicated would be good, whether in this PR or a new one :+1:

PR #744, opened 2020-07-28
migration: Support API migration for muted topics

preetmishra:
This reflects the expected response change for muted_topics, from `[stream_name, topic]` to `[stream_name, topic, date_muted]` in Zulip version 3.0, Feature level 1 (see https://github.com/zulip/zulip/commit/9340cd1a0b7af7af9b83a622e37d02f1a2340609).

Tests amended.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -752,6 +752,28 @@ def stream_dict(streams_fixture):
     return {stream['stream_id']: stream for stream in streams_fixture}
 
 
+@pytest.fixture(params=[
+        {
+            ('Stream 1', 'muted stream muted topic'): None,
+            ('Stream 2', 'muted topic'): None,
+        },
+        {
+            ('Stream 1', 'muted stream muted topic'): 1530129122,
+            ('Stream 2', 'muted topic'): 1530129122,
+        },
+    ],
+    ids=[
+        'zulip_feature_level:None',
+        'zulip_feature_level:1',
+    ]
+)
+def processed_muted_topics(request):
+    """
+    Locally processed muted topics data (see _muted_topics in Model.__init__).
+    """
+    return request.param
+
+
 @pytest.fixture
 def classified_unread_counts():
     """
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1578,17 +1578,18 @@ class TestModel:
         assert model.is_muted_stream(stream_id) == is_muted
 
     @pytest.mark.parametrize('topic, is_muted', [
-        ((1, 'stream muted & unmuted topic'), True),
+        ((1, 'stream muted & unmuted topic'), False),
         ((2, 'muted topic'), True),
         ((1, 'muted stream muted topic'), True),
         ((2, 'unmuted topic'), False),
     ])
     def test_is_muted_topic(self, topic, is_muted, stream_dict, model):
         model.stream_dict = stream_dict
-        model.muted_streams = [1]
         model.muted_topics = [
             ['Stream 2', 'muted topic'],
             ['Stream 1', 'muted stream muted topic'],
         ]
-        assert model.is_muted_topic(stream_id=topic[0],
-                                    topic=topic[1]) == is_muted
+
+        return_value = model.is_muted_topic(stream_id=topic[0], topic=topic[1])
+
+        assert return_value == is_muted
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -72,6 +72,35 @@ class TestModel:
         self.classify_unread_counts.assert_called_once_with(model)
         assert model.unread_counts == []
 
+    @pytest.mark.parametrize(['server_response', 'locally_processed_data',
+                              'zulip_feature_level'], [
+            (
+                [['Stream 1', 'muted stream muted topic']],
+                {('Stream 1', 'muted stream muted topic'): None},
+                None,
+            ),
+            (
+                [['Stream 2', 'muted topic', 1530129122]],
+                {('Stream 2', 'muted topic'): 1530129122},
+                1,
+            ),
+        ],
+        ids=[
+            'zulip_feature_level:None',
+            'zulip_feature_level:1',
+        ]
+    )
+    def test_init_muted_topics(self, mocker, initial_data, server_response,
+                               locally_processed_data, zulip_feature_level):
+        mocker.patch('zulipterminal.model.Model.get_messages', return_value='')
+        initial_data['zulip_feature_level'] = zulip_feature_level
+        initial_data['muted_topics'] = server_response
+        self.client.register = mocker.Mock(return_value=initial_data)
+
+        model = Model(self.controller)
+
+        assert model._muted_topics == locally_processed_data
+
     def test_init_InvalidAPIKey_response(self, mocker, initial_data):
         # Both network calls indicate the same response
         mocker.patch('zulipterminal.model.Model.get_messages',
@@ -1583,12 +1612,10 @@ class TestModel:
         ((1, 'muted stream muted topic'), True),
         ((2, 'unmuted topic'), False),
     ])
-    def test_is_muted_topic(self, topic, is_muted, stream_dict, model):
+    def test_is_muted_topic(self, topic, is_muted, stream_dict, model,
+                            processed_muted_topics):
         model.stream_dict = stream_dict
-        model._muted_topics = [
-            ['Stream 2', 'muted topic'],
-            ['Stream 1', 'muted stream muted topic'],
-        ]
+        model._muted_topics = processed_muted_topics
 
         return_value = model.is_muted_topic(stream_id=topic[0], topic=topic[1])
 
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1585,7 +1585,7 @@ class TestModel:
     ])
     def test_is_muted_topic(self, topic, is_muted, stream_dict, model):
         model.stream_dict = stream_dict
-        model.muted_topics = [
+        model._muted_topics = [
             ['Stream 2', 'muted topic'],
             ['Stream 1', 'muted stream muted topic'],
         ]
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -2679,34 +2679,24 @@ class TestTopicButton:
         assert topic_button.stream_id == stream_id
         assert topic_button.topic_name == title
 
-    @pytest.mark.parametrize(['stream_name', 'title',
-                              'is_muted_topic_return_value',
-                              'is_muted_called'], [
-        ('Django', 'topic1', True, True),
-        ('Django', 'topic2', False, False),
-        ('GSoC', 'topic1', False, False),
-    ], ids=[
-        # Assuming 'Django', 'topic1' is muted via muted_topics.
-        'stream_and_topic_match',
-        'topic_mismatch',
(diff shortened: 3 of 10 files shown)


preetmishra:
Logged #745 to track related improvements.

preetmishra:
@neiljp Thanks for the review. I have updated the PR as per our discussion in the stream. This is larger than what we started with, but I believe the refactoring is worthwhile.

I have also added the proposed extension in #745.

zulipterminal/model.py
```diff
@@ -423,7 +427,14 @@ def is_in_muted_topics(self, stream_id: int, topic: str) -> bool:
         """
         stream_name = self.stream_dict[stream_id]['name']
         topic_to_search = [stream_name, topic]  # type: List[str]
-        return topic_to_search in self.muted_topics
+        # The intent is to use the following data structure locally for lookups
+        # without having server version & feature level checks throughout the
+        # codebase.
+        muted_topics_lookup = [
```
sumanthvrao, line 433: This is a nice idea to unify both versions! 👍

zulipterminal/ui_tools/buttons.py
```diff
@@ -242,8 +242,8 @@ def __init__(self, stream_id: int, topic: str,
                          width=width,
                          count=count)
 
-        topic_pair = [self.stream_name, self.topic_name]
-        if topic_pair in controller.model.muted_topics:
```
neiljp, line 246: I'll emphasize that we should add a leading underscore to `Model.muted_topics` - it'll mark it as 'private' (by convention at least), and any hidden uses of it are made evident too :)

tests/ui/test_utils.py
```diff
@@ -83,8 +82,23 @@ def test_is_muted(mocker, msg, narrow, muted_streams, muted_topics, muted):
...
+        1: {
+            'name': 'foo',
+        },
+        2: {
+            'name': 'boo',
+        },
+        3: {
+            'name': 'zoo',
+        },
+    }
```
neiljp, line 95: Would using an existing stream fixture make sense here?

tests/ui/test_utils.py
```diff
@@ -83,8 +82,23 @@ def test_is_muted(mocker, msg, narrow, muted_streams, muted_topics, muted):
...
+    if msg['type'] == 'stream':
+        model.is_in_muted_topics = mocker.Mock(return_value=(
+            [model.stream_dict[msg['stream_id']]['name'],
+             msg.get('subject', '')] in muted_topics
+        ))
```
neiljp, line 100: This seems strange since it's basically reimplementing the function?

tests/model/test_model.py
```diff
@@ -1516,12 +1516,20 @@ def test_is_muted_stream(self, muted_streams, stream_id, is_muted,
...
+        if server_version == '3.0' and server_feature_level >= 1:
+            model.muted_topics = [
+                ['Stream 2', 'muted topic', 1530129122],
+                ['Stream 1', 'muted stream muted topic', 1530129122],
+            ]
+        elif server_version == '2.1':
+            model.muted_topics = [
+                ['Stream 2', 'muted topic'],
+                ['Stream 1', 'muted stream muted topic'],
+            ]
```
neiljp, line 1532: I'd prefer that we had this version-dependent data in a fixture, if possible. Also can we depend upon just the feature level?

zulipterminal/model.py
```diff
@@ -423,7 +427,14 @@ def is_in_muted_topics(self, stream_id: int, topic: str) -> bool:
...
+            [stream_name, topic]
+            for stream_name, topic, *_ in self.muted_topics
+        ]
+        return topic_to_search in muted_topics_lookup
```
neiljp, line 437: This approach does work, but it would be beneficial to have some defensive programming (assert on type of the incoming data), and I wonder if we can generate and use just one data-structure, like a dict?

zulipterminal/model.py
```diff
@@ -115,8 +115,12 @@ def __init__(self, controller: Any) -> None:
         (self.stream_dict, self.muted_streams,
          self.pinned_streams, self.unpinned_streams) = stream_data
 
+        # NOTE: The expected response has been upgraded from
+        # [stream_name, topic] to [stream_name, topic, date_muted] in
+        # 3.0, Feature level 1.
```
neiljp, line 120: It'd be useful to standardize our notes in the code for this; we could add that to the development section - this would make it easier to check for changes at different server versions and feature levels :)

preetmishra:
@neiljp The first two commits make modification in the existing code to facilitate the migration. The last commit is the migration commit (should we call it a bugfix?). Note that the first two commits do not introduce any 'new' feature.

zulipterminal/model.py
```diff
@@ -417,9 +417,10 @@ def fetch_all_topics(self, workers: int) -> None:
     def is_muted_stream(self, stream_id: int) -> bool:
         return stream_id in self.muted_streams
 
-    def is_muted_topic(self, stream_id: int, topic: str) -> bool:
-        if stream_id in self.muted_streams:
-            return True
+    def is_in_muted_topics(self, stream_id: int, topic: str) -> bool:
+        """
+        Returns True if the topic_name is muted via muted_topics.
+        """
```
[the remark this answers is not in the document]
preetmishra: The reason why I pivoted from `is_muted_topic` is that it is slightly confusing given that we could call a topic muted even if it belongs to a muted stream (?) and in this case, we are more interested in whether the topic belongs to the `muted_topics` data structure.

tests/ui/test_utils.py
preetmishra: The stream names used in its parametrize are different from `stream_dict`. I could change the stream names in a separate commit.

preetmishra:
@neiljp Thanks for the review. In the latest update, I have simplified the parameterize blocks to avoid unnecessary mocking wherever feasible, made `muted_topics` protected with an underscore, added `_muted_topics_dict` along with an assert, and addressed other in-line suggestions (excluding the standard comment that we are discussing in [zulip-terminal > Comments for server versions](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Comments.20for.20server.20versions)).

preetmishra:
(Rebased origin/master.)

preetmishra:
(Updated to rebase onto upstream/master.)

preetmishra:
Updated to resolve conflicts.

zulipterminal/model.py
```diff
@@ -116,8 +116,20 @@ def __init__(self, controller: Any) -> None:
...
+            self.initial_data['muted_topics']
+        )  # type: List[List[Union[str, int]]]
+        assert all(len(sub_list) in (2, 3) for sub_list in self._muted_topics)
+        # The intent is to use the following data structure locally to avoid
+        # server version & feature level checks throughout the codebase.
+        self._muted_topics_dict = {
+            (stream_name, topic): (None if self.server_feature_level is None
+                                   else date_muted[0])
+            for stream_name, topic, *date_muted in self._muted_topics
+        }
```
neiljp: I'm glad you understood what I was suggesting here - this looks much simpler :+1:

I'm not sure we still need to store `self._muted_topics` though? Can we just name the final variable that instead?
(I know we have `_dict` on other variables, but including the type in the name is often superfluous)

tests/conftest.py
```diff
@@ -752,6 +752,25 @@ def stream_dict(streams_fixture):
...
+            ('Stream 2', 'muted topic'): None,
+            ('Stream 1', 'muted stream muted topic'): None,
+        },
+    ],
+    ids=[
+        'server_feature_level:1',
+        'server_feature_level:None',
+    ]
+)
+def muted_topics_dict(request):
```
neiljp, line 770: This is the parametrization of muted_topics in the Model only - we should likely have one for initial data from server, and a test for initial data -> model data, which we don't have? (that was in the previous iteration?)

tests/model/test_model.py
```diff
@@ -1585,7 +1585,7 @@ def test_is_muted_stream(self, muted_streams, stream_id, is_muted,
     ])
     def test_is_muted_topic(self, topic, is_muted, stream_dict, model):
         model.stream_dict = stream_dict
-        model.muted_topics = [
+        model._muted_topics = [
```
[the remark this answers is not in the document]
preetmishra: Hmm. I am unable to find any other `muted_topics` instance in the code. Could you point out where?

Re commit, I am inclined towards having it here since we replace the old references with `is_muted_topic()` and it is easier to follow with less number of changes.

tests/conftest.py
preetmishra: I didn't have that test in the previous iteration. Thanks for pointing that out.

I have added the test in the latest update with a local `parametrize` block in `Model.py`. It wouldn't be needed in `conftest` as we'd only test against `processed_muted_topics` everywhere else.

tests/ui/test_ui_tools.py
```diff
@@ -2677,22 +2679,25 @@ def test_init_calls_top_button(self, mocker, width, count, title,
...
+    @pytest.mark.parametrize(['stream_name', 'title', 'is_muted_topic_return',
                               'is_muted_called'], [
-        ('Django', 'topic1', [['Django', 'topic1']], True),
-        ('Django', 'topic2', [['Django', 'topic1']], False),
-        ('GSoC', 'topic1', [['Django', 'topic1']], False),
+        ('Django', 'topic1', True, True),
+        ('Django', 'topic2', False, False),
+        ('GSoC', 'topic1', False, False),
     ], ids=[
+        # Assuming 'Django', 'topic1' is muted via muted_topics.
```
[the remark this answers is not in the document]
preetmishra: Yes, it is needed here to understand why the values are set to `True`/`False` for `is_muted_topic_return_value`. Though, I have added an extra refactor commit to simplify it.

The `is_muted` util test is simplified in this commit itself. We might want to discuss `test_classify_unread_counts` - how would we like it to be handled: Maybe as a separate issue.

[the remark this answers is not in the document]
preetmishra: Yes, it is needed here to understand why the values are set to `True`/`False` for `is_muted_topic_return_value`. Though, I have added an additional refactor commit to simplify it.

The `is_muted` util test is simplified in this commit itself. We might want to discuss `test_classify_unread_counts` - how would we like it to be handled? Maybe as a separate issue.

preetmishra:
@neiljp Thanks for the review. This is ready with improved assertion against server response, improved tests, an additional refactor commit and other in-line suggestions that you mentioned.

tests/model/test_model.py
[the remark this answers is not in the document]
neiljp: I can't find another instance now, so this looks good :+1:

tests/conftest.py
neiljp: I'm actually thinking we might move the server response fixtures into a separate fixture file specific to the model in a similar way. What do you think?

preetmishra: The ones that are already in `conftest.py` or also the ones that are used in `test_model.py`?

preetmishra: In any case, I believe it is reasonable.

neiljp:
@preetmishra Thanks for this bugfix and compatibility change (as well as the refactors) :+1: As per PMs I've merged the first four :tada:

I think we can save the last commit for a separate set of changes to review separately. Other than unblocking the other PR, this also could be followed by:
* a server-response fixture refactor (after discussion)
* feature-level comments (we briefly discussed already)

PR #746, opened 2020-07-28
Recipient box typeahead for streams and topics

kaustubh-nair:


zulipterminal/helper.py
```diff
@@ -490,6 +490,11 @@ def match_emoji(emoji: str, text: str) -> bool:
     return emoji.lower().startswith(text.lower())
 
 
+def match_stream_names(stream_names: List[str], search_text: str) -> List[str]:
+    return [name for name in stream_names
+            if name.lower().startswith(search_text.lower())]
+
+
```
kaustubh-nair, line 502: There are no pinned stream considerations here - I'm not using the existing function.
Refactoring `match_streams` looks a little hard, and pinning order is not super important - so I'd like to deal with this in a follow-up.

kaustubh-nair: Though one hacky way to use the current `match_streams` would be to pass `[(stream_name, stream_name)..]` in data`.

neiljp:
@kaustubh-nair Is this PR supposed to be functional at this point?

zulipterminal/ui_tools/boxes.py
```diff
@@ -63,9 +64,11 @@ def private_box_view(self, button: Any=None, email: str='') -> None:
         ]
         self.focus_position = 1
 
-    def stream_box_view(self, caption: str='', title: str='') -> None:
+    def stream_box_view(self, stream_id: int, caption: str='', title: str=''
+                        ) -> None:
         self.set_editor_mode()
         self.to_write_box = None
+        self.stream_id = stream_id
```
neiljp, line 71: This gets set only on set-up, so if the stream is changed before the message is sent, the topic autocomplete uses the stream used at initialization. This is the same issue present with another PR, maybe?

zulipterminal/helper.py
kaustubh-nair: They are certainly an improvement, but I don't see it as being a necessity for this particular feature. I'll make a temp fix but open an issue and make a follow-up PR for refactoring `match_stream`

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Yes, I can fix it now that #738 is merged

zulipterminal/ui_tools/boxes.py
```diff
@@ -144,13 +144,13 @@ def generic_autocomplete(self, text: str, state: Optional[int]
...
             typeahead = fewer_typeaheads[state]  # type: Optional[str]
-            if typeahead:
-                typeahead = text[:prefix_index] + typeahead
         else:
             typeahead = None
             state = None
         self.view.set_typeahead_footer(reduced_suggestions,
                                        state, is_truncated)
+        if typeahead:
+            typeahead = text[:prefix_index] + typeahead
```
neiljp: Is this change necessary, given the next commit? Aren't you moving it back?

kaustubh-nair: The git diff is a bit weird so it looks a little strange, but this statement stays in the same method.

tests/ui_tools/test_boxes.py
```diff
@@ -251,6 +251,39 @@ def test_generic_autocomplete_emojis(self, write_box, text,
...
+        write_box.view.set_typeahead_footer = mocker.patch(
+                                'zulipterminal.ui.View.set_typeahead_footer')
+
+        typeahead = write_box._process_typeaheads(typeaheads, state,
+                                                  suggestions)
+
+        assert typeahead == expected_typeahead
+        write_box.view.set_typeahead_footer.assert_called_once_with(
+                            suggestions[:10], expected_state, is_truncated)
+
```
kaustubh-nair, line 340: I'm not sure what to do about the existing `generic_autocomplete` tests now, since now they span two different methods. Changing them to check for `_process_typeaheads` would be hard because the parameters would get complex. (Because we'd have to create lists of suggestions)

zulipterminal/ui_tools/boxes.py
```diff
@@ -103,6 +108,19 @@ def stream_box_view(self, stream_id: int, caption: str='', title: str='',
         ]
         self.contents = write_box
 
+    def _stream_box_autocomplete(self, text: str, state: Optional[int]
+                                 ) -> Optional[str]:
+        streams_list = self.view.pinned_streams + self.view.unpinned_streams
+        streams = [stream[0]
+                   for stream in streams_list]
+
+        stream_data = list(zip(streams, streams))
```
kaustubh-nair, line 134: Is this okay, or should `match_stream` be refactored?

zulipterminal/ui_tools/boxes.py
```diff
@@ -103,6 +108,19 @@ def stream_box_view(self, stream_id: int, caption: str='', title: str='',
...
+        matched_streams = match_stream(stream_data, text,
+                                       self.view.pinned_streams)
+
+        return self._process_typeaheads(matched_streams[0], state,
+                                        matched_streams[1])
```
kaustubh-nair, line 139: `matched_stream[0]` and `matched_stream[1]` have the same data, but I've used both of them for the sake of consistency.

kaustubh-nair:
@neiljp Thanks for the review! The problem you see for stream autocomplete is also the same bug caused due to readline's ability to autocomplete only words. It shows all possible suggestions for a an empty search string.
I've pushed a newer version with the following updates:
* Added tests with cases for failing whitespace errors.
* Fixed for changing topics by dynamically loading them when needed.
* New commit for improving invalid stream error message.

zulipterminal/ui_tools/boxes.py
neiljp: This is resolved by the change at the end  of this commit?

zulipterminal/ui_tools/boxes.py
```diff
@@ -134,10 +133,19 @@ def generic_autocomplete(self, text: str, state: Optional[int]
...
             autocomplete_func(text[prefix_index:], prefix)
         )
+
+        typeahead = self._process_typeaheads(typeaheads, state,
+                                             suggestions)
+        if typeahead:
+            typeahead = text[:prefix_index] + typeahead
+        return typeahead
+
+    def _process_typeaheads(self, typeaheads: List[str], state: Optional[int],
```
neiljp, line 181: This is a very general method name - is there anything more specific we could use?

tests/ui_tools/test_boxes.py
```diff
@@ -251,6 +251,34 @@ def test_generic_autocomplete_emojis(self, write_box, text,
...
 
+    @pytest.mark.parametrize(['suggestions', 'state', 'expected_state',
+                              'expected_typeahead', 'is_truncated'], [
+      (['zero', 'one', 'two'], 1, 1, '*one*', False),
+      (['zero', 'one', 'two'] * 4, 1, 1, '*one*', True),
+      (['zero', 'one', 'two'], None, None, None, False),
+      (['zero', 'one', 'two'], 5, None, None, False),
+      (['zero', 'one', 'two'], -5, None, None, False),
+     ], ids=[
+        'less_than_10_typeaheads',
```
neiljp, line 316: minor: These are 'countable', so 'fewer' would be more accurate than 'less'

tests/ui_tools/test_boxes.py
```diff
@@ -251,6 +251,34 @@ def test_generic_autocomplete_emojis(self, write_box, text,
...
+        'more_than_10_typeaheads',
+        'invalid_state-None',
+        'invalid_state-greater_than_possible_index',
+        'invalid_state-less_than_possible_index',
```
neiljp, line 320: While accurate, are these literally invalid state? Do they represent internal state that should not happen? If we're testing the range of inputs, should we cover those that the typeahead footer understands/expects?

It's also difficult knowing whether to test this method directly, rather than just cases where it's used, which I think you're getting the feeling of from your other review comment.

tests/ui_tools/test_boxes.py
```diff
@@ -251,6 +251,34 @@ def test_generic_autocomplete_emojis(self, write_box, text,
...
+    ])
+    def test__process_typeaheads(self, write_box, suggestions, state,
+                                 expected_state, expected_typeahead,
+                                 is_truncated, mocker):
+        write_box.view.set_typeahead_footer = mocker.patch(
+                                'zulipterminal.ui.View.set_typeahead_footer')
+        typeaheads = ['*{}*'.format(s) for s in suggestions]
```
neiljp, line 327: This is an example formatting rather than specific, right? If so, worth a comment.

zulipterminal/model.py
```diff
@@ -645,6 +645,13 @@ def toggle_stream_pinned_status(self, stream_id: int) -> bool:
         response = self.client.update_subscription_settings(request)
         return response['result'] == 'success'
 
+    def stream_ids_from_name(self, stream_name: str) -> int:
+        for stream_id, stream in self.stream_dict.items():
+            if stream['name'] == stream_name:
+                return stream_id
+
+        raise RuntimeError("Invalid stream name specified.")
```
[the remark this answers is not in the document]
kaustubh-nair: I'm using this just to deal with the mypy error of the method not having a return value. Do I return something like `''` instead?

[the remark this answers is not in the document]
neiljp: Well, we could do various things in case of error (empty string, `None`, custom exception, general exception), as long as we "handle" them somehow. mypy only knows return types, so exceptions act as a fix for this at some level from its perspective - and not always in a wrong way. As I said, having this exception is fine as long as we expect all stream names to be valid and otherwise it being an internal error - so here we don't expect to handle it. Later we might change it to be something that callers are expected to handle.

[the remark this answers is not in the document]
kaustubh-nair: Okay, this should be fine for now because the recipient validity check ensures that stream names are always valid.

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Yes that should keep `stream_id` in sync

tests/ui_tools/test_boxes.py
kaustubh-nair: @neiljp I'm not sure whose internal state you're referring to - is that zt or readline? They are invalid, in the sense that they are indexes that do not exist for the array under consideration.

We don't have to cover all cases for footer since it is tested completely in `test_ui`?

I think it is okay to test method in isolation, as long as we cover all the cases for invalid states. This would mean that other tests for autocomplete methods will require fewer test parameters, but they would be difficult to write because we'd have to use lists instead, unless of course we use really small lists.

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: This was merged via the commit in #772, so this should be good for now.

tests/ui_tools/test_boxes.py
kaustubh-nair: Fixed

tests/ui_tools/test_boxes.py
kaustubh-nair: Added

zulipterminal/ui_tools/boxes.py
kaustubh-nair: I kept it general because there are multiple things happening here - returning typeahead after checking for state constraints, and setting suggestions to footer.

tests/ui_tools/test_boxes.py
```diff
@@ -251,6 +251,38 @@ def test_generic_autocomplete_emojis(self, write_box, text,
...
+                     'Stream 1', 'Stream 2']),
+        ('', 1, [['Stream 2']], ['Stream 2', 'Secret stream',
+                                 'Some general stream', 'Stream 1']),
+        ('St', 1, [], ['Stream 1', 'Stream 2', 'Secret stream',
+                       'Some general stream']),
+        ('St', 1, [['Stream 2']], ['Stream 2', 'Stream 1',
+                                   'Secret stream', 'Some general stream']),
+        pytest.param('Some gen', 1, [], ['Some general stream'],
+                     marks=pytest.mark.xfail(
+                         reason="Lacking urwid-readline support")),
```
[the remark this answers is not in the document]
kaustubh-nair: Ohh, since this is a readline problem, we'd have to test for the actual writebox contents instead of testing the autocomplete methods.

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Added

zulipterminal/ui_tools/boxes.py
kaustubh-nair: Added

zulipterminal/ui_tools/boxes.py
```diff
@@ -109,6 +117,15 @@ def stream_box_view(self, stream_id: int, caption: str='', title: str='',
...
+    def _topic_box_typeahead(self, text: str, state: Optional[int]
+                             ) -> Optional[str]:
+        topic_names = self.model.topics_in_stream(self.stream_id)
+
+        topic_typeaheads = match_topics(topic_names, text)
+
+        return self._process_typeaheads(topic_typeaheads, state,
+                                        topic_typeaheads)
+
     def _stream_box_autocomplete(self, text: str, state: Optional[int]
```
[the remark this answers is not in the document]
kaustubh-nair: FIxed

zulipterminal/ui_tools/boxes.py
```diff
@@ -306,7 +306,12 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                     if header.focus_col == 0:
                         stream_name = header[0].edit_text
                         if not self.model.is_valid_stream(stream_name):
-                            self.view.set_footer_text("Invalid stream name", 3)
+                            invalid_stream_error = (
+                                'Invalid stream name.'
+                                ' Use {} to autocomplete.'
+                                .format(keys_for_command('AUTOCOMPLETE').pop())
```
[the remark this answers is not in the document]
kaustubh-nair: Fixed

tests/ui_tools/test_boxes.py
```diff
@@ -251,6 +251,53 @@ def test_generic_autocomplete_emojis(self, write_box, text,
...
+        _process_typeaheads = mocker.patch(BOXES
+                                           + '.WriteBox._process_typeaheads')
+
+        write_box._stream_box_autocomplete(text, state)
+
+        _process_typeaheads.assert_called_once_with(matching_streams, state,
+                                                    matching_streams)
+
+    @pytest.mark.parametrize('text, expected_text', [
+        ('Som', 'Some general stream'),
```
kaustubh-nair, line 283: Added a normal case to make sure test works properly

kaustubh-nair:
@neiljp Thanks for the review! PR updated.

README.md
```diff
@@ -172,31 +172,31 @@ source ~/.zshenv
...
-| Paste last kill                                       | <kbd>Ctrl</kbd> + <kbd>Y</kbd>                |
-| Undo last action                                      | <kbd>Ctrl</kbd> + <kbd>_</kbd>                |
-| Jump to previous line                                 | <kbd>Ctrl</kbd> + <kbd>P</kbd> / <kbd>↑</kbd> |
-| Jump to next line                                     | <kbd>Ctrl</kbd> + <kbd>N</kbd> / <kbd>↓</kbd> |
-| Clear compose box                                     | <kbd>Ctrl</kbd> + <kbd>L</kbd>                |
+| Command                                                   | Key Combination                               |
+| -----------------------------------------------------     | --------------------------------------------- |
+| Cycle through recipient and content boxes                 | <kbd>tab</kbd>                                |
+| Send a message                                            | <kbd>Alt Enter</kbd> / <kbd>Ctrl d</kbd>      |
+| Autocomplete @mentions, #stream_names, :emoji: and topics | <kbd>Ctrl</kbd> + <kbd>f</kbd>                |
```
kaustubh-nair, line 179: I hope this is okay, we don't have wrapping text for topics, nor do we support typeahead for it in message box

PR #748, opened 2020-07-28
bugfix: ui: Do not update footer if new text is same as current.

### LINKED ISSUE #647: Footer gets updated as many times as the edit key is pressed on any uneditable message

kaustubh-nair:
This prevents rapid footer updates if an error message is triggered
multiple times. An example of this would be a user holding-down/pressing
too often the edit message key on a message that cannot be edited.

This fix does not account for updates that might happen for different
messages, but we can safely assume that the time difference between such
messages would not be too small and such a bug would most likely not
occur.

Test added.

Fixes #647.

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -76,6 +76,15 @@ class TestView:
         assert view.users_view == right_view()
         assert return_value == (line_box(), right_tab())
 
+    def test_set_footer_text_same_test(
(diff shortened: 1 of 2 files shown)


kaustubh-nair:
This is an alternative to #725

neiljp:
@kaustubh-nair Rebased and merged this great simple fix with minor tweaks, thanks! :tada:

PR #749, opened 2020-07-29
Markup markdown text in stream description

### LINKED ISSUE #743: Markup markdown text in stream description
We could potentially extract the markup methods that we currently use for marking-up the message content in `MessageBox` for this purpose.

For comparison, see [#integration](https://chat.zulip.org/#narrow/stream/127-integrations) or [#feedback](https://chat.zulip.org/#narrow/stream/137-feedback) description on the web app.

preetmishra:
While the PR's ultimate intent is to markup the stream description, the first commit has changes to decouple our markup methods.

#### Commits
* The first commit makes our primary markup methods, `transform_content`, `soup2markup` and `indent_quoted_content`, independent of their class object.
* The second commit uses `transform_content` in `StreamInfoView` for markup.
* The third commit makes `footlinks_view` static and more configurable.

Fixes #743.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -713,30 +713,46 @@ class MessageBox(urwid.Pile):
         message_links: 'OrderedDict[str, Tuple[str, int, bool]]',
         *,
         footlinks_enabled: bool,
-    ) -> Any:
+        padded: bool,
+        wrap: str
+    ) -> Tuple[Any, int]:
+        """
+        Returns a Tuple that consists footlinks view (widget) and its required
+        width.
+        """
         # Return if footlinks are disabled by the user.
         if not footlinks_enabled:
-            return None
+            return None, 0
 
         footlinks = []
+        footlinks_width = 0
         for link, (text, index, show_footlink) in message_links.items():
             if not show_footlink:
                 continue
 
-            footlinks.extend([
+            styled_footlink = [
                 ('msg_link_index', '{}:'.format(index)),
-                ' ',
+                (None, ' '),
                 ('msg_link', link),
-                '\n',
-            ])
+            ]
+            footlinks_width = max(
+                footlinks_width,
+                sum([len(text) for style, text in styled_footlink])
+            )
+            footlinks.extend([*styled_footlink, '\n'])
 
         if not footlinks:
-            return None
+            return None, 0
 
         footlinks[-1] = footlinks[-1][:-1]  # Remove the last newline.
-        return urwid.Padding(urwid.Text(footlinks, wrap='ellipsis'),
-                             align='left', left=8, width=('relative', 100),
-                             min_width=10, right=2)
+
+        text_widget = urwid.Text(footlinks, wrap=wrap)
+        if padded:
+            return urwid.Padding(text_widget, align='left', left=8,
+                                 width=('relative', 100), min_width=10,
+                                 right=2), footlinks_width
+        else:
+            return text_widget, footlinks_width
 
     @classmethod
     def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
@@ -1080,9 +1096,11 @@ class MessageBox(urwid.Pile):
         reactions = self.reactions_view(self.message['reactions'])
 
         # Footlinks.
-        footlinks = self.footlinks_view(
+        footlinks, _ = self.footlinks_view(
             self.message_links,
             footlinks_enabled=self.model.controller.footlinks_enabled,
+            padded=True,
+            wrap='ellipsis',
         )
 
         # Build parts together and return
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -735,12 +735,17 @@ class MessageBox(urwid.Pile):
                              align='left', left=8, width=('relative', 100),
                              min_width=10, right=2)
 
-    def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
+    @classmethod
+    def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
+                    **state: Any
+                    ) -> Tuple[List[Any],
+                               'OrderedDict[str, Tuple[str, int, bool]]',
+                               List[Tuple[str, str]]]:
         # Ensure a string is provided, in case the soup finds none
         # This could occur if eg. an image is removed or not shown
         markup = ['']  # type: List[Union[str, Tuple[Optional[str], Any]]]
         if soup is None:  # This is not iterable, so return promptly
-            return markup
+            return markup, metadata['message_links'], metadata['time_mentions']
         unrendered_tags = {  # In pairs of 'tag_name': 'text'
             # TODO: Some of these could be implemented
             'br': '',  # No indicator of absence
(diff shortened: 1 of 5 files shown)


sumanthvrao (commented):
@preetmishra Left some thoughts on the first commit here.

preetmishra:
@sumanthvrao Thanks for the review. I have amended the tests and made the suggested amendments in the latest update.

preetmishra:
Updated to resolve conflicts.

neiljp (commented):
@preetmishra This appears to work pretty well :+1:

I'm not sure how busy you are at the moment to do any final work on this; my inline points should be fairly minor though we did just migrate the popup tests which has led to the conflict in `test_ui_tools.py`.

Is the first commit just a refactor? (if large - I'm not sure how easy it'd be to split it?)

Maybe originally I thought this was just an exploratory PR and put off reviewing it, and perhaps wondered if you planned to move the adjusted methods into a separate file for processing rendered content? It is effective as it stands, and we can always make the latter migration later - it should be much easier with the decoupling changes you'll have made here :+1:

zulipterminal/ui_tools/boxes.py
```diff
@@ -619,18 +619,23 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
 
     # Use quotes as a workaround for OrderedDict typing issue.
     # See https://github.com/python/mypy/issues/6904.
+    @staticmethod
     def footlinks_view(
-        self, message_links: 'OrderedDict[str, Tuple[str, int, bool]]',
-    ) -> Any:
+        message_links: 'OrderedDict[str, Tuple[str, int, bool]]',
+        footlinks_enabled: bool, padded: bool, wrap: str,
+    ) -> Tuple[Any, int]:
```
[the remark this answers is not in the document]
preetmishra: Yes, it seems clearer. I have separated the last commit into two commits as suggested. Thanks!

preetmishra:
@neiljp Thanks for the review. I have resolved the conflict and made the suggested amendments.

Yes, the first commit is a refactor (a bit tricky to separate further).

While our ultimate intent was to move the rendering methods to a separate file, we decided to decouple the methods first and do the migrations later if needed.

preetmishra:
(Travis is having some trouble building the script for OS X. I will try restarting the job again in a while.)

preetmishra:
(Fixed.)

zulipterminal/ui_tools/boxes.py
```diff
@@ -672,30 +672,42 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
-                             align='left', left=8, width=('relative', 100),
-                             min_width=10, right=2)
+
+        text_widget = urwid.Text(footlinks, wrap=wrap)
+        if padded:
+            return urwid.Padding(text_widget, align='left', left=8,
+                                 width=('relative', 100), min_width=10,
+                                 right=2), footlinks_width
+        else:
+            return text_widget, footlinks_width
```
neiljp, line 710: This is the only part that depends on `wrap` or `padded`? That makes it feel separate to the function, ie. the conditional belongs outside of the calling points. The padding is clearly just a wrapper, that is used in one place, and the use of the wrap parameter is less clear - it depends if we're set on returning a widget vs `footlinks`.

zulipterminal/ui_tools/boxes.py
```diff
@@ -672,30 +672,42 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
     def footlinks_view(
         message_links: 'OrderedDict[str, Tuple[str, int, bool]]',
         footlinks_enabled: bool,
-    ) -> Any:
+        padded: bool,
+        wrap: str,
+    ) -> Tuple[Any, int]:
```
neiljp, line 677: Given the extra return value, a docstring may be helpful.

zulipterminal/ui_tools/boxes.py
```diff
@@ -695,12 +695,19 @@ def footlinks_view(
                              align='left', left=8, width=('relative', 100),
                              min_width=10, right=2)
 
-    def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
+    @classmethod
+    def soup2markup(cls, soup: Any, server_url: str, metadata: Dict[str, Any],
+                    message_links: 'OrderedDict[str, Tuple[str, int, bool]]',
```
neiljp, line 714: Is this quoting since `OrderedDict` doesn't have a `typing` version?

neiljp (commented):
@preetmishra Thanks for keeping this updated :+1: 

I continue to like the move of the htmlsoup->urwidmarkup machinery into a state where we could extract it, though looking at this again I'm a little concerned that we've gone very verbose in the recursive calls with the extra parameters. Did we discuss other ways of achieving this? Can we use `metadata`? Along similar lines, perhaps we could have a "constant" dict (eg. metadata) and keep dynamic data in the `state` dict?

Most of my feedback otherwise is minor, I'd just like to keep the recursive code as clean as possible, if we can :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -695,12 +695,19 @@ def footlinks_view(
                              align='left', left=8, width=('relative', 100),
                              min_width=10, right=2)
 
-    def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
+    @classmethod
+    def soup2markup(cls, soup: Any, server_url: str, metadata: Dict[str, Any],
```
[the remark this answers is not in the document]
preetmishra: Yes, `metadata` is mutable but it is being used to track _immutable_ objects between recursive calls. I changed the commit description on purpose.

zulipterminal/ui_tools/boxes.py
```diff
@@ -744,7 +750,8 @@ def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
                     markup.append(unrendered_template.format(text))
             elif element.name in ('p', 'del'):
                 # PARAGRAPH, STRIKE-THROUGH
-                markup.extend(self.soup2markup(element))
+                markup.extend(cls.soup2markup(element, server_url, metadata,
+                                              message_links, time_mentions)[0])
```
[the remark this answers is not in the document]
preetmishra: This has been fixed with pushing other entities to `metadata`.

zulipterminal/ui_tools/views.py
```diff
@@ -1095,12 +1098,23 @@ def __init__(self, controller: Any, stream_id: int) -> None:
         urwid.connect_signal(pinned_setting, 'change',
                              self.toggle_pinned_status)
 
+        self.markup_desc, *_ = MessageBox.transform_content(
+            rendered_desc,
+            self.controller.model.server_url,
+        )
+        desc = urwid.Text(self.markup_desc)
```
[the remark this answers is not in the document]
preetmishra: Thanks for pointing it out. Fixed!

zulipterminal/ui_tools/boxes.py
preetmishra: Yes, mypy doesn't work well with `OrderedDict`. We have a comment for it in `footlinks_view` as well.

zulipterminal/ui_tools/boxes.py
preetmishra: This does feel out of place. We could return only the markup from `footlinks_view` (+ rename it to `markup_footlinks`) and then add wrappers as needed in the caller functions. Thoughts?

preetmishra:
@neiljp Thanks for the review and pointers. I wasn't inclined towards verbose recursive calls as well. I have moved the data that I want to track amidst the recursive calls into `metadata` to reduce the verbosity. I have also made other amendments as per your in-line comments.

preetmishra:
Updated to resolve conflicts.

neiljp:
@preetmishra I just rebased this to take into account recent changes including the footlinks threshold work and manually pushed this to main ending at 1be52c508c3e030a1293a03648490e57d1e4f992 :tada:

All seems to work fine - thanks for your work on this and the rebasing!

PR #750, opened 2020-07-29
Delegate link text for raw links in a better way

### LINKED ISSUE #742: Delegate link text for raw links in a better way
Currently, we use the last segment of a link for this purpose. It works well for links that have an image or a file. For instance, `https://foo.com/image.png` gets rendered as `image.png`. However, it does have a few caveats.

A route, suggested by @sumanthvrao, that we could take is the following:
* If a link is an internal link, show its relevant part.
* If a link is pointing to an image or a file, show its filename/ending.
* If a link is a general link, show its domain name.

preetmishra:
Currently, we use the last segment of a link for this purpose. It works well for links that have an image or a file. For instance,
https://foo.com/image.png gets rendered as image.png, but it does have a few caveats.

This shows the ending for a filename, the domain name for a general link and the relative URL for an internal link.

Test amended.

Fixes #742.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -580,11 +580,19 @@ class MessageBox(urwid.Pile):
                     for data in [link, text]
                 ]   # Split on '://' is for cases where text == link.
                 if link_without_scheme == text_without_scheme:
-                    segment = text.split('/')[-1]
-                    # Replace text with its last segment if the segment has
-                    # something significant than simply the 'domain name'.
-                    if segment != text_without_scheme:
-                        text = segment
+                    last_segment = text.split('/')[-1]
+                    if '.' in last_segment:
+                        new_text = last_segment  # Filename.
+                    elif text.startswith(self.model.server_url):
+                        # Relative URL.
+                        new_text = text.split(self.model.server_url)[-1]
+                    else:
+                        new_text = (
+                            parsed_link.netloc if parsed_link.netloc
+                            else text.split('/')[0]
+                        )  # Domain name.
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -580,11 +580,21 @@ def soup2markup(self, soup: Any, **state: Any) -> List[Any]:
...
+                    elif text.startswith(self.model.server_url):
+                        # Relative URL.
+                        new_text = text.split(self.model.server_url)[-1]
+                    else:
+                        new_text = (
+                            parsed_link.netloc if parsed_link.netloc
+                            else text.split('/')[0]
+                        )  # Domain name.
+                    # Replace text with new_text if new_text is different from
+                    # text_without_scheme.
```
preetmishra, line 595: This comment feels redundant now. Should I remove it?

sumanthvrao: Agreed. The actual code is pretty simple to follow here.

sumanthvrao:
@preetmishra Thanks for implementing this idea. I tried this out manually and this feels like is a nice improvement especially when encountering complicated links. PR looks good as well 👍

preetmishra:
@sumanthvrao Thanks for the review! I have removed the comment in this update.

neiljp:
@preetmishra Thanks for refining the handling here :+1: I am a little torn by the last-segment vs domain choice, but I do think if we're not going to show the full link (the only other alternative?) then a domain name is definitely better than having what is sometimes a cryptic last part of a URL.

@sumanthvrao Thanks for the review!

Merging this now :tada:

PR #751, opened 2020-07-30
Upgrade core dependencies

neiljp:
This sets dependencies as generally more flexible:
* 2.1.x (x>=1) for urwid, to allow bugfixes
* '>=' current latest versions for zulip, beautifulsoup4, urwid-readline and lxml

Feedback welcome!

Note that the urwid upgrade fixes 256-color support for at least `mosh` and `urxvt`, is likely to do so for `Terminal.app` (macos) and may for others too. If anyone tests those and has feedback for this in particular, we can update the `FAQ` to improve those notes on terminal emulator support :+1:

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,7 +6,7 @@ name = "pypi"
 [packages]
 urwid = "==2.1.0"
 zulip = ">=0.7.0"
-urwid-readline = "==0.11"
+urwid-readline = ">=0.11"
 beautifulsoup4 = ">=4.9.0"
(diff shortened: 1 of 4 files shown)


neiljp:
Latest update includes FAQ changes to indicate urwid upgrade means colors should work properly on mosh and [u]rxvt, and an extra color section which mentions why colors can look strange in emulators with custom colors, and that the colors were improved following the urwid upgrade for those emulators.

PR #757, opened 2020-07-31
model/boxes: Allow users to post stream messages without a topic.

### LINKED ISSUE #754: Deal with no topic messages
We don't allow sending of messages with an empty topic, however the web-app allows this by adding the messages to a topic named `(no topic)`. We should simulate something similar for ZT and care must be taken that this is consistent with what the web-app does.

kaustubh-nair:
This follows the convention used by other Zulip clients to
automatically post a message without a topic, or a topic with only
whitespaces, to a topic named '(no topic)'.

Tests amended.

Fixes #754.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -1,3 +1,4 @@
+import re
 from collections import OrderedDict, defaultdict
 from datetime import date, datetime
 from sys import platform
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -197,25 +197,25 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                 self.view.set_footer_text()
```
neiljp, line 198: Re first commit: it would be useful to describe more of the reason for the refactor in the title and/or summary. That's made clearer in the second commit but just reading this commit it's less clear.

zulipterminal/ui_tools/boxes.py
```diff
@@ -198,16 +199,21 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
 
         if is_command_key('SEND_MESSAGE', key):
             if not self.to_write_box:
+                if re.search(r'^\s*$', self.title_write_box.edit_text):
```
neiljp, line 202: Perhaps we should use `re.fullmatch` here to simplify?

kaustubh-nair:
@neiljp Thanks for the review! This is ready

neiljp:
@kaustubh-nair Your first commit claims to have added a test, but there is only a test in the second commit. Did you mean to have a simple test in that first commit and then extend it in the second?

kaustubh-nair:
@neiljp That's what I had planned before, but the logic in the first commit is too simple to require a test, so I thought I'd just add one for topic matching in the second one. It would just involve checking if the model methods are called with mocked parameters.

neiljp:
@kaustubh-nair Thanks - I just wanted clarification on the test before merging, though I also combined some test lines in the mocking manually. In any case, another good merge! :tada:

PR #758, opened 2020-07-31
Move unread_count from StreamButton updates to Model.

### LINKED ISSUE #722: Move stream muting logic for unread counts from to Model from StreamButton
It would be cleaner to remove from the UI and have model methods for this.

kaustubh-nair:
Currently, logic for updating unread counts after muting/unmuting
streams is present in StreamButton. This commit moves this logic to the
Model.

unread_counts['streams'] has been changed to defaultdict(int) so that it
returns 0 when a stream id is not present.

Tests amended.

Fixes #722.

### DIFF
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -168,20 +168,13 @@ class StreamButton(TopButton):
 
     def mark_muted(self) -> None:
         self.update_widget('M')
-        self.model.unread_counts['all_msg'] -= self.count
         self.view.home_button.update_count(
             self.model.unread_counts['all_msg'])
 
-    def mark_unmuted(self) -> None:
-        if self.stream_id in self.model.unread_counts['streams']:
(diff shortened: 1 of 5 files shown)


zulipterminal/ui_tools/buttons.py
```diff
@@ -168,20 +168,13 @@ def __init__(self, properties: List[Any],
...
-            self.model.unread_counts['all_msg'] += self.count
-            self.view.home_button.update_count(
-                self.model.unread_counts['all_msg'])
-        else:
-            # All messages in this stream are read.
-            self.update_count(0)
+    def mark_unmuted(self, unread_count: int) -> None:
+        self.update_count(unread_count)
+        self.view.home_button.update_count(
+            self.model.unread_counts['all_msg'])
```
kaustubh-nair, line 177: This is now simple enough to not require a test

neiljp:
@kaustubh-nair Thanks - this looks good for keeping the code adjacent to the related data :tada:

PR #759, opened 2020-08-01
Fetch topics on demand to prevent heavy load on /topics

### LINKED ISSUE #493: Fix heavy load on the /topics endpoint when the terminal app is started up
See https://github.com/zulip/zulip/issues/13726 for some background on the performance impact of the `GET /topics` query.

The current behavior of the terminal app is to fetch the full topic history of every subscribed stream on startup using the `GET /topics` endpoint.  This is fairly expensive, and not in line with how the webapp works (where we only fetch full/old topic history via "more topics").  I'm not sure what the right solution is (I hesitate a bit to recommend that the mobile app copy the webapp's behavior of fetching the most recent 1K+ messages of overall history and use the topics from those), but we should figure out a way to make terminal app startup better.

preetmishra:
This is the v1 that we discussed in [zulip-terminal > Topics on demand](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Topics.20on.20demand/near/967222).

Now, topics are only fetched when they required for the `topics_view()` (for the topic toggle) or for the `_update_topic_index()`.

Tests amended with a few spelling corrections.

Fixes #493.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -124,8 +124,6 @@ class Model:
 
         self.unread_counts = classify_unread_counts(self)
 
-        self.fetch_all_topics(workers=5)
-
         self.new_user_input = True
         self._start_presence_updates()
 
@@ -392,36 +390,21 @@ class Model:
                 return response['msg']
         return ""
 
+    def topics_in_stream(self, stream_id: int) -> List[str]:
+        """
+        Returns a list of topic names for stream_id from the index.
+        """
+        if not self.index['topics'][stream_id]:
+            self._fetch_topics_in_streams([stream_id])
+
+        return self.index['topics'][stream_id]
+
     @staticmethod
     def exception_safe_result(future: 'Future[str]') -> str:
         try:
             return future.result()
         except zulip.ZulipError as e:
             return str(e)
-
-    def fetch_all_topics(self, workers: int) -> None:
-        """
-        Distribute stream ids across threads in order to fetch
-        topics concurrently.
-        """
-        with ThreadPoolExecutor(max_workers=workers) as executor:
-            list_of_streams = list(self.stream_dict.keys())
-            thread_objects = {
-                i: executor.submit(self._fetch_topics_in_streams,
-                                   list_of_streams[i::workers])
-                for i in range(workers)
-            }  # type: Dict[int, Future[str]]
-            wait(thread_objects.values())
-
-        results = {
-            str(name): self.exception_safe_result(thread_object)
-            for name, thread_object in thread_objects.items()
-        }  # type: Dict[str, str]
-        if any(results.values()):
-            failures = ['fetch_topics[{}]'.format(name)
-                        for name, result in results.items()
-                        if result]
-            raise ServerConnectionFailure(", ".join(failures))
 
     def is_muted_stream(self, stream_id: int) -> bool:
         return stream_id in self.muted_streams
@@ -853,7 +836,7 @@ class Model:
         Update topic order in index based on incoming message.
         Helper method called by _handle_message_event
         """
-        topic_index = self.index['topics'][stream_id]
+        topic_index = self.topics_in_stream(stream_id)
         for topic_iterator, topic in enumerate(topic_index):
             if topic == topic_name:
                 topic_index.insert(0, topic_index.pop(topic_iterator))
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -376,7 +376,7 @@ class Model:
         display_error_if_present(response, self.controller)
         return response['msg']
 
-    def get_topics_in_stream(self, stream_list: Iterable[int]) -> str:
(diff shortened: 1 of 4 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -767,6 +767,11 @@ def streams_view(self) -> Any:
 
     def topics_view(self, stream_button: Any) -> Any:
         stream_id = stream_button.stream_id
+
+        # Fetch topics and index them if they have not been fetched yet.
+        if not self.model.index['topics'][stream_id]:
+            self.model.get_topics_in_stream([stream_id])
```
neiljp, line 773: This looks very similar to the code added in the model, and this seems like a good opportunity to limit direct access to the index via a model method. Here we access the updated data below in any case.

zulipterminal/model.py
```diff
@@ -853,6 +827,11 @@ def _update_topic_index(self, stream_id: int, topic_name: str) -> None:
         Update topic order in index based on incoming message.
         Helper method called by _handle_message_event
         """
+        # Fetch topics before updating the index if they have not been fetched
+        # yet via the topic toggle.
+        if not self.index['topics'][stream_id]:
+            self.get_topics_in_stream([stream_id])
```
neiljp, line 833: This is not specific to your change here, but reading this code reminds me that `get_` is unclear just from the name - this is actually an API call, whereas `get_` can express any case where you're gathering data through an object, from a resource or otherwise (particularly if you come from languages where "getters" and "setters" are standard). In this context I might expect any external access of the topic list for a stream to be hidden behind a `get_*` (or no prefix), with any potential API call (`fetch_*`) generally happening behind the scenes.

To clarify, this is really another issue, since we do this in a few cases.

preetmishra:
@neiljp Thanks for the review. I have updated the PR to address your suggestions.

zulipterminal/model.py
preetmishra: Agreed. It would make it easier to depict what the method actually does.

zulipterminal/model.py
```diff
@@ -853,6 +835,10 @@ def _update_topic_index(self, stream_id: int, topic_name: str) -> None:
         Update topic order in index based on incoming message.
         Helper method called by _handle_message_event
         """
+        # Fetch and index topics before updating the index if they have not
+        # been fetched yet via the topic toggle.
+        self.set_initial_topics_index(stream_id)
+
         topic_index = self.index['topics'][stream_id]
```
neiljp, line 842: Using a method in this iteration is a good first step, but note that in both cases the code is setting the initial index - ensuring it exists or is fetched - then looking it up for a particular stream. This is a standard pattern where one object handles caching/synchronizing internally, but the external users of that code don't need to consider how that happens, or even be aware of it - they just want the list of topics! Of course, there can be cases where the external code might want this (eg. observer pattern) but we don't need it here, at least yet.

preetmishra:
@neiljp Thanks for the exposition. :+1: I have updated the commit to include the lookup and limit access to `index`. I have also added another commit to encapsulate `get_topics_in_stream()`.

preetmishra:
@neiljp Thanks for further comments. I have updated the PR based on your suggestion and our discussion in PM.

zulipterminal/model.py
```diff
@@ -841,14 +841,17 @@ def _update_topic_index(self, stream_id: int, topic_name: str) -> None:
...
         Helper method called by _handle_message_event
         """
-        topic_index = self.topics_in_stream(stream_id)
-        for topic_iterator, topic in enumerate(topic_index):
+        topic_list = self.topics_in_stream(stream_id)
+        for topic_iterator, topic in enumerate(topic_list):
             if topic == topic_name:
-                topic_index.insert(0, topic_index.pop(topic_iterator))
+                topic_list.insert(0, topic_list.pop(topic_iterator))
                 return
```
neiljp, line 848: This doesn't update the data now - that's what the code says, but it took a while to figure out why. The UI still works since we update them independently. We should probably comment in the caller that it works this way, and whether we want to determine if if it's important to keep this synchronized, or how.

preetmishra:
@neiljp Thanks for catching the synchronization bit. I have updated the PR to address that and other in-line suggestions.

preetmishra:
Updated with an identity assert for `topics_in_stream()`.

neiljp:
@preetmishra Thanks for this contribution - this closes an important bug from the perspective of the server :+1:

@timabbott I mentioned this briefly on czo (#z-t>topics on demand), but now that this approach is complete, what do you think about a slower hybrid approach where topics are loaded slowly in the background?

timabbott:
I think there's probably a broader API discussion we should have about whether the server should offer a "Recent topics for all streams" API feature in /feature; I think we're going to end up needing something like that for the "Recent topics" feature in the webapp.  @amanagr FYI.

PR #760, opened 2020-08-02
Split API call and traceback logging into different log files

neiljp:
Previously the zulip-terminal-traceback.log file was polluted with urllib3/requests debug messages. This separates these two sources.

In future, we could enable API call logging only in a debug or development mode.

We could potentially do the latter sooner rather than later - thoughts?

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -18,8 +18,21 @@ from zulipterminal.model import ServerConnectionFailure
 from zulipterminal.version import ZT_VERSION
 
 
-LOG_FILENAME = 'zulip-terminal-tracebacks.log'
-logging.basicConfig(filename=LOG_FILENAME, level=logging.DEBUG)
+TRACEBACK_LOG_FILENAME = 'zulip-terminal-tracebacks.log'
+API_CALL_LOG_FILENAME = 'zulip-terminal-API-requests.log'
+
+# Create a logger for this application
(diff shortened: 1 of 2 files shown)


zulipterminal/cli/run.py
```diff
@@ -18,8 +18,20 @@
 from zulipterminal.version import ZT_VERSION
 
 
-LOG_FILENAME = 'zulip-terminal-tracebacks.log'
-logging.basicConfig(filename=LOG_FILENAME, level=logging.DEBUG)
+ZT_LOG_FILENAME = 'zulip-terminal-tracebacks.log'
```
preetmishra, line 21: How about `TRACEBACK_LOG_FILENAME`?

zulipterminal/cli/run.py
```diff
@@ -18,8 +18,20 @@
...
+API_CALL_LOG_FILENAME = 'API-requests.log'
```
preetmishra, line 22: We might want to add `zulip-terminal` prefix here as well given that we may end up putting in it `/tmp` later.

neiljp:
@preetmishra Thanks for the review :+1:

Updated to use the more specific file and variable names, and only show API calls in debug mode.

neiljp:
@preetmishra I have quite some extra files so .gitignore doesn't show a clean tree often, so thanks for the reminder :+1:

PR #761, opened 2020-08-02
bugfix: helper: Check view is rendered before displaying errors.

sumanthvrao:
A quick oneliner which hopefully fixes some errors which are caused trying to display an error in the footer, before the view is loaded.


neiljp:
@sumanthvrao Thanks for this - merging this promptly so that it allows people to use the git version for now! in case they run into this race. As we discussed on the stream, these issues should be resolved post-#667.

PR #763, opened 2020-08-03
README: Remove instructions for unsupported WSL notifications

andersk:
Commit ea8321d66900c06a0b6dc348fa31ff0af4f855c2 removed the corresponding functionality for security reasons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,15 +89,6 @@ commands can be found for other linux systems as well.
 sudo apt-get install libnotify-bin
 ```
 
-#### WSL
-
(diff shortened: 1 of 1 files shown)


neiljp:
@andersk I agree this should be updated if the WSL notifications are not put back very soon, but I think in the meantime we should rather have a WSL section indicating they are not currently supported, or words to that effect? We explicitly mention the other platforms we list supporting, so an absence could be taken to indicate implicit integrated support for notifications on WSL.

neiljp:
Merged #835 in preference to this.

PR #768, opened 2020-08-09
model: Ignore peer subscription events for unsubscribed streams.

kaustubh-nair:
Since Model.stream_dict does not contain data for unsubscribed streams,
trying to handle this event raises exceptions.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1566,6 +1566,23 @@ class TestModel:
         new_subscribers = model.stream_dict[event['stream_id']]['subscribers']
         assert new_subscribers == expected_subscribers
 
+    @pytest.mark.parametrize('event', [
+        ({'type': 'subscription', 'op': 'peer_add',
+          'stream_id': 462, 'user_id': 12}),
+        ({'type': 'subscription', 'op': 'peer_remove',
+          'stream_id': 462, 'user_id': 12}),
+    ], ids=[
+        'peer_subscribed_to_stream_that_user_is_unsubscribed_to',
+        'peer_unsubscribed_from_stream_that_user_is_unsubscribed_to',
+    ])
+    def test__handle_subscription_event_subscribers_to_unsubscribed_streams(
+                            self, model, mocker, stream_dict, event):
+        model.stream_dict = deepcopy(stream_dict)
+
+        model._handle_subscription_event(event)
+
+        assert model.stream_dict == stream_dict
+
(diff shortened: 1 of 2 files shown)


zulipterminal/model.py
```diff
@@ -698,10 +698,12 @@ def get_stream_by_id(streams: List[List[Any]], stream_id: int
...
-        elif event['op'] == 'peer_add':
+        elif(event['op'] == 'peer_add'
+             and event['stream_id'] in self.stream_dict):
             subscribers = self.stream_dict[event['stream_id']]['subscribers']
             subscribers.append(event['user_id'])
-        elif event['op'] == 'peer_remove':
+        elif(event['op'] == 'peer_remove'
+             and event['stream_id'] in self.stream_dict):
             subscribers = self.stream_dict[event['stream_id']]['subscribers']
             subscribers.remove(event['user_id'])
```
kaustubh-nair, line 708: I chose not to create a new variable for `event[stream_id]` because I'm not sure if all events will contain `stream_id`

neiljp: Do you mean refactoring it above the if/elif chain?

Could we add spaces in there after the keywords again?

kaustubh-nair: Yes, I meant making a new variable outside the if/elif block.

Isn't there a rule available for the spaces? I couldn't find any

kaustubh-nair:
@neiljp Thanks for the review PR updated.

tests/model/test_model.py
```diff
@@ -1566,6 +1566,23 @@ def test__handle_subscription_event_subscribers(self, model, mocker,
...
+        'peer_unsubscribed_from_stream_that_user_is_unsubscribed_to',
+    ])
+    def test__handle_subscription_event_subscribers_to_unsubscribed_streams(
+                            self, model, mocker, stream_dict, event):
+        model.stream_dict = stream_dict
+
+        model._handle_subscription_event(event)
+
+        # No exceptions should be raised - tested without any assertion.
+
```
kaustubh-nair, line 1585: I'm not sure if there's a correct way to test this. The missing check previously raised a `KeyError` because the stream id was missing in `stream_dict`. So I need to test if a `KeyError` is not raised, which can be tested without an assertion?

Regardless, the call to `_handle_subscription_event` does not do anything at all so there isn't anything to be asserted in the first place.

neiljp:
@kaustubh-nair Merged with extra commit text - thanks! :+1:

PR #771, opened 2020-08-12
Show time mentions (with source text) in MsgInfoView

preetmishra:
This adds an instance attribute, `time_mentions`, to the class `MessageBox` to log time mentions (with their source text) and eventually show them in `MsgInfoView`.


neiljp:
@preetmishra Thanks for the quick follow-up :+1: This looks fine - we could have waited for this in the previous PR, but this is a straightforward clean extension :tada:

It's quite useful to see what syntax works and doesn't in the time mentions!

PR #777, opened 2020-08-16
Support development using makefile

neiljp:
This has been in the works since before 0.5.2, so here's a PR for wider review.

The main feature is how it makes setting up, maintaining and resetting a development venv even easier - simply run `make` and activate the venv. If `setup.py` changes, then `make` sets up the environment accordingly on the next run. This latter approach is adapted from a number of examples I've seen. If an install fails or gets too big, then it can just be wiped and made afresh.

I've also added and applied some explicit 'fixing' tools. We can iterate on these, or drop these in the first instance.

If we move to zulint in time, then we could still maintain this approach and swap-out linting/fixing commands.

I've updated some documentation, but there may be other parts I've missed.


README.md
```diff
@@ -277,6 +284,11 @@ Once you have a development environment set up, you might find the following use
 | Run all tests | `pytest` | `pipenv run pytest` |
 | Build test coverage report | `pytest --cov-report html:cov_html --cov=./` | `pipenv run pytest --cov-report html:cov_html --cov=./` |
 
+If using make with pip, you should also be able to use the following:
+* `make` - ensure the development environment is up to date with specified dependencies
+* `make lint` - equivalent to `tools/lint-all`
+* `make test` - equivalent to running `make lint` and pytest
+
```
kaustubh-nair: We should add `make pytest` here too. I run the linters only after all my tests pass which could take a few iterations.

makefile
```diff
@@ -0,0 +1,34 @@
+.PHONY: install-devel test lint pytest check-clean-tree fix force-fix venv
+
+# Customize your venv name by running make as "ZT_VENV=my_venv_name make <command>"
+
+BASEPYTHON?=python3
+ZT_VENV?=zt_venv
```
kaustubh-nair, line 6: Should we add the default directory to `.gitignore`?

neiljp: Thanks - yes, this was on my to-do but I forgot to add it :+1:

README.md
neiljp: I wasn't sure how much to duplicate the table - some things don't run via make, and others are equivalent except for testing venv. Prefixing `pytest` with `make` each time felt like maybe too much. Happy to include it though.

'fix' should likely depend on `venv` too.

preetmishra: I am wary of `make` with `pytest` as well given that writing `make pytest` takes more time and hinders tab completion.

makefile
```diff
@@ -0,0 +1,34 @@
+.PHONY: install-devel test lint pytest check-clean-tree fix force-fix venv
+
+# Customize your venv name by running make as "ZT_VENV=my_venv_name make <command>"
```
preetmishra, line 3: We should mention that the `ZT_VENV=...` would not only be required for the first time but every time it is run.

Additionally, perhaps rename `<command>` to `<zt-command>`?

Also, should we rather include this in the README so that it is easier to find?

tests/model/test_model.py
```diff
@@ -283,15 +283,15 @@ def test_set_narrow_not_already_set(self, model, initial_narrow, narrow,
             }
         }, {0, 1}),
         ([['stream', 'FOO'],
-         ['topic', 'BOO']], {
+          ['topic', 'BOO']], {
```
preetmishra, line 286: The fact that `make fix` made formatting changes in files which were _already_ passing the linting tests is concerning. Do we have a plan to enforce `make fix` promptly?

Otherwise, it could potentially accumulate a bunch of unrelated changes in PRs which build on a `master` commit that didn't run `make fix`.

makefile
```diff
@@ -0,0 +1,34 @@
...
+	@tools/lint-all
+
+pytest: venv
+	@pytest
+
+### VENV SETUP ###
+# Short name for file dependency
+venv: $(ZT_VENV)/bin/activate
+
+# If setup.py is updated or activate script doesn't exist, update virtual env
```
preetmishra, line 44: We should perhaps include the fact that the `makefile` will continue to make a new `venv` even if a _custom_ `venv` exists.

README.md
neiljp: You're meaning for pytest options?

makefile
neiljp: I had specifically left these notes in the makefile for now as they seemed more 'advanced', and may be scope for further extension - they work, but for now it's easier to just have one environment with a standard name.

I'm not sure we gain from `zt-command` - what matters is running the different makefile 'targets', which are more just specific to those in the makefile.

tests/model/test_model.py
neiljp: I believe the difference is that autopep8 fixes many pep8 rules as specified, but also some additional ones as here: https://pypi.org/project/autopep8/#features

We could add this to CI without `--in-place` and with `--diff`, as a linter. Also autoflake. Ideally this would be part of flake8, so we wouldn't need another checker :/

makefile
neiljp: You mean if you don't use a custom environment name for all `make` invocations, it will build the standard one?

neiljp:
Just pushed with the .gitignore.

README.md
preetmishra: No, I meant in general. It is easier to do `pyte + <TAB>` than write `make pytest`.

tests/model/test_model.py
preetmishra: Ah, I see. :/ Nonetheless, we should perhaps enforce this with our CI for consistency.

makefile
preetmishra: Yes, it seems to build the standard one without that.

preetmishra: On second thought, since the custom one is a bit awkward to run and manage, perhaps we should only promote the standard one.

makefile
```diff
@@ -0,0 +1,34 @@
+.PHONY: install-devel test lint pytest check-clean-tree fix force-fix venv
+
+# Customize your venv name by running make as "ZT_VENV=my_venv_name make <command>"
+
+BASEPYTHON?=python3
```
sumanthvrao, line 5: I think the only restriction this poses is being able to choose the exact python version during venv creation. This isn't a limitation on its own, but if we are going for an "advanced" setup option, we should consider this being customizable.

README.md
sumanthvrao: To clarify, we can still run `make pytest` and it will work (because of the way its listed in makefile) . Are we considering here whether to only mention it README or not?

sumanthvrao (commented):
@neiljp As mentioned previously I think this will be very useful to ppl who want to get a quick ZT dev environment setup. Manually the PR works fine and the code changes look good as well.  :+1:

(As I side note, the docker PR I had open would end up being much simpler now as well.)

neiljp:
Following discussion on czo:
* targets renamed: test -> check, pytest -> test
* note added to makefile indicating ZT_VENV and BASEPYTHON are advanced undocumented features
* `make fix` documentation removed from README for now
* `make` documentation simplified to just indicate it's useful for keeping environment up to date and when
* documentation added to indicate when `make check` is useful

Also fixed .phony targets for 'fix', which are moved to the correct commit.

Does this address the outstanding feedback?

kaustubh-nair (approved):
@neiljp This looks good to me!

sumanthvrao:
@neiljp  Thanks for the updating this. Looks good to me!

neiljp:
Thanks for the reviews everyone! :tada:

Issue #773, opened 2020-08-13
Allow a threshold configuration for footlinks


preetmishra:
Currently, the footlinks can either be configured to `enabled` or `disabled` via the `zuliprc`.

However, we could potentially extend it to establish a middle-ground where only a certain number (threshold) of footlinks appear in the `MessageBox`.

Conversation link: https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Configurable.20footlinks.20by.20threshold.

shanukun:
@zulipbot claim

zulipbot:
Welcome to Zulip, @shanukun! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

zulipbot:
Hello @shanukun, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

Abhirup-99:
@zulipbot claim

neiljp:
@Abhirup-99 zulipbot is maybe not working right now, but note that someone else was working on a PR for this previously.

@shanukun Could you comment as to whether you plan to continue with that PR, or if someone else can pick up where you left off?

Abhirup-99:
> @Abhirup-99 zulipbot is maybe not working right now, but note that someone else was working on a PR for this previously.
> 
> @shanukun Could you comment as to whether you plan to continue with that PR, or if someone else can pick up where you left off?

Ok, I would wait for his reply before starting then. Thanks for the reply.

[...]

PR #780, opened 2020-08-19
[WIP] footlinks: Add threshold configuration for footlinks.

shanukun:
This enables user to configure footlinks threshold settings
using footlinks_threshold in zuliprc.

#773 .

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -294,13 +294,12 @@ def main(options: Optional[List[str]]=None) -> None:
         print("   footlinks setting '{}' specified {}."
               .format(*zterm['footlinks']))
         # For binary settings
-        # Specify setting in order True, False
+        # Specify setting in order 1, 0
         valid_settings = {
             'autohide': ['autohide', 'no_autohide'],
             'notify': ['enabled', 'disabled'],
-            'footlinks': ['enabled', 'disabled'],
         }
-        boolean_settings = dict()  # type: Dict[str, bool]
+        int_settings = dict()  # type: Dict[str, int]
         for setting, valid_values in valid_settings.items():
             if zterm[setting][0] not in valid_values:
                 print("Invalid {} setting '{}' was specified {}."
@@ -310,7 +309,12 @@ def main(options: Optional[List[str]]=None) -> None:
                     print("  ", option)
                 print("Specify the {} option in zuliprc file.".format(setting))
                 sys.exit(1)
-            boolean_settings[setting] = (zterm[setting][0] == valid_values[0])
+            int_settings[setting] = int(zterm[setting][0] == valid_values[0])
+
+        # For threshold settings
+        footlinks_threshold = zterm['footlinks_threshold'][0]
+        if footlinks_threshold.isnumeric():
+            int_settings['footlinks_threshold'] = int(footlinks_threshold)
 
         color_depth = int(args.color_depth)
         if color_depth == 1:
@@ -321,7 +325,7 @@ def main(options: Optional[List[str]]=None) -> None:
         Controller(zuliprc_path,
                    theme_data,
                    int(args.color_depth),
-                   **boolean_settings).main()
+                   **int_settings).main()
     except ServerConnectionFailure as e:
         print(in_color('red',
                        "\nError connecting to Zulip server: {}.".format(e)))
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -28,14 +28,15 @@ class Controller:
     the application.
     """
 
-    def __init__(self, config_file: str, theme: ThemeSpec,
-                 color_depth: int,
-                 autohide: bool, notify: bool, footlinks: bool) -> None:
+    def __init__(
+        self, config_file: str, theme: ThemeSpec, color_depth: int,
+        autohide: int, notify: int, footlinks_threshold: int
+    ) -> None:
         self.theme = theme
         self.color_depth = color_depth
         self.autohide = autohide
         self.notify_enabled = notify
-        self.footlinks_enabled = footlinks
+        self.footlinks_threshold = footlinks_threshold
 
         self._editor = None  # type: Optional[Any]
 
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -531,11 +531,15 @@ class MessageBox(urwid.Pile):
         self, message_links: 'OrderedDict[str, Tuple[str, int, bool]]',
     ) -> Any:
         # Return if footlinks are disabled by the user.
-        if not self.model.controller.footlinks_enabled:
+        if self.model.controller.footlinks_threshold < 1:
             return None
 
         footlinks = []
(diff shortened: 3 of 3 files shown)


shanukun:
@preetmishra can you please review this and suggest the changes regarding linting.
- Also refer to this question link: https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Configurable.20footlinks.20by.20threshold.

zulipterminal/core.py
```diff
@@ -28,14 +28,16 @@ class Controller:
...
-                 autohide: bool, notify: bool, footlinks: bool) -> None:
+    def __init__(
+        self, config_file: str, theme: ThemeSpec, color_depth: int,
+        autohide: bool, notify: bool, footlinks: bool, footlinks_threshold: int
+    ) -> None:
         self.theme = theme
         self.color_depth = color_depth
         self.autohide = autohide
         self.notify_enabled = notify
         self.footlinks_enabled = footlinks
```
preetmishra, line 39: We would likely want to only have `footlinks_threshold`.

zulipterminal/ui_tools/boxes.py
```diff
@@ -535,7 +535,13 @@ def footlinks_view(
             return None
 
         footlinks = []
-        for link, (text, index, show_footlink) in message_links.items():
+        for limit, message_link in enumerate(message_links.items()):
+            # print("Value:", self.model.controller.footlinks_threshold)
```
preetmishra, line 539: Let's remove the print comment. :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -535,7 +535,13 @@ def footlinks_view(
...
+            if (limit == self.model.controller.footlinks_threshold):
+                break
+
+            link, (text, index, show_footlink) = message_link
```
preetmishra, line 543: Perhaps iterate over the key, `link`, and unpack its value here?

zulipterminal/ui_tools/boxes.py
```diff
@@ -535,7 +535,13 @@ def footlinks_view(
             return None
 
         footlinks = []
-        for link, (text, index, show_footlink) in message_links.items():
+        for limit, message_link in enumerate(message_links.items()):
```
preetmishra, line 538: `limit` feels like a misnomer. Would `index` be better?

zulipterminal/cli/run.py
```diff
@@ -312,6 +312,12 @@ def main(options: Optional[List[str]]=None) -> None:
                 sys.exit(1)
             boolean_settings[setting] = (zterm[setting][0] == valid_values[0])
 
+        # For threshold settings
+        int_settings = dict()
+        footlinks_threshold = zterm['footlinks_threshold'][0]
+        if footlinks_threshold.isnumeric():
+            int_settings['footlinks_threshold'] = int(footlinks_threshold)
+
```
preetmishra, line 320: See review.

zulipterminal/cli/run.py
```diff
@@ -321,7 +327,8 @@ def main(options: Optional[List[str]]=None) -> None:
         Controller(zuliprc_path,
                    theme_data,
                    int(args.color_depth),
-                   **boolean_settings).main()
+                   **boolean_settings,
+                   **int_settings).main()
```
preetmishra, line 331: See the second point in the review.

preetmishra (commented):
@shanukun Thanks for working on this. :+1: It seems to work well with `footlinks=enabled`. However, we'd likely want to only have `footlinks_threshold`.

The linting error is a [mypy bug](https://github.com/python/mypy/issues/4708). It was triggered due to the double `dict` unpacking in the `Controller` instantiation.

The code in `run.py` doesn't seem to be in-line with how the other existing boolean setting is handled. You might want to read the existing code to explore it further. That said, we could also refactor the existing code if you think we could handle it in a better way.

See also https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Configurable.20footlinks.20by.20threshold/near/993363.

zulipterminal/ui_tools/boxes.py
shanukun: Yeah. Definitely.

zulipterminal/ui_tools/boxes.py
shanukun: We are already using `index` for storing `message_link` index. Using `index` will not affect the code but it will definitely be confusing.
Maybe `pos` or `position`?

zulipterminal/cli/run.py
shanukun: So I should remove `footlinks=enabled` setting and keep only `footlinks_threshold` and also should change all `bool` settings to `int`?

zulipterminal/ui_tools/boxes.py
preetmishra: Could we rather not change this line and use the existing `index` for the check? That would be a lot cleaner and would not require https://github.com/zulip/zulip-terminal/pull/780#discussion_r473230293 either.

zulipterminal/cli/run.py
preetmishra: Yes, that's what I was suggesting - remove `footlinks` and only have `footlinks_threshold`.

I believe the current set of boolean settings are user-friendly with `enabled` and `disabled`. We shouldn't change all the settings to `int`.

preetmishra:
@shanukun I was meaning to put this in the last review but it slipped my mind!

You could include `Fixes #issue_index` in the commit description to let GitHub detect and automatically close the issue as soon as the commit is merged.

You can read more about Zulip's commit guidelines here: https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages.

zulipterminal/ui_tools/boxes.py
shanukun: Yeah, that sounds about right.

zulipterminal/cli/run.py
shanukun: - I meant storing them in `Dict()` as `int`.

preetmishra:
@shanukun It appears that we still do not have a way to replicate the 'enabled' mode with the new threshold setting. Let's continue the discussion in [#zulip-terminal>Configurable footlinks by threshold](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Configurable.20footlinks.20by.20threshold/near/993368) before moving ahead with the pull request.

neiljp:
@shanukun @preetmishra Where do we stand on this PR?

preetmishra:
@neiljp Primarily, it needs a response from Shanu re the implementation plan that he has as the PR currently doesn't have a way to replicate the original 'enabled' mode, where all the footlinks are shown.

neiljp:
@preetmishra Thanks for the summary :+1: I think @shanukun may be contributing elsewhere, but of course is welcome to pick this back up.

neiljp:
@shanukun Thanks for your work on starting this feature :+1: @Abhirup-99 has extended this in #841 so I'm closing this now.

PR #784, opened 2020-08-21
Invalidate leading spaces (in panel boxes)

neiljp:
This is an alternative to a commit from #700 - instead of allowing (leading) spaces in the text box, leading to "   x" being valid, they're just not accepted, by use of `valid_char()`.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -1,4 +1,5 @@
 import pytest
+from pytest import param
 
 from zulipterminal.config.keys import keys_for_command
 from zulipterminal.ui_tools.boxes import PanelSearchBox, WriteBox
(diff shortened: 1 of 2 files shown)


neiljp:
Code improved and better-documented, with tests added :)

@preetmishra (due to #700) or others welcome to review!

tests/ui_tools/test_boxes.py
```diff
@@ -530,6 +531,22 @@ def test_reset_search_text(self, panel_search_box):
         assert panel_search_box.caption == ""
         assert panel_search_box.edit_text == panel_search_box.search_text
 
+    @pytest.mark.parametrize("search_text, entered_string, expected_result", [
+        param('', 'backspace', False, id="no_text-disallow_urwid_backspace"),
+        param('', 'x', True, id="no_text-allow_entry_of_x"),
+        param('', ' ', False, id="no_text-disallow_entry_of_space"),
+        param('x', ' ', True, id="text-allow_entry_of_space"),
+        # backspace is not *entered*, but treated specially when passed in
```
preetmishra, line 539: Could you clarify what do you mean by 'backspace is not *entered*'? The backspace does seem to get entered into the search box when some text is already present and get validated by `return super().valid_char(ch)`.

preetmishra (commented):
@neiljp Thanks for a cleaner alternative to #700. This seems to work well locally and the tests help in achieving further reliance. The commit looks good too. :+1: 

I have left a minor comment.

tests/ui_tools/test_boxes.py
neiljp: If you experiment with `valid_char` just returning `True`, you'll notice that special urwid keys like `enter` and `backspace` are entered as text too, rather than treated as unicode characters. This is maybe mentioned more in the code than the test?

preetmishra: Ah! Thanks for the clarification.

neiljp:
@preetmishra I just pushed an updated version of this with more text in the code and commit, and extra test cases.

preetmishra (approved):
@neiljp Thanks for the extra comments and extended commit description. This works well and the commit looks good too. :+1:

neiljp:
@preetmishra Thanks for the feedback! :+1: Merging now :tada:

PR #787, opened 2020-08-25
Add an explore mode to run the app without marking messages as read

### LINKED ISSUE #585: Review/Explore/Child mode (don't mark as read)
If I understand the proposal correctly from the summary, this is less of an issue in ZT right now compared to the webapp, but there has been discussion of having a mode in the webapp to 'see messages without marking them as read':
https://chat.zulip.org/#narrow/stream/137-feedback/topic/Trick.20to.20see.20messages.20without.20marking.20them.20as.20read

While I don't think this is a priority, this should be straightforward to implement, but we should discuss it first.

preetmishra:
This introduces an optional argument, `--explore/-e`, to run the app in an explore mode where messages are not marked as read intentionally.

Tests amended and added.

Fixes #585.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,8 @@ If it doesn't find this file, you have two options:
 
    Your personal zuliprc file can be obtained from Zulip servers in your account settings in the web application, which gives you all the permissions you have there. Bot zuliprc files can be downloaded from a similar area for each bot, and will have more limited permissions.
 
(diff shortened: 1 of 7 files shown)


preetmishra:
@neiljp Thanks! I have updated the name from `explore_mode` to `in_explore_mode`.

neiljp:
@preetmishra Merging this now :tada: Thanks for the documentation update too, though I'm considering if later we should make this more prominent.

PR #789, opened 2020-08-29
FAQ: Add entry discussing when messages are marked as read.

neiljp:
This is a slightly more detailed follow-up after #787, discussing how ZT handles marking messages read, since it varies slightly from the webapp.

This was motivated after reading threads of people asking about this in the webapp, and of course we handle it differently - so we should document this!

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -29,6 +29,18 @@ Issues have been reported with the following:
 
 Please let us know if you have feedback on the success or failure in these or any other terminal emulator!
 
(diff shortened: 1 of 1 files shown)


preetmishra (approved):
@neiljp Thanks for working on this. This helps in explicating how we currently handle reading messages thoroughly. 

Definitely a valuable addition. :+1:

neiljp:
@preetmishra Thanks for the feedback :+1: This is a straightforward documentation change so I'm going to merge this now.

PR #790, opened 2020-08-31
model: Support standard field 'op' in update_message_flags events.

### LINKED ISSUE #786: Use `op` rather than `operation` when present in `update_message_flags` events
See https://github.com/zulip/zulip/pull/16168.  We'll need to check `feature_level` for which field to look at, but this should let us eventually delete the server-side compatibility code to send `operation` as well on these events.

neiljp:
All other events use 'op' instead of 'operation'. As of feature level 32 this is also the case in the update_message_flags event.

Tests amended to use new fixture over feature levels None/31/32.

Fixes #786.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1252,9 +1252,25 @@ class TestModel:
         model._handle_reaction_event(response)
         assert len(model.index['messages'][1]['reactions']) == 1
 
-    def test_update_star_status_no_index(self, mocker, model):
+    @pytest.fixture(params=[
+        ('op', 32),  # At server feature level 32, event uses standard field
+        ('operation', 31),
+        ('operation', None),
+    ])
+    def update_message_flags_operation(self, request):
+        return request.param
+
+    def test_update_star_status_no_index(self, mocker, model,
+                                         update_message_flags_operation):
+        operation, model.server_feature_level = update_message_flags_operation
+
         model.index = dict(messages={})  # Not indexed
-        event = dict(messages=[1], flag='starred', all=False, operation='add')
+        event = {
+            'messages': [1],
+            'flag': 'starred',
+            'all': False,
+            operation: 'add'
+        }
         mocker.patch('zulipterminal.model.Model._update_rendered_view')
         set_count = mocker.patch('zulipterminal.model.set_count')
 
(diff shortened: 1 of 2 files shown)


zulipterminal/model.py
```diff
@@ -1012,25 +1012,31 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
         """
         Handle change to message flags (eg. starred, read)
         """
+        if (self.server_feature_level is None
+                or self.server_feature_level < 32):
+            operation_key = 'operation'  # type: Literal['operation', 'op']
+        else:
+            operation_key = 'op'
```
timabbott, line 1019: It might cleaner to have this set `operation = event['operation']` or `operation = event['op']` and remove the `operation_key` variable.

zulipterminal/model.py
```diff
@@ -37,7 +37,7 @@
     'rendered_content': str,
     # update_message_flags:
     'messages': List[int],
-    'operation': str,
+    'operation': str,  # NOTE: deprecated in Zulip 4.0 / ZFL 32
```
preetmishra, line 40: We should perhaps mention that the modern version is `op` in the comment.

tests/model/test_model.py
```diff
@@ -1264,13 +1280,17 @@ def test_update_star_status_no_index(self, mocker, model):
         model._update_rendered_view.assert_not_called()
         set_count.assert_not_called()
 
-    def test_update_star_status_invalid_operation(self, mocker, model):
+    def test_update_star_status_invalid_operation(
+            self, mocker, model, update_message_flags_operation
```
preetmishra, line 1284: [minor] We could add a trailing comma at the end.

zulipterminal/model.py
preetmishra, line 40: We should perhaps mention that the updated version is `op` in this comment.

preetmishra (commented):
@neiljp Thanks for the migration. This seems to work well locally and the commit looks good too. I have left two minor in-line comments. :+1:

neiljp:
@timabbott @preetmishra Thanks for the reviews; good points taken into account in the second version :+1:

preetmishra (approved):
@neiljp Thanks for the amendments. The changes look good to me. :+1:

kaustubh-nair (approved):
Looks good to me!

neiljp:
Thanks for the the review and feedback everyone - merging this now :tada:
(fully prepped for future removal of `operation`)

PR #795, opened 2020-09-06
bugfix: Ensure subscription status checked before muting status.

neiljp:
This causes a traceback for me when I do a specific search, since checking muted topic status assumes the stream is subscribed.

The test is updated but is rather implementation-specific.

I'd like to refactor the code further, so this is a bugfix which would benefit from being addressed better in future.


neiljp:
I hit this bug again recently in master, so merging now :)

PR #798, opened 2020-09-25
Display '(you)' beside logged-in-user in user list.

### LINKED ISSUE #437: Display '(You)' beside logged-in-user in user list.

Astronag:
Fixes #437 
This PR displays '(you)' in white color beside the current user's username in user list.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1128,7 +1128,8 @@ class TestRightColumnView:
                 view=self.view,
                 width=width,
                 color='user_' + self.view.users[0]['status'],
-                count=1
+                count=1,
+                is_current_user=False
             )
         users_view.assert_called_once_with(self.view.controller,
                                            right_col_view.users_btn_list)
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -53,6 +53,7 @@ DEF_base = dict(
     dark_blue='#24a',
     dark_cyan='#088',
     dark_gray='#666',
+    light_gray='#ccc',
     light_red='#f00',
     light_green='#0f0',
     dark_green='#080',
@@ -166,6 +167,8 @@ THEMES = {
          None,             DEF['yellow'],             DEF['black']),
         ('edit_time',      'light blue',              'black',
          None,             DEF['light_blue'],         DEF['black']),
+        ('current_user',   'white',                   'black',
+         None,             DEF['white'],              DEF['black']),
     ],
     'gruvbox_dark': [
         # default colorscheme on 16 colors, gruvbox colorscheme
@@ -242,6 +245,8 @@ THEMES = {
          None,             YELLOW,            BLACK),
         ('edit_time',      'light blue',      'black',
          None,             LIGHTBLUE,         BLACK),
+        ('current_user',   'white',           'black',
+         None,             WHITE,             BLACK),
     ],
     'zt_light': [
         (None,             'black',           'white'),
@@ -280,6 +285,7 @@ THEMES = {
         ('edit_tag',       'white',           'dark gray'),
         ('edit_author',    'dark green',      'white'),
         ('edit_time',      'dark blue',       'white'),
+        ('current_user',   'dark gray',       'white'),
     ],
     'zt_blue': [
         (None,             'black',           'light blue'),
@@ -318,6 +324,7 @@ THEMES = {
         ('edit_tag',       'white',           'dark blue'),
         ('edit_author',    'dark gray',       'light blue'),
         ('edit_time',      'dark blue',       'light blue'),
+        ('current_user',   'light gray',      'light blue'),
     ]
 }  # type: Dict[str, ThemeSpec]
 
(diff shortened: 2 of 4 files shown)


neiljp (commented):
@Astronag Thanks for working on this! We've had a few people look into this - are you building on their work?

This does work fine, but I've left some feedback inline which I think could make this simpler and cleaner. Code-wise the user-specific code doesn't really belong in the general TopButton, particularly with the approach you're taking with replacing the count by specific user-specific text - so I'd suggest customizing only the UserButton behavior.

While your current multiple commits do break up the additions, we'll typically merge commits such that each individually is valid on it's own - ie. it runs fine, passes tests and so on (that's partly since we don't squash commits before merging, and we merge a sequence of commits rather than a merge commit - ie. we use a rebase-based workflow). That's not currently the case - if I load your PR and `git checkout HEAD~3` then this crashes for me, and I've not checked `make check` for each commit (additions/changes to tests adjusting for code would go into the same commit, typically).

By taking a look at the merged commit history you should be able to see the structure we use for commit text - also see `gitlint` in the developer section of the README as a useful tool for giving some hints in this regard. The general zulip git guide is useful too, and regarding the "Fixes" lines, we tend to place them in the commit text body rather than title.

If you'd like to discuss this in a different format then you're welcome to join us on chat.zulip.org in #**zulip-terminal** :)

Astronag:
I looked at other people's approach and got an idea on how it can be done.
I'll make the suggested changes.

Astronag:
@neiljp I've made all the changes that you have suggested. Are there any changes?

neiljp:
@Astronag You've added some commits which look closer to our regular style :+1: However, as you can see here on github, you still have the previous commits (ie. changes) present. You should be able to see this locally with a raw `git log -p`, though I prefer tools like `tig` in full screen, but you can look at them individually on github too.

What we look for is a sequence of intentional logical commits with the code matching the commit text, in an order that flows and doesn't break at each step - this is almost always not the way one initially makes commits, as one might want to go back and adjust them at a later time before they're ready to merge, sometimes dramatically!

Here your new commits actually remove some of the code you added in the earlier commits. If you've not come across it already, this is just the situation that an interactive-rebase works well for.

You might find this part of the zulip git guide helpful (and these ideas in general), which talks broadly about this and ideas for resources:  
https://zulip.readthedocs.io/en/latest/git/using.html#examine-and-tidy-your-commit-history

I've held off on examining the code further at this point, given how your later commits adjust from the baseline of your earlier code rather than before that. Generally reviews are based on the differences within each commit, rather than the state of the code after each one,

neiljp:
@Astronag Let me know if you need any help - or come chat with the Zulip community on chat.zulip.org, including in the #**zulip-terminal** stream :)

Astronag:
> @Astronag Let me know if you need any help - or come chat with the Zulip community on chat.zulip.org, including in the #**zulip-terminal** stream :)

Sure. 
I have changed the previous commits using interactive-rebase. Please check whether they are fine.

neiljp:
@Astronag I can see what you've tried to do here :+1:

However, it's not quite right in terms of content in commits - if you look on GitHub, you'll see you have 3 commits now (not 2), with the second undoing some of the 1st. That's not unheard of in a sequence of commits, but here you seem to have built upon your original commit? I tried squashing the 1st and 2nd commits to make a commit for just the Tuple style change, but that leaves some changes including theme additions in that combined commit which belong in the last one (which is where we actually start using the new style and a few other things).

I'd suggest you try squashing the first two like I did, and figuring out what code doesn't belong in that first combined 'Tuple style' commit. You can then split that commit into two - there are many resources for that, though you might also find something like `git add -e` useful too - and if that code is relevant to the last commit, later squash *those* commits together :)

Astronag:
Okay. Finally, there has to be 2 commits, one for Tuple styling and the other for displaying '(you)' right?

neiljp:
@Astronag That was my suggestion in https://github.com/zulip/zulip-terminal/pull/798#discussion_r495219413

Astronag:
@neiljp Suggested changes have been made to the commits.

neiljp:
@Astronag Thanks for completing the final polish on this :+1: After making only some very minor textual changes I just merged this :tada: (final commit as merged is https://github.com/zulip/zulip-terminal/commit/ad8130a3f7e6d1bb21ce177c663a28c8ef634fa2)

@manojkestur This could be useful for #802 if we do update the styling of the counts or symbol (if we keep the symbol, you could extract it into `symbols.py` as a separate commit)

Astronag:
Thank you. Learnt a lot while working on this.

manojkestur:
Sure, i will look into it.

Issue #797, opened 2020-09-23
Invalid Credentials when logging in from docker


fredrikekre:
I get the following error:
```
Invalid Credentials, Please try again!
```
when I try to log in to https://julialang.zulipchat.com from the alpine docker image. I have verified that the credentials work when loggin in from e.g. the browser. Here is the complete "session":
```
$ git clone https://github.com/zulip/zulip-terminal zulip-terminal
[...]

$ cd zulip-terminal/docker/

$ docker build -t zulip-terminal:alpine -f Dockerfile.alpine .
[...]
Successfully tagged zulip-terminal:alpine

$ mkdir $HOME/.zulip

$ docker run -ti -v $HOME/.zulip:/.zulip zulip-terminal:alpine 
zuliprc file was not found at /.zulip/zuliprc
Please enter your credentials to login into your Zulip organization.

NOTE: The Zulip URL is where you would go in a web browser to log in to Zulip.
It often looks like one of the following:
   your-org.zulipchat.com (Zulip cloud)
   zulip.your-org.com (self-hosted servers)
   chat.zulip.org (the Zulip community server)
Zulip URL: https://julialang.zulipchat.com
Email: ekrefredrik@gmail.com
Password: 

Invalid Credentials, Please try again!
```
I also tried with the buster image, but that just crashes on startup:
```
$ docker run -ti -v $HOME/.zulip:/.zulip zulip-terminal:buster 
Traceback (most recent call last):
  File "/home/zulip/zulip-terminal-venv/bin/zulip-term", line 5, in <module>
    from zulipterminal.cli.run import main
  File "/home/zulip/zulip-terminal-venv/lib/python3.7/site-packages/zulipterminal/cli/run.py", line 10, in <module>
    from urwid import display_common, set_encoding
  File "/home/zulip/zulip-terminal-venv/lib/python3.7/site-packages/urwid/__init__.py", line 26, in <module>
    from urwid.widget import (FLOW, BOX, FIXED, LEFT, RIGHT, CENTER, TOP, MIDDLE,
  File "/home/zulip/zulip-terminal-venv/lib/python3.7/site-packages/urwid/widget.py", line 27, in <module>
    from urwid.util import (MetaSuper, decompose_tagmarkup, calc_width,
  File "/home/zulip/zulip-terminal-venv/lib/python3.7/site-packages/urwid/util.py", line 61, in <module>
    detected_encoding = detect_encoding()
[...]

PR #800, opened 2020-09-25
docker: fix locale setting in Debian Buster Dockerfile.

fredrikekre:
Fixes the locale issue mentioned in https://github.com/zulip/zulip-terminal/issues/797#issue-707661389.

### DIFF
--- a/docker/Dockerfile.buster
+++ b/docker/Dockerfile.buster
@@ -14,6 +14,7 @@ FROM python:3.7-slim-buster
 RUN useradd --user-group --create-home zulip
 COPY --from=builder --chown=zulip:zulip /home/zulip /home/zulip
 USER zulip
+ENV LC_ALL=C
 WORKDIR /home/zulip
 ENTRYPOINT ["/home/zulip/zulip-terminal-venv/bin/zulip-term"]
 CMD ["--config-file", "/.zulip/zuliprc"]


fredrikekre:
Ah, just noted that the same ENV var is set in #571 and looks like the answer to https://github.com/zulip/zulip-terminal/pull/571#discussion_r419318527 is "Yes".

neiljp:
@fredrikekre Thanks for this - this makes the buster build work for you, I assume?

I'm aware you didn't have this issue with the other dockerfile, but do you think we want it in both dockerfiles?

While this may fix the issue for you, my concern is that this would pin the locale too far (ie. we basically lose all localization with `C`?). However, if users can override `ENV` when running the image, or of course adjust the docker file, it seems more important to get this working first. That said does setting `LANG` instead fix this?

@sumanthvrao Given you added this to #571, do you have thoughts on these points?

fredrikekre:
> this makes the buster build work for you, I assume?

Yes.

> do you think we want it in both dockerfiles?

Not sure, given that it works without it seems unnecessary to add it, and from my very limited research, it seems like musl libc handles locales in a somewhat different way, so this fix might not even be applicable.

> That said does setting `LANG` instead fix this?

Doesn't look like it.

---

I have to say that I didn't really do any extensive research about this. I dug into this a bit more, and I found https://github.com/urwid/urwid/commit/fd010f8336407ad1a42522b424d2ea7b7800ebc4 which likely fixes this (the call to [`locale.setlocale`](https://github.com/urwid/urwid/blob/4c739b6be21b0c98324e7b7780d8e712f3ad6db3/urwid/util.py#L59) is what throws, but that error is now [caught](https://github.com/urwid/urwid/blob/4c739b6be21b0c98324e7b7780d8e712f3ad6db3/urwid/util.py#L60). https://github.com/urwid/urwid/commit/fd010f8336407ad1a42522b424d2ea7b7800ebc4 is in [urwid@2.1.2](https://github.com/urwid/urwid/releases/tag/release-2.1.2) (tagged 2 hours ago, what a timing!), so this PR is probably redundant when upgrading urwid.

fredrikekre:
Closing in favor of upgrading urvid.

PR #802, opened 2020-09-27
Displays unread message counts of muted streams and topics.

### LINKED ISSUE #209: Left-sidebar doesn't show muted topic/stream unread count
Instead, it always shows muted streams to be fully read. The screenshot of webapp and ZT can be found in https://chat.zulip.org/#narrow/stream/206-zulip-terminal/subject/left-sidebar.20doesn't.20show.20muted.20topic.2Fstream.20unread.20count/near/674408.

Steps to reproduce:
- I always find them to be the case when installed from the latest master

manojkestur:
This PR displays unread message count of muted streams and topics.
Names of muted streams and topics are displayed in a different color instead of 'M'.

Fixes #209

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -194,7 +194,8 @@ def _set_count_in_view(controller: Any, new_count: int,
             continue
 
         msg_type = message['type']
-        add_to_counts = True
+        text_color = None
+        unread_count_color = 'unread_count'
         if 'mentioned' in message['flags']:
             unread_counts['all_mentions'] += new_count
             all_mentioned.update_count(unread_counts['all_mentions'])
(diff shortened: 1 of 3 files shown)


manojkestur:
@neiljp Can you review this PR

neiljp (commented):
@manojkestur Hey! Thanks for taking an interest in the project!

You're addressing part of quite an old issue, so the scope of what to do is perhaps less clear than it was, making this more challenging than it might be!

Your git commit text matches our current style :+1: However, it would be useful to separate each change to the muted stream/topic text and counts (UI) into separate commits, and definitely keep that work separate from any change to the logic. That will aid in assessing which of these aspects to go forward with, and understanding more clearly what changes you're making in each case.

I've tried the PR:
* This is how I noticed the varying unread counts values; having it in a separate commit it might be clearer what is intended there
* Indicating the muted elements with the text styling is definitely an improvement :+1: Though, I think it might need tuning in some themes since we likely want to have the text faded slightly, but not as unreadable as it seems in eg. the light theme?
* I'm less clear on just showing the counts of muted streams, as it looks like they are contributing towards the unread totals being in the same style. I know this was suggested in the original issue, so we should discuss that - we could style the muted counts differently, or maybe use a different character than `M`? I'm not personally completely convinced it's important to show a count when it's muted though.

I'm going to start a conversation at chat.zulip.org in #**zulip-terminal** regarding this - you're welcome to join in and comment :)

manojkestur:
@neiljp Thanks for the review. I will make the suggested changes.

manojkestur:
I have to do 2 separate commits for dimming muted streams and displaying counts right?

neiljp:
@manojkestur That would be preferable, as we can always then pick just one commit to apply easily, separate from the other. If commits are earlier in a branch then they are less likely to have dependencies upon your other commits and easier to 'cherry pick' separately.

neiljp:
@manojkestur I'm not sure if you're ready for a review of this yet, but from a quick scan I can see you have a few unrelated changes in commits, and you replaced the `M` in one commit by an empty string only to replace that in the next commit (it'd be cleaner to leave it as 'M' and change it in the second).

Re the mute symbol, it looks a little strange here, particularly with a smaller font, though I've not found a better unicode symbol. Did you look at many? Are you available to discuss on chat.zulip.org some time?

manojkestur:
@neiljp I thought only dimming of muted streams and topic without M. Now i will change that to M.
I saw same unicode symbols with different styling but when i pasted it, it's taking basic unicode symbol only.
Yes, i am available to discuss on chat.zulip.org.

manojkestur:
@neiljp I have done the suggested changes.Let me know if any changes required.

manojkestur:
collage

manojkestur:
final display

manojkestur:
@neiljp I have resolved the conflicts and have done the required changes.

neiljp:
@manojkestur Thanks for rebasing and updating the last commit :+1: 

I do like the possibility of this remaining change, though as I indicated when we discussed it in the stream, I also like the idea of not being distracted by whatever the counts are in those muted streams. Having the option of being like the webapp does appear useful in any case, though I've started a discussion in #general to better understand the motivation in the webapp.

This commit does seem to render OK at first glance, but doesn't respond to muting events properly? I also don't understand at first reading why you are adjusting the helper code. The internal data should be being updated properly, but please let us know if you've found a bug in the logic there - though we should keep that different from changing how the UI is styled.

The key methods - at least for streams - are `mark_muted` and `mark_unmuted`. Can we just update those to use the `update_coiunt` method, perhaps with an optional parameter for the muted style, the same as we have for the text color?

manojkestur:
@neiljp I have checked the muted events multiple times it is working fine. When topic is muted and stream is not muted, unread messages were not added to all messages but when we mute the stream where muted topic resides the all messages count was reducing to fix this bug i have done little changes in model.py and helper.
Unread count was not updated when stream was muted so i changed that in helper.py.
I have kept unread_count as parameter i have done the changes.

PR #805, opened 2020-10-05
keymapping: Add new 'a' key for all messages.

### LINKED ISSUE #804: Migrate 'all messages' key to 'a' in UI (retaining current key)
The webapp is likely to move to a default view of recent topics in the future, meaning that all-messages will acquire a dedicated key since it is not the default view (associated with `esc`) any longer.

While ZT does not have an interface for recent-topics, it would be beneficial to at least support the new key for all-messages. Even if the webapp chooses not to take this path, it could be useful to migrate away from using `esc` for so many actions, and improve (partially automatic) documentation of how to narrow to all messages.

This will require updating the keybindings and ensuring that this works for both `esc` and `a`, and checking that the UI renders for `a` only. This key may be hardcoded right now, so a new keybinding entry may be needed (to appear in the correct place in the help menu), and referred to in appropriate places in the code. The README should also be updated accordingly.

The intention is that users migrate to this keybinding in the near term via the advertised key. At a later stage we might remove the `esc` key completely, and we could open an issue for this following resolution of this issue.

adwait-thattey:
A New Keymapping 'ALL_MESSAGES' is added.
This maps to 'a' key.

This replaces the 'GO_BACK' ('ESC') key currently
being used for displaying all messages.

Currently both 'GO_BACK' and 'ALL_MESSAGES' work.

Fixes #804

Signed-off-by: Adwait Thattey <coderdude1999@gmail.com>

Screenshot: 
Screenshot_20201005_103853

### DIFF
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -324,12 +324,12 @@ def is_command_key(command: str, key: str) -> bool:
         raise InvalidCommand(command)
 
 
-def keys_for_command(command: str) -> Set[str]:
+def keys_for_command(command: str) -> List[str]:
     """
     Returns the actual keys for a given mapped command
     """
     try:
-        return set(KEY_BINDINGS[command]['keys'])
+        return sorted(KEY_BINDINGS[command]['keys'])
     except KeyError as exception:
         raise InvalidCommand(command)
 
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -151,6 +151,11 @@ KEY_BINDINGS = OrderedDict([
         'help_text': 'Toggle topics in a stream',
         'key_category': 'stream_list',
     }),
+    ('ALL_MESSAGES', {
+        'keys': {'a', 'esc'},
+        'help_text': 'Narrow to all messages',
+        'key_category': 'navigation',
+    }),
     ('ALL_PM', {
         'keys': {'P'},
         'help_text': 'Narrow to all private messages',
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -94,8 +94,8 @@ class TopButton(urwid.Button):
 
 class HomeButton(TopButton):
     def __init__(self, controller: Any, width: int, count: int=0) -> None:
-        button_text = ("All messages   ["
-                       + keys_for_command("GO_BACK").pop()  # FIXME
+        button_text = ("All messages     ["
+                       + keys_for_command("ALL_MESSAGES").pop(0)
                        + "]")
         super().__init__(controller, button_text,
                          controller.show_all_messages, count=count,
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ source ~/.zshenv
 | Scroll up                                             | <kbd>PgUp</kbd> / <kbd>K</kbd>                |
 | Scroll down                                           | <kbd>PgDn</kbd> / <kbd>J</kbd>                |
 | Go to the last message                                | <kbd>G</kbd> / <kbd>end</kbd>                 |
+| Narrow to all messages                                | <kbd>a</kbd> / <kbd>esc</kbd>                 |
 | Narrow to all private messages                        | <kbd>P</kbd>                                  |
 | Narrow to all starred messages                        | <kbd>f</kbd>                                  |
 | Narrow to messages in which you're mentioned          | <kbd>#</kbd>                                  |
--- a/tests/config/test_keys.py
+++ b/tests/config/test_keys.py
@@ -23,7 +23,7 @@ def invalid_command(request):
 
 
 def test_keys_for_command(valid_command):
-    assert (keys.KEY_BINDINGS[valid_command]['keys']
+    assert (sorted(keys.KEY_BINDINGS[valid_command]['keys'])
             == keys.keys_for_command(valid_command))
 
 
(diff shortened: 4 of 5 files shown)


adwait-thattey:
@neiljp As written above, both 'Esc' and 'a' are working right now and will display all messages. Please let me know if this needs to change.

Also, I could not find if any tests need to be added or updated.

neiljp (commented):
@adwait-thattey Thanks for working on this :) This gets a good way there according to the original issue :+1:

While this does keep the original behavior, this is by hardcoding it as an option, whereas it is more flexible if we can rely upon extending the keybindings, ie. adding to the set of keys for a command.

I should note that there is limited space next to each 'button' to show one shortcut key, otherwise we could show both keys, which is why it's fine for the help and the `[a]` to differ.

The other thing mentioned in the issue is adding this to the README, as we have these hotkeys listed there too.

Regarding tests, this passes ones we currently have. If you do want to investigate improving them with this, you could check if there are any hard-coded 'esc' in the tests as that could indicate how they could be broadened to use your new `ALL_MESSAGES` token.

adwait-thattey:
@neiljp  I made the changes.
There might be one issue. As the keys are a set, in the top-left buttons, both 'a' or 'esc' can appear depending on which gets popped. 
Is this fine?

neiljp:
Just to note here: I responded elsewhere today that we should likely move to a list for the external interface for the keys, for more reproducible behavior generally - this would be a good commit to have before the existing one.

adwait-thattey:
Yes. On it

adwait-thattey:
@neiljp I have added 2 commits. 
The first one changes behaviour to return a sorted list from the keys_for_command method and modifies tests by changing types there too

2nd one adds the 'a' key

zulipterminal/config/keys.py
```diff
@@ -324,12 +324,12 @@ def is_command_key(command: str, key: str) -> bool:
...
 
 
-def keys_for_command(command: str) -> Set[str]:
+def keys_for_command(command: str) -> List[str]:
     """
     Returns the actual keys for a given mapped command
     """
     try:
-        return set(KEY_BINDINGS[command]['keys'])
+        return sorted(set(KEY_BINDINGS[command]['keys']))
```
neiljp, line 337: We can remove the `set` here, since sorted returns a new list.

zulipterminal/ui_tools/buttons.py
```diff
@@ -95,7 +95,7 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
 class HomeButton(TopButton):
     def __init__(self, controller: Any, width: int, count: int=0) -> None:
         button_text = ("All messages   ["
-                       + keys_for_command("GO_BACK").pop()  # FIXME
+                       + keys_for_command("GO_BACK").pop(0)  # FIXME
```
neiljp: `pop` defaults to `pop(-1)` and there's only one element in the set/list right now, so this behavior change should be in the next commit?

zulipterminal/ui_tools/buttons.py
```diff
@@ -94,8 +94,8 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
 
 class HomeButton(TopButton):
     def __init__(self, controller: Any, width: int, count: int=0) -> None:
-        button_text = ("All messages   ["
-                       + keys_for_command("GO_BACK").pop(0)  # FIXME
+        button_text = ("All messages     ["
+                       + keys_for_command("ALL_MESSAGES").pop(0)  # FIXME
```
neiljp, line 98: I checked the git log and we can remove this FIXME, which I think I added to indicate that `GO_BACK` should be replaced with something more specific, just like you're doing :+1:

neiljp (commented):
@adwait-thattey This looks good :+1: I've left some comments inline.

There's one mention of `set` in another test (docstring) which you could update.

My only other concern is that we do use `keys_for_command` in many places (`git grep keys_for_command`) and that something may fall through the gaps.

Once you've updated the commits, I'd suggest checking out gitlint and our style in git log, for matching our existing style, though we could also do this before merging.

adwait-thattey:
@neiljp I made the changes. 
I searched through all the instances of `keys_for_command` but couldn't find any other place that may need change.
While many actions have multiple keys attached to them, only "All_MESSAGES" is used with pop(), needing to change to pop(0). 

pop is used in many other places and I wonder if they will need to be changed to pop(0) too... considering if more than 1 keys are added to those actions later

neiljp:
@adwait-thattey Thanks! I've merged these two commits with minor changes to the commit text :tada:
They're visible at https://github.com/zulip/zulip-terminal/commit/17ee342b59c5c9e6d5ec809cc74d157c431941f7 and the previous commit.

neiljp:
@adwait-thattey As I mentioned, a good follow-up would be to migrate the system to take a standard element (first, or last?) from lists of keys for each command (and document that in the file), and then ensure that the UI treats that in the correct way.

adwait-thattey:
@neiljp Thanks a lot :)
> As I mentioned, a good follow-up would be to migrate the system to take a standard element (first, or last?) from lists of keys for each command (and document that in the file), and then ensure that the UI treats that in the correct way.

Yes, I will try this. Should I open another issue for this?

neiljp:
@adwait-thattey If you plan to work on it, you can just open a PR when you're ready :)

adwait-thattey:
@neiljp I opened a followup PR #806 and made some comments there.
Please check 
Thanks :)

PR #806, opened 2020-10-11
keymapping: replace pop with prime_key method

adwait-thattey:
keys_for_command currently returns a sorted_list
and pop is used to get the key that is to be displayed in the UI

A new method prime_key_for_command is introduced, that always
returns first element from the sorted list

pop is rpelaced everywhere with this new method

tests are also added to assert this behaviour

A followup to PR #805

### DIFF
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -1,11 +1,11 @@
 from collections import OrderedDict
-from typing import List, Set
+from typing import List
 
 from typing_extensions import TypedDict
 
 
 KeyBinding = TypedDict('KeyBinding', {
-    'keys': Set[str],
+    'keys': List[str],
     'help_text': str,
     'excluded_from_random_tips': bool,
     'key_category': str,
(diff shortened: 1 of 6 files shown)


adwait-thattey:
@neiljp 
I have introduced a new method `prime_key_for_command` that always returns the first element from whatever, `keys_for_command` returns (currently sorted list)
Please review the changes 
The commit message is temporary. I will change it later

I also have another idea, what if we convert keys everywhere in key_bindings dict from set to a simple list.
This will allow us to use always the first element, rather than using the first element from the sorted list.
Anyone who adds new keys can decide which key they want to display by just keeping it first in the list rather than always having the key first in alphabetical order to appear first...

zulipterminal/config/keys.py
```diff
@@ -339,6 +343,14 @@ def keys_for_command(command: str) -> List[str]:
         raise InvalidCommand(command)
 
 
+def prime_key_for_command(command: str) -> str:
+    """
+    Returns the primary key for a command,
+    currently defined as the first key
+    """
```
neiljp, line 350: Is this the `prime` or `primary` key? :) Primary may be clearer?

If the docstring then repeats the function description, we can trim it to avoid duplication.
For a user of this function 'first key' has no meaning?
A good alternative might be to say again what it means, and/or for what application? (what is 'primary key', why might it be useful?)

neiljp (commented):
@adwait-thattey I like the approach of using a method here as it avoids encoding the underlying data type - like we do with `pop` right now :+1:

Regarding your point about set vs list, I definitely think we should change the data-structure for this - when specifying the keys for a command in the data-structure, we want to be able to specify the 'prime' key, and I'm not sure we can right now (it's the first in the sorted list?)

The alternative would be to maintain extra entries for each token to indicate if there is a prime key, which I think is too much work to maintain and risks inconsistency with duplication, even if we attempt to minimize that with tests.

You could treat the datastructure change as a separate commit, as long as the behavior doesn't change.

I've not checked the tests, but it's possible that they just haven't been generalized to use all possible keys there, so a direct translation from cases of using`pop` may not be meaningful.

adwait-thattey:
@neiljp  I made the changes in the name and changed the data structure holding the keys.
Also, I thought setting keys as Tuples might be better than Lists as they are immutable?
I will change it back to List if needed

zulipterminal/config/keys.py
```diff
@@ -1,304 +1,304 @@
...
+    'keys': Tuple[str, ...],
     'help_text': str,
     'excluded_from_random_tips': bool,
     'key_category': str,
 }, total=False)
 
 KEY_BINDINGS = OrderedDict([
     ('HELP', {
-        'keys': {'?'},
+        'keys': ('?',),
```
neiljp, line 19: While `mypy` should check for this during tests for development/released versions, if a user customizes this then this becomes a potential issue if they miss a trailing comma - for that reason I think a list would be simpler?

I do agree that immutable would be preferable, but the data types just don't always lend themselves to it, unfortunately.

neiljp (commented):
@adwait-thattey Thanks for updating this :+1:

I had a scan through your commits and the pop cases I was concerned with in the tests should be fine - this is due to the limitation in urwid-readline to only have one key for autocompleting (or in reverse). I think may be an outstanding issue about testing for this.

After using a list instead of tuple (see below) I think this should be good to go :+1:

adwait-thattey:
@neiljp Made the changes
Thanks :)

neiljp:
@adwait-thattey Thanks! I've merged this with minor commit text adjustments to match our style a little more closely :tada:

The commits as merged end here: 4eca34c1e608448d3f34ce3f88ed8e6eaa8b7d49

adwait-thattey:
@neiljp Thanks :)

PR #807, opened 2020-10-12
refactor/enhancement: Added "[default theme]" suffix to the --list-themes command-line option and cleaned up the duplicate code.

sumagnadas:
This PR adds the suffixes `[default theme]` to the listed themes. The first suffix is applied to the theme `zt_dark` which is the default theme for Zulip Terminal. 

This PR also combines the two places with a function which lists themes hence cleaning up the duplicate code present because of the addition of the `--list-themes` command-line option.

A follow up to PR #803.

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -225,6 +225,18 @@ def parse_zuliprc(zuliprc_str: str) -> Dict[str, Any]:
     return settings
 
 
+def list_themes() -> None:
+    available_themes = all_themes()
+    print("The following themes are available:")
+    for theme in available_themes:
+        suffix = ""
+        if theme == "zt_dark":
+            suffix += "[default theme]"
+        print("  ", theme, suffix)
+    print("Specify theme in zuliprc file or override "
+          "using -t/--theme options on command line.")
+
+
 def main(options: Optional[List[str]]=None) -> None:
     """
     Launch Zulip Terminal.
@@ -253,10 +265,7 @@ def main(options: Optional[List[str]]=None) -> None:
         sys.exit(0)
 
     if args.list_themes:
-        available_themes = all_themes()
-        print('Themes available:')
-        for theme in available_themes:
-            print('    {}'.format(theme))
+        list_themes()
         sys.exit(0)
 
     if args.config_file:
@@ -284,11 +293,7 @@ def main(options: Optional[List[str]]=None) -> None:
         if not is_valid_theme:
             print("Invalid theme '{}' was specified {}."
                   .format(*theme_to_use))
-            print("The following themes are available:")
-            for theme in available_themes:
-                print("  ", theme)
-            print("Specify theme in zuliprc file or override "
-                  "using -t/--theme options on command line.")
+            list_themes()
             sys.exit(1)
         if theme_to_use[0] not in available_themes:
             # theme must be an alias, as it is valid


zulipterminal/cli/run.py
```diff
@@ -281,15 +281,23 @@ def main(options: Optional[List[str]]=None) -> None:
...
+                exit_code = 1
             print("The following themes are available:")
             for theme in available_themes:
-                print("  ", theme)
+                suffix = ""
+                if theme == "zt_dark":
+                    suffix += "[default theme]"
+                if theme_to_use[0] == theme:
+                    suffix += "[user preference]"
+                print("  ", theme, suffix)
```
neiljp, line 290: This doesn't get run for just `--list-themes` - there is still a block of code earlier in the file that is triggered earlier, which I see you've fixed in the last commit. I'd expected it in the same commit.

I think users may want to just run `zulip-term --list-themes` without necessarily having/specifying a config file, so I'm not sure we can actually move the code down here, though it would be good to have the same code called - so an ideal place to use a function. That would allow the user-preference to be optional, if we don't know it - if we don't have access to the file.

docs/FAQ.md
```diff
@@ -76,6 +76,23 @@ We think of Zulip Terminal as a keyboard-centric client. Consequently, while fun
 
 If any of the hotkeys don't work for you, feel free to open an issue or discuss it on [#zulip-terminal](https://chat.zulip.org/#narrow/stream/206-zulip-terminal).
 
+## Are there any themes available, other than the default one?
```
neiljp, line 79: This is great :+1: I cherry-picked this commit and merged it :tada: I moved the section further up in the file as it's a higher priority than some others, and also made some other minor textual adjustments.

neiljp (commented):
@sumagnadas Thanks for the follow-up. As below, I've merged one commit with minor adjustments already, but it would be useful to not have a valid config file in order just to list themes.

zulipterminal/cli/run.py
sumagnadas: I have removed this part and will commit it later.

sumagnadas:
I kept the `user_preference` suffix to denote the theme which is either specified on the command-line or the `zuliprc` config file. So its there because of the command-line option specification.

sumagnadas:
I also removed the FAQ commit to squash the initial commit and the "code refactor" commit into one.

zulipterminal/cli/run.py
sumagnadas: What code do you want to put in the new function?

neiljp: I would like `zulip-term --list-themes -c invalid` to still work, which requires some code to remain at the top, as it was before.

However, we also want to run the same type of code in the lower section, so to avoid duplication we can use a function to avoid duplication.

if we have extra information then the lower version can do a slightly different output. This could depend on a parameter to the new function - or depending if the function outputs the information or returns a data-structure, the data could be modified before output in the lower section instead.

zulipterminal/cli/run.py
```diff
@@ -252,18 +267,15 @@ def main(options: Optional[List[str]]=None) -> None:
...
-        sys.exit(0)
-
     if args.config_file:
         zuliprc_path = args.config_file
     else:
         zuliprc_path = '~/zuliprc'
 
+    if args.list_themes and not path.exists(path.expanduser(zuliprc_path)):
+        list_themes(('No theme', 'with no config'), False)
+        sys.exit(0)
```
neiljp, line 277: This code is moved around over the two commits, it should be fine where it was :)

The conditional can remain the same as before - I think we had confusion over the 'invalid' config file, I meant that `--list-themes` should be independent of whether we have a config file or not. For now it's fine to keep all the logic using args.list_themes in this top area.

neiljp (commented):
@sumagnadas Thanks for taking the function approach. This is looking closer to the use of common code for `--list-themes` and the 'bad theme' case now that I had envisaged in an earlier response :+1:

With the current way that the rest of the code is structured in this file - the prompting for server details and so on being inside other functions, if the configuration file isn't found to be valid - then I think we should leave the `has_config` and specified-theme parameters for now:
* `--list-themes` being above, can't have this information without adjusting the way these other functions work
* in the case of an invalid theme, then passing the theme won't achieve anything, as by definition it is invalid!

So, we can still use the new function at the top (just for args.list_themes) and below (just for the invalid theme), but completely separately for now.

zulipterminal/cli/run.py
neiljp: Yes, I think we can leave the user-preference part for now, for the reasons I gave in the review summary.

sumagnadas: I will update the function and squash the change with the previous commit then.

neiljp: That sounds good :+1:

sumagnadas:
I also edited the first commit to not include the `[user preference]` because, in the second commit, it was removed.

sumagnadas:
@neiljp Anything I should change in any of the commits?

neiljp:
@sumagnadas I think it'd work best with the two commits squashed together with a reworded commit text. The overall code looks good in the 'files changed' tab of the PR, and the commit text can reflect that :+1:

My reasoning for suggesting squashing is based on the overall result of this PR in the 'files changed' tab - each current commit doesn't seem to build towards it.

sumagnadas:
See if this works. Sorry if the commit text is not appropriate. I am still trying to get the hang of working with Github repos, commits and the other stuff of open source contribution.

neiljp:
@sumagnadas Thanks for working on this :+1: I've slightly reworded the commit text and merged it otherwise as-is :tada:

If you want to see the differences in the commit title/description and aren't sure how to go about that, I'd recommend learning about `git range-diff`.

I also merged #810 - if you have any questions or other feedback on if that would have helped with writing the commits, we'd be happy to know :) (see mainly the `GitLint` section)

PR #808, opened 2020-10-13
enhancement: Support for setting default color depth in zuliprc files added.

### LINKED ISSUE #601: Support setting default color depth in zuliprc files
Following #563, we can further add support to set the default color depth via zuliprc files.

This touches only a small part of the codebase, though may require adapting if the tutorial work is merged.

sumagnadas:
Fixes #601.


zulipterminal/cli/run.py
```diff
@@ -330,15 +331,19 @@ def main(options: Optional[List[str]]=None) -> None:
                 sys.exit(1)
             boolean_settings[setting] = (zterm[setting][0] == valid_values[0])
 
-        color_depth = int(args.color_depth)
+        if args.color_depth:
+            color_depth = int(args.color_depth)
+        else:
+            color_depth = int(zterm['color-depth'][0])
```
neiljp, line 337: We should likely do some validation here - this crashes if it is not one of the valid values.

neiljp (commented):
@sumagnadas This works broadly but we definitely need to validate the color depth specified in the config file..

Other things that would definitely be good to include:
* Demonstrating this configuration option in the README alongside the others
* Outputting the color depth being used when loading (like we do with theme/autohide/footlinks)

For the last point, you could consider an extra commit to indicate the status of notifications when loading, which we appear to not output either.

zulipterminal/cli/run.py
sumagnadas: WIll do that..

sumagnadas:
Should i squash the last commit and the previous commit together?

zulipterminal/cli/run.py
sumagnadas: And should the program exit if it is not one of those values or just tell the user that "this is not one of the valid values" and start zulip terminal with the default of 256?

neiljp: It's a little unclear, but my suggestion would be that since we validate color depth on the command line (via argparse choices) and reject bad values before running, we should likely do the same before running via a config-file option. If a user has specified a value in the config, then it's likely they have (tried to) do it intentionally, and running anyhow may give the impression the config file value is valid and correct. Feedback much like from argparse with wrong values, indicating good values, should be reasonable.

sumagnadas: So the program should not quit if a bad value is given in the variable `color-depth`?

neiljp: If there is a bad value from the command-line, argparse gives a message (give it a try!)

If there is a bad value from the config-file, we should give a similar message.

sumagnadas:
Yea I saw that. And after shoeing the message, the program quits. So I am
asking "should the program quit in this case also? Or should it tell the
user that he/she has given the wrong value for color_depth so the program
is going to use the default value of 256?"

neiljp:
@sumagnadas Yes, it should quit. For reference, see the behavior if the values of other parameters are incorrect in the config file.

neiljp (commented):
@sumagnadas Thanks for the update - this seems to work better now. With fixed tests and the readme updated, this will be great :+1:

zulipterminal/cli/run.py
```diff
@@ -311,12 +317,15 @@ def main(options: Optional[List[str]]=None) -> None:
               .format(*zterm['autohide']))
         print("   footlinks setting '{}' specified {}."
               .format(*zterm['footlinks']))
+        print("   color depth setting '{}' specified {}."
+              .format(*zterm['color-depth']))
```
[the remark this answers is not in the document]
sumagnadas: Done that. Just have to update the README now.

neiljp:
@sumagnadas Thanks for the update and work on this :+1: I slightly adjusted the commit text and have merged it manually :tada:

PR #811, opened 2020-10-20
Migrate popup tests out of general ui_tools file

neiljp:
We discussed this in the past in #**zulip-terminal>Splitting UI tests** - now seems a good time to make this change.

This moves all popup-related tests, resulting in ~400 lines moved, making the general test file change from ~2.8klines to ~2.4klines - and simplifying imports.

**I'm not sure we have particular open PRs which should be impacted by this, but please let me know if there are specific ones you know are impacted.**

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -7,7 +7,7 @@ import pytz
 from bs4 import BeautifulSoup
 from urwid import Columns, Divider, Padding, Text
 
-from zulipterminal.config.keys import is_command_key, keys_for_command
+from zulipterminal.config.keys import keys_for_command
 from zulipterminal.config.symbols import (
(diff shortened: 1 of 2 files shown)


tests/ui/test_ui_tools.py
```diff
@@ -17,12 +17,11 @@
     StreamButton, TopButton, TopicButton, UserButton,
 )
 from zulipterminal.ui_tools.views import (
-    AboutView, EditModeView, HelpView, LeftColumnView, MessageView,
-    MiddleColumnView, ModListWalker, MsgInfoView, PopUpConfirmationView,
-    PopUpView, RightColumnView, StreamInfoView, StreamsView,
-    StreamsViewDivider, TopicsView, UsersView,
+    EditModeView, HelpView, LeftColumnView, MessageView, MiddleColumnView,
```
sumanthvrao: minor: I know this makes the file lengthier, but we may want to move towards having one import per line. That way we reduce the diffs in rearranging to adjust for line length limit.

PR #814, opened 2020-10-24
Adjust subscription op=peer_* events for ZFL35+

neiljp:
Previously these events contained single stream and user ids, but now instead contain lists of each to reduce the numbers of events required for cases where users join or leave large numbers of streams.

This is already merged into master as of https://github.com/zulip/zulip/commit/7ff3859136638defff01abafcaa8db51a2e2eb54, so pending zulip/zulip#16628, this should limit breakage for those using czo.


neiljp:
Merging this now that travis is more reliable again :tada:

PR #817, opened 2020-10-30
Use varying symbols for user status.

neiljp:
This moves from having a single small bullet to a style similar to the
webapp, with circles of varying degrees of fullness:
* filled (active)
* bottom-filled (idle)
* empty (offline)
* small (inactive)

Tests updated.

NOTE: I'll move the unicode to symbols.py before this is merged.

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -10,7 +10,8 @@ from zulipterminal.config.keys import (
     HELP_CATEGORIES, KEY_BINDINGS, is_command_key, keys_for_command,
 )
 from zulipterminal.config.symbols import (
-    CHECK_MARK, LIST_TITLE_BAR_LINE, PINNED_STREAMS_DIVIDER,
+    CHECK_MARK, LIST_TITLE_BAR_LINE, PINNED_STREAMS_DIVIDER, STATUS_ACTIVE,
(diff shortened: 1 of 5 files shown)


neiljp:
Just pushed an updated version with the unicode in symbols.py.

The change is broadly as follows (this is with a search on czo, to show all the statuses):
user_icons_old - OLD

user_icons_new - NEW

zulipterminal/ui_tools/views.py
```diff
@@ -651,10 +652,19 @@ def users_view(self, users: Any=None) -> Any:
...
+        state_icon = {
+            'active': STATUS_ACTIVE,
+            'idle': STATUS_IDLE,
+            'offline': STATUS_OFFLINE,
+            'inactive': STATUS_INACTIVE,
+        }
+
         users_btn_list = list()
         for user in users:
+            status = user['status']
```
sumanthvrao, line 665: You seem to have done a small refactor here, but I guess it is small enough to integrate with the other change?

sumanthvrao:
These symbol changes look nice! :+1:

zulipterminal/ui_tools/views.py
neiljp: This is indeed small, and within a function it should be safe enough :+1:

neiljp:
Thanks for the feedback! :+1: Let's get this merged :)

PR #820, opened 2020-11-03
Doc updates: multiple servers (FAQ), pytest tips (README:dev)

neiljp:


neiljp:
@zee-bit wrote in #821:
> @neiljp This is a very nice addition. Although the list of commands is more than enough for testing ZT code-changes, it would be better if we could add the link to official `pytest` docs there too.

Thanks for reading this - and it's a very good point, that we don't link to the pytest docs. I've added a quick sentence at the end of the new section.

neiljp:
Merging this now since no other similar updates are currently incoming.

PR #822, opened 2020-11-05
search_box: Improved search box query handler.

aryanshridhar:
Earlier , The query typed in search box is erased when the user navigates to a new stream .
As suggested by @neiljp <a href = 'https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Sticky.20message.20search.20.23T822/near/1058839'>here</a> , Added a extra keybinding of <code>meta /</code>( <code>alt /</code>) which retains the search query and carries it forward whenever the user changes narrows .

It includes : 
<ul>
<li>If the user wants to perform a <code>non-sticky</code> search across narrows , the usual keybinding of <code>/</code> would make the job done .</li>
<li>But , if user prefers for a <code>sticky-search</code> , activating the searchbox with <code>meta /</code> is useful.</li>
</ul>

<strong>Non-Sticky Search : </strong>(<code>/</code>)

<strong>Sticky Search :</strong>(<code>meta /</code>)

<br/>
Would love an review @neiljp .

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -39,6 +39,7 @@ class Controller:
         self.footlinks_enabled = footlinks
         self.message_search = None
         self.stream_button = None
+        self.sticky_search = False
 
         self._editor = None  # type: Optional[Any]
 
@@ -241,10 +242,12 @@ class Controller:
         else:
             self.view.message_view.log.extend(w_list)
 
-        if self.message_search is not None:
-            self.view.search_box.text_box.set_edit_text(self.message_search)
-        else:
-            self.view.search_box.text_box.set_edit_text('')
+        if self.sticky_search:
+            if self.message_search is not None:
+                self.view.search_box.text_box.set_edit_text(
+                    self.message_search)
+            else:
+                self.view.search_box.text_box.set_edit_text('')
 
         self.exit_editor_mode()
 
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -37,6 +37,8 @@ class Controller:
         self.autohide = autohide
         self.notify_enabled = notify
         self.footlinks_enabled = footlinks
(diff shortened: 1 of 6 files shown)


zulipterminal/core.py
```diff
@@ -37,6 +37,8 @@ def __init__(self, config_file: str, theme: ThemeSpec,
         self.autohide = autohide
         self.notify_enabled = notify
         self.footlinks_enabled = footlinks
+        self.message_search = None
+        self.stream_button = None
```
aryanshridhar, line 41: Adding this here kind of feels inappropriate , Would love any suggestions regarding the same .

neiljp (commented):
@aryanshridhar This doesn't appear to resolve the issue in the way that we discussed, at least by my interpretation and some manual testing.

For example, if I do `alt /` in a stream (or topic), type in a search, then `z` to toggle into topic and back out to stream, neither of those views is a search narrow, even if the text is preserved in the text box - and I can tell since the content doesn't all include the searched-for text.

I know I pointed to _narrow_to, but note that search narrows are currently handled by a different method.

Including the sticky search status in the Controller seems reasonable as a first try, but note that the narrow state, including search status, should be accessible via the model, though it is perhaps debatable whether it belongs there in the long term.

I'd suggest you work from a first commit where sticky search is enabled by default (though via that boolean in the controller), so that you don't need to consider updating it at first, but rather focus on the new behavior, while retaining the previous behavior in different conditional branches in the code. Then you can introduce the toggling behavior in a later commit.

I think we also need to consider, when we allow configuration via `/` or `alt /`, whether we're currently in a regular or sticky search; that could be in a followup to the first commit, once there is the possibility of changing between the modes.

neiljp:
@aryanshridhar Do you intend to continue this soon? I don't believe we have an issue for this?

PR #823, opened 2020-11-11
Autocomplete with exact matches in stream & topic compose boxes

### LINKED ISSUE #776: Consider spaces in autocomplete
Currently, `urwid-readline` only considers the last word for autocomplete and ignores any whitespaces preceding it. This would be needed in long topic names that have spaces or completing full names of users.

(https://github.com/rr-/urwid_readline/issues/13)

neiljp:
This should partially or fully fix #776, but is pending on https://github.com/rr-/urwid_readline/pull/16 and assumes there will be a release of urwid-readline. For now my branch can be used to test this behavior; this allows 2 xfail-ing tests to be set to act normally.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,7 +6,7 @@ name = "pypi"
 [packages]
 urwid = "~=2.1.2"
 zulip = ">=0.7.0"
-urwid-readline = ">=0.11"
+urwid-readline = ">=0.12"
 beautifulsoup4 = ">=4.9.0"
 lxml = ">=4.5.2"
 typing_extensions = ">=3.7"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,11 @@
 urwid~=2.1.2
 zulip>=0.7.0
+urwid_readline>=0.11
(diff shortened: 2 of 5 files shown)


neiljp:
I've tidied `requirements.txt` to reflect the order in `setup.py` - though we could perhaps just remove requirements.txt?

The previous iteration also didn't update the dependency in requirements.txt and Pipfile, which this now does.

`urwid-readline` 0.12 has now been released, so this is more easily tested and should work fine.

kaustubh-nair (approved):
@neiljp This looks good!
We'd also want to incorporate this with the normal typeahead in the compose box?

neiljp:
@kaustubh-nair That would be another project/issue, as we can't just have no delimiters like here, as we'd match the entire box again. The logic available in the library may replace some of what we have already, but we may be able to dynamically set the autocomplete-delimiters, as long as we remember to reset them! If you follow what I'm saying, please open an issue to describe what could be done so we don't forget :)

I'll merge this now.

PR #825, opened 2020-11-14
model: Fetch emoji using events

mkp6781:
Fetching emoji using events ensures that emoji updates are not missed.
Initial state of emoji is updated from initial_data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,18 @@ Other gitlint options are available; for example it is possible to apply it to a
 
 Generally it is best to refer to the git log to get an idea of the general style to follow, and in particular look for similar types of commits to the ones you are writing.
 
+### Tips for working with tests (pytest)
+
+We currently have thousands of tests which get checked upon running `pytest`. While it is dependent on your system capability, this should typically take less than one minute to run. However, during debugging you may still wish to limit the scope of your tests, to improve the turnaround time:
+* If lots of tests are failing in a very verbose way, you might try the `-x` option (eg. `pytest -x`) to stop tests after the first failure; due to parametrization of tests and test fixtures, many apparent errors/failures can be resolved with just one fix! (try eg. `pytest --maxfail 3` for a less-strict version of this)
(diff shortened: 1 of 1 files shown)


zulipterminal/model.py
```diff
@@ -1086,6 +1089,13 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
             set_count(list(message_ids_to_mark & indexed_message_ids),
                       self.controller, -1)
 
+    def _handle_update_emoji_event(self, event: Event) -> None:
+        """
+        Handle update of emoji
+        """
+        realm_emojis = event['realm_emoji']
+        self.active_emoji_data.update(realm_emojis)
```
neiljp, line 1097: As I mentioned previously, events are not in the same format as the original data, so a dict update is not feasible here.

zulipterminal/model.py
```diff
@@ -145,7 +148,7 @@ def __init__(self, controller: Any) -> None:
         for name, data in unicode_emoji_data.items():
             data['type'] = 'unicode_emoji'
         typed_unicode_emoji_data = cast(NamedEmojiData, unicode_emoji_data)
-        custom_emoji_data = self.fetch_custom_emojis()
+        custom_emoji_data = self.initial_data['realm_emoji']
```
neiljp, line 151: This sets `custom_emoji_data` to a different data format than before; note the processing done in `fetch_custom_emojis` - we could rename and adjust that to process the realm emoji in the initial data instead.

In principle we could adjust the code to handle original and emoji events to be in the same format, but we still need to integrate those with the format of the `active_emoji_data` upon update.

neiljp (commented):
@mkp6781 Thanks for starting to put together the ideas from the issues - it's clearer to see what you have here and give feedback :+:1

This shows you have the scaffold for this, but note that the realm_emoji data you're adding is in a different format to that previously - that's at least one reason why you have so many test failures.

Note that if you update the fetch_custom_emojis and the associated locations and tests, then this will likely make adapting the tests easier.

I'm not sure why the tests aren't running yet :confused:

PR #827, opened 2020-11-16
model: Fetch realm emoji using events

mkp6781:
Continues work on #825. 
Processed `initial_data['realm_emoji]` and set `custom_emoji_data` according to the schema given by `NamedEmojiData`.
Similar processing for `realm_emoji_update` before updating `active_emoji_data` in `_handle_update_emoji_event`.
Also, the deactivated realm emojis are removed from `active_emoji_data`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -208,6 +208,13 @@ def unicode_emojis():
     ])
 
 
+@pytest.fixture
+def zulip_emoji():
+    return OrderedDict([
+        ('zulip', {'code': 'zulip', 'type': 'zulip_extra_emoji'})
+    ])
+
+
 stream_msg_template = {
     'id': 537286,
     'sender_full_name': 'Foo Foo',
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -185,13 +185,44 @@ def streams_fixture():
 
 
 @pytest.fixture
-def custom_emojis():
+def realm_emojis():
+    # Omitting source_url, author_id (server version 3.0),
+    # author (server version < 3.0) since they are not used.
+    return {
+        "1": {
+            "deactivated": True,
+            "id": "1",
+            "name": "green_tick",
+        },
+        "202020": {
+            "deactivated": False,
+            "id": "202020",
+            "name": "joker",
+        },
+        "2": {
+            "deactivated": True,
+            "id": "2",
+            "name": "joy_cat",
+        },
+        "3": {
+            "deactivated": False,
+            "id": "3",
+            "name": "singing",
+        },
+        "4": {
+            "deactivated": False,
+            "id": "4",
+            "name": "zulip",
+        }
+    }
+
+
+@pytest.fixture
+def realm_emojis_data():
     return OrderedDict([
-        ('urwid', {'code': '100', 'type': 'realm_emoji'}),
-        ('dancing', {'code': '3', 'type': 'realm_emoji'}),
-        ('snape', {'code': '20', 'type': 'realm_emoji'}),
         ('joker', {'code': '202020', 'type': 'realm_emoji'}),
-        ('zulip', {'code': '12345', 'type': 'realm_emoji'}),
+        ('singing', {'code': '3', 'type': 'realm_emoji'}),
+        ('zulip', {'code': '4', 'type': 'realm_emoji'}),
     ])
 
 
@@ -395,7 +426,7 @@ def topics():
 
 
 @pytest.fixture
-def initial_data(logged_on_user, users_fixture, streams_fixture):
+def initial_data(logged_on_user, users_fixture, streams_fixture, realm_emojis):
     """
     Response from /register API request.
     """
@@ -547,6 +578,7 @@ def initial_data(logged_on_user, users_fixture, streams_fixture):
             }
         },
         'twenty_four_hour_time': True,
+        'realm_emoji': realm_emojis,
         'last_event_id': -1,
         'muted_topics': [],
         'realm_user_groups': [],
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -34,7 +34,7 @@ class TestModel:
 
     @pytest.fixture
     def model(self, mocker, initial_data, user_profile,
-              unicode_emojis, custom_emojis):
+              unicode_emojis):
         mocker.patch('zulipterminal.model.Model.get_messages',
                      return_value='')
         self.client.register.return_value = initial_data
@@ -47,13 +47,11 @@ class TestModel:
         self.client.get_profile.return_value = user_profile
         mocker.patch('zulipterminal.model.unicode_emojis',
                      EMOJI_DATA=unicode_emojis)
-        mocker.patch('zulipterminal.model.Model.fetch_custom_emojis',
-                     return_value=custom_emojis)
         model = Model(self.controller)
         return model
 
     def test_init(self, model, initial_data, user_profile,
-                  unicode_emojis, custom_emojis, zulip_emoji, stream_dict):
+                  unicode_emojis, realm_emojis_data, zulip_emoji, stream_dict):
         assert hasattr(model, 'controller')
         assert hasattr(model, 'client')
         assert model.narrow == []
@@ -80,9 +78,11 @@ class TestModel:
         self.classify_unread_counts.assert_called_once_with(model)
         assert model.unread_counts == []
         assert model.active_emoji_data == OrderedDict(sorted(
-            {**unicode_emojis, **custom_emojis, **zulip_emoji}.items(),
+            {**unicode_emojis, **realm_emojis_data, **zulip_emoji}.items(),
             key=lambda e: e[0]
         ))
+        # Deactivated emoji is removed from active emojis set
+        assert 'green_tick' not in model.active_emoji_data
         # Custom emoji replaces unicode emoji with same name.
         assert model.active_emoji_data['joker']['type'] == 'realm_emoji'
         # zulip_extra_emoji replaces all other emoji types for 'zulip' emoji.
@@ -185,6 +185,7 @@ class TestModel:
             'realm_user',
             'realm_user_groups',
             'update_display_settings',
+            'realm_emoji',
             'zulip_version',
         ]
         model.client.register.assert_called_once_with(
@@ -2110,57 +2111,24 @@ class TestModel:
             model.user_name_from_id(user_id)
 
     def test_generate_all_emoji_data(self, mocker, model, zulip_emoji,
-                                     unicode_emojis, custom_emojis):
-        all_emoji_data = model.generate_all_emoji_data()
+                                     unicode_emojis, realm_emojis_data,
+                                     realm_emojis):
+        all_emoji_data = model.generate_all_emoji_data(realm_emojis)
 
         assert all_emoji_data == OrderedDict(sorted(
-            {**unicode_emojis, **custom_emojis, **zulip_emoji}.items(),
+            {**unicode_emojis, **realm_emojis_data, **zulip_emoji}.items(),
             key=lambda e: e[0]
         ))
-        assert all_emoji_data['urwid']['type'] == 'realm_emoji'
+        # custom emoji added to active_emoji_data
+        assert all_emoji_data['singing']['type'] == 'realm_emoji'
+        # Deactivated custom emoji is removed from active emojis set
+        assert 'green_tick' not in all_emoji_data
+        # deactivated custom emoji which replaced unicode emoji with same name
+        assert all_emoji_data['joy_cat']['type'] == 'unicode_emoji'
         # Custom emoji replaces unicode emoji with same name.
         assert all_emoji_data['joker']['type'] == 'realm_emoji'
         # zulip_extra_emoji replaces all other emoji types for 'zulip' emoji.
         assert all_emoji_data['zulip']['type'] == 'zulip_extra_emoji'
-
-    @pytest.mark.parametrize('response', [
-        {
-            # Omitting source_url, author_id (server version 3.0),
-            # author (server version < 3.0) from response since they
-            # are not used.
-            'emoji': {
-                '100': {
-                    'deactivated': False,
-                    'id': '100',
-                    'name': 'urwid',
-                },
-                '20': {
-                    'deactivated': False,
-                    'id': '20',
-                    'name': 'snape',
-                },
-                '3': {
-                    'deactivated': False,
-                    'id': '3',
-                    'name': 'dancing',
-                },
-                '81': {  # Not included in custom_emojis because deactivated.
-                    'deactivated': True,
-                    'id': '81',
-                    'name': 'green_tick',
-                },
-            },
-            'msg': '',
-            'result': 'success'
-        }
-    ])
-    def test_fetch_custom_emojis(self, mocker, model, custom_emojis,
-                                 response):
-        self.client.get_realm_emoji.return_value = response
-
-        fetched_custom_emojis = model.fetch_custom_emojis()
-
-        assert fetched_custom_emojis == custom_emojis
 
     # Use LoopEnder with raising_event to cause the event loop to end without
     # processing the event
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -174,6 +174,7 @@ class TestModel:
             'typing',
             'update_message_flags',
             'update_display_settings',
+            'realm_emoji',
         ]
         fetch_event_types = [
             'realm',
@@ -2129,6 +2130,65 @@ class TestModel:
         assert all_emoji_data['joker']['type'] == 'realm_emoji'
         # zulip_extra_emoji replaces all other emoji types for 'zulip' emoji.
         assert all_emoji_data['zulip']['type'] == 'zulip_extra_emoji'
+
+    @pytest.mark.parametrize(['to_vary_in_realm_emoji',
+                              'emoji_should_be_active'], [
+            ({
+                "deactivated": False,
+                "id": "20",
+                "name": "joy_cat",
+            }, True),
+            ({
+                "deactivated": True,
+                "id": "202020",
+                "name": "joker",
+            }, True),
+            ({
+                "deactivated": False,
+                "id": "22",
+                "name": "zulip",
+            }, True),
+            ({
+                "deactivated": True,
+                "id": "4",
+                "name": "zulip",
+            }, True),
+            ({
+                "deactivated": False,
+                "id": "23",
+                "name": "new_emoji",
+            }, True),
+            ({
+                "deactivated": True,
+                "id": "3",
+                "name": "singing",
+            }, False),
+        ],
+        ids=[
+            'realm_emoji_with_same_name_as_unicode_emoji_added',
+            'realm_emoji_with_same_name_as_unicode_emoji_removed',
+            'realm_emoji_with_name_as_zulip_added',
+            'realm_emoji_with_name_as_zulip_removed',
+            'realm_emoji_added',
+            'realm_emoji_removed',
+        ]
+    )
+    def test__handle_update_emoji_event(self, mocker, model, realm_emojis,
+                                        emoji_should_be_active,
+                                        to_vary_in_realm_emoji):
+        emoji_name = to_vary_in_realm_emoji['name']
+        realm_emojis[to_vary_in_realm_emoji['id']] = to_vary_in_realm_emoji
+        event = {
+            'type': 'realm_emoji',
+            'realm_emoji': realm_emojis
+        }
+
+        model._handle_update_emoji_event(event)
+
+        if emoji_should_be_active:
+            assert emoji_name in model.active_emoji_data
+        else:
+            assert emoji_name not in model.active_emoji_data
 
     # Use LoopEnder with raising_event to cause the event loop to end without
     # processing the event
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -53,7 +53,7 @@ class TestModel:
         return model
 
     def test_init(self, model, initial_data, user_profile,
-                  unicode_emojis, custom_emojis, stream_dict):
+                  unicode_emojis, custom_emojis, zulip_emoji, stream_dict):
         assert hasattr(model, 'controller')
         assert hasattr(model, 'client')
         assert model.narrow == []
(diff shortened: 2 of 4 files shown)


zulipterminal/model.py
```diff
@@ -1086,6 +1088,23 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
...
+        for emoji in realm_emoji_update:
+            if realm_emoji_update[emoji]['deactivated']:
+                # Removes emojis that have been deleted
+                # from the active_emoji_data
+                _ = self.active_emoji_data.pop(
+                    realm_emoji_update[emoji]['name'], "")
+        realm_emojis = {emoji['name']: {'code': emoji_code,
+                                        'type': 'realm_emoji'}
+                        for emoji_code, emoji in realm_emoji_update.items()
+                        if not emoji['deactivated']}  # type: NamedEmojiData
```
neiljp, line 1105: Note that this is functionally equivalent to `fetch_custom_emojis`, so given that function no longer actually does a fetch but rather just processes data, let's rename and reuse it.

zulipterminal/model.py
```diff
@@ -1086,6 +1088,23 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
...
+        """
+        Handle update of emoji
+        """
+        realm_emoji_update = event['realm_emoji']
+        for emoji in realm_emoji_update:
+            if realm_emoji_update[emoji]['deactivated']:
+                # Removes emojis that have been deleted
+                # from the active_emoji_data
+                _ = self.active_emoji_data.pop(
+                    realm_emoji_update[emoji]['name'], "")
```
neiljp, line 1101: Let's adjust this loop to avoid repeated indexing of `realm_emoji_update[emoji]` - note that you're only cycling through the keys of the dict right now.

neiljp (commented):
@mkp6781 Thanks for working on this, it looks like this is on the right path :+1:

Did you open this to avoid having the other PR based on your master branch?

We may need special behavior with the zulip emoji - I'm not sure if that can be replaced and if we need to handle that specially locally or not.

The tests are running - and failing right now. This looks like only one or two issues, so you might find this easy to fix, either via travis or running pytest locally (see the pytest tips section).

It would be great if you could write some tests for this - these events aren't really that common to test manually on a normal server.

mkp6781:
I have fixed the tests by adding `realm_emoji` to the `initial_data` in conftest.py.

>It would be great if you could write some tests for this - these events aren't really that common to test manually on a normal server. 

I am working on writing the test for `_handle_update_emoji_event`. Will update the PR soon.

zulipterminal/model.py
```diff
@@ -1086,6 +1088,22 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
...
 
+    def _handle_update_emoji_event(self, event: Event) -> None:
+        """
+        Handle update of emoji
+        """
+        realm_emoji_update = event['realm_emoji']
+        for emoji_code, emoji in realm_emoji_update.items():
+            if emoji['deactivated']:
+                # Removes emojis that have been deleted
+                # from the active_emoji_data
```
neiljp, line 1099: This comment applies to the entire loop? If so, let's move it up.

neiljp (commented):
@mkp6781 Thanks for the update - you can just add a note here requesting review, or in principle adjusting the labels (if zulipbot is working). This feedback is a little longer due to giving some ideas about commit structure. I've only reviewed the net changes, and will review in more detail in a later iteration.

You have 6 commits now, but myself and other contributors don't really need to see the intermediate steps you're taking (unless you want to discuss those intermediate stages while working on it), particularly if you're adding/removing debug statements or correcting some other change from a previous commit. Given you have the same commit title for each commit, did you mean to have them all as one commit? If so you may want to investigate `--amend` in future. However, there is nothing wrong with multiple commits with temporary names that you can later bring together - it's easier to squash commits together than split them apart, generally.

For now, in terms of adjusting multiple commits, I'd recommend an interactive rebase to adjust your commits to tell a "story" of the changes you're making and why. As a pointer here, it'd be useful to have commits which make minimum changes to achieve:
* migrate from using fetch to using the value possible in the initial_data
* add event handling

After each commit linting/tests should pass, so the tests can be updated and extended for any new functionality in each commit.

You could also consider separate additional commits to refactor the code, to prepare for the related change, such as generalizing the name/parameterization of the emoji code.

If commits are functionally isolated in a sequence like this, this also means that independent groups of commits can sometimes be merged separately first, from the same PR, which can make discussion of the later commits and functionality easier.

mkp6781:
@zulipbot claim

zulipterminal/model.py
mkp6781: I have not reused it because the return value of `        mocker.patch('zulipterminal.model.Model.process_realm_emoji_data',
                     return_value=custom_emojis)` has been hardcoded. As a result my `test__handle_update_emoji_event` fails.

tests/model/test_model.py
```diff
@@ -42,7 +42,7 @@ def model(self, mocker, initial_data, user_profile,
         self.client.get_profile.return_value = user_profile
         mocker.patch('zulipterminal.model.unicode_emojis',
                      EMOJI_DATA=unicode_emojis)
-        mocker.patch('zulipterminal.model.Model.fetch_custom_emojis',
+        mocker.patch('zulipterminal.model.Model.process_realm_emoji_data',
                      return_value=custom_emojis)
```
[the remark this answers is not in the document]
mkp6781: Yes it does. 
I did not modify the code in `test_process_realm_emoji_data`. I merely renamed it from `fetch_custom_emoji`.  It seems like this piece of code would never fail given that the return value has been hardcoded.

neiljp:
@mkp6781 I've assigned the issue to you manually for now, since zulipbot appears to be unresponsive.

mkp6781:
Hi 
@neiljp Should I make more changes to the tests to make sure that we can use `process_realm_emoji` function again to update `active_emoji_data`?

mkp6781:

zulipterminal/model.py
```diff
@@ -380,13 +383,12 @@ def update_stream_message(self, topic: str, message_id: int,
...
-        response = self.client.get_realm_emoji()
+    def process_realm_emoji_data(self) -> NamedEmojiData:
         custom_emojis = {emoji['name']: {'code': emoji_code,
                                          'type': 'realm_emoji'}
-                         for emoji_code, emoji in response['emoji'].items()
+                         for emoji_code, emoji in
+                         self.initial_data['realm_emoji'].items()
                          if not emoji['deactivated']}  # type: NamedEmojiData
-        display_error_if_present(response, self.controller)
         return custom_emojis
```
neiljp, line 392: This is difficult to test and reuse since you're not allowing the function to vary based on some other data - ie. passing in a parameter, like initial_data emoji vs event emoji.

zulipterminal/model.py
```diff
@@ -1086,6 +1088,22 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
...
+    def _handle_update_emoji_event(self, event: Event) -> None:
+        """
+        Handle update of emoji
+        """
+        realm_emoji_update = event['realm_emoji']
+        # Removes emojis that have been deleted
+        # from the active_emoji_data
+        for emoji_code, emoji in realm_emoji_update.items():
+            if emoji['deactivated']:
+                _ = self.active_emoji_data.pop(emoji['name'], "")
```
neiljp, line 1100: Note that if the deactivated emoji had the same name as a unicode emoji, then this is the wrong behavior, as it doesn't bring back the unicode emoji with that name.

We could fully regenerate the active emoji in a function, in the first instance, like we do in `__init__`

zulipterminal/model.py
neiljp: See my note about parameterising the function.

neiljp (commented):
@mkp6781 Thanks for your work on this! I've left some comments below - let me know if you need more guidance :+1:

zulipterminal/model.py
mkp6781: When a realm emoji and unicode emoji have the same name, the unicode emoji is replaced by the realm emoji.
According to what you have mentioned here, the code should bring back this unicode emoji once the user deactivates the realm emoji with the same name.
Correct me if I'm wrong

neiljp: The order goes unicode<realm<zulip, which is why we have the dict unpacking of each individual dict in the way we do, in `__init__`.

mkp6781:

mkp6781:
> As a pointer here, it'd be useful to have commits which make minimum changes to achieve:
> 
>     * migrate from using fetch to using the value possible in the initial_data
> 
>     * add event handling
> 
> 
> After each commit linting/tests should pass, so the tests can be updated and extended for any new functionality in each commit.

Thank you for the elaborate description regarding the commit structure. I have modified my commits accordingly. I have also made the changes mentioned.

mkp6781:

mkp6781:
Please ignore the last 3 force pushes.

zulipterminal/model.py
```diff
@@ -128,18 +128,9 @@ def __init__(self, controller: Any) -> None:
...
-        zulip_extra_emoji: NamedEmojiData = {
-                'zulip': {'code': 'zulip', 'type': 'zulip_extra_emoji'}
-        }
-        all_emoji_data = {**typed_unicode_emoji_data,
-                          **custom_emoji_data,
-                          **zulip_extra_emoji}.items()
-        self.active_emoji_data = OrderedDict(sorted(all_emoji_data,
+        all_emoji_data = self.generate_all_emoji_data(
+                                        self.initial_data['realm_emoji'])
+        self.active_emoji_data = OrderedDict(sorted(all_emoji_data.items(),
```
neiljp, line 134: Can we include this in the generation function?

neiljp (commented):
@mkp6781 Thanks for pushing ahead with this :+1:

The first commit is looking a bit large - it's both moving the code from `__init__` and using initial data instead. Splitting them into two commits doing first one and then the other may make the changes clearer?

zulipterminal/model.py
```diff
@@ -425,17 +416,28 @@ def update_stream_message(self, topic: str, message_id: int,
...
+                'zulip': {'code': 'zulip', 'type': 'zulip_extra_emoji'}
         }
-        display_error_if_present(response, self.controller)
+        all_emoji_data = {**typed_unicode_emoji_data,
+                          **custom_emoji_data,
+                          **zulip_extra_emoji}
+        return all_emoji_data
+
+    def process_realm_emoji_data(self,
+                                 emoji_data: Dict[str, Any]) -> NamedEmojiData:
```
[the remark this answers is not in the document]
mkp6781: Actually its also being used in `test_init` in test_model.py to convert `initial_data['realm_emoji']` to NamedEmojiData type.

mkp6781:

mkp6781:
@zulipbot  remove "PR needs update"

zulipterminal/model.py
```diff
@@ -438,6 +425,22 @@ def fetch_custom_emojis(self) -> NamedEmojiData:
...
+        for name, data in unicode_emoji_data.items():
+            data['type'] = 'unicode_emoji'
+        typed_unicode_emoji_data = cast(NamedEmojiData, unicode_emoji_data)
+        custom_emoji_data: NamedEmojiData = self.fetch_custom_emojis()
+        zulip_extra_emoji: NamedEmojiData = {
+                'zulip': {'code': 'zulip', 'type': 'zulip_extra_emoji'}
+        }
+        all_emoji_data = {**typed_unicode_emoji_data,
+                          **custom_emoji_data,
+                          **zulip_extra_emoji}
```
neiljp, line 437: If you leave the `.items()` here then the 'move diff' will be smaller.

In case you didn't see my note about noting colored blocks try looking at the diff for just this commit using this:
https://git-scm.com/docs/diff-options#Documentation/diff-options.txt---color-movedltmodegt
(or the equivalent in a .gitconfig or similar)

tests/model/test_model.py
```diff
@@ -2015,6 +2016,46 @@ def test_user_name_from_id_invalid(self, model, user_id):
...
+            "zulip_feature_level": 2,
+            "zulip_version": "2.1.0"
+        })
+    ])
+    def test__handle_update_emoji_event(self, mocker, model,
+                                        response):
+        model._handle_update_emoji_event(response)
+        # For an already active emoji which has been deactivated
+        with pytest.raises(KeyError):
+            model.active_emoji_data['urwid']
```
neiljp, line 2052: Are you trying to achieve the equivalent of `assert "urwid" not in model.active_emoji_data`? If so, that's likely clearer - it's not a specific exception we care about, but whether the key is present (subtle, perhaps, but still).

zulipterminal/api_types.py
```diff
@@ -107,11 +107,17 @@ class UpdateMessageFlagsEvent(TypedDict):
     all: bool
 
 
+class UpdateRealmEmojiEvent(TypedDict):
+    type: Literal['realm_emoji']
+    realm_emoji: Dict[str, Any]
```
neiljp, line 112: This may expand the scope of this PR too far, but if you wished to do a refactor and define and apply a `TypedDict` for the `Any` part of this in the API, that would be useful. I hope it's the same format for the initial_data and events :)

I'm happy to look into this myself if you'd prefer, just let me know.

neiljp (commented):
@mkp6781 Thanks for the update! The structure is definitely clearer now :+1:

Having some good tests for this will help both of us (and others!) understand what the methods do, given the initial state - that last commit is a good example of this :)

tests/model/test_model.py
```diff
@@ -2008,45 +2015,6 @@ def test_user_name_from_id_invalid(self, model, user_id):
...
-                    'deactivated': True,
-                    'id': '81',
-                    'name': 'green_tick',
-                },
-            },
-            'msg': '',
-            'result': 'success'
-        }
-    ])
-    def test_fetch_custom_emojis(self, mocker, model, custom_emojis,
```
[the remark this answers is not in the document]
mkp6781: Wouldn't this check be same as the check in test_init if I am to add the test in the first commit.

zulipterminal/model.py
```diff
@@ -1149,6 +1150,14 @@ def formatted_local_time(self, timestamp: int,
         else:
             return local_time.strftime('%a %b %d %H:%M')
 
+    def _handle_update_emoji_event(self, event: Event) -> None:
+        """
+        Handle update of emoji
+        """
+        assert event['type'] == "realm_emoji"
+        self.active_emoji_data = self.generate_all_emoji_data(
+                                                    event['realm_emoji'])
```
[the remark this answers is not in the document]
mkp6781: From my understanding, the response gives us a complete list of all realm emoji that have been added or removed by the organisation. This is what I observed on flushing the output.

tests/model/test_model.py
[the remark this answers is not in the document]
neiljp: Broadly the idea is that we could test init for loading all the initial data from various sources matches what we expect after processing here and elsewhere. In contrast, we can test the processing functions like the one you're adding for various forms of input - numbers of realm-emoji, overlaps with unicode/zulip emoji, deactivated or not, and so on - which we wouldn't want to do in a more general test like test_init.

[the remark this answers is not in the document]
mkp6781: Got it. 👍🏼 I'll add the tests.

mkp6781:

zulipterminal/api_types.py
mkp6781: I have defined `RealmEmojiData` to specify value types of the `Any` part.

mkp6781:

neiljp (commented):
@mkp6781 This older PR mainly needs a rebase and could benefit from some cleaning up of the tests using fixtures - otherwise we're going to end up with separate emoji data unused or duplicated - but otherwise it looks good to go :+1:

The only aspect that comes to mind is that I'm unsure if there are consequences to removing emoji (eg. modifying messages to remove them being used on old messages, or just limiting them on new ones?) - I'm not sure if this is documented in the API or elsewhere.

tests/model/test_model.py
```diff
@@ -2043,58 +2044,65 @@ def test_user_name_from_id_invalid(self, model, user_id):
...
         assert all_emoji_data['zulip']['type'] == 'zulip_extra_emoji'
 
-    @pytest.mark.parametrize('response', [
-        {
-            # Omitting source_url, author_id (server version 3.0),
-            # author (server version < 3.0) from response since they
-            # are not used.
-            'emoji': {
-                '100': {
-                    'deactivated': False,
```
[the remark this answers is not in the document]
mkp6781: Now there is a fixture for `realm_emojis` in `conftest.py`.

zulipterminal/model.py
```diff
@@ -468,9 +469,8 @@ def update_stream_message(self, topic: str, message_id: int,
 
         return response['result'] == 'success'
 
-    def generate_all_emoji_data(self,
-                                custom_emoji: Dict[str,
-                                                   Any]) -> NamedEmojiData:
+    def generate_all_emoji_data(self, custom_emoji:
+                                Dict[str, RealmEmojiData]) -> NamedEmojiData:
```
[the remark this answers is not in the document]
mkp6781:
```
    def generate_all_emoji_data(self, custom_emoji: Dict[str,
                                RealmEmojiData]) -> NamedEmojiData:
```
Is this how it should be because otherwise character limit gets exceeded.

tests/model/test_model.py
```diff
@@ -2103,6 +2104,51 @@ def test_generate_all_emoji_data(self, mocker, model, zulip_emoji,
...
+                    "id": "202020",
+                    "name": "joker",
+                    "deactivated": True
+                },
+                "39": {
+                    "id": "39",
+                    "name": "zulip",
+                    "deactivated": False
+                },
+            },
```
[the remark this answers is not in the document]
mkp6781: I'm not sure about implementing a template for the emoji data. But I have modified the test to use the `realm_emoji` fixture and the `to_vary*` parameter. 

> This response itself can be built in the test from a template
Could you explain this a bit more!

mkp6781:
I have rebased the PR and made changes to the tests. Maybe there are a few more changes to be made which can be done in the next iteration.

> The only aspect that comes to mind is that I'm unsure if there are consequences to removing emoji (eg. modifying messages to remove them being used on old messages, or just limiting them on new ones?) - I'm not sure if this is documented in the API or elsewhere.

I will think about this tomorrow.

tests/model/test_model.py
```diff
@@ -73,13 +73,6 @@ def test_init(self, model, initial_data, user_profile,
         assert model.users == []
         self.classify_unread_counts.assert_called_once_with(model)
         assert model.unread_counts == []
-        zulip_emoji = {
-                'zulip': {'code': 'zulip', 'type': 'zulip_extra_emoji'}
-        }
-        assert model.active_emoji_data == OrderedDict(sorted(
-            {**unicode_emojis, **custom_emojis, **zulip_emoji}.items(),
-            key=lambda e: e[0]
-        ))
```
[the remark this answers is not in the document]
neiljp: We can use the `zulip_emoji` fixture now though.

tests/model/test_model.py
```diff
@@ -2138,6 +2139,69 @@ def test_generate_all_emoji_data(self, mocker, model, zulip_emoji,
...
+        ]
+    )
+    def test__handle_update_emoji_event(self, mocker, model, zulip_emoji,
+                                        unicode_emojis, realm_emojis,
+                                        to_vary_in_realm_emoji):
+        realm_emojis[to_vary_in_realm_emoji['id']] = to_vary_in_realm_emoji
+        response = {
+            'type': 'realm_emoji',
+            'realm_emoji': realm_emojis
+        }
```
neiljp: This was broadly what I meant by a template. I might consider naming this `event` or similar; there's no request/response pair exactly, and we're testing an event-handler.

zulipterminal/api_types.py
```diff
@@ -148,9 +148,17 @@ class UpdateDisplaySettings(TypedDict):
     setting: bool
 
 
+class RealmEmojiData(TypedDict):
+    id: str
+    name: str
+    source_url: str
+    deactivated: bool
+    author_id: int  # NOTE: new in Zulip 3.0 / ZFL 7
```
neiljp, line 156: This is a change, not just an addition; check the api notes.

zulipterminal/model.py
```diff
@@ -480,9 +481,8 @@ def update_stream_message(self, topic: str, message_id: int,
 
         return response['result'] == 'success'
 
-    def generate_all_emoji_data(self,
-                                custom_emoji: Dict[str,
-                                                   Any]) -> NamedEmojiData:
+    def generate_all_emoji_data(self, custom_emoji: Dict[str,
+                                RealmEmojiData]) -> NamedEmojiData:
```
neiljp, line 485: You don't need to specify the refactor location - this is about refining the typing information from the server.

neiljp (commented):
@mkp6781 The test structure looks much better with these changes :+1: A few minor points to consider, but this is essentially ready :)

For the final point in my last review, I'm not sure what the webapp does, if anything, but I'd suggest that could be something to research for a follow-up. If you don't want to take it up straight away, please make sure to file an issue for

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: See the comment in the next review.

tests/conftest.py
```diff
@@ -185,13 +185,43 @@ def streams_fixture():
...
+
+@pytest.fixture
+def realm_emojis_data():
     return OrderedDict([
-        ('urwid', {'code': '100', 'type': 'realm_emoji'}),
-        ('dancing', {'code': '3', 'type': 'realm_emoji'}),
-        ('snape', {'code': '20', 'type': 'realm_emoji'}),
         ('joker', {'code': '202020', 'type': 'realm_emoji'}),
-        ('zulip', {'code': '12345', 'type': 'realm_emoji'}),
+        ('singing', {'code': '3', 'type': 'realm_emoji'}),
```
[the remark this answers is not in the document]
mkp6781: I have now added 'zulip' to the `realm_emojis_data` fixture.

mkp6781:
> For the final point in my last review, I'm not sure what the webapp does, if anything, but I'd suggest that could be something to research for a follow-up. If you don't want to take it up straight away, please make sure to file an issue for

The webapp doesn't make any changes to old messages when a realm emoji is added/removed and ZT follows the same behaviour.

mkp6781:
@zulipbot add "PR needs review "

mkp6781:

neiljp:
@mkp6781 Just pushed a few minor changes and merging now :tada:

Ideas for follow-up:
* We could tighten up the last event test to indicate the expected type rather than whether it'll be present by broadening the expectation from bool to Optional[unicode/realm/zulip]?
* A useful refactor could be to extract the type in the latter into api_types as EmojiType?

PR #831, opened 2020-11-21
GitHub action migration

neiljp:
This is a starting point to transition from Travis CI to GH Actions.

It is not guaranteed that this will be merged, but this provides a potential route in case of issues with Travis.

Current unfinished elements:
* fails to run pytest with pypy3
* unclear how the build environment compares on macos-latest to what we have in travis
* migrate badges to refer to Actions, not Travis
* travis config not removed
* documentation may need updating

**NOTE** The pypy3 issue is the most significant to resolve, and possibly understanding the macos issue further - without at least the first, Travis cannot be fully replaced.


.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,69 @@
...
+  mypy:
+    runs-on: ubuntu-latest
+    name: Lint - Type consistency (mypy)
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Install with linting tools
+        run: pip install .[linting]
```
sumanthvrao, line 15: It might be too much, but we could also consider adding separate setups for mypy, isort, and flake8, so that they can be installed with `pip install .[mypy]` and so on, instead of the `[.linting]` each time.

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,69 @@
...
+          python-version: ${{ matrix.test-python-version }}
+      - name: Upgrade pip
+        run: python -m pip install --upgrade pip
+      - name: Ensure regular package installs from checkout
+        run: pip install .
+      - name: Install test dependencies
+        run: pip install .[testing]
+      - name: Run tests with pytest
+        run: pytest 
+      - name: Run codecov
```
sumanthvrao, line 77: Should we run `codecov` in every python version? We could consider checking a conditional here and running it on one python version.

.github/workflows/lint-and-test.yml
neiljp: The balance in existing CI and here too is intended to be between keeping dependencies centralized (`setup.py` only; some systems list their deps as explicit installs in CI) and not too fine-grained for clutter in that file. I agree we could in principle add more extras. I think one original motivation for having linting separate was just that `mypy` doesn't work so well under some testing environments, so `.[testing]` would fail - and incidentally this was fewer dependencies to install in each CI run.

.github/workflows/lint-and-test.yml
neiljp: Good question - I think it's run in every version currently, but I'm not sure if we should simplify this. Feel free to investigate!

neiljp:
This has been partially merged given the issues we've had with Travis, but this has meant that we've lost support for PyPy and macos for now, which we should resolve if possible by determining how to incorporate them via GitHub Actions.

neiljp:
The last updates should have resolved the remaining issues, so any feedback on this would be welcome.

We might mention the need for external libraries for pypy3 in the notes, given it's necessary here, but that may be a more general issue to investigate too?

@rht You helped with the Travis macos setup I believe - do you have any feeling on whether the Actions version is the best translation of what we had or what we should aim for?

I'm inclined to merge this relatively soon in any case, given that this is now passing.

rht:
The Actions version for macos looks good. The only relevant parameter is the Python version, but Homebrew already provides 3.9 in their formula.

.github/workflows/lint-and-test.yml
```diff
@@ -44,21 +44,33 @@ jobs:
...
   pytest:
-    runs-on: ubuntu-latest
     strategy:
       # Not failing fast allows all matrix jobs to try & finish even if one fails early
       fail-fast: false
       matrix:
-        test-python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
-    name: Install & pytest - CPython ${{ matrix.test-python-version }}
+        env:
+          - {PYTHON: 3.5, OS: ubuntu-latest, NAME: "CPython 3.5 (ubuntu)"}
```
rht, line 52: Why is 3.5 kept though?

neiljp: There's a separate PR for migrating away from 3.5.

.github/workflows/lint-and-test.yml
neiljp: This was resolved in #869.

neiljp:
OK, I'm going to merge this now; the 3.5 removal is in #838.

PR #838, opened 2020-12-01
Migrate to minimum of python3.6 (aka remove python3.5)

neiljp:
Status so far:
* [x] Remove python 3.5 support in installations and CI
* [x] Upgrade isort & simplify config
* [x] Migrate from % and .format string formatting to f-strings where reasonable
* [x] Migrate `TypedDict`s to class syntax
* [ ] Use inline type annotations


neiljp:
The last planned bullet-point for this PR is now listed separately in #901.

Merging this now.

Issue #545, opened 2020-03-17
Check if a user's package versions are consistent with those specified in requirements


kaustubh-nair:
Users running the development version of the app might have outdated packages if the packages are upgraded in subsequent commits. 
While starting up, ZT should check if the package versions are the same, and display a warning/error if there's some mismatch between the installed and required versions.

sundargs2000:
@zulipbot claim

sundargs2000:
Hey. So, I was working on this and I was wondering whether to use the requirement parser package (https://pypi.org/project/requirements-parser/) or just implement a regex parser. Please let me know @kaustubh-nair @amanagr.

neiljp:
@sundargs2000 Thanks for taking the initiative to look into this issue :+1:

While this is marked as a good first issue - and it doesn't require heavy understanding of the project internals to explore - there are a few points to be aware of:
* If we implemented the check so as to be present in the published version, we would want it to be as close as possible to a no-op when things were OK, to minimize startup time (ie. to not impact users not in a development environment, or dev users on one branch)
* While our requirements.txt and Pipfile should be up to date, we mainly publish to PyPI so setup.py would be the preferred source of truth (also they have other dependencies, which we don't strictly need for purely running ZT)
* We'd need to make sure we have the list of dependencies available at run-time, which I'm not sure if we do? 
* In May we added a development alternative to use a makefile to set up the venv; this doesn't resolve the issue through checking at app startup, though each make invocation should ensure this as a pre-requisite.

As above, the makefile approach doesn't resolve this issue as originally specified, so there is still an option to follow-through with that, but if included in the distributed source there are quite a few extra constraints to consider. I appreciate you've worked on the PR and you're welcome to continue with that - noting that we may have a high threshold to merge it. I'd suggest you also explore the current makefile option in your development environment, as we'd also welcome feedback on that, particularly if you used a different method of installation previously.

sundargs2000:
Hey @neiljp. Thank you for the feedback.
- So about making it no-op, does seem very tough as the functions I have used are what pip uses as well during installation. I don't have complete clarity and will look into it. 
- Setup.py would be a better source of truth. Also, I need to just check testing_deps, linting_deps and dev_helper_deps since running with inconsistent install_requires gives an error. This warning is mainly aimed at developers who would pull the repository and not install through pip.
I will look into whether there is any other better method. Any suggestions would be helpful as well. Thank you!

neiljp:
[...]

PR #839, opened 2020-12-03
Checks if a user's package versions are consistent with those specified in requirements.

sundargs2000:
Hey guys. This is in reference to #545. Please go through and let me know if a different implementation was expected. I'll write the tests as well.

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -239,24 +239,24 @@ def list_themes() -> None:
           "using -t/--theme options on command line.")
 
 def check_packages_version() -> None:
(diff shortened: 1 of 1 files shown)


neiljp:
I left comments on the issue regarding the approach, so I'm marking this as needing update based on the discussion there. We can discuss the implementation here.

sundargs2000:
@neiljp Since we can't import from setup.py, the only way I found out was to use `ast` to get `install_requires`. Is that fine or is there any other method?

neiljp:
@sundargs2000 I'm not sure of the best approach here; it's an interesting problem, but you might consider other issues for now.

PR #841, opened 2020-12-05
footlinks: Add threshold configuration for footlinks.

### LINKED ISSUE #773: Allow a threshold configuration for footlinks

Abhirup-99:
This enables user to configure footlinks threshold settings
using footlinks_threshold in zuliprc, retiring the previous
footlinks. Tests have been updated to reflect the change.
Fixes #773.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,13 @@ theme=default
 # Autohide defaults to 'no_autohide', but can be set to 'autohide' to hide the left & right panels except when focused.
 autohide=autohide
 # Footlinks default to 'enabled', but can be set to 'disabled' to hide footlinks.
+# disabled won't show any footlinks.
+# enabled would show the first 3.
 footlinks=disabled
+# If you need more flexibility, use maximum-footlinks.
+# Maximum footlinks to be shown, defaults to 3, but can be set to any value>=0.
+# Cannot be used along with footlinks, use either of them.
+maximum-footlinks=3
 # Notify defaults to 'disabled', but can be set to 'enabled' to display notifications (see next section).
 notify=enabled
 # Color depth defaults to 256, but can be set to 1 (for monochrome) or 16.
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -110,7 +110,7 @@ def test_valid_zuliprc_but_no_connection(capsys, mocker, minimal_zuliprc,
         "Loading with:",
         "   theme 'zt_dark' specified with no config.",
         "   autohide setting 'no_autohide' specified with no config.",
-        "   footlinks setting 'enabled' specified with no config.",
+        "   maximum footlinks value '3' specified with no config.",
         "   color depth setting '256' specified with no config.",
         "\x1b[91m",
         ("Error connecting to Zulip server: {}.\x1b[0m".
@@ -150,7 +150,7 @@ def test_warning_regarding_incomplete_theme(capsys, mocker, monkeypatch,
         "      (you could try: {}, {})"
         "\x1b[0m".format('a', 'b'),
         "   autohide setting 'no_autohide' specified with no config.",
-        "   footlinks setting 'enabled' specified with no config.",
+        "   maximum footlinks value '3' specified with no config.",
         "   color depth setting '256' specified with no config.",
         "\x1b[91m",
         ("Error connecting to Zulip server: {}.\x1b[0m".
@@ -262,9 +262,103 @@ def test_main_cannot_write_zuliprc_given_good_credentials(
     assert lines[-1] == expected_line
 
 
-@pytest.mark.parametrize('error_code, helper_text', [
-    (1, ""),
-    (2, "helper"),
+@pytest.fixture
+def parameterized_zuliprc(tmpdir):
+    def func(config):
+        zuliprc_path = str(tmpdir) + "/zuliprc"
+        with open(zuliprc_path, "w") as f:
+            f.write("[api]\n\n")  # minimal to avoid Exception
+            f.write("[zterm]\n")
+            for key, value in config.items():
+                f.write(f"{key}={value}\n")
+        os.chmod(zuliprc_path, 0o600)
+        return zuliprc_path
+    return func
+
+
+@pytest.mark.parametrize("config_key,config_value,footlinks_output", [
+        ("footlinks", "disabled",
+            "   maximum footlinks value '0' "
+            "specified in zuliprc file from footlinks."),
+        ("footlinks", "enabled",
+            "   maximum footlinks value '3' "
+            "specified in zuliprc file from footlinks."),
+        ("maximum-footlinks", "3",
+            "   maximum footlinks value '3' specified in zuliprc file."),
+        ("maximum-footlinks", "0",
+            "   maximum footlinks value '0' specified in zuliprc file.")
+    ],
+    ids=[
+        "footlinks_disabled",
+        "footlinks_enabled",
+        "maximum-footlinks_3",
+        "maximum-footlinks_0"
+    ])
+def test_successful_main_function_with_config(
+        config_key, config_value,
+        capsys, mocker, parameterized_zuliprc,
+        footlinks_output
+                                             ):
+    config = {
+        "theme": "default",
+        "autohide": "autohide",
+        "notify": "enabled",
+        "color-depth": "256"
+    }
+    config[config_key] = config_value
+    zuliprc = parameterized_zuliprc(config)
+    mocker.patch("zulipterminal.core.Controller.__init__",
+                 return_value=None)
+    mocker.patch("zulipterminal.core.Controller.main",
+                 return_value=None)
+    with pytest.raises(SystemExit):
+        main(["-c", zuliprc])
+
+    captured = capsys.readouterr()
+    lines = captured.out.strip().split("\n")
+    expected_lines = [
+        'Loading with:',
+        "   theme 'zt_dark' specified in zuliprc file (by alias 'default').",
+        "   autohide setting 'autohide' specified in zuliprc file.",
+        footlinks_output,
+        "   color depth setting '256' specified in zuliprc file."
+                     ]
+    assert lines == expected_lines
+
+
+@pytest.mark.parametrize("zulip_config,err_message", [
+    ({
+        "footlinks": "enabled",
+        "maximum-footlinks": "3"
+    },
+     "Footlinks property is not allowed alongside maximum-footlinks"
+    ),
+    ({"maximum-footlinks": "-3"},
+     "Minimum value allowed for maximum-footlinks is 0"
+     )
+    ])
+def test_main_error_with_invalid_zuliprc_options(
+        parameterized_zuliprc, zulip_config, capsys,
+        mocker, err_message
+                                                ):
+    zuliprc = parameterized_zuliprc(zulip_config)
+    mocker.patch("zulipterminal.core.Controller.__init__",
+                 return_value=None
+                 )
+    mocker.patch("zulipterminal.core.Controller.main",
+                 return_value=None)
+    with pytest.raises(SystemExit) as e:
+        main(["-c", zuliprc])
+    assert str(e.value) == "1"
+    captured = capsys.readouterr()
+    lines = captured.out.strip()
+    expected_lines = f"\033[91m{err_message}\033[0m"
+    assert lines == expected_lines
+
+
+@pytest.mark.parametrize('error_code,helper_text', [
+    (0, ""),
+    (1, "helper")
 ])
 def test_exit_with_error(error_code, helper_text,
                          capsys, error_message="some text"):
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -2073,7 +2073,7 @@ class TestMessageBox:
     )
     def test_footlinks_view(self, message_fixture, message_links,
                             expected_text, expected_attrib):
-        self.model.controller.footlinks_enabled = True
+        self.model.controller.maximum_footlinks = 3
         msg_box = MessageBox(message_fixture, self.model, None)
 
         footlinks = msg_box.footlinks_view(message_links)
@@ -2085,16 +2085,17 @@ class TestMessageBox:
             assert footlinks is None
             assert not hasattr(footlinks, 'original_widget')
 
-    @pytest.mark.parametrize('footlinks_enabled, expected_instance', [
-        (False, type(None)),
-        (True, Padding),
+    @pytest.mark.parametrize('maximum_footlinks, expected_instance', [
+        (0, type(None)),
+        (1, Padding),
+        (3, Padding),
     ])
-    def test_footlinks_enabled(self, message_fixture, footlinks_enabled,
-                               expected_instance):
+    def test_footlinks_limit(self, message_fixture, maximum_footlinks,
+                             expected_instance):
         message_links = OrderedDict([
             ('https://github.com/zulip/zulip-terminal', ('ZT', 1, True)),
         ])
-        self.model.controller.footlinks_enabled = footlinks_enabled
+        self.model.controller.maximum_footlinks = maximum_footlinks
         msg_box = MessageBox(message_fixture, self.model, None)
 
         footlinks = msg_box.footlinks_view(message_links)
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -25,6 +25,8 @@ from zulipterminal.version import ZT_VERSION
 
 TRACEBACK_LOG_FILENAME = 'zulip-terminal-tracebacks.log'
 API_CALL_LOG_FILENAME = 'zulip-terminal-API-requests.log'
+ZULIPRC_CONFIG = 'in zuliprc file'
+NO_CONFIG = 'with no config'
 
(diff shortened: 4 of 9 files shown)


Abhirup-99:
@zulipbot add PR needs review

Abhirup-99:
@neiljp pr is updated, please take a look now.

neiljp (commented):
@Abhirup-99 This seems to work well - and I'd forgotten how many places we use footlinks now!

Other than the inline comments, some other general thoughts:
* You have updated tests, but it would be great to see one for the new behavior, unless I've missed this being covered - preferably at least testing how the controller parameter affects the number of shown footlinks, and potentially that the new value gets passed in from the config file?
* You've removed the functionality we previously had for that config option, which we have in 0.5.2; it'd be useful to allow the old option to be used, perhaps only as an alternative, but without allowing both in the file at the same time.
* The README doesn't document the new/alternative option
* We could add a FAQ entry about enabling/disabling/maximum footlinks, particularly in terms of how the old/new option interacts?

Abhirup-99:

tests/cli/test_run.py
```diff
@@ -246,3 +246,124 @@ def test_main_cannot_write_zuliprc_given_good_credentials(
...
+                                                            ):
+    modified_zuliprc = parameterized_zuliprc(minimal_zuliprc, zulip_config)
+    mocker.patch('zulipterminal.core.Controller.__init__',
+                 side_effect=SystemExit)
+    with pytest.raises(SystemExit) as e:
+        main(['-c', modified_zuliprc])
+        assert str(e.value) == "1"
+    captured = capsys.readouterr()
+    lines = captured.out.strip()
+    expected_lines = "Maximum value of 3 is allowed"
```
neiljp, line 368: This wasn't the logic you had before? The maximum (threshold) is intended to be the maximum number of footlinks shown with message content in the message list, and I'm not convinced we need to constrain users to a particular number? Less than 0 makes little sense, though we could treat negative as unlimited.

tests/ui/test_ui_tools.py
```diff
@@ -2079,16 +2079,16 @@ def test_footlinks_view(self, message_fixture, message_links,
             assert footlinks is None
             assert not hasattr(footlinks, 'original_widget')
 
-    @pytest.mark.parametrize('footlinks_enabled, expected_instance', [
-        (False, type(None)),
-        (True, Padding),
+    @pytest.mark.parametrize('maximum_footlinks, expected_instance', [
+        (0, type(None)),
+        (3, Padding),
```
neiljp, line 2084: I think it would be reasonable to add another case here, which doesn't correspond to the enabled/disabled pair.

zulipterminal/cli/run.py
```diff
@@ -213,7 +213,7 @@ def parse_zuliprc(zuliprc_str: str) -> Dict[str, Any]:
         'theme': ('zt_dark', NO_CONFIG),
         'autohide': ('no_autohide', NO_CONFIG),
         'notify': ('disabled', NO_CONFIG),
-        'footlinks': ('enabled', NO_CONFIG),
```
neiljp, line 216: Do we want to leave this option in?

tests/cli/test_run.py
```diff
@@ -246,3 +246,124 @@ def test_main_cannot_write_zuliprc_given_good_credentials(
...
+        "notify": "enabled",
+        "color-depth": "256"
+    },
+    {
+        "theme": "default",
+        "autohide": "autohide",
+        "maximum-footlinks": "3",
+        "notify": "enabled",
+        "color-depth": "256"
+    }])
```
neiljp, line 302: These are the defaults, but given this PR I'd like to see a test for the max=0 and footlinks=enabled cases too. If the other values are the same, you can simplify the parametrization in any case and build the dict in the test, which would make the intent of this test clearer in any case :)

neiljp (commented):
@Abhirup-99 This has become quite complex for a 'good first issue', so thanks for sticking with it!
(most notably to handle the overlapping old and new config options, as well as the tests with configs)

The `parameterized_zuliprc` looks to be really useful for adding other tests for our config handling later :+1:

We should be able to simplify the tests later if we refactor main, which you might have mentioned in the stream?

tests/cli/test_run.py
```diff
@@ -246,3 +246,124 @@ def test_main_cannot_write_zuliprc_given_good_credentials(
...
+                f.write("{key}={value}\n".format(
+                    key=key,
+                    value=value
+                ))
+        return zuliprc
+    return func
+
+
+@pytest.mark.parametrize('zulip_config', [{}])
+def test_successful_main_function_with_no_config(
```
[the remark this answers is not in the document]
Abhirup-99: Yes, this test is to make sure that the main function is completely executed and only terminated on the last line when the controller is initialized with `SystemExit`.

tests/cli/test_run.py
```diff
@@ -246,3 +246,124 @@ def test_main_cannot_write_zuliprc_given_good_credentials(
...
+        "color-depth": "256"
+    },
+    {
+        "theme": "default",
+        "autohide": "autohide",
+        "maximum-footlinks": "3",
+        "notify": "enabled",
+        "color-depth": "256"
+    }])
+def test_successful_main_function_with_config(
```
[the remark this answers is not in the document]
Abhirup-99: Yes, just a default value check, we can remove it though. Was writing a bunch of tests, so thought one which checks the default flow would also be useful.

tests/cli/test_run.py
```diff
@@ -246,3 +246,124 @@ def test_main_cannot_write_zuliprc_given_good_credentials(
...
+        parameterized_zuliprc, zulip_config, minimal_zuliprc,
+        capsys, mocker
+                                             ):
+    modified_zuliprc = parameterized_zuliprc(minimal_zuliprc, zulip_config)
+    mocker.patch('zulipterminal.core.Controller.__init__',
+                 side_effect=SystemExit)
```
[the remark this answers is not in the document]
Abhirup-99: Yes, without the side effect, the functions would never terminate, as it would go one with `Controller` initialization.

tests/cli/test_run.py
Abhirup-99: I inherited this part of the logic from the previous pr. That was reviewed, so I thought this was a limiting scenario.

zulipterminal/cli/run.py
Abhirup-99: We can, a bit of change in the logic. Would update it soon.

Abhirup-99:

tests/cli/test_run.py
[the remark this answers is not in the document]
neiljp: I think we have a number of tests which might not terminate on failing; I'd like to resolve that, but doesn't the validation failure happen before the Controller init?

tests/cli/test_run.py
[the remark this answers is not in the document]
neiljp: I think having the set of defaults passing would be a good test case, but we could integrate it with the similar pre-existing one, perhaps? I'd have to take another look.

tests/cli/test_run.py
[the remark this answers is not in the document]
Abhirup-99: The control would finish in each case whether it goes to completion with the SystemExit in Controller init or it throws errors and stops, or it stops with SystemExit in Controller init and the test fails due to the assert statements after the completion. The SystemExit are the safety nets that would ensure that in any of the above cases the control finishes.

tests/cli/test_run.py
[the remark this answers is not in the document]
Abhirup-99: Ok sure, but I have updated it to make it more of a successful main completion with various variations of maximum-footlinks and footlinks values.

zulipterminal/cli/run.py
```diff
@@ -285,6 +286,19 @@ def main(options: Optional[List[str]]=None) -> None:
...
             theme_to_use = zterm['theme']
 
+        ZULIPRC_CONFIG = 'in zuliprc file'
+        if(zterm['footlinks'][1] == ZULIPRC_CONFIG
+           and zterm['maximum-footlinks'][1] == ZULIPRC_CONFIG):
+            print(in_color(color='red',
+                  text='Footlinks property is not allowed '
+                  'alongside maximum-footlinks'))
+            sys.exit(1)
+        elif(zterm['maximum-footlinks'][1] == ZULIPRC_CONFIG
```
neiljp, line 296: This is clearer as just an `if`.

Please add a space between `if`/`elif` and the conditional part.

zulipterminal/cli/run.py
```diff
@@ -285,6 +286,19 @@ def main(options: Optional[List[str]]=None) -> None:
         else:
             theme_to_use = zterm['theme']
 
+        ZULIPRC_CONFIG = 'in zuliprc file'
```
neiljp, line 289: I'd prefer we didn't rely on testing against this - this is intended for output purposes.

Perhaps we can move the validation into `parse_zuliprc`?

zulipterminal/cli/run.py
```diff
@@ -352,7 +369,20 @@ def main(options: Optional[List[str]]=None) -> None:
...
+
+        if zterm['footlinks'][1] == ZULIPRC_CONFIG:
+            if zterm['footlinks'][0] == 'disabled':
+                maximum_footlinks = 0
+            else:
+                # Random Large Figure which practically
+                # wouldn't be there in the messages
+                maximum_footlinks = 1000
```
neiljp, line 380: This should be 3; see my general feedback.

zulipterminal/cli/run.py
```diff
@@ -321,16 +335,19 @@ def main(options: Optional[List[str]]=None) -> None:
...
         print("   autohide setting '{}' specified {}."
               .format(*zterm['autohide']))
-        print("   footlinks setting '{}' specified {}."
-              .format(*zterm['footlinks']))
+        if zterm['footlinks'][1] == ZULIPRC_CONFIG:
+            print("   footlinks setting '{}' specified {}."
+                  .format(*zterm['footlinks']))
+        else:
+            print("   maximum footlinks setting '{}' specified {}."
+                  .format(*zterm['maximum-footlinks']))
```
neiljp, line 343: We output this as just a maximum value in the about popup; we should make them consistent.

We could output the maximum and append an element to the last part of the zterm tuple to indicate it came from the footlinks option?

neiljp (commented):
@Abhirup-99 The parametrized_zuliprc is great :+1:

To clarify what I said before, we want disabled=0 and enabled=3 for the new maximum levels (not 1000). If the user wants other values then they can use maximum-footlinks=[integer>=0]. Perhaps the maximum=3 was confused with enabled=3 in the previous version?

Regarding the refactor commit, I was referring to extracting something like `exit_with_error("some text")`. This would simplify the code slightly (include some red color, exit appropriately), but also ease testing as you can
* mock it
* assert on the text passed in
* check it raises

For ensuring the Controller `__init__` doesn't happen, if we mock it without a side-effect then it does nothing? We can assert that it does or doesn't get called, too, if we're supposed to exit before that point.

zulipterminal/cli/run.py
Abhirup-99: But `parse_zuliprc` is specifically for parsing the zuliprc file, rather than validation. All other validations take place in the `main` function.

zulipterminal/cli/run.py
Abhirup-99: That could be done, but then it would force another if clause during the traversing of zterm, and again another to check the length of the tuple during printing. Wouldn't a bit of hard-coded values and an if-else during printing than that approach?

Abhirup-99:

Abhirup-99:

Abhirup-99:

zulipterminal/cli/run.py
neiljp: Ahem. After running with the non-PR dev version quite a bit, I see lots of footlinks in some cases, so 1000 seems like a reasonable limit after all. Ideally we wouldn't use a fixed high number but something more programmatic, eg rather than comparing x to 1000 in a conditional we could wrap it and have a bool be True or False in this case, and dependent on the number otherwise.

tests/cli/test_run.py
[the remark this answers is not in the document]
Abhirup-99: I have updated the logic to mock the calls now, rather than using SystemExit.

zulipterminal/cli/run.py
```diff
@@ -339,12 +341,38 @@ def main(options: Optional[List[str]]=None) -> None:
...
+
+        elif(zterm['maximum-footlinks'][1] == ZULIPRC_CONFIG
+             and int(zterm['maximum-footlinks'][0]) < 0):
+            exit_with_error(
+                error_message='Minimum value allowed '
+                'for maximum-footlinks is 0',
+                helper_text='',
+                error_code=1)
+
+        elif zterm['footlinks'][1] == ZULIPRC_CONFIG:
```
neiljp, line 360: This and the `elif` above can both be just `if`, as the earlier blocks both have an exit condition in them.

tests/cli/test_run.py
```diff
@@ -262,9 +262,124 @@ def test_main_cannot_write_zuliprc_given_good_credentials(
...
+    lines = captured.out.strip()
+    expected_lines = "\033[91m{}\033[0m".format(
+        "Footlinks property is not allowed alongside maximum-footlinks")
+    assert lines == expected_lines
+
+
+@pytest.mark.parametrize("zulip_config", [
+        {"maximum-footlinks": "-3"}
+    ])
+def test_minimum_footlinks(
```
neiljp, line 360: The body of these two tests looks similar enough to combine through parametrization?

Maybe a test name starting `test_main_error_`? This isn't really a success we're testing for :)

tests/cli/test_run.py
```diff
@@ -262,9 +262,124 @@ def test_main_cannot_write_zuliprc_given_good_credentials(
...
+                                                                      ):
+    zuliprc = parameterized_zuliprc(zulip_config)
+    mocker.patch("zulipterminal.core.Controller.__init__",
+                 return_value=None
+                 )
+    mocker.patch("zulipterminal.core.Controller.main",
+                 return_value=None)
+    with pytest.raises(SystemExit) as e:
+        main(["-c", zuliprc])
+        assert str(e.value) == "1"
```
neiljp, line 349: This needs to be outside the context-manager.

README.md
```diff
@@ -101,12 +101,15 @@ site=https://realm.zulipchat.com
...
-footlinks=disabled
+# Maximum footlinks to be shown, defaults to 3, but can be set to any value>=0
+maximum-footlinks=3
 # Notify defaults to 'disabled', but can be set to 'enabled' to display notifications (see next section).
 notify=enabled
 # Color depth defaults to 256, but can be set to 1 (for monochrome) or 16.
 color-depth=256
+# Footlinks default to 'enabled', but can be set to 'disabled' to hide footlinks.
+# Cannot be used along side maximum-footlinks
+footlinks=disabled
```
neiljp, line 112: I'd leave this in the same location, with the new option right below it, with something like "if you need more flexibility...".

README.md
```diff
@@ -101,12 +101,15 @@ site=https://realm.zulipchat.com
 theme=default
 # Autohide defaults to 'no_autohide', but can be set to 'autohide' to hide the left & right panels except when focused.
 autohide=autohide
-# Footlinks default to 'enabled', but can be set to 'disabled' to hide footlinks.
-footlinks=disabled
+# Maximum footlinks to be shown, defaults to 3, but can be set to any value>=0
+maximum-footlinks=3
```
neiljp, line 105: Worth noting equivalent values to 'footlinks', ie include that zero is disabled?

tests/cli/test_run.py
```diff
@@ -262,9 +262,124 @@ def test_main_cannot_write_zuliprc_given_good_credentials(
...
+    captured = capsys.readouterr()
+    lines = captured.out.strip()
+    expected_lines = "\033[91m{}\033[0m".format(
+        "Footlinks property is not allowed alongside maximum-footlinks")
```
neiljp, line 353: f-string opportunity, particularly with a parametrized test :)

neiljp (commented):
@Abhirup-99 This is looking really ready :)

These should all be fairly straightforward changes, but let me know if not :+1:

Abhirup-99:

Abhirup-99:

neiljp:
@Abhirup-99 Thanks for your extended work on this - I've just merged this manually as 0eeb6ea6c84677e10070b2c14788d2768e2f7499 :tada:

As we discussed this does slightly change the previous meaning/behavior, which will need addressing in the release notes, or we could adjust later to be more backwards compatible.

I adjusted the general layout in various places including in the tests and made a small simplification, removed some apparent holdover code from the previous commit we already merged and clarified the commit text. I'd suggest looking into `git range-diff` if you haven't already for seeing what's different, though this is simple enough for a `git diff` between branches to see what I tweaked.

Issue #609, opened 2020-04-21
Render italic text differently than bold text


preetmishra:
Currently, we use `msg_bold` for bold and italic text. While this does differentiate the two from the rest of the message, the distinction between these two remains unaddressed.

We could try using 'italics' setting from `urwid`. However, if we choose not to go with that (due to support concerns), we could try adding a new style in `themes.py` that provides a notable distinction.

preetmishra:
@zulipbot add "message rendering" "UI" "good first issue"

neiljp:
Before we implement this, note that the theming is possibly more complex than initially expected: this needs to cope with not just bold *or* italic, but also bold *and* italic - and so potentially the style needs to easily overlap with bold, or we would need two extra themes (which is a bit less clean).

Ezio-Sarthak:
Hi @neiljp . Would you like to start discussing the issue? I think it is possible to make a font italic *and* bold. What I tried simply was to add both bold and italics in the palette.
![Screenshot from 2020-12-07 16-20-59](https://user-images.githubusercontent.com/55916430/101342150-2b79dd00-38a8-11eb-86b9-d1c0498bfa87.png)
Please let me know if I am missing something.

neiljp:
@Ezio-Sarthak It may work in one terminal, but I'm not sure of the widespread support for this, or how well it displays if it is not supported. I started a topic on czo for this: #**zulip-terminal>Italic styling #T609**.

PR #844, opened 2020-12-08
Simple program for styling discussion (issue #609)

Ezio-Sarthak:
This PR is opened to discuss regarding #609

### DIFF
--- a/tools/font-styles-test.py
+++ b/tools/font-styles-test.py
@@ -1,0 +1,20 @@
+#!/usr/bin/env python3
+
+import urwid
+
+def exit_on_q(key):
+	if key in ('q', 'Q'):
+		raise urwid.ExitMainLoop()
+
+palette = [
+	('normal         ', 'white', 'black'),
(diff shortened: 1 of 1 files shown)


neiljp (commented):
@Ezio-Sarthak We used this a little already, and I think this would be good to keep around as a tool in case we want to explore this in future, particularly if you're open to tidying it up a little:
* Turning it from a module to an executable script in `tools/`
* we should set a background and foreground color for styles (I noted this problem in the chat?)
* similarly adding a 'normal' style
* extract the common text into a variable (hello world)
* perhaps simplify building the UI by cycling through the palette and construct the pile automatically?

Ezio-Sarthak:
Sure @neiljp, I'll do that :)

neiljp:
@Ezio-Sarthak Thanks for updating this :+1: I removed the trailing `.py` from the filename and added more details to the commit text and have just merged this as 896112728430c56e4347f5c9f6f397cb2f02a5a7 :tada:

PR #848, opened 2020-12-10
Feature: Add user info view

### LINKED ISSUE #511: Feature: Enable view of user info

Ezio-Sarthak:
**What does this PR do?!**
- Adds a basic interpolable user info view from web app.
- Custom fields omitted for now.
- Press `i` in the users list to see user info.

**Commit flow**
1. Create UI <=> API mappings for a realm user.
2. Adds a method to return tidied user info, so that it becomes displayable.
3. Support backward compatibility of API user response.
4. Adds hotkey `i` to trigger the popup.
5. Structures the user info popup to be shown.
6. Adds UI elements for toggling the popup.
7. Adds keypress events to trigger the popup.

Fixes #511

**Screenshots/GIF**
Screenshot from 2021-04-27 19-58-50

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -971,8 +971,12 @@ class TestModel:
             ({"role": 300}, "role", 300),
             ({"role": 600}, "role", 600),
             ({}, "role", 400),
+            ({"is_owner": True}, "role", 100),
+            ({"is_admin": True}, "role", 200),
+            ({"is_guest": True}, "role", 600),
             ({"is_bot": True}, "is_bot", True),
             ({"bot_owner_id": 12}, "bot_owner_name", "Human 2"),
+            ({"bot_owner": "person2@example.com"}, "bot_owner_name", "Human 2"),
             ({}, "bot_owner_name", ""),
         ],
         ids=[
@@ -991,8 +995,12 @@ class TestModel:
             "user_is_moderator:Zulip_4.0+_ZFL60",
             "user_is_guest:Zulip_4.0+_ZFL59",
             "user_is_member",
+            "user_is_owner:Zulip_3.0+",
+            "user_is_admin:preZulip_4.0",
+            "user_is_guest:preZulip_4.0",
             "user_is_bot",
             "user_bot_has_owner:Zulip_3.0+_ZFL1",
+            "user_bot_has_owner:preZulip_3.0",
             "user_bot_has_no_owner",
         ],
     )
@@ -1001,12 +1009,14 @@ class TestModel:
         model,
         mocker,
         _all_users_by_id,
+        user_dict,
         to_vary_in_each_user,
         key,
         expected_value,
     ):
         _all_users_by_id[11] = dict({"user_id": 11}, **to_vary_in_each_user)
         model._all_users_by_id = _all_users_by_id
+        model.user_dict = user_dict
 
         assert model.get_user_info(11)[key] == expected_value
 
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -949,6 +949,76 @@ class TestModel:
         }
         assert user_group_names == ["Group 1", "Group 2", "Group 3", "Group 4"]
 
+    @pytest.mark.parametrize(
+        ["to_vary_in_each_user", "key", "expected_value"],
+        [
+            ({"full_name": "Test User"}, "full_name", "Test User"),
+            ({}, "full_name", "(No name)"),
+            ({"email": "person1@example.com"}, "email", "person1@example.com"),
+            ({}, "email", ""),
+            (
+                {"date_joined": "2021-02-28T19:58:29.035543+00:00"},
+                "date_joined",
+                "2021-02-28T19:58:29.035543+00:00",
+            ),
+            ({}, "date_joined", ""),
+            ({"timezone": "Asia/Kolkata"}, "timezone", "Asia/Kolkata"),
+            ({}, "timezone", ""),
+            ({"bot_type": 1}, "bot_type", 1),
+            ({}, "bot_type", 0),
+            ({"role": 100}, "role", 100),
+            ({"role": 200}, "role", 200),
+            ({"role": 300}, "role", 300),
+            ({"role": 600}, "role", 600),
+            ({}, "role", 400),
+            ({"is_bot": True}, "is_bot", True),
+            ({"bot_owner_id": 12}, "bot_owner_name", "Human 2"),
+            ({}, "bot_owner_name", ""),
+        ],
+        ids=[
+            "user_full_name",
+            "user_empty_full_name",
+            "user_email",
+            "user_empty_email",
+            "user_date_joined",
+            "user_empty_date_joined",
+            "user_timezone",
+            "user_empty_timezone",
+            "user_bot_type",
+            "user_empty_bot_type",
+            "user_is_owner:Zulip_4.0+_ZFL59",
+            "user_is_admin:Zulip_4.0+_ZFL59",
+            "user_is_moderator:Zulip_4.0+_ZFL60",
+            "user_is_guest:Zulip_4.0+_ZFL59",
+            "user_is_member",
+            "user_is_bot",
+            "user_bot_has_owner:Zulip_3.0+_ZFL1",
+            "user_bot_has_no_owner",
+        ],
+    )
+    def test_get_user_info(
+        self,
+        model,
+        mocker,
+        _all_users_by_id,
+        to_vary_in_each_user,
+        key,
+        expected_value,
+    ):
+        _all_users_by_id[11] = dict({"user_id": 11}, **to_vary_in_each_user)
+        model._all_users_by_id = _all_users_by_id
+
+        assert model.get_user_info(11)[key] == expected_value
+
+    def test_get_user_info_USER_NOT_FOUND(self, model):
+        assert model.get_user_info(-1) is None
+
+    def test_get_user_info_sample_response(
+        self, model, _all_users_by_id, tidied_user_info_response
+    ):
+        model._all_users_by_id = _all_users_by_id
+        assert model.get_user_info(12) == tidied_user_info_response
+
     def test_get_all_users(self, mocker, initial_data, user_list, user_dict, user_id):
         mocker.patch(MODEL + ".get_messages", return_value="")
         self.client.register.return_value = initial_data
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -15,6 +15,7 @@ from zulipterminal.ui_tools.views import (
     PopUpView,
     StreamInfoView,
     StreamMembersView,
+    UserInfoView,
 )
 from zulipterminal.version import MINIMUM_SUPPORTED_SERVER_VERSION, ZT_VERSION
 
@@ -203,6 +204,130 @@ class TestAboutView:
         assert len(about_view.feature_level_content) == (
             1 if server_feature_level else 0
         )
+
+
+class TestUserInfoView:
+    @pytest.fixture(autouse=True)
+    def mock_external_classes(self, mocker, tidied_user_info_response):
+        self.controller = mocker.Mock()
+        mocker.patch.object(
+            self.controller, "maximum_popup_dimensions", return_value=(64, 64)
+        )
+        mocker.patch(MODULE + ".urwid.SimpleFocusListWalker", return_value=[])
+
+        self.user_data = tidied_user_info_response
+
+        mocker.patch.object(
+            self.controller.model, "get_user_info", return_value=self.user_data
+        )
+        mocker.patch.object(
+            self.controller.model,
+            "formatted_local_time",
+            return_value="Tue Mar 13 10:55 AM",
+        )
+
+        self.user_info_view = UserInfoView(
+            self.controller, 10000, "User Info (up/down scrolls)"
+        )
+
+    @pytest.mark.parametrize(
+        [
+            "to_vary_in_each_user",
+            "expected_key",
+            "expected_value",
+        ],
+        [
+            ({}, "Email", "person2@example.com"),
+            ({"email": ""}, "Email", None),
+            ({"date_joined": "2021-03-18 16:52:48"}, "Date joined", "2021-03-18"),
+            ({}, "Date joined", None),
+            ({"timezone": "America/Los_Angeles"}, "Timezone", "America/Los Angeles"),
+            ({}, "Timezone", None),
+            (
+                {"is_bot": True, "bot_type": 1, "bot_owner_name": "Test Owner"},
+                "Owner",
+                "Test Owner",
+            ),
+            ({}, "Owner", None),
+            (
+                {"last_active": "Tue Mar 13 10:55:22"},
+                "Last active",
+                "Tue Mar 13 10:55:22",
+            ),
+            ({}, "Last active", None),
+            ({"is_bot": True, "bot_type": 1}, "Role", "Generic Bot"),
+            ({"is_bot": True, "bot_type": 2}, "Role", "Incoming Webhook Bot"),
+            ({"is_bot": True, "bot_type": 3}, "Role", "Outgoing Webhook Bot"),
+            ({"is_bot": True, "bot_type": 4}, "Role", "Embedded Bot"),
+            ({"role": 100}, "Role", "Owner"),
+            ({"role": 200}, "Role", "Administrator"),
+            ({"role": 300}, "Role", "Moderator"),
+            ({"role": 600}, "Role", "Guest"),
+            ({"role": 400}, "Role", "Member"),
+        ],
+        ids=[
+            "user_email",
+            "user_empty_email",
+            "user_date_joined",
+            "user_empty_date_joined",
+            "user_timezone",
+            "user_empty_timezone",
+            "user_bot_owner",
+            "user_empty_bot_owner",
+            "user_last_active",
+            "user_empty_last_active",
+            "user_is_generic_bot",
+            "user_is_incoming_webhook_bot",
+            "user_is_outgoing_webhook_bot",
+            "user_is_embedded_bot",
+            "user_is_owner",
+            "user_is_admin",
+            "user_is_moderator",
+            "user_is_guest",
+            "user_is_member",
+        ],
+    )
+    def test__fetch_user_data(
+        self, mocker, to_vary_in_each_user, expected_key, expected_value
+    ):
+        data = dict(self.user_data, **to_vary_in_each_user)
+
+        mocker.patch.object(self.controller.model, "get_user_info", return_value=data)
+
+        display_data = self.user_info_view._fetch_user_data(self.controller, 1)
+
+        assert display_data.get(expected_key, None) == expected_value
+
+    def test__fetch_user_data_USER_NOT_FOUND(self, mocker):
+        mocker.patch.object(self.controller.model, "get_user_info", return_value=dict())
+
+        display_data = self.user_info_view._fetch_user_data(self.controller, 1)
+
+        assert display_data["Name"] == "(Unavailable)"
+        assert display_data["Error"] == "User data not found"
+
+    @pytest.mark.parametrize(
+        "key", {*keys_for_command("GO_BACK"), *keys_for_command("USER_INFO")}
+    )
+    def test_keypress_exit_popup(self, key, widget_size):
+        size = widget_size(self.user_info_view)
+        self.user_info_view.keypress(size, key)
+        assert self.controller.exit_popup.called
+
+    def test_keypress_exit_popup_invalid_key(self, widget_size):
+        key = "a"
+        size = widget_size(self.user_info_view)
+        self.user_info_view.keypress(size, key)
+        assert not self.controller.exit_popup.called
+
+    def test_keypress_navigation(
+        self, mocker, widget_size, navigation_key_expected_key_pair
+    ):
+        key, expected_key = navigation_key_expected_key_pair
+        size = widget_size(self.user_info_view)
+        super_keypress = mocker.patch(MODULE + ".urwid.ListBox.keypress")
+        self.user_info_view.keypress(size, key)
+        super_keypress.assert_called_once_with(size, expected_key)
 
 
 class TestEditHistoryView:
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -264,6 +264,11 @@ KEY_BINDINGS: 'OrderedDict[str, KeyBinding]' = OrderedDict([
         'help_text': 'Quit',
         'key_category': 'general',
     }),
+    ('USER_INFO', {
+        'keys': ['i'],
+        'help_text': 'View user information (From Users list)',
+        'key_category': 'general',
+    }),
     ('BEGINNING_OF_LINE', {
         'keys': ['ctrl a'],
         'help_text': 'Jump to the beginning of line',
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -63,6 +63,7 @@ REQUIRED_STYLES = {
     'area:msg'        : 'standout',
     'area:stream'     : 'standout',
     'area:error'      : 'standout',
+    'area:user'       : 'standout',
     'search_error'    : 'standout',
     'task:success'    : 'standout',
     'task:error'      : 'standout',
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -29,6 +29,7 @@ from zulipterminal.ui_tools.views import (
     PopUpConfirmationView,
     StreamInfoView,
     StreamMembersView,
+    UserInfoView,
 )
 from zulipterminal.version import ZT_VERSION
 
@@ -284,6 +285,12 @@ class Controller:
                 maximum_footlinks=self.maximum_footlinks,
             ),
             "area:help",
+        )
+
+    def show_user_info(self, user_id: int) -> None:
+        self.show_pop_up(
+            UserInfoView(self, user_id, "User Information (up/down scrolls)"),
+            "area:user",
         )
 
     def show_edit_history(
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -17,6 +17,7 @@ from typing import (
     Iterable,
     Iterator,
     List,
+    Optional,
     Set,
     Tuple,
     TypeVar,
@@ -49,6 +50,20 @@ class EmojiData(TypedDict):
 
 
 NamedEmojiData = Dict[str, EmojiData]
+
+
+class TidiedUserInfo(TypedDict):
+    full_name: str
+    email: str
+    date_joined: str
+    timezone: str
+    role: Optional[int]
+    last_active: str
+
+    is_bot: bool
+    # Below fields are only meaningful if is_bot == True
+    bot_type: int
+    bot_owner_name: str
 
 
 class Index(TypedDict):
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -39,6 +39,7 @@ from zulipterminal.helper import (
     Message,
     NamedEmojiData,
     StreamData,
+    TidiedUserInfo,
     asynch,
     canonicalize_color,
     classify_unread_counts,
@@ -706,6 +707,49 @@ class Model:
                 if stream["name"] == stream_name
                 if sub != self.user_id
             ]
+
+    def get_user_info(self, user_id: int) -> Optional[TidiedUserInfo]:
+        api_user_data: Optional[RealmUser] = self._all_users_by_id.get(user_id, None)
+
+        if not api_user_data:
+            return None
+
+        # TODO: Add custom fields later as an enhancement
+        user_info: TidiedUserInfo = dict(
+            full_name=api_user_data.get("full_name", "(No name)"),
+            email=api_user_data.get("email", ""),
+            date_joined=api_user_data.get("date_joined", ""),
+            timezone=api_user_data.get("timezone", ""),
+            is_bot=api_user_data.get("is_bot", False),
+            # Role `None` for triggering servers < Zulip 4.1 (ZFL 59)
+            role=api_user_data.get("role", None),
+            bot_type=api_user_data.get("bot_type", 0),
+            bot_owner_name="",  # Can be non-empty only if is_bot == True
+            last_active="",
+        )
+
+        if user_info["role"] is None:
+            # Default role is member
+            user_info["role"] = 400
+
+        bot_owner: Optional[RealmUser] = None
+
+        if api_user_data.get("bot_owner_id", None):
+            bot_owner = self._all_users_by_id.get(api_user_data["bot_owner_id"], None)
+
+        user_info["bot_owner_name"] = bot_owner["full_name"] if bot_owner else ""
+
+        if self.initial_data["presences"].get(user_info["email"], None):
+            timestamp = self.initial_data["presences"][user_info["email"]][
+                "aggregated"
+            ]["timestamp"]
+
+            # Take 24h vs AM/PM format into consideration
+            user_info["last_active"] = self.formatted_local_time(
+                timestamp, show_seconds=True
+            )
+
+        return user_info
 
     def get_all_users(self) -> List[Dict[str, Any]]:
         # Dict which stores the active/idle status of users (by email)
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -35,6 +35,7 @@ from zulipterminal.api_types import (
     Subscription,
 )
 from zulipterminal.config.keys import primary_key_for_command
+from zulipterminal.config.ui_mappings import ROLE_BY_ID
 from zulipterminal.helper import (
     Message,
     NamedEmojiData,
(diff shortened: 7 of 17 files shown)


Ezio-Sarthak:
The PR is ready for review :)

Pipfile
```diff
@@ -12,7 +12,7 @@ lxml = ">=4.5.2"
 typing_extensions = ">=3.7"
 python-dateutil = ">=2.8.1"
 tzlocal = ">=2.1"
-pytz>=2019.3
+pytz = ">=2019.3"
```
neiljp, line 15: Please split the "mistakes" up and put them into the appropriate commits themselves. We wouldn't normally merge a commit that was like this. Let us know if you want help with git.

zulipterminal/ui_tools/views.py
```diff
@@ -1016,6 +1016,26 @@ def __init__(self, controller: Any, title: str, *, zt_version: str,
         super().__init__(controller, widgets, 'ABOUT', popup_width, title)
 
 
+class UserInfoView(PopUpView):
+    def __init__(self,
+                controller: Any,
+                display_data: dict,
```
neiljp, line 1022: Can we improve this type?

neiljp (commented):
@Ezio-Sarthak This looks like a good start :+1:

I like that you've got the separate commits, though we may want to squash some together to represent the changes that should only be visible at the same time.

I do like `u` on a message to some extent, I'm just wary of using keys for actions and users getting used to them, only for us to run out of keys to use later :) I think a `u` from message info, or `i` from the user list might be good triggers for the popup, until we can agree about `u` from the message list.

I often find myself wondering whether someone may be awake, so I look forward to the next iteration :)

Pipfile
```diff
@@ -12,6 +12,7 @@ lxml = ">=4.5.2"
 typing_extensions = ">=3.7"
 python-dateutil = ">=2.8.1"
 tzlocal = ">=2.1"
+pytz>=2019.3
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yeah, I guess the `tz_local` is specifically for obtaining local user's timezone :)
Sure, I'll do that.

zulipterminal/config/keys.py
```diff
@@ -305,6 +305,11 @@
         'help_text': 'Clear compose screen',
         'key_category': 'msg_compose',
     }),
+    ('USER_INFO', {
+        'keys': ['u'],
+        'help_text': 'View user description',
+        'key_category': 'msg_actions',
```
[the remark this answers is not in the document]
Ezio-Sarthak: I totally feel you :) Just like the edit history view in the message info, I'll add the key for the user info view (`i`) so that we're clear with the keys.
Yeah, I'll hook the README and keys commit together :)

zulipterminal/ui_tools/views.py
```diff
@@ -1016,6 +1016,26 @@ def __init__(self, controller: Any, title: str, *, zt_version: str,
...
+                title: str) -> None:
+
+        user_details = []
+        for key, value in display_data.items():
+            if(key == 'Name'):
+                continue
+            user_details.append((key,value))
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yeah, absolutely. I'll make it cleaner and simpler :)

zulipterminal/core.py
```diff
@@ -213,6 +215,34 @@ def show_about(self) -> None:
...
 
+    def show_user_info(self, user_id:int) -> None:
+        response = self.client.get_user_by_id(user_id)
+        # TODO: add custom fields later as an enhancement
+        display_data = {}
+        res_data = response['user']
+        display_data['Name'] = res_data['full_name']
+        display_data['Email'] = res_data['email']
+        display_data['Date joined'] = res_data['date_joined'][:10]
+        display_data['Timezone'] = res_data['timezone']
```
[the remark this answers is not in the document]
Ezio-Sarthak: Gotcha. I'll initialize this and also update its type in `helper.py` :)

zulipterminal/core.py
```diff
@@ -213,6 +215,34 @@ def show_about(self) -> None:
...
+        try:
+            display_data['Local time'] = datetime.now(pytz.timezone(str(res_data['timezone']))).strftime("%H:%M")
+        except:
+            display_data['Local time'] = "Unknown timezone"
+        presence = self.client.get_user_presence(str(res_data['email']))
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yeah, I think so (not completely sure though).
By the GET request `client.get_user_by_id(user_id)`, I got the following dict:
```
{
 'result': 'success',
 'msg': '',
'user': 
  {
    'email': 'user16790@chat.zulip.org', 'user_id': 16790, 'avatar_version': 2,
    'is_admin': False, 'is_owner': False, 'is_guest': False, 'is_bot': False,
    'full_name': 'Matt Hall', 'timezone': 'America/Los_Angeles',
    'is_active': True, 'date_joined': '2020-07-24T21:53:55.770840+00:00',
     'avatar_url': '/user_avatars/2/848f00cf86435f7a12c7f6dafb476e47a7ee5665.png?x=x&version=2'
  }
}
```
Seems like we *have* to fetch the timezone of a user in any case. Please drop any suggestions if you have :)

zulipterminal/ui_tools/views.py
Ezio-Sarthak: Definitely. I'll update `helper.py` as mentioned above :)

Pipfile
Ezio-Sarthak: Yup. Apologies for that :sweat_smile:. I'll put the responsible code changes in separate commits :)

Ezio-Sarthak:
@neiljp Thanks for the review! I'll update and amend all the changes mentioned, and certainly make the PR more cleaner one :)

zulipterminal/core.py
```diff
@@ -213,6 +215,34 @@ def show_about(self) -> None:
...
+            display_data['Last active'] = "Active now"
+        else:
+            display_data['Last active'] = str(datetime.fromtimestamp(presence['presence']['aggregated']['timestamp']))
+
+        if res_data['is_admin']:
+            display_data['Role'] = "Admin"
+        elif res_data['is_guest']:
+            display_data['Role'] = "Guest"
+        else:
+            display_data['Role'] = "Member"
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yeah! I just pulled over some code from the previous [PR](https://github.com/zulip/zulip-terminal/pull/519), so another iteration will fix this :)

Ezio-Sarthak:
PR updated :)

zulipterminal/core.py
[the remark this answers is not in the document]
neiljp: Right, well the user data should be available in the model from what I can tell. It would benefit from refactoring, but the initial data is in `Model.initial_data['realm_users']`.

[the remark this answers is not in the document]
Ezio-Sarthak: Thanks @neiljp for reminding me of the users realm, I've updated the PR accordingly :)

neiljp:
@Ezio-Sarthak This is one of many in the queue to review, but you could check over your commit titles.

Ezio-Sarthak:
Commit titles improved :)

neiljp (commented):
@Ezio-Sarthak I just gave this a go again :+1: Broadly it looks fine, but being a new area there are UI and data aspects which we've not yet had to consider too, which makes this more complex.

Practically:
* As a user I find it strange to have `i` cancel the message info popup, perhaps we should have `s` for sender in that popup? I'm wondering if we should generalize the approach of the link buttons and make the sender 'actionable' too (or instead), but we can discuss on the stream.
* Is a different popup style for this popup reasonable? It's not really help/about, but it's not really message either?
* I'm not sure the roles are quite right; you may want to check the server for the ordering, and I'm not sure if bot is strictly mutually exclusive with the others?
* When there's no timezone, it seems strange to not show the time? That output could be clearer or combined?

Code-wise:
* Please put the tests in the same commit, eg. for `UserInfoView`
* Is the last commit inspired/translated from webapp code? If so, please indicate from where so we have a reference. For the tests you may want to consider some edge cases.
* A lot of show_user_info seems like it digs into the model, or perhaps performs functions that belong there; can we add a model method to provide the data instead?
* Similarly, can we use presence data from the model that's there already?
* In the other direction, does some processing in show_user_info belong in the actual popup? If not, then a lot of our popups have differently named classes but seem very similar, and we may want to simplify later, or as a refactor somewhere.

Ezio-Sarthak:
@neiljp Thanks for a detailed review of the PR! Here's my take on clearing up some of the issues/bugs with the PR:

1. Change the `i` key to `s` key for showing the sender's info for now. (I will discuss on czo about this shortly).
2. Re style of the popup, makes sense. I opted for the message area, just because it popped up in the message info view. We could try adding another palette like `info_user`?
3. I've gone through some of the server and client code for `user_info_popover`, and as you said, the bot conditioning + timezone needs to be handled more appropriately. I'll work on it as well.
4. I'll do the commit squashing/re-ordering/rewording as mentioned.
5. Thanks for suggesting the logic shift to `Model`! I'll update the corresponding commit. (+ the presence status too)
6. I seem to be unclear about what you meant by 'processing belonging in actual popup', but I guess shifting some intermediate extractions to `Model` may reduce this. For the refactoring of popup class names, we could potentially discuss in detail on czo :)

I hope I'm not overlooking something. I'll appreciate any corrections in my concerns :)

Ezio-Sarthak:

tests/model/test_model.py
```diff
@@ -744,6 +745,40 @@ def test__group_info_from_realm_user_groups(self, model,
...
 
+    @pytest.mark.parametrize('presence, time_in_sec, expected_status', [
+        ('active',  0,       'Active now'),
+        ('active',  110,     'Active now'),
+        ('idle',    110,     'Just now'),
+        ('idle',    190,     '3 minutes ago'),
+        ('idle',    3600,    'An hour ago'),
+        ('idle',    10000,   '2 hours ago'),
+        ('idle',    100000,  'Yesterday'),
+        ('idle',    200000,  '2 days ago'),
```
neiljp, line 756: Something here isn't working, as I'm getting `-1 days ago` for idle users in the user list.

zulipterminal/model.py
```diff
@@ -660,14 +661,49 @@ def get_user_info(self, user_id: int) -> Dict[str, Any]:
...
+                user_info['Last active'] = self.get_status(data['email'])
+            else:
+                # There are situations where the client has incomplete
+                # presence history on a user. This can happen when users are
+                # deactivated, or when they just haven't been present in a
+                # long time (and we may have queries on presence that go back
+                # only N weeks).
+                # We give the somewhat vague status of "Unknown" for these
+                # users.
+                user_info['Last active'] = 'More than 2 weeks ago'
```
neiljp, line 677: This doesn't seem to match; could you explain?

neiljp (commented):
@Ezio-Sarthak There's a lot of work here, so note that it takes more work to refine and think of the flow of commits and work. This does seem clearer now.

If we're keeping the popup for the user info view *and* the sender, my preference would be `i` from the user info view and `s` from the message popup (the latter maybe temporary). I understand that may complicate the UI code - the reason I prefer the concept is to unify the concept of "I press `i` to see information on this item".

minor: The format of the dependencies commit is slightly different from previous - check the git log for "requirements". If you think that could be better documented in the README please suggest a correction :)

zulipterminal/helper.py
```diff
@@ -124,6 +124,13 @@ class UnreadCounts(TypedDict):
     'change_all': 'Also change previous and following messages to this topic',
 }
 
+bot_types = {
+    1: 'Generic Bot',
+    2: 'Incoming Webhook Bot',
+    3: 'Outgoing Webhook Bot',
+    4: 'Embeded Bot',
+}
```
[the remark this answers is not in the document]
Ezio-Sarthak: Cool. I'll update that. Given the PR is loaded with commits right now, could I make another PR that refactors and converts all the constant metadata to full-caps name?

zulipterminal/model.py
```diff
@@ -614,6 +617,57 @@ def get_other_subscribers_in_stream(self, stream_id: Optional[int]=None,
...
+
+        else:
+            user_info['Role'] = next((r for r, is_role in user_role.items()
+                                      if is_role), 'Member')
+
+            if self.initial_data['presences'].get(data['email'], None):
+                presence = (self.initial_data['presences'][data['email']]
+                            ['aggregated']['timestamp'])
+                user_info['Last active'] = (datetime.fromtimestamp(presence)
+                                            .strftime("%d-%b-%Y %H:%M:%S"))
```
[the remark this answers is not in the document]
Ezio-Sarthak: Certainly. I didn't rebase this branch for a while so was unaware of that method :sweat_smile: 

Are you suggesting shifting the `formatted_local_time` method in `helper`? If so then I might do this refactor in a prior commit? (Assuming the aspect of this commit unrelated)

>However, in general I'd suggest this method doesn't aim as much for priming the data just for "output", so much as doing error-checking and refining - so that time function could sit in the UI, potentially?

I'm unclear about which method you mean by `this method`. Could you clarify this a bit? :sweat_smile:

tests/model/test_model.py
Ezio-Sarthak: Strange :thinking:. It's working fine for me locally.
There may be some timezone data mishandling in the Model method (on first look),
Or, there could be some unlooked loopholes in the below logic (in `helper`, last commit):
```python
    days = (datetime.now() - datetime.fromtimestamp(
            timestamp)).days
```
Thanks for reporting, as this could be highly misleading :+1:

zulipterminal/model.py
Ezio-Sarthak: Ah, that needs to be amended. Actually, I picked that whole comment (L670-L676) while referring to the web app's implementation :sweat_smile: 

All that means (AFAIK) is that the Webapp declares the status `Active More than two weeks ago` for all those users whose either account is being deactivated, or are not active for a very long time :)

zulipterminal/helper.py
```diff
@@ -708,3 +709,36 @@ def get_unused_fence(content: str) -> str:
...
+
+    minutes = int((datetime.now() - datetime.fromtimestamp(
+                  timestamp)).seconds / 60)
+    if minutes <= 2:
+        return 'Just now'
+    if minutes < 60:
+        return f"{minutes} minutes ago"
+
+    days = (datetime.now() - datetime.fromtimestamp(
+            timestamp)).days
```
Ezio-Sarthak, line 730: Re my reply mentioning this code snippet, I think this isn't the right way to get the days. There could be a difference in the `datetime.now()` on L722 and here, and that's simply unreliable. I'll change it to get days by the minutes variable above :+1:

zulipterminal/model.py
neiljp: Does the webapp implementation use that text, or "Unknown"?

tests/model/test_model.py
neiljp: The 'ago' part can be left until later I suspect - ie. we could leave it as absolute date/time for now?

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: Re the new/centralized method, the date/time output is centralized - and after #917 will be user-customized via the user account setting on the server - so the user likely should expect that format. So instead of using datetime directly, just use that new method? I'm not sure where helper comes into it?

Re this method, I meant the model one you're adding to get the user info - there are fine lines between the data from the server, the clarified/checked internal data, and that presented to the user. However, for now we can work on getting it working in a "reasonable" place and iterate from there - it's a bit like optimization: take care over a sensible starting point, but let's not debate for too long over how to improve it rather than getting it done :)

zulipterminal/helper.py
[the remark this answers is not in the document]
neiljp: If you're adding code, you can write it correctly when you add it - there's no point adding "bad" code if you know you're going to replace it :)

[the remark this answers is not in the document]
Ezio-Sarthak: Thanks! I Will remember that from now :sweat_smile:

zulipterminal/model.py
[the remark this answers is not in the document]
Ezio-Sarthak: Thanks for the clarification!

zulipterminal/model.py
Ezio-Sarthak: I guess they used it just to symbolize that the data of a user inactive for more than 2 weeks ago is undefined.

[This](https://chat.zulip.org/#narrow/stream/9-issues/topic/Last.20Online.3A.202.20Weeks.20Ago/near/727386) message on czo might be useful

neiljp: Going as per the webapp seems good - and referencing that discussion or the issue may be relevant too.

Ezio-Sarthak:
Thanks for all the suggestions! Sorry for the delay :sweat_smile: 

I've skipped adding tests for the fetching methods in Model and class for this iteration so that once we put things in the right place, we can add them after that if that makes sense :)

PS - I found adding tests for the model method I modified in this iteration to be a bit confusing. I'll appreciate any validation tips :sweat_smile:

Ezio-Sarthak:

Ezio-Sarthak:

tests/model/test_model.py
Ezio-Sarthak: Yes

zulipterminal/core.py
```diff
@@ -236,6 +237,12 @@ def show_about(self) -> None:
             'area:help'
         )
 
+    def show_user_info(self, user_id: int) -> None:
+        self.show_pop_up(
+            UserInfoView(self, user_id, 'User Information (up/down scrolls)'),
+            'area:help'
```
neiljp, line 243: Let's definitely add a new area for this - if nothing else, this is confusing to read.

zulipterminal/ui_tools/buttons.py
```diff
@@ -244,6 +245,12 @@ def _narrow_with_compose(self, button: Any) -> None:
             recipient_user_ids=[self.user_id]
         )
 
+    def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
+        # Shows user info on pressing 's'
```
neiljp, line 249: This comment seems wrong - and likely unnecessary?

zulipterminal/ui_tools/views.py
```diff
@@ -1034,6 +1038,54 @@ def __init__(self, controller: Any, title: str, *, zt_version: str,
...
+        # Convert the data to presentation form
+        self.display_data['Name'] = data['full_name']
+        if data['email']:
+            self.display_data['Email'] = data['email']
+        if data['date_joined']:
+            self.display_data['Date joined'] = data['date_joined'][:10]
+        if data['timezone']:
+            self.display_data['Timezone'] = data['timezone']
+            self.display_data['Local time'] = datetime.now(
+                pytz.timezone(str(data['timezone']))).strftime("%H:%M")
```
neiljp, line 1075: This seems related to the data processing (so model-focused?), and also should likely take into account the local preferences for 24h vs am/pm - though the latter need not be centralized as long as we at least add a comment regarding that.

neiljp (commented):
@Ezio-Sarthak This is pretty close and will be useful :+1: (I like that we'll have a standard `i` for each column after this!)

There are a few areas we can tidy (see below), but this is definitely functional.

The areas I'm most concerned about are that I could trigger a crash when I looked at a user (`github test`), and so generally I think we could improve the stability of this new feature with model (and maybe UI) tests.

zulipterminal/model.py
```diff
@@ -610,6 +610,41 @@ def get_other_subscribers_in_stream(self, stream_id: Optional[int]=None,
...
 
+    def get_user_info(self, user_id: int) -> Dict[str, Any]:
+        # TODO: Add custom fields later as an enhancement
+        data = next((i for i in self.initial_data['realm_users']
+                     if i['user_id'] == user_id), Dict[str, Any])
+
+        if not data:
+            # If the user is a cross realm bot (e.g., Welcome Bot)
+            data = next((i for i in self.initial_data['cross_realm_bots']
+                         if i['user_id'] == user_id), Dict[str, Any])
```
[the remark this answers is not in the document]
Ezio-Sarthak: In `self.user_dict`, some parameters like `date_joined`, `timezone`, roles etc. do not exist.
Re the commit text, thanks for the suggestion. I'll update the commit message :sweat_smile:

zulipterminal/ui_tools/buttons.py
Ezio-Sarthak: Oops. That's not required. Thanks for pointing out :+1:

zulipterminal/core.py
Ezio-Sarthak: Sure. I'll propose `area:info` in my coming iteration

Ezio-Sarthak:
@neiljp Thanks for the valuable comments :) I overlooked some important aspects, and I hope most of them are addressed in this iteration :sweat_smile:.

Re the traceback, I've fixed it in this iteration by a quick debug. Re model + UI testing, I'll definitely add/amend some tests in the coming iteration :+1:

zulipterminal/model.py
```diff
@@ -610,6 +611,54 @@ def get_other_subscribers_in_stream(self, stream_id: Optional[int]=None,
...
+            timezone=data.get('timezone', ''),
+            local_time='',
+            is_bot=data.get('is_bot', False),
+            role=[
+                ('is_owner', data.get('is_owner',   False)),
+                ('is_admin', data.get('is_admin',   False)),
+                ('is_guest', data.get('is_guest',   False)),
+            ],
+            bot_type=data.get('bot_type', -1),
+            bot_owner_id=data.get('bot_owner_id', -1),
```
neiljp, line 637: Does `None` have a meaning already here? If not, this may be clearer?

zulipterminal/core.py
neiljp: We already have areas for more specific names, so I'd suggest `area:user` may be more appropriate?

zulipterminal/config/themes.py
```diff
@@ -276,6 +279,8 @@
          None,             BLACK,             LIGHTBLUE),
         ('area:error',     'white',           'dark red',
          None,             WHITE,             DARKRED),
+        ('area:info',      'white',           'dark blue',
+         None,             DEF['white'],      DEF['dark_blue']),
```
neiljp, line 283: This is gruvbox, not DEFault color scheme.

tests/ui_tools/test_popups.py
```diff
@@ -185,6 +186,55 @@ def test_feature_level_content(self, mocker, zulip_version):
...
+            last_active='Active now',
+        )
+        mocker.patch.object(self.controller.model, 'get_user_info',
+                            return_value=user_data)
+
+        self.user_info_view = UserInfoView(self.controller, 10000,
+                                           'User Info (up/down scrolls)')
+
+    @pytest.mark.parametrize('key', {*keys_for_command('GO_BACK'),
+                                     *keys_for_command('USER_INFO')})
```
neiljp, line 217: I'd move the USER_INFO addition closer to this commit; we might squash the commits together later, but this is the first commit where it gets used in any case.

zulipterminal/model.py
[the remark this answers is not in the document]
Ezio-Sarthak: Re the annotation, I'll update it in the coming update. Thanks for the heads up :)

Re the last point, can we improve the `user_name_from_id` to search in `realm_users` too. I'm not sure how runtime-friendly it would be, however.

zulipterminal/model.py
```diff
@@ -702,6 +703,55 @@ def get_other_subscribers_in_stream(self, stream_id: Optional[int]=None,
...
+            full_name=data.get('full_name', '(No name)'),
+            email=data.get('email', ''),
+            date_joined=data.get('date_joined', ''),
+            timezone=data.get('timezone', ''),
+            local_time='',
+            is_bot=data.get('is_bot', False),
+            role=[
+                ('is_owner', data.get('is_owner',   False)),
+                ('is_admin', data.get('is_admin',   False)),
+                ('is_guest', data.get('is_guest',   False)),
```
neiljp, line 723: This PR has hung around just long enough that we could now add `is_moderator`?

zulipterminal/model.py
```diff
@@ -702,6 +703,55 @@ def get_other_subscribers_in_stream(self, stream_id: Optional[int]=None,
                     if stream['name'] == stream_name
                     if sub != self.user_id]
 
+    def get_tidied_user_info(self, user_id: int) -> Dict[str, Any]:
+        data = self._user_info_by_id(user_id)
```
neiljp, line 707: Minor: we could emphasize that this is "raw" data by naming this eg. `api_user_data`, or something shorter?

zulipterminal/model.py
```diff
@@ -702,6 +703,55 @@ def get_other_subscribers_in_stream(self, stream_id: Optional[int]=None,
                     if stream['name'] == stream_name
                     if sub != self.user_id]
 
+    def get_tidied_user_info(self, user_id: int) -> Dict[str, Any]:
```
neiljp, line 706: I'm not sure if we need "tidied" here, though I appreciate the intent to make it clearer. Potentially using a TypedDict (helper.py perhaps, since it's not an api type) would make it clear to users from the internal method even without an underscore, though the name itself does need to be distinct at runtime of course.

zulipterminal/ui_tools/views.py
```diff
@@ -1041,6 +1045,66 @@ def __init__(self, controller: Any, title: str, *, zt_version: str,
...
+            display_data['Email'] = data['email']
+        if data['date_joined']:
+            display_data['Date joined'] = data['date_joined'][:10]
+        if data['timezone']:
+            display_data['Timezone'] = data['timezone']
+        if data['local_time']:
+            display_data['Local time'] = data['local_time'][11:]
+
+        if data['is_bot']:
+            display_data['Role'] = BOT_TYPE_BY_ID[data['bot_type']]
```
neiljp, line 1093: To be forward-compatible, we likely want to make this into a `.get` with some kind of 'unknown' text?

Possibly the same for USER_ROLE in terms of the default value, as eg. 'moderator' would not be identified right now - which I've tested since I now am one :)

tests/ui_tools/test_popups.py
```diff
@@ -187,6 +188,131 @@ def test_feature_level_content(self, mocker, zulip_version):
...
 
 
+class TestUserInfoView:
+    @pytest.fixture(autouse=True)
+    def mock_external_classes(self, mocker):
+        self.controller = mocker.Mock()
+        mocker.patch.object(self.controller, 'maximum_popup_dimensions',
+                            return_value=(64, 64))
+        mocker.patch(VIEWS + '.urwid.SimpleFocusListWalker', return_value=[])
+        self.user_data = dict(
```
neiljp, line 198: This is essentially a fixture for the output of `get_tidied_user_info`?

I understand for getting this going that a dict internal to the test is easier, but it would be good if we could unify a "sample output" from that method, ie. relate it to our current standard test users in some way. If we change the output from the model method then we need to change the internals of this test otherwise - or eg. handling adding is_moderator.

tests/ui_tools/test_popups.py
```diff
@@ -187,6 +188,131 @@ def test_feature_level_content(self, mocker, zulip_version):
...
+    def test__fetch_user_data(self, mocker, initial_key_val, expected_key_val):
+        data = self.user_data
+
+        if initial_key_val:
+            i_key, i_val = initial_key_val
+            data[i_key] = i_val
+            if i_key == 'bot_type':
+                data['is_bot'] = True
+                data['bot_owner'] = 'Test Owner'
+        e_key, e_val = expected_key_val
```
neiljp, line 277: If we used something like the `dict(fixture_copy, **to_vary_in_fixture)` approach as used in other tests, would this be clearer? (also your other similar tests)
(the bot parts can be part of the to_vary element?)

neiljp (commented):
@Ezio-Sarthak This looks tidier than I remember :+1:

We're essentially making new data available here, rather than just plumbing through to existing prepared data, so don't be too concerned this is taking a bit longer than other PRs. In particular the project has ended up with a lot of duplicate test data in the past which became quite challenging to update, so I'm keen to keep this as clean as we can, within reason :)

zulipterminal/model.py
```diff
@@ -695,6 +695,14 @@ def get_other_subscribers_in_stream(self, stream_id: Optional[int]=None,
                     if stream['name'] == stream_name
                     if sub != self.user_id]
 
+    def _user_info_by_id(self, user_id: int) -> Optional[Dict[str, Any]]:
+        data = self.all_users_by_id.get(user_id, None)
+        if not data:
+            # If the user is a cross realm bot (e.g., Welcome Bot)
+            data = self.cross_realm_bots_by_id.get(user_id, None)
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yes. I think I could remove this commit now that I could access all users via a single `get` command. What do you think?

zulipterminal/model.py
```diff
@@ -132,6 +132,8 @@ def __init__(self, controller: Any) -> None:
         # Register to the queue before initializing further so that we don't
         # lose any updates while messages are being fetched.
         self._fetch_initial_data()
+        self.all_users_by_id: Dict[int, Dict[str, Any]] = {}
+        self.cross_realm_bots_by_id: Dict[int, Dict[str, Any]] = {}
```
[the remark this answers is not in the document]
Ezio-Sarthak: Re the 2nd point, I like this idea. I reckon the API responses for `realm_users` and `cross_realm_bots` are mostly the same, except maybe the `bot_type`, `bot_owner_id` params.

However, I'm getting a weird mypy error while using `get` to check if (say) `bot_type` exists.
It says: `error: TypedDict "RealmUser" has no key 'bot_type'`. (Here RealmUser, CrossRealmBot, and UserRealmData are the TypedDicts for realm_users, cross_realm_bots, and all users (Union) respectively)

I presume this is [python/mypy#7151](https://github.com/python/mypy/issues/7151).

That said, I wonder if we could keep it a single TypedDict for now?

zulipterminal/model.py
Ezio-Sarthak: `api_user_data` sgtm :+1:

zulipterminal/ui_tools/views.py
Ezio-Sarthak: Yeah, that makes sense. I wish to propose that as I'm returning a custom-made `role` param from `get_tidied_user_info`, I could forcefully pass `is_moderator` (with `False` as default) to it, and then handle it easily. I suppose it would make sense with regards to backward compatibility?

Noting in my above proposal, if `is_moderator` is false, it wouldn't affect the default `Member` value as well, due to the internal ordering in `role` :)

zulipterminal/ui_tools/views.py
```diff
@@ -1044,6 +1046,64 @@ def __init__(self, controller: Any, title: str, *, zt_version: str,
...
+        display_data = {'Name': data['full_name']}
+
+        if data['email']:
+            display_data['Email'] = data['email']
+        if data['date_joined']:
+            display_data['Date joined'] = data['date_joined'][:10]
+        if data['timezone']:
+            display_data['Timezone'] = data['timezone']
+        if data['local_time']:
+            display_data['Local time'] = data['local_time'][11:]
```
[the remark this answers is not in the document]
Ezio-Sarthak: Agreed. I understand the timezone strings might be unaltered due to internal calculations, and the server would expect to let the client-side handle it. This should be straightforward with `replace` in our case. Could you elaborate (either here or in stream) exactly what similarity you observed regarding the buggy underscore behavior with webapp ?:sweat_smile:

zulipterminal/model.py
```diff
@@ -671,6 +672,71 @@ def get_other_subscribers_in_stream(self, stream_id: Optional[int]=None,
...
+            ],
+            bot_type=data.get('bot_type', 0),
+            bot_owner='',  # Non-empty only for valid bot types
+            last_active='',
+        )  # type: Dict[str, Any]
+
+        if user_info['timezone']:
+            # Converting all timestamps to UTC
+            tz = pytz.timezone(user_info['timezone'])
+            time = datetime.datetime.now(tz).replace(tzinfo=None).timestamp()
```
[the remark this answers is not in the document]
Ezio-Sarthak: Right. I'll consider this and your other comment mentioning `now()` declaration in UI.

zulipterminal/api_types.py
```diff
@@ -85,6 +85,24 @@ class Subscription(TypedDict):
     stream_weekly_traffic: Optional[int]
 
 
+class RealmUserData(TypedDict):
```
Ezio-Sarthak, line 88: This is an API response for a general realm user. I tried separating parameters unique to cross-realm bots, but found almost all parameters overlapping to the general API response. Let me know if we could further distinguish this TypedDict.

neiljp (commented):
@Ezio-Sarthak I'm keen to merge this, so after having merged (split) the first commit with a slight adjustment as we discussed, I've left a few comments and I'd like to get this in ASAP.

After reviewing the data we get from the server today, I think we're not quite handling the role correctly, and there's also an older-style bot_owner field to consider.

I think we need to integrate the user-id code better, but it's in now for use in other PRs.

tests/model/test_model.py
```diff
@@ -952,6 +952,78 @@ def test__group_info_from_realm_user_groups(self, model, user_groups_fixture):
...
+            ),
+            ({}, "date_joined", ""),
+            ({"timezone": "Asia/Kolkata"}, "timezone", "Asia/Kolkata"),
+            ({}, "timezone", ""),
+            ({"bot_type": 1}, "bot_type", 1),
+            ({}, "bot_type", 0),
+            ({"is_bot": True}, "is_bot", True),
+            ({"is_owner": True}, "role", 100),
+            ({"is_admin": True}, "role", 200),
+            ({"is_moderator": True}, "role", 300),
```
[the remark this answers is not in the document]
Ezio-Sarthak: Are you suggesting to just re-order the tests, or remove these ones?

tests/model/test_model.py
```diff
@@ -952,6 +952,78 @@ def test__group_info_from_realm_user_groups(self, model, user_groups_fixture):
...
+            ({"is_guest": True}, "role", 600),
+            ({"role": 100}, "role", 100),
+            ({"role": 200}, "role", 200),
+            ({"role": 300}, "role", 300),
+            ({"role": 600}, "role", 600),
+            ({}, "role", 400),
```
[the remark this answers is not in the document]
Ezio-Sarthak: Could you elaborate on this? I'm trying to basically unify both responses to a single role id one.

neiljp:
@Ezio-Sarthak Thanks for the fixture work; I squashed those fixup commits from yesterday and also added a tiny test to use that new fixture early on, making a few tiny commit text changes, resulting in what's on the PR now - which I'm merging now! :tada:

We should really test those new model attributes we added, though I was ~ok with merging that commit early, there's nothing validating them right now.

Issue #684, opened 2020-06-12
Enhance the general popup styling


preetmishra:
We could improve the general popup styling to make popups stand out a bit more from the rest of the view.

This may involve a bit of experimentation, exploration and discussion with the team. Feel free to drop a message in [#zulip-terminal > Popup Styling](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Popup.20Styling) about your design proposal. The related conversation should be in the mentioned topic.

The relevant code should be in `zulipterminal/config/themes.py` and `zulipterminal/ui_tools/views.py`.

neiljp:
Anyone interested in this is also welcome to drop here some screenshots showing alternative designs which we can discuss and give feedback on.

vinitwadgaonkar:
i am intrested

neiljp:
Fixed by #853 :tada: Thanks @Ezio-Sarthak :+1:

PR #849, opened 2020-12-11
[WIP] enhancement: General popup styling

Ezio-Sarthak:
This PR aims the following:
- Border styling of popups now tries to blend with the background.
- Popup titles are also now styled.
- Concerning issue #684.
- Open for further discussion. :)

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -43,8 +43,22 @@ required_styles = {  # style-name: monochrome-bit-depth-style
     'edit_tag': 'standout',
     'edit_author': 'bold',
     'edit_time': 'bold',
+    'popup_border': 'standout',
+    'popup_title': 'bold',
 }
 
+# Pop-up border styles for different popup views
+popup_borders = {
+    'help': dict(tlcorner='▛', tline='▀', trcorner='▜', rline='▐',
+                 lline='▌', blcorner='▙', bline='▄', brcorner='▟'),
+    'info': dict(tlcorner='╔', tline='═', trcorner='╗', rline='║',
+                 lline='║', blcorner='╚', bline='═', brcorner='╝'),
+    'stream_info': dict(tlcorner='▘', tline='▀', trcorner='▝', rline='▐',
+                        lline='▌', blcorner='▖', bline='▄', brcorner='▗'),
+    # TODO: A different border styling may be specified
+    'about': dict(tlcorner='▘', tline='▀', trcorner='▝', rline='▐',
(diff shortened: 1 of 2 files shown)


neiljp (commented):
@ezio-sarthak Thanks for working on this :+1: This WIP explores and demonstrates lots of options we could take!

Design is a bit personal, so I'm definitely open to discussion on this, but I think it's important that we look at the range of changes you've explored here - this is my own take on them:
* I really like the clean edge of the help menu, though it doesn't resolve the current situation that the title doesn't look part of the menu - maybe the top border (or "title text" only?) could be a thinner box edge, with the actual text in a content row below it? Essentially this would be like having two borders, ie. an extra line only at the top.
* I appreciate the exploration of different box-shape edges, like the one with the corners sticking out (log cabin corners?), but I definitely think that exploring one at a time would be simpler, and having seen the ones you have here I'd suggest trying the help-menu style only first.
* For the semantic styling, I was thinking something more like colors (ie. urwid styles/palette-entries) rather than box-shapes; also I don't think there's much utility in using specific popup names as semantic keys which we have to keep extending in the lookup dict - I was thinking more along the lines of categories, perhaps message-related, stream-related, error/warning... just some initial ideas there.
* It would be good to separate the border change as above from the color change in the body, ideally in a separate commit. I'm not sure about changing the central popup colors generally, as with a dark theme I'd probably want a mostly-dark popup. The contrast between the popup and general UI here is good, but doesn't necessarily fit the themes. When separating the colors, you could add internal blocking/padding of border color -> internal color, like a sort of inverse of the help menu border, though we could explore that as a separate commit to see what it might look like separately.

As you can tell from the above, I'm personally most interested in moving from the double-line border to your help-menu style border (for all popups), initially retaining the previous internal styles (though possibly with that inverse padding), while also incorporating the title properly "inside" the popup by having some kind of line above it too. **[*]**

I think this is a great exploratory PR demonstrating what is possible, and it'd be useful for others to be able to see these designs in future. On that basis if you'd like to follow through with work on **[*]** above, I'd suggest you do it in another PR so that we don't lose the ability to see these explored options here, while you can work on specific smaller  improvements to consider individually.

Ezio-Sarthak:
Thanks for having look at the designs :)
I'll surely clean it up by opting only help-menu scheme for now

Ezio-Sarthak:
Re Cleaned up this PR by squashing similar commits and fixing linting errors :)

neiljp:
@Ezio-Sarthak I'm closing this now, as we've merged most of the main ideas, and one of the commits links back to this in any case.

PR #851, opened 2020-12-15
streams: Handle adding/removing stream events.

### LINKED ISSUE #816: Adding new stream throws an Keyerror
Whenever an admin adds an extra stream in the server , an error is thrown if a message in received the newly added stream as : 
<a href = 'https://user-images.githubusercontent.com/53977614/97470053-15a1ff80-196d-11eb-9f21-c6a483bf4b2f.png'>Screenshot</a>

<strong>Traceback :</strong>
<pre>
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
  File "/home/aryanshridhar/Desktop/zulip-terminal-env/zulip-terminal/zulipterminal/model.py", line 1194, in poll_for_events
    self.event_actions[event['type']](event)
  File "/home/aryanshridhar/Desktop/zulip-terminal-env/zulip-terminal/zulipterminal/model.py", line 918, in _handle_message_event
    set_count([message['id']], self.controller, 1)
  File "/home/aryanshridhar/Desktop/zulip-terminal-env/zulip-terminal/zulipterminal/helper.py", line 249, in set_count
    _set_count_in_view(controller, new_count, changed_messages, unread_counts)
  File "/home/aryanshridhar/Desktop/zulip-terminal-env/zulip-terminal/zulipterminal/helper.py", line 214, in _set_count_in_view
    if controller.model.is_muted_topic(stream_id, msg_topic):
  File "/home/aryanshridhar/Desktop/zulip-terminal-env/zulip-terminal/zulipterminal/model.py", line 474, in is_muted_topic
KeyError: 263064

</pre>

<strong>To reproduce :</strong>

<ul>
<li>Create a new stream in the server .</li>
<li>Add (<code>Reply/New topic</code>) an message in the newly created stream .</li>
<li>Checkout the zulip terminal (which was active throughout the process) .</li>
</ul>

Abhirup-99:
Fixes #816

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -7,7 +7,7 @@ import pytest
 from zulip import ZulipError
 
 from zulipterminal.helper import initial_index, powerset
-from zulipterminal.model import Model, ServerConnectionFailure
+from zulipterminal.model import Model, ServerConnectionFailure, sort_streams
 
 
 class TestModel:
@@ -46,14 +46,33 @@ class TestModel:
         model = Model(self.controller)
         return model
 
+    @pytest.fixture
+    def initial_pinned_streams(self, model, stream_dict):
+        pinned_streams = []
+        for _, stream in stream_dict.items():
+            if stream['pin_to_top']:
+                pinned_streams.append(model.make_reduced_stream_data(stream))
+        return pinned_streams
+
+    @pytest.fixture
+    def initial_unpinned_streams(self, model, stream_dict):
+        pinned_streams = []
+        for _, stream in stream_dict.items():
+            if not stream['pin_to_top']:
+                pinned_streams.append(model.make_reduced_stream_data(stream))
+        return pinned_streams
+
     def test_init(self, model, initial_data, user_profile,
-                  unicode_emojis, custom_emojis, stream_dict):
+                  unicode_emojis, custom_emojis, stream_dict,
+                  initial_pinned_streams, initial_unpinned_streams):
         assert hasattr(model, 'controller')
         assert hasattr(model, 'client')
         assert model.narrow == []
         assert model._have_last_message == {}
         assert model.stream_id is None
         assert model.stream_dict == stream_dict
+        assert model.pinned_streams == initial_pinned_streams
+        assert model.unpinned_streams == initial_unpinned_streams
         assert model.subscribed_streams == {1, 2, 99, 1000}
         assert model.recipients == frozenset()
         assert model.index == initial_index
@@ -778,7 +797,7 @@ class TestModel:
         assert all(msg_id == msg['stream_id']
                    for msg_id, msg in model.stream_dict.items())
         assert model.muted_streams == muted
-        assert model.pinned_streams == []  # FIXME generalize/parametrize
+        assert len(model.pinned_streams)  # FIXME generalize/parametrize
         assert len(model.unpinned_streams)  # FIXME generalize/parametrize
 
     def test__handle_message_event_with_Falsey_log(self, mocker,
@@ -2089,6 +2108,142 @@ class TestModel:
         fetched_custom_emojis = model.fetch_custom_emojis()
 
         assert fetched_custom_emojis == custom_emojis
+
+    @pytest.mark.parametrize('event', [
+        {
+            'op': 'add',
+            'type': 'subscription',
+            'subscriptions': [
+                {
+                    'name': 'all',
+                    'stream_id': 1,
+                    'description': '',
+                    'color': '#95a5fd',
+                    'pin_to_top': False,
+                    'invite_only': False,
+                    'in_home_view': True
+                }
+            ]
+        },
+        {
+            'op': 'add',
+            'type': 'subscription',
+            'subscriptions': [
+                {
+                    'name': 'design',
+                    'stream_id': 2,
+                    'description': '',
+                    'color': '#95a5fd',
+                    'pin_to_top': True,
+                    'invite_only': False,
+                    'in_home_view': True
+                }
+            ]
+        },
+        {
+            'op': 'add',
+            'type': 'subscription',
+            'subscriptions': [
+                {
+                    'name': 'all',
+                    'stream_id': 3,
+                    'description': '',
+                    'color': '#95a5fd',
+                    'pin_to_top': True,
+                    'invite_only': False,
+                    'in_home_view': False
+                }
+            ]
+        },
+    ], ids=[
+        'add_1',
+        'add_and_pin_2',
+        'add_and_mute_3'
+    ])
+    def test__handle_subscription_event_add_stream(
+                                    self, model, mocker,
+                                    event,
+                                    initial_pinned_streams,
+                                    initial_unpinned_streams
+    ):
+
+        model.pinned_streams = deepcopy(initial_pinned_streams)
+        model.unpinned_streams = deepcopy(initial_unpinned_streams)
+        model._handle_subscription_event(event)
+        is_stream_pinned = event['subscriptions'][0]['pin_to_top']
+        if is_stream_pinned:
+            new_pinned_stream = (model
+                                 .make_reduced_stream_data(
+                                  event['subscriptions'][0]))
+            expected_pinned_streams = (initial_pinned_streams
+                                       + [new_pinned_stream])
+            expected_unpinned_streams = initial_unpinned_streams
+            sort_streams(expected_pinned_streams)
+        else:
+            new_unpinned_stream = (model
+                                   .make_reduced_stream_data(
+                                    event['subscriptions'][0]))
+            expected_unpinned_streams = (initial_unpinned_streams
+                                         + [new_unpinned_stream])
+            expected_pinned_streams = initial_pinned_streams
+            sort_streams(expected_unpinned_streams)
+
+        assert model.pinned_streams == expected_pinned_streams
+        assert model.unpinned_streams == expected_unpinned_streams
+
+        (model.controller.view.left_panel.update_stream_view
+         .assert_called_once_with())
+        model.controller.update_screen.assert_called_once_with()
+
+    @pytest.mark.parametrize('event', [
+        {
+            'op': 'remove',
+            'type': 'subscription',
+            'subscriptions': [{
+                'name': 'Secret stream',
+                'stream_id': 99
+            }]
+        },
+        {
+            'op': 'remove',
+            'type': 'subscription',
+            'subscriptions': [{
+                'name': 'Some general stream',
+                'stream_id': 1000
+            }]
+        },
+    ], ids=[
+        'remove_99',
+        'remove_1000'
+    ])
+    def test__handle_subscription_event_remove_stream(
+                                    self, model, mocker,
+                                    event,
+                                    initial_pinned_streams,
+                                    initial_unpinned_streams,
+                                    stream_dict
+    ):
+        model.pinned_streams = deepcopy(initial_pinned_streams)
+        model.unpinned_streams = deepcopy(initial_unpinned_streams)
+        model.stream_dict = stream_dict
+        stream_id = event['subscriptions'][0]['stream_id']
+        if model.is_pinned_stream(stream_id):
+            stream = model._get_stream_by_id(initial_pinned_streams, stream_id)
+            initial_pinned_streams.remove(stream)
+        else:
+            stream = (model
+                      ._get_stream_by_id(initial_unpinned_streams, stream_id))
+            initial_unpinned_streams.remove(stream)
+        model._handle_subscription_event(event)
+        expected_pinned_streams = initial_pinned_streams
+        expected_unpinned_streams = initial_unpinned_streams
+
+        assert model.pinned_streams == expected_pinned_streams
+        assert model.unpinned_streams == expected_unpinned_streams
+
+        update_left_panel = model.controller.view.left_panel.update_stream_view
+        update_left_panel.assert_called_once_with()
+        model.controller.update_screen.assert_called_once_with()
 
     # Use LoopEnder with raising_event to cause the event loop to end without
     # processing the event
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -54,6 +54,7 @@ class TestModel:
         assert model._have_last_message == {}
         assert model.stream_id is None
         assert model.stream_dict == stream_dict
+        assert model.subscribed_streams == {1, 2, 99, 1000}
         assert model.recipients == frozenset()
         assert model.index == initial_index
         model.get_messages.assert_called_once_with(num_before=30,
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -2021,6 +2021,30 @@ class TestModel:
 
         assert return_value == expected_response
 
+    @pytest.mark.parametrize('stream_id, expected_response,', [
+            (99, False),
+            (101, True),
+            (2, False),
+        ],
+        ids=[
+            'subscribed_pinned_stream',
+            'subscribed_unpinned_stream',
+            'unsubscribed_stream',
+        ]
+    )
+    def test_is_user_in_unsubscribed_stream(self, model,
+                                            stream_id, expected_response,
+                                            initial_pinned_streams,
+                                            initial_unpinned_streams):
+
+        model.pinned_streams = deepcopy(initial_pinned_streams)
+        model.unpinned_streams = deepcopy(initial_unpinned_streams)
+        return_value = model.is_user_in_unsubscribed_stream(stream_id)
+
+        assert model.pinned_streams == initial_pinned_streams
+        assert model.unpinned_streams == initial_unpinned_streams
+        assert return_value == expected_response
+
     @pytest.mark.parametrize('response', [{
         'result': 'success',
         'msg': '',
--- a/zulipterminal/api_types.py
+++ b/zulipterminal/api_types.py
@@ -121,6 +121,7 @@ class SubscriptionEvent(TypedDict):
 
     value: bool
     message_ids: List[int]  # Present when subject of msg(s) is updated
+    subscriptions: List[Subscription]  # Present when stream is added/removed
 
 
 class TypingEvent(TypedDict):
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -102,6 +102,14 @@ class Controller:
             self._critical_exception = True
         os.write(self._exception_pipe, b'1')
 
+    def check_for_invalid_operation(self, stream_id: int) -> bool:
+        if self.model.is_user_in_unsubscribed_stream(stream_id):
+            self.model.controller.view.set_footer_text(
+                " This is not allowed in unsubscribed streams.",
+                3)
+            return True
+        return False
+
     def is_in_editor_mode(self) -> bool:
         return self._editor is not None
 
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -824,6 +824,9 @@ class Model:
     def is_user_subscribed_to_stream(self, stream_id: int) -> bool:
         return stream_id in self.stream_dict
 
+    def is_user_in_unsubscribed_stream(self, stream_id: int) -> bool:
+        return not(stream_id in self.subscribed_streams)
+
     def _get_stream_by_id(self, streams: List[StreamData], stream_id: int
                           ) -> StreamData:
         for stream in streams:
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -831,12 +831,36 @@ class Model:
                 return stream
         raise RuntimeError("Invalid stream id.")
 
+    def _unsubscribe_to_streams(self,
+                                stream_info_list: List[Subscription]) -> None:
+        is_pinned_stream_removed = False
+        is_unpinned_stream_removed = False
+        for stream_info in stream_info_list:
+            stream_id = stream_info['stream_id']
+            if self.is_pinned_stream(stream_id):
+                stream = self._get_stream_by_id(self.pinned_streams,
+                                                stream_id)
+                self.pinned_streams.remove(stream)
+                is_pinned_stream_removed = True
+            else:
+                stream = self._get_stream_by_id(self.unpinned_streams,
+                                                stream_id)
+                self.unpinned_streams.remove(stream)
+                is_unpinned_stream_removed = True
+            if stream_id in self.muted_streams:
+                self.muted_streams.remove(stream_id)
+            self.subscribed_streams.remove(stream_id)
+
     def _handle_subscription_event(self, event: Event) -> None:
         """
         Handle changes in subscription (eg. muting/unmuting,
                                         pinning/unpinning streams)
         """
         assert event['type'] == "subscription"
+
+        def _update_sidebar() -> None:
+            self.controller.view.left_panel.update_stream_view()
+            self.controller.update_screen()
 
         if event['op'] == 'update':
             if hasattr(self.controller, 'view'):
@@ -880,8 +904,16 @@ class Model:
                             self.unpinned_streams.append(stream)
                     sort_streams(self.unpinned_streams)
                     sort_streams(self.pinned_streams)
-                    self.controller.view.left_panel.update_stream_view()
-                    self.controller.update_screen()
+                    _update_sidebar()
+
+        elif event['op'] == 'add':
+            self._subscribe_to_streams(event['subscriptions'])
+            _update_sidebar()
+
+        elif event['op'] == 'remove':
+            self._unsubscribe_to_streams(event['subscriptions'])
+            _update_sidebar()
+
         elif event['op'] in ('peer_add', 'peer_remove'):
             # NOTE: ZFL 35 commit was not atomic with API change
             #       (ZFL >=35 can use new plural style)
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -737,15 +737,15 @@ class Model:
 
         return self.user_dict[user_email]['full_name']
 
-    def _subscribe_to_streams(self, subscriptions: List[Subscription]) -> None:
-        def make_reduced_stream_data(stream: Subscription) -> StreamData:
-            # stream_id has been changed to id.
-            return StreamData({'name': stream['name'],
-                               'id': stream['stream_id'],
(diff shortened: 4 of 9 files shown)


zulipterminal/model.py
```diff
@@ -785,6 +786,21 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int
...
+                return StreamData({'name': stream['name'],
+                                   'id': stream['stream_id'],
+                                   'color': stream['color'],
+                                   'invite_only': stream['invite_only'],
+                                   'description': stream['description']})
+
+            subscription = event['subscriptions'][0]
+            stream_data = make_reduced_stream_data(subscription)
+            self.stream_dict[subscription['stream_id']] = stream_data
+            self.pinned_streams.append(stream_data)
```
neiljp, line 801: The new stream is not necessarily pinned.

zulipterminal/model.py
```diff
@@ -785,6 +786,21 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int
                     sort_streams(self.pinned_streams)
                     self.controller.view.left_panel.update_stream_view()
                     self.controller.update_screen()
+        elif event['op'] == 'add':
+            def make_reduced_stream_data(stream: Any) -> StreamData:
+                # stream_id has been changed to id.
+                return StreamData({'name': stream['name'],
+                                   'id': stream['stream_id'],
+                                   'color': stream['color'],
```
neiljp, line 794: This may need simplifying to the form we use elsewhere.

neiljp (commented):
@Abhirup-99 Thanks for giving this a go :+1:

There are a few issues with the PR as it stands - how have you tested this? I tried subscribing to a stream in the webapp and the list of streams in ZT doesn't update.

I would recommend checking what the existing code does for subscriptions when they are loaded at startup, ie. any translation of the data, as well as how the stream list is updated and re-rendered in other cases.

Regarding your question in the stream, adding remove events would be good to include in the same PR.

zulipterminal/model.py
Abhirup-99: > @Abhirup-99 Thanks for giving this a go 
> 
> There are a few issues with the PR as it stands - how have you tested this? I tried subscribing to a stream in the webapp and the list of streams in ZT doesn't update.
> 
> I would recommend checking what the existing code does for subscriptions when they are loaded at startup, ie. any translation of the data, as well as how the stream list is updated and re-rendered in other cases.
> 
> Regarding your question in the stream, adding remove events would be good to include in the same PR.

Sorry for the embarrassing mistake, I forgot to push 1 line in the code, I was removing all print statements and additional stuff and might have deleted that as well. I am pushing this shortly along with the tests for this part.

zulipterminal/model.py
Abhirup-99: Can you explain what you mean by simplifying? I added this part from the ` _stream_info_from_subscriptions` method? The value returned was added directly to the stream dictionary after being extracted out from this function.

zulipterminal/model.py
Abhirup-99: > The new stream is not necessarily pinned.
Fixed in the new push.

zulipterminal/model.py
Abhirup-99: I looked through the code, the other parts use this same piece of code. I picked up the pinned and unpinned streams while testing, and they use the same structure as above.
```
[
 {
   'name': 'announce', 
   'id': 1,
   'color': '#e9b', 
   'invite_only': False,
   'description': 'Stuff everyone should see. (Everyone is subscribed to this by default.)
 '},
...
 ]
```

Abhirup-99:
@neiljp 
Sorry, the pr became a bit longer than I had anticipated it to be. This happened because of the addition of the tests, the incoming events from add_stream event is quite large so that primarily contributed to it, also handling of some race around conditions when the user is removed from a stream but is that stream is on view. I disabled some of the keypress events when that condition occurs. Any other thoughts/suggestions would be helpful.

zulipterminal/model.py
```diff
@@ -730,6 +730,15 @@ def toggle_stream_pinned_status(self, stream_id: int) -> bool:
...
         return stream_id in self.stream_dict
 
+    def is_user_view_in_removed_stream(self, stream_id: int) -> bool:
+        for stream in self.pinned_streams:
+            if stream['id'] == stream_id:
+                return False
+        for stream in self.unpinned_streams:
+            if stream['id'] == stream_id:
+                return False
+        return True
```
neiljp, line 740: Note that this should be equivalent to the (simpler) method just above it, as long as the stream_dict is updated appropriately - which we really should do (in addition to the muted stream tracking).

Currently stream_dict is just subscribed streams, though is the original data from the server as opposed to the smaller/simpler data used by the UI (StreamData).

Note that if we did use this method, the name is much more specific than is warranted by the logic - I'd want a simpler cleaner name :)

neiljp (commented):
@Abhirup-99 This looks more detailed than the previous version and I can see that it does add and remove streams from the list :+1:

The issues I raised earlier are still present however, but I hope my comments will make the color/sorting issues clearer re the addition of streams. We should also ensure other data in the model is updated appropriately, eg. at least muted streams and stream_dict (that I noted so far in this review).

Once you understand the above, you might notice some common duplicated code. It would likely be useful to make a refactoring, extracting the current processing of initial subscriptions (including colors, storing muting state) into an internal method that can process additional streams later too - and would centralize the processing of the data. Possibly we could do another which would handle the StreamData updates, including eg. the sorting. That would likely be better as a preparatory commit (or two) retaining the current behavior but generating methods you can use in the add/remove commits.

It's good to see you covering the removal of streams, but I'd suggest we split the UI warnings into a separate commit; this is important work, but the behavior needs clarifying. For example, while we should limit unnecessary server calls if they're going to fail in any case, I'm not entirely sure if posting/editing to an unsubscribed stream is limited - and maybe other similar aspects too.

zulipterminal/model.py
Abhirup-99: The stream_dict can't be updated in case of removal of streams. This causes any other action related to that stream to fail, like editing, or adding a message. I added the UI warnings to counter this in the first place. I have updated the commits and the UI warnings are now part of a separate commit.

zulipterminal/model.py
```diff
@@ -805,6 +814,21 @@ def make_reduced_stream_data(stream: Any) -> StreamData:
...
             self.controller.update_screen()
 
+        elif event['op'] == 'remove':
+            for stream_info in event['subscriptions']:
+                stream_id = stream_info['stream_id']
+                try:
+                    stream = get_stream_by_id(self.pinned_streams,
+                                              stream_id)
+                    self.pinned_streams.remove(stream)
+                except RuntimeError:
```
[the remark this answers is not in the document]
Abhirup-99: I have updated the function name :), but we have to keep it there.

zulipterminal/model.py
```diff
@@ -785,6 +794,17 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int
...
                     self.controller.update_screen()
+        elif event['op'] == 'add':
+            for subscription in event['subscriptions']:
+                stream_data = make_reduced_stream_data(subscription)
+                self.stream_dict[subscription['stream_id']] = subscription
+                if(subscription['pin_to_top']):
+                    self.pinned_streams.append(stream_data)
+                else:
+                    self.unpinned_streams.append(stream_data)
+            self.controller.view.left_panel.update_stream_view()
```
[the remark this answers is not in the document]
Abhirup-99: Fixed now :)

tests/model/test_model.py
```diff
@@ -1735,6 +1735,114 @@ def set_from_list_of_dict(data):
...
+                                        'invite_only': False,
+                                        'description': 'abc'
+                                    }],
+                                    initial_unpinned_streams=[{
+                                        'name': 'test2',
+                                        'id': 101,
+                                        'color': '#48b',
+                                        'invite_only': False,
+                                        'description': 'lets test'
+                                    }]
```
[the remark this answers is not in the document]
Abhirup-99: Created fixtures to address this.

Abhirup-99:

neiljp (commented):
@Abhirup-99 This still works well and the data is updated much more cleanly for the most part :+1:

Event handling can often expose challenges with the data model, and this PR shows how we can improve it further. So, we could tidy up what you have a little (updating for at least the muted removal and stream_dict vs StreamData issue), or use some of the new code you have as a basis for quite some refactoring where we combine the generation of the model data from initial_data with how it is updated in these events. Let me know how you want to proceed.

On the basis of the latter being the case, I wanted to get some comments back to you, and have focused mainly on the first commit. Other than one function, you've got some new functions in that commit, but also some that are duplicated from elsewhere, I think? To avoid that duplication it would be great to see commits where when you add those methods for use in the new code, that you remove them from the other place(s) and call the new method. In such 'refactors' we move code around (or otherwise adjust it) so that it can be adapted or reused more easily in a later commit - and when tests don't break after a refactor commit, we hopefully know things are still good too!

zulipterminal/model.py
```diff
@@ -729,17 +730,55 @@ def toggle_stream_pinned_status(self, stream_id: int) -> bool:
     def is_user_subscribed_to_stream(self, stream_id: int) -> bool:
         return stream_id in self.stream_dict
 
+    def make_reduced_stream_data(self, stream: Any) -> StreamData:
+        # stream_id has been changed to id.
+        return StreamData({'name': stream['name'],
+                           'id': stream['stream_id'],
+                           'color': stream['color'],
+                           'invite_only': stream['invite_only'],
+                           'description': stream['description']})
```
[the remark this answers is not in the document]
Abhirup-99: This is not a duplication, I moved it outside the method where it was previously used and added it in the class level functions so that all other methods can access it.

zulipterminal/model.py
```diff
@@ -729,17 +730,55 @@ def toggle_stream_pinned_status(self, stream_id: int) -> bool:
...
+                return stream
+        raise RuntimeError("Invalid stream id.")
+
+    def update_sidebar(self) -> None:
+        self.controller.view.left_panel.update_stream_view()
+        self.controller.update_screen()
+
+    def add_stream(self, stream_info: Any) -> None:
+        stream_data = self.make_reduced_stream_data(stream_info)
+        self.stream_dict[stream_info['stream_id']] = stream_info
```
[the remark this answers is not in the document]
Abhirup-99: We can't use an assert over here, considering the scenarios when the user has been unsubscribed and then subscribed to a stream from web UI. This would cause the assert to fail.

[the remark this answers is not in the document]
neiljp: I agree, based on the point you made in the stream.

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: Based on this commit, the other code is not removed, so it's a copy (duplication), not a move.

neiljp:
@Abhirup-99 You replied to some points, but do you want to do some big refactoring first, or go for a simpler version? I know what I'd like to do for the refactoring, but we could get this fixed first.

Abhirup-99:
@zulipbot add "pr needs Review"

Abhirup-99:

Abhirup-99:

zulipterminal/ui_tools/views.py
```diff
@@ -558,13 +558,22 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             return key
 
         elif is_command_key('STREAM_MESSAGE', key):
+            if(self.model.is_user_in_unsubscribed_stream(
+               self.model.stream_id)):
+                self.model.controller.view.set_footer_text(
+                        " You can't write messages from unsubscribed streams.",
+                        3)
```
neiljp, line 565: This seems like a sufficiently duplicated check/string to have a general message of the type as indicated in another comment? Or even a helper method in the controller.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1412,13 +1418,21 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
-            self.model.controller.view.middle_column.set_focus('footer')
+            stream_id = cast(int, self.stream_id)
+            if(not self.model.is_user_in_unsubscribed_stream(stream_id)):
+                self.keypress(size, 'enter')
+                mention = f"@**{self.message['sender_full_name']}** "
+                controller_write_box = self.model.controller.view.write_box
+                controller_write_box.msg_write_box.set_edit_text(
+                    mention)
+                controller_write_box.msg_write_box.set_edit_pos(
+                    len(mention))
```
neiljp, line 1429: This refactoring is different from the conditional you're adding; when you're making so many changes, keep them focused on the key intention of the commit.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1412,13 +1418,21 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
             )
         elif is_command_key('MENTION_REPLY', key):
-            self.keypress(size, 'enter')
-            mention = f"@**{self.message['sender_full_name']}** "
-            self.model.controller.view.write_box.msg_write_box.set_edit_text(
-                mention)
-            self.model.controller.view.write_box.msg_write_box.set_edit_pos(
-                len(mention))
-            self.model.controller.view.middle_column.set_focus('footer')
+            stream_id = cast(int, self.stream_id)
```
neiljp, line 1421: This doesn't seem right. The value you're casting to an int can be None, and the method you're passing it to happens to pass it to another method which treats None OK - but that's not guaranteed. Types are supposed to help us (via eg. mypy), but they can't if we do this.

neiljp (commented):
@Abhirup-99 This is looking good :+1: I almost overlooked the intent of the first commit since you'd not renamed the method before reusing it in the penultimate commit; if you can include the intent (why) in the commit text then this may have been clearer.

Writing larger PRs can be complex - this is looking promising :)

tests/model/test_model.py
```diff
@@ -2039,6 +2059,223 @@ def test_fetch_custom_emojis(self, mocker, model, custom_emojis,
...
+            [{
+                'name': 'test2',
+                'id': 101,
+                'color': '#9af',
+                'invite_only': False,
+                'description': 'lets test'
+             }]
+        )], ids=[
+            'add_2',
+            'add_and_pin_4',
```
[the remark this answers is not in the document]
Abhirup-99: I have them ids over there. I have updated them to be serialized.

zulipterminal/model.py
```diff
@@ -784,11 +784,40 @@ def toggle_stream_pinned_status(self, stream_id: int) -> bool:
     def is_user_subscribed_to_stream(self, stream_id: int) -> bool:
         return stream_id in self.stream_dict
 
+    def _update_sidebar(self) -> None:
```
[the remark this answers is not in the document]
Abhirup-99: Moved inside.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1368,6 +1368,12 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                     stream_id=self.stream_id,
                 )
         elif is_command_key('STREAM_MESSAGE', key):
+            stream_id = cast(int, self.stream_id)
+            if(not self.model.is_user_in_unsubscribed_stream(stream_id)):
+                self.model.controller.view.set_footer_text(
+                        " You can't write messages to unsubscribed streams.",
```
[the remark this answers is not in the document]
Abhirup-99: Writing after unsubscribing is not allowed in webapp.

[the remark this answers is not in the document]
Abhirup-99: screenshot-chat zulip org-2021 02 18-21_08_10

Abhirup-99:

Abhirup-99:

zulipterminal/model.py
```diff
@@ -821,12 +821,41 @@ def _get_stream_by_id(self, streams: List[StreamData], stream_id: int
...
+                self.pinned_streams.remove(stream)
+                is_pinned_stream_removed = True
+            else:
+                stream = self._get_stream_by_id(self.unpinned_streams,
+                                                stream_id)
+                self.unpinned_streams.remove(stream)
+                is_unpinned_stream_removed = True
+                sort_streams(self.unpinned_streams)
+            if not self.stream_dict[stream_id]:
+                self.muted_streams.remove(stream_id)
```
neiljp, line 846: Could you explain what this is intended to do? This won't avoid a `KeyError`.

zulipterminal/model.py
```diff
@@ -821,12 +821,41 @@ def _get_stream_by_id(self, streams: List[StreamData], stream_id: int
...
+                stream = self._get_stream_by_id(self.pinned_streams,
+                                                stream_id)
+                self.pinned_streams.remove(stream)
+                is_pinned_stream_removed = True
+            else:
+                stream = self._get_stream_by_id(self.unpinned_streams,
+                                                stream_id)
+                self.unpinned_streams.remove(stream)
+                is_unpinned_stream_removed = True
+                sort_streams(self.unpinned_streams)
```
neiljp, line 844: Is this redundant given the overall sort below?

neiljp (commented):
@Abhirup-99 Thanks for sticking with this - this will be great when we can merge it, in many ways :)

zulipterminal/api_types.py
```diff
@@ -76,6 +76,34 @@ class ReactionEvent(TypedDict):
...
 
 
+class Subscription(TypedDict):
+    name: str
+    stream_id: int
+    description: str
+    rendered_description: str
+    invite_only: bool
+    is_web_public: bool
+    stream_host_policy: int
```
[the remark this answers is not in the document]
Abhirup-99: https://github.com/zulip/zulip/blob/d74266ac7503d97ed528ab0bf567197b5611da5c/zerver/openapi/zulip.yaml#L496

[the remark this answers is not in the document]
neiljp: Sure - I'm not sure what `stream_host_policy` is :)

zulipterminal/api_types.py
```diff
@@ -76,6 +76,34 @@ class ReactionEvent(TypedDict):
     message_id: int
 
 
+class Subscription(TypedDict):
+    name: str
+    stream_id: int
+    description: str
+    rendered_description: str
+    invite_only: bool
```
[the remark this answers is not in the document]
Abhirup-99: I followed the YAML file ordering mostly, patching it up with the ones I found on the networks tab. I would add a small comment above it, stating the source of the keys and the ordering I chose.

zulipterminal/model.py
```diff
@@ -723,40 +727,43 @@ def user_name_from_id(self, user_id: int) -> str:
...
+            self.stream_dict[subscription['stream_id']] = subscription
+            streamData = make_reduced_stream_data(subscription)
+            if subscription['pin_to_top']:
+                new_pinned_streams.append(streamData)
+            else:
+                new_unpinned_streams.append(streamData)
+            if not subscription['in_home_view']:
+                new_muted_streams.add(subscription['stream_id'])
+
+        # To reduce unneccesary sorts
```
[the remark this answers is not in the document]
Abhirup-99: I meant not making sorts if the `pinned_streams` or `unpinned_streams` are empty.  The blank line was before my commit, so I kept it as it is. I have removed the comment. 👍🏽

zulipterminal/ui_tools/views.py
```diff
@@ -1128,7 +1132,9 @@ def __init__(self, controller: Any, stream_id: int) -> None:
                                            stream_id),
                                        checked_symbol=CHECK_MARK)
         urwid.connect_signal(muted_setting, 'change', self.toggle_mute_status)
-        pinned_state = controller.model.is_pinned_stream(stream_id)
+        pinned_state = (controller.model
+                        .is_user_in_stream(stream_id,
+                                           controller.model.pinned_streams))
```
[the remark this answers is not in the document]
Abhirup-99: I agree it does seem like a backward change when viewed isolatedly, but the thought process behind the change was I won't have access to the `stream_dict` to check if the user is a part of the stream or either pinned or unpinned streams, so I have to check it looping through the two arrays `pinned_streams` or `unpinned_stream` depending on the use case, so the idea to generalize this into a function, which can be called by passing `unpinned_streams` or `pinned_streams` depending on the use case seemed a better approach.

zulipterminal/ui_tools/views.py
```diff
@@ -556,13 +556,17 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
             self.body.keypress(size, 'c')
             # For new streams with no previous conversation.
             if self.footer.focus is None:
-                stream_id = self.model.stream_id
                 stream_dict = self.model.stream_dict
                 self.footer.stream_box_view(
-                    caption=stream_dict[stream_id]['name'])
+                    caption=stream_dict[stream_id]['name'],
+                    stream_id=stream_id
+                    )
```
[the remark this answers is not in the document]
Abhirup-99: Removed, this originally was not part of the pr, might have crept in during a faulty rebase.

zulipterminal/model.py
Abhirup-99: Updated the logic.

Abhirup-99:
@zulipbot add "PR needs review" remove "PR awaiting update"

Abhirup-99:

zulipterminal/model.py
```diff
@@ -824,13 +824,41 @@ def _get_stream_by_id(self, streams: List[StreamData], stream_id: int
...
+                                                stream_id)
+                self.unpinned_streams.remove(stream)
+                is_unpinned_stream_removed = True
+            if stream_id in self.muted_streams:
+                self.muted_streams.remove(stream_id)
+
+        if is_pinned_stream_removed:
+            sort_streams(self.pinned_streams)
+        if is_unpinned_stream_removed:
+            sort_streams(self.unpinned_streams)
```
neiljp, line 853: I know we improved this code in the past, and it does read well, but... if we just remove elements from a sorted container, isn't it still sorted?

zulipterminal/model.py
```diff
@@ -817,6 +817,10 @@ def toggle_stream_pinned_status(self, stream_id: int) -> bool:
     def is_user_subscribed_to_stream(self, stream_id: int) -> bool:
         return stream_id in self.stream_dict
 
+    def is_user_in_unsubscribed_stream(self, stream_id: int) -> bool:
+        return not(self.is_stream_in_list(stream_id, self.pinned_streams)
+                   or self.is_stream_in_list(stream_id, self.unpinned_streams))
```
neiljp, line 822: I think we need to refine what `stream_dict` represents, as intuitively I expected these two methods to be the inverse of each other - and so we wouldn't need both.

However, right now I noticed we're not removing entries from stream_dict, so it's going to become a "streams we might be subscribed to" at this point. That doesn't seem overly useful, so I'd suggest we either make it represent "all streams" and check when we look it up in various places that we aren't assuming they're subscribed (through another datastructure), or keep it as subscribed streams only. We would need something like this later for a subscription interface in any case, but for now it should also simplify this case.

zulipterminal/core.py
```diff
@@ -102,6 +102,15 @@ def raise_exception_in_main_thread(self,
             self._critical_exception = True
         os.write(self._exception_pipe, b'1')
 
+    def check_for_invalid_operation(self, stream_id: Optional[int]) -> bool:
+        assert stream_id is not None
+        if self.model.is_user_in_unsubscribed_stream(stream_id):
```
neiljp, line 107: The use of the assert here is confusing - why is it not just int?

I understand extracting the common code, but the function name seems very general.

neiljp (commented):
@Abhirup-99 I've pushed a slight adjustment of the first refactor commit since it's definitely a solid improvement and foundation for the other work.

There should be less to resolve here now after the few cherry-picks, and while a visual/functional change is maybe not yet evident, I hope you can see the code foundation shifting to make this easier. The challenge is often uncovering the data-structures underneath which need updating first!

zulipterminal/model.py
Abhirup-99: Ahh 😅 , I should have thought that. Thanks for pointing it out, would remove it shortly.

zulipterminal/core.py
Abhirup-99: `stream_id` is `Optional[int]` currently, that's why the `assert` was necessary. When I first wrote this PR, `stream_id` which is actually `self.stream_id` was `int`. This function is called in 2 or 3 places, in each of them I had to put an `assert`, so thought of just pushing it in this function.

zulipterminal/model.py
Abhirup-99: Would update this shortly.

zulipterminal/model.py
```diff
@@ -801,14 +801,15 @@ def stream_id_from_name(self, stream_name: str) -> int:
                 return stream_id
         raise RuntimeError("Invalid stream name.")
 
-    def is_pinned_stream(self, stream_id: int) -> bool:
-        return stream_id in [stream['id'] for stream in self.pinned_streams]
+    def is_stream_in_list(self, stream_id: int,
+                          stream_list: List[StreamData]) -> bool:
+        return stream_id in [stream['id'] for stream in stream_list]
```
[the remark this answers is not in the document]
Abhirup-99: Reverted to it's previous state.

tests/model/test_model.py
```diff
@@ -46,6 +46,48 @@ def model(self, mocker, initial_data, user_profile,
...
+        return [{
+                    'name': 'test2',
+                    'id': 100,
+                    'color': '#9af',
+                    'invite_only': False,
+                    'description': 'lets test'
+                }]
+
+    @pytest.fixture
+    def initial_stream_dict(self):
```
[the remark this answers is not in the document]
Abhirup-99: Updated to use the conftest.py fixtures.

zulipterminal/model.py
Abhirup-99: Added another data structure.

zulipterminal/core.py
LoopThrough-i-j: I think `stream_id` should be `int` in the function declaration itself.

Issue #782, opened 2020-08-20
Support a variety of footer notification styling


neiljp:
We can consider styling of footer notifications, eg. to distinguish errors/warnings from simple notifications like #781, and we could even add a positive one for "message sent" type notifications.

Ezio-Sarthak:
I presume no one's working on this currently. Since I am familiar with that part of the codebase, I would love to work on this issue!

PR #853, opened 2020-12-18
enhancement: Further popup styling improvements.

### LINKED ISSUE #684: Enhance the general popup styling

Ezio-Sarthak:
This is a follow-up PR from #849. It aims the following:

- [x] Popup border and title styling.

- [x] Popup color styling.

Fixes #684 (for now) :)
The palettes added with this PR might help in styling of footer notifications in #782

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -43,6 +43,8 @@ required_styles = {  # style-name: monochrome-bit-depth-style
     'edit_tag': 'standout',
     'edit_author': 'bold',
     'edit_time': 'bold',
+    'popup_border': 'bold',
+    'popup_title': 'bold',
 }
 
 
@@ -171,6 +173,10 @@ THEMES = {
          None,             DEF['white'],              DEF['black']),
         ('muted',          'light blue',              'black',
          None,             DEF['light_blue'],         DEF['black']),
+        ('popup_title',    'black',                   'white',
+         None,             DEF['black'],              DEF['white']),
+        ('popup_border',   'white',                   'black',
+         None,             DEF['white'],              DEF['black']),
     ],
     'gruvbox_dark': [
         # default colorscheme on 16 colors, gruvbox colorscheme
@@ -251,6 +257,10 @@ THEMES = {
          None,             WHITE,             BLACK),
         ('muted',          'light blue',      'black',
          None,             LIGHTBLUE,         BLACK),
+        ('popup_title',    'black',           'white',
+         None,             BLACK,             WHITE),
+        ('popup_border',   'white',           'black',
+         None,             WHITE,             BLACK),
     ],
     'zt_light': [
         (None,             'black',           'white'),
@@ -291,6 +301,8 @@ THEMES = {
         ('edit_time',      'dark blue',       'white'),
         ('current_user',   'dark gray',       'white'),
         ('muted',          'dark gray',       'white'),
+        ('popup_title',    'white',           'black'),
+        ('popup_border',   'black',           'white'),
     ],
     'zt_blue': [
         (None,             'black',           'light blue'),
@@ -331,6 +343,8 @@ THEMES = {
         ('edit_time',      'dark blue',       'light blue'),
         ('current_user',   'light gray',      'light blue'),
         ('muted',          'light gray',      'light blue'),
+        ('popup_title',    'light blue',      'white'),
+        ('popup_border',   'white',           'light blue'),
     ]
 }  # type: Dict[str, ThemeSpec]
 
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -44,7 +44,10 @@ required_styles = {  # style-name: monochrome-bit-depth-style
     'edit_author': 'bold',
     'edit_time': 'bold',
     'popup_border': 'bold',
-    'popup_title': 'bold',
+    'area:help': 'standout',
+    'area:msg': 'standout',
+    'area:stream': 'standout',
+    'area:error': 'standout',
 }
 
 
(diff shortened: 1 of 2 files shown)


zulipterminal/core.py
```diff
@@ -162,19 +162,22 @@ def maximum_popup_dimensions(self) -> Tuple[int, int]:
...
+                    header=urwid.BoxAdapter(urwid.AttrMap(
+                        urwid.Filler(text), 'popup_title'), height=1),
+                    body=to_show),
+                              **border_lines), 'popup_border'),
+                bottom_w=self.view,
+                align='center',
+                valign='middle',
+                # +2 to both of the following, due to LineBox
+                width=to_show.width + 2,
+                height=to_show.height + 2,
```
neiljp, line 180: You may need to adjust this, due to the extra title height - for example, message-info boxes scroll needlessly.

zulipterminal/core.py
```diff
@@ -162,19 +162,22 @@ def maximum_popup_dimensions(self) -> Tuple[int, int]:
...
-            align='center',
-            valign='middle',
-            # +2 to both of the following, due to LineBox
-            width=to_show.width + 2,
-            height=to_show.height + 2,
+                urwid.AttrMap(urwid.LineBox(urwid.Frame(
+                    header=urwid.BoxAdapter(urwid.AttrMap(
+                        urwid.Filler(text), 'popup_title'), height=1),
+                    body=to_show),
+                              **border_lines), 'popup_border'),
```
neiljp, line 174: This could be expressed a lot clearer. I know it may change if we move widgets around, but this is quite difficult to follow. Intermediate variables would help.

neiljp (commented):
@Ezio-Sarthak This is a nice self-contained improvement compared to the other PR :+1:

There is currently an issue with the popup height, but more importantly I'd really also like to avoid the gaps at the top - perhaps we could just put a linebox around the scrolling content to avoid the difference at the edge of that area and the title. We could also use the top of that linebox to make a slight half-box line above the content to give a small gap at the top between the title and the first line of content.

I also noticed that this looks strange in color-depth=1 at the top.

zulipterminal/config/themes.py
```diff
@@ -44,7 +44,10 @@
     'edit_author': 'bold',
     'edit_time': 'bold',
     'popup_border': 'standout',
-    'popup_title': 'bold',
+    'help_related': 'standout',
+    'msg_related': 'standout',
+    'stream_related': 'standout',
+    'error_related': 'standout',
```
neiljp, line 50: I think these would work better with a common prefix, possibly even `area:help`, `area:msg`, ...?

zulipterminal/core.py
```diff
@@ -161,12 +161,12 @@ def maximum_popup_dimensions(self) -> Tuple[int, int]:
                                  self.loop.screen.get_cols_rows())
         return max_cols, max_rows
 
-    def show_pop_up(self, to_show: Any) -> None:
+    def show_pop_up(self, to_show: Any, category: str='msg_related') -> None:
```
neiljp, line 164: Could you explain the use of the default value here?

neiljp (commented):
@Ezio-Sarthak This looks cleaner now :+1:

Layout-wise I think the remaining issue is that the title looks off-center vertically, along with the merging of the title into the background in 1-color mode. Perhaps a half-bar of border color above it?

zulipterminal/core.py
```diff
@@ -162,19 +162,23 @@ def maximum_popup_dimensions(self) -> Tuple[int, int]:
...
-            height=to_show.height + 2,
+                urwid.AttrMap(urwid.Frame(header=title, body=content),
+                              'popup_border'),
+                self.view,
+                align='center',
+                valign='middle',
+                # +2 to both of the following, due to LineBox
+                # +1 to height, due to title enhancement
+                width=to_show.width + 2,
+                height=to_show.height + 3,
```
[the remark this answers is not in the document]
Ezio-Sarthak: Oops, that's not needed :+1:

zulipterminal/config/themes.py
```diff
@@ -251,6 +256,10 @@
          None,             WHITE,             BLACK),
         ('muted',          'light blue',      'black',
          None,             LIGHTBLUE,         BLACK),
+        ('popup_title',    'black, bold',     'white',
+         None,             DEF['black:bold'], DEF['white']),
+        ('popup_border',   'white',           'black',
+         None,             DEF['white'],      DEF['black']),
```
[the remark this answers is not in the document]
Ezio-Sarthak: Thanks! I'll use equivalent to those already defined in gruvbox-dark theme

zulipterminal/core.py
Ezio-Sarthak: Yeah, that's not needed, sorry for confusion :sweat_smile:

Ezio-Sarthak:
Thanks for the suggestions!

I've updated the PR. I've corrected the small mistakes I did :)

neiljp (commented):
@Ezio-Sarthak Thanks for addressing my points :+1:

While you've made the header higher, the title block looks taller too, ie. a half-line of eg. green below the line of white. Could we move the 'white' half-line down to centre the title text vertically between two white bars? I'm not sure what style to use, maybe a different character with the border style? That may need investigating with respect to 1-color rendering again.

Does this look similar on all themes? I checked 'light' and the title doesn't look quite the same compared to 'default' (zt_light vs zt_dark)

zulipterminal/core.py
```diff
@@ -161,18 +161,18 @@ def maximum_popup_dimensions(self) -> Tuple[int, int]:
                                  self.loop.screen.get_cols_rows())
         return max_cols, max_rows
 
-    def show_pop_up(self, to_show: Any) -> None:
+    def show_pop_up(self, to_show: Any, category: str) -> None:
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yeah that's understandable :+1:

neiljp:
@Ezio-Sarthak Merged with adjusted commit text as e08fbebe8ccc9c3d27b751d6431cde424cd47f03 and prior commit :tada:

I also slightly adjusted the 1-bit style in the first commit to match that used later (title -> area), and moved the top border style & character from the second to the first commit. If you haven't already I'd recommend giving `git range-diff` a try to show this kind of change :)

Ezio-Sarthak:
`git range-diff` seems great to distinguish between different commits!

PR #854, opened 2020-12-21
boxes/helper: Updated to new quoting style.

### LINKED ISSUE #514: Use new quoting style
The webapp now adds an extra line when quoting, by default, with who wrote the quotation and a link to the original message. This is added automatically as extra (editable) markdown/bugdown, so we see this in ZT when someone quotes in the webapp - but we should really do something similar from ZT for interoperability.

zee-bit:
This PR aims to bring ZT one step more closer to its web counterpart. The quoting style is changed to include the name of the user whose message was quoted(and his `user_id`) along with a link that would narrow to the quoted message. This is the new markdown/bugdown syntax of a quoted message:
````
@**Full Name|1234(id)** said[link_to_msg]:
```quote
    quoted_msg
```
````

For interoperability, the link to the quoted message is same as that of the webapp, so narrowing to the quote in ZT would not be possible, except in webapp.

Also, the URL generation and encoding part of the code has been shifted to a new module named `server_url.py` to separate it from the rest of the code inside `helper.py` and make it easier for constructing and deconstructing server URL in the future.

This PR should fix #514 and #151(partially).

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -5,7 +5,7 @@ import time
 from collections import OrderedDict, defaultdict
 from functools import wraps
 from itertools import chain, combinations
-from re import ASCII, match
+from re import ASCII, MULTILINE, findall, match
 from threading import Thread
 from typing import (
     Any, Callable, DefaultDict, Dict, FrozenSet, Iterable, List, Set, Tuple,
@@ -655,3 +655,21 @@ def hash_util_decode(string: str) -> str:
     # Acknowledge custom string replacements in zulip/zulip's
     # zerver/lib/url_encoding.py before unquote.
     return unquote(string.replace('.', '%'))
+
+
+def get_unused_fence(content: str) -> str:
+    """
+    Generates fence for quoted-message based on regex pattern
+    of continuous back-ticks. Referred and translated from
+    zulip/static/shared/js/fenced_code.js.
+    """
+    fence_length_regex = '^ {0,3}(`{3,})'
+    max_length_fence = 3
+
+    matches = findall(fence_length_regex, content,
+                      flags=MULTILINE)
+    if len(matches) != 0:
+        max_length_fence = max(max_length_fence,
+                               len(max(matches, key=len)) + 1)
+
+    return '`' * max_length_fence
--- a/zulipterminal/server_url.py
+++ b/zulipterminal/server_url.py
@@ -1,0 +1,87 @@
+import urllib.parse
+
+from zulipterminal.helper import Message
+
+
+def hash_util_encode(string: str) -> str:
+    """
+    Hide URI-encoding by replacing '%' with '.'
+    urllib.quote is equivalent to encodeURIComponent in JavaScript.
+    Referred from zerver/lib/url_encoding.py
+    """
+    # `safe` has a default value of "/", but we want those encoded, too.
+    return urllib.parse.quote(
+        string, safe=b"").replace(".", "%2E").replace("%", ".")
+
+
+def encode_stream(stream_id: int, stream_name: str) -> str:
+    """
+    Encodes stream_name with stream_id and replacing any occurence
+    of whitespace to '-'. This is the format of message representation
+    in webapp. Referred from zerver/lib/url_encoding.py.
+    """
+    stream_name = stream_name.replace(' ', '-')
+    return str(stream_id) + '-' + hash_util_encode(stream_name)
+
+
+def near_stream_message_url(server_url: str, message: Message) -> str:
+    """
+    Returns the complete encoded URL of a message from #narrow/stream.
+    Referred from zerver/lib/url_encoding.py.
+    """
+    message_id = str(message['id'])
+    stream_id = message['stream_id']
+    stream_name = message['display_recipient']
+    topic_name = message['subject']
+    encoded_stream = encode_stream(stream_id, stream_name)
+    encoded_topic = hash_util_encode(topic_name)
+
+    parts = [
+        server_url,
+        '#narrow',
+        'stream',
+        encoded_stream,
+        'topic',
+        encoded_topic,
+        'near',
+        message_id,
+    ]
+    full_url = '/'.join(parts)
+    return full_url
+
+
+def near_pm_message_url(server_url: str, message: Message) -> str:
+    """
+    Returns the complete encoded URL of a message from #narrow/pm-with.
+    Referred from zerver/lib/url_encoding.py.
+    """
+    message_id = str(message['id'])
+    str_user_ids = [
+        str(recipient['id'])
+        for recipient in message['display_recipient']
+    ]
+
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: zulipterminal/ui_tools/boxes.py [hunk 1]
```diff
@@ -1186,7 +1186,21 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+            #   ```
+            quote = '@_**{}|{}** '.format(self.message['sender_full_name'],
+                                          self.message['sender_id'])
+
+            # FIXME: The type and generation of link needs to be discussed.
+            # For interoperability webapp link should be given, but
+            # that would not narrow to the quoted message inside ZT.
+            absolute_url = absolute_url_of_message(self.model, self.message)
+            quote += '[said]({}):\n'.format(absolute_url)
+            quote += '```quote\n' + self.model.client.get_raw_message(
```
andersk, line 1203: This will fail to correctly quote a message that already contains a quote or a code block. You need to quote ```` ``` ```` with ````` ```` `````, ````` ```` ````` with `````` ````` ``````, and so on. See [`get_unused_fence`](https://github.com/zulip/zulip/blob/160cc5120aae0870c68e6e3a12a391035bd25843/static/shared/js/fenced_code.js#L241-L251) (called [here](https://github.com/zulip/zulip/blob/160cc5120aae0870c68e6e3a12a391035bd25843/static/js/compose_actions.js#L444)) in the web app.

zee-bit: Thanks for the review @andersk. I probably missed that part of code logic, from the webapp. Nevertheless, I am now supposed to refactor the whole code along with some server code into a new module `server_urls.py` as instructed by Neil in CZO. I'll keep this review in mind while refactoring.

neiljp (commented):
@zee-bit Initial testing showed this to render fine, though possibly some topics with `/` in them don't work?

You have the updated quote style ok, but as we discussed in the stream the challenge is how to handle url generation. I'd suggest decoupling the message-to-url part (helper.py is fine for now) as you have done here (from the url-generation part, in another file). The server has `url_encoding.py`, but it uses other types connected to the server implementation like `Realm` and it uses f-strings (though we plan to to move to that soon with python3.6). My suggestion would be to use the server file as a basis for a module in our repo which handles chat server urls like this, and we can incorporate your and @preetmishra 's work if it builds upon that in a useful way.

Other than interoperability and tracking changes in the server, the reason I'd like this in a separate file is since we both construct and deconstruct server URLs from narrows and it'd be useful to have it together, and also is related to my hope that we could later integrate this into the python API package.

We can discuss more in the stream if you like.

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
zee-bit: I have fixed the issue pointed by Ander's in a separate commit and refactored some parts of the code as mentioned at the beginning of this comment too. So, I guess we can close this conversation as resolved.

zee-bit:

zulipterminal/ui_tools/boxes.py
```diff
@@ -1200,16 +1200,16 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
-
-            # FIXME: The type and generation of link needs to be discussed.
-            # For interoperability webapp link should be given, but
-            # that would not narrow to the quoted message inside ZT.
             absolute_url = absolute_url_of_message(self.model, self.message)
-            quote += '[said]({}):\n'.format(absolute_url)
-            quote += '```quote\n' + self.model.client.get_raw_message(
-                self.message['id'])['raw_content'] + '\n```\n'
+            fence = get_unused_fence(self.model.client.get_raw_message(
+                self.message['id'])['raw_content'])
```
neiljp, line 1207: If you want to put this addition as the first commit we can merge this separately.

However, it would be useful to have some tests of the introduced function (and perhaps the integrated case) in the same commit too.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1200,16 +1200,16 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+
+            quote = '@_**{0}|{1}** [said]({2}):\n{3}quote\n{4}\n{3}\n'.format(
+                        self.message['sender_full_name'],
+                        self.message['sender_id'], absolute_url, fence,
+                        self.model.client.get_raw_message(
+                            self.message['id'])['raw_content'])
```
neiljp, line 1213: Separate lines for each parameter and extracting the raw message calculation would make this easier to read. We can use a smaller indent too?

This won't apply if you put this commit first, but when you have both the fencing and improving quoting, note that we want to avoid calling the server twice for the get_raw_message.

zulipterminal/helper.py
```diff
@@ -657,6 +657,19 @@ def hash_util_decode(string: str) -> str:
     return unquote(string.replace('.', '%'))
 
 
+def get_unused_fence(content: Any) -> Any:
```
neiljp, line 660: You don't have types specified here - can we improve this?

Does this come from a translation of a function in the server? We should reference it as a comment, if so, and perhaps when. Also if you used some particular logic to translate it.

zulipterminal/server_url.py
```diff
@@ -0,0 +1,79 @@
+import urllib
```
neiljp, line 1: This commit doesn't seem to just *move* the code across, but also changes it... so doesn't this really replace the first commit? Doesn't this make the first commit meaningless? Which version are we including? Since this is the start of this code, we can just add it to the right file in the first place :)

As per my note elsewhere, if we bring code across from the server (which this looks like), we should indicate details, and what has changed or not - as well as anything else that you used in the process that's relevant.

neiljp (commented):
@zee-bit This is promising, but as per another PR we can tidy this up a little more in terms of commit structure :)

It'll be great to have the bugfix, but as with the other code I'd like to see text of how it was translated across from the server (if it was), and also some tests should help with indicating and checking behavior :+1:

zee-bit:

zulipterminal/helper.py
zee-bit: I have improved type annotations to `str`.
Not necessarily, its referred from `zulip/static/shared/js/fenced_code.js`, if I am correct in assuming by server you mean `zulip/zerver`. I have added short comments addressing this function.

zulipterminal/server_url.py
zee-bit: You are correct, the last commit ultimately makes the first commit useless. I have removed the first commit. :)

I have added brief comments for each new function added. If you want I can reword those comments to be more verbose.

neiljp:
@zee-bit I manually merged this just now with slightly adjusted commit text, those extra test cases I mentioned, and squashing the last two commits together. The last commit of the set is 71052402430bce1e551ae1e96e92b047c2404392 :tada:

PR #856, opened 2020-12-28
views: Updated Stream Info.

### LINKED ISSUE #753: Add member list to stream info view
Follow-up to #730 - which stores subscriber info for streams. We can display this, along with weekly message count for streams in the stream settings view.

zee-bit:
This PR is divided into two parts:
- Show Stream Members in Stream Info view - The First commit leading this PR adds the feature to view Members subscribed to a stream from inside the Stream info view on pressing `m` key.
- Show estimated messages per week for streams - The Second commit will add another submenu inside Stream info which will show the Estimated messages per week gathered inside that stream.

Tests amended.
This PR fixes #753.

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -1081,8 +1081,16 @@ class StreamInfoView(PopUpView):
         stream_marker = (STREAM_MARKER_PRIVATE if stream['invite_only']
                          else STREAM_MARKER_PUBLIC)
         title = '{} {}'.format(stream_marker, stream['name'])
+        total_members = len(stream['subscribers'])
+        keys = ', '.join(map(repr, keys_for_command('STREAM_MEMBERS')))
         desc = stream['description']
-        stream_info_content = [('', [desc]), ('Stream settings', [])]
+        stream_info_content = [('', [desc]),
+                               ('Stream Details', [
+                                   ('Stream Members',
+                                    '[{}] Press {} to view'.format(
+                                        total_members, keys))
+                               ]),
+                               ('Stream settings', [])]
         popup_width, column_widths = self.calculate_table_widths(
             stream_info_content, len(title))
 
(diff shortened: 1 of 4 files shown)


zee-bit:

zulipterminal/ui_tools/views.py
```diff
@@ -1112,6 +1119,43 @@ def toggle_mute_status(self, button: Any, new_state: bool) -> None:
...
+class StreamMembersView(PopUpView):
+    def __init__(self, controller: Any, stream_id: int) -> None:
+        self.stream_id = stream_id
+        self.controller = controller
+        model = controller.model
+
+        user_ids = model.get_other_subscribers_in_stream(stream_id=stream_id)
+        user_names = [model.user_name_from_id(id) for id in user_ids]
+        sorted_user_names = sorted(user_names)
+        # sorted_user_names.insert(0, model.user_full_name)
```
neiljp: This commented code should be in this commit for showing members, right?

We should sort this in a similar way as we do the presence list (though probably not per active status). Note that showing stream members in this way might just be a temporary feature until we add filtering of the user list.

zulipterminal/ui_tools/views.py
```diff
@@ -1134,11 +1137,13 @@ def __init__(self, controller: Any, stream_id: int) -> None:
...
         sorted_user_names = sorted(user_names)
-        # sorted_user_names.insert(0, model.user_full_name)
+        sorted_user_names.insert(0, model.user_full_name)
+        total_members = len(sorted_user_names)
         title = 'Stream Members (up/down scrolls)'
 
-        stream_users_content = [('', [
-            (name, '') for name in sorted_user_names]
+        stream_users_content = [(
+            'Subscribed members({})'.format(total_members), [
```
neiljp, line 1145: I'm not sure the members list needs a category, as there's just a list of users.

For the number of members, we can place this in the stream info popup, before the `Press m to view` text.

zulipterminal/ui_tools/views.py
```diff
@@ -1077,12 +1077,15 @@ def __init__(self, controller: Any, stream_id: int) -> None:
...
+            'stream_weekly_traffic']
         keys = ', '.join(map(repr, keys_for_command('STREAM_MEMBERS')))
         title = '# {}'.format(stream['name'])
         desc = stream['description']
         stream_info_content = [('', [desc]),
-                               ('Stream Info', [
-                                   ('Estimated messages per week', str(93)),
+                               ('Stream Details', [
+                                   ('Weekly Message Count',
+                                    str(weekly_msg_count)),
```
neiljp, line 1088: All of this is different from the previous commit? Let's just add all the message count details into this commit.

neiljp (commented):
@zee-bit Thanks for working on this - it opens up a path for showing more stream data, which is great :+1:

This functions fairly well and is a great starting point, and I've left some feedback on mainly the code (not tests) inline.

Right now the commit structure is a little confusing - ideally each commit would add a specific element or a particular change, and code associated with that change would be only in that commit. It seems like your idea is that the first commit adds stream members and the second adds the weekly counts, but the first one is named like the PR rather than focusing on the specific change. There are also changes in each commit that are changed from one to the other (eg. stream counts are added as a fixed value in the first then added properly in the second, and the stream member link is named differently in the second weekly-count commit). It would make it easier to follow, review and ultimately merge if changes are isolated cleanly into their own commits :)

As I note inline, the stream members popup may be superseded later by filtering of the user list, which will provide native support for things like searching/filtering for users, so this could be useful in the interim but I think we shouldn't put too much effort into it.

Once these commits are firmed up (or the PR merged), note that there are other stream properties which we could list, based on other available stream data.

zee-bit:
@neiljp Sorry, for messing up the whole commit history and structure(I had a very embarrassing perception about PR's and commit's that led to this mess :sweat_smile: ). There is another PR where I might have done this same mistake(I'll fix that too).

Now, I have updated the code keeping in mind all the reviews above. I have used git interactive-rebase feature to amend these mistakes. The commits are now much cleaner and separate from each other. Hoping to get further review soon. :blush:

neiljp:
@zee-bit I have many PRs to review, but you seem to have the commit structure better, so if you want to update any other PRs like this that'd be great :+1:

neiljp:
@zee-bit As per discussion in PMs I've merged this with a few minor simplifications as the commits ending with 1d81a6a49c53ef7e3e6a33927c4c9f2ce8860aa1 - thanks! :tada:

PR #858, opened 2020-12-28
gitignore: Add zulip-terminal-thread-exceptions.log.

prah23:
Added `zulip-terminal-thread-exceptions.log` to the `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,12 +6,10 @@ cov_html/*
 .mypy_cache
 Pipfile.lock
 zulipterminal.egg-info
-debug.log
 dist/*
 build/*
 zulip_term.egg-info
-zulip-terminal-tracebacks.log
-zulip-terminal-API-requests.log
+*.log
 zt_venv/*
 zulip-terminal-venv/*
 


.gitignore
```diff
@@ -12,6 +12,7 @@ build/*
 zulip_term.egg-info
 zulip-terminal-tracebacks.log
 zulip-terminal-API-requests.log
+zulip-terminal-thread-exceptions.log
```
neiljp, line 15: Given the common naming we have for log files, this might be simpler covering all logfiles with one pattern?

neiljp (commented):
@prah23 Thanks for catching this :+1:

Thanks for following the git commit style we use; I left one comment since we could simplify this given our standard log naming.

.gitignore
prah23: That makes sense. Since we exclude all logfiles from being pushed, I can think of `*.logs`. Any logfiles being added in the future would also be excluded.

prah23: That makes sense. Since we exclude all logfiles from being pushed, I can think of `*.log`. Any logfiles being added in the future would also be excluded.

neiljp: I was originally thinking of `zulip-terminal-*.log`, but that might be cleaner :+1:

We could better split the gitignore into sections and maybe add comments, but it's not that large and time is likely better spent elsewhere :)

neiljp:
@prah23 Thanks for the suggestion and PR :+1: I just merged this with an extended description in the commit text as 7800d049aa83ad30127c04ad33fc5d4120a1a35d :tada:

PR #864, opened 2021-01-02
boxes: Disable cycling using arrow keys if recipient is invalid.

### LINKED ISSUE #779: Disallow cycling from recipient box using arrow keys if recipient is invalid
Follow-up from #738, we should also disable switching from recipient box using arrow keys, in addition to currently supported `CYCLE_COMPOSE_BOX` key, if the recipient is invalid.

zee-bit:
This PR disables cycling from the recipient box to any other box inside the `WriteBox` view using arrow keys, if the recipient name is invalid. This is very similar to `CYCLE_COMPOSE_FOCUS` using <kbd>Tab</kbd> key. Tests have also been updated and amended.

This PR fixes #779.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -177,8 +177,7 @@ class TestWriteBox:
         write_box.to_write_box.edit_text = raw_recipients
         write_box.to_write_box.set_edit_pos(len(raw_recipients))
         expected_lines = [
-            "Invalid recipient(s) - " + invalid_recipients,
-            " - Use ",
+            "Invalid recipient(s) - " + invalid_recipients + ". Use ",
             ("footer_contrast", primary_key_for_command("AUTOCOMPLETE")),
             " or ",
             ("footer_contrast", primary_key_for_command("AUTOCOMPLETE_REVERSE")),
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -1044,6 +1044,7 @@ class TestWriteBox:
             "box_type",
             "msg_body_edit_enabled",
             "message_being_edited",
+            "key",
             "expected_focus_name",
             "expected_focus_col_name",
         ],
@@ -1054,9 +1055,10 @@ class TestWriteBox:
                 "stream",
                 True,
                 False,
+                "tab_key",
                 "CONTAINER_HEADER",
                 "HEADER_BOX_TOPIC",
-                id="stream_name_to_topic_box",
+                id="stream_name_to_topic_box_tab",
             ),
             case(
                 "CONTAINER_HEADER",
@@ -1064,9 +1066,10 @@ class TestWriteBox:
                 "stream",
                 True,
                 False,
+                "tab_key",
                 "CONTAINER_MESSAGE",
                 "MESSAGE_BOX_BODY",
-                id="topic_to_message_box",
+                id="topic_to_message_box_tab",
             ),
             case(
                 "CONTAINER_HEADER",
@@ -1074,9 +1077,10 @@ class TestWriteBox:
                 "stream",
                 False,
                 True,
+                "tab_key",
                 "CONTAINER_HEADER",
                 "HEADER_BOX_EDIT",
-                id="topic_edit_only-topic_to_edit_mode_box",
+                id="topic_edit_only-topic_to_edit_mode_box_tab",
             ),
             case(
                 "CONTAINER_HEADER",
@@ -1084,9 +1088,10 @@ class TestWriteBox:
                 "stream",
                 False,
                 True,
+                "tab_key",
                 "CONTAINER_HEADER",
                 "HEADER_BOX_TOPIC",
-                id="topic_edit_only-edit_mode_to_topic_box",
+                id="topic_edit_only-edit_mode_to_topic_box_tab",
             ),
             case(
                 "CONTAINER_MESSAGE",
@@ -1094,9 +1099,10 @@ class TestWriteBox:
                 "stream",
                 True,
                 False,
+                "tab_key",
                 "CONTAINER_HEADER",
                 "HEADER_BOX_STREAM",
-                id="message_to_stream_name_box",
+                id="message_to_stream_name_box_tab",
             ),
             case(
                 "CONTAINER_HEADER",
@@ -1104,9 +1110,10 @@ class TestWriteBox:
                 "stream",
                 True,
                 True,
+                "tab_key",
                 "CONTAINER_HEADER",
                 "HEADER_BOX_TOPIC",
-                id="edit_box-stream_name_to_topic_box",
+                id="edit_box-stream_name_to_topic_box_tab",
             ),
             case(
                 "CONTAINER_HEADER",
@@ -1114,9 +1121,10 @@ class TestWriteBox:
                 "stream",
                 True,
                 True,
+                "tab_key",
                 "CONTAINER_HEADER",
                 "HEADER_BOX_EDIT",
-                id="edit_box-topic_to_edit_mode_box",
+                id="edit_box-topic_to_edit_mode_box_tab",
             ),
             case(
                 "CONTAINER_HEADER",
@@ -1124,9 +1132,10 @@ class TestWriteBox:
                 "stream",
                 True,
                 True,
+                "tab_key",
                 "CONTAINER_MESSAGE",
                 "MESSAGE_BOX_BODY",
-                id="edit_box-edit_mode_to_message_box",
+                id="edit_box-edit_mode_to_message_box_tab",
             ),
             case(
                 "CONTAINER_MESSAGE",
@@ -1134,9 +1143,10 @@ class TestWriteBox:
                 "stream",
                 True,
                 True,
+                "tab_key",
                 "CONTAINER_HEADER",
                 "HEADER_BOX_STREAM",
-                id="edit_box-message_to_stream_name_box",
+                id="edit_box-message_to_stream_name_box_tab",
             ),
             case(
                 "CONTAINER_HEADER",
@@ -1144,9 +1154,10 @@ class TestWriteBox:
                 "private",
                 True,
                 False,
+                "tab_key",
                 "CONTAINER_MESSAGE",
                 "MESSAGE_BOX_BODY",
-                id="recipient_to_message_box",
+                id="recipient_to_message_box_tab",
             ),
             case(
                 "CONTAINER_MESSAGE",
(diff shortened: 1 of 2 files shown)


neiljp:
@zee-bit Just some quick feedback:
* is there a lot of duplication here? Could we reuse parts to make this a lot simpler?
* the tests look large and complex too
* is there some way we can trigger on 'losing focus', to avoid having all these cases? (that might cover the mouse as well, as a minor side-effect)

preetmishra (commented):
@zee-bit Hey, thanks for working on this. I pulled your changes locally and it seems to work well functionally. :+1:

However, echoing what @neiljp mentioned, we could benefit from a general approach if feasible. In any case, we would want to reduce the apparent redundancy.

zee-bit:
@neiljp Regarding your feedback:
- I have made two functions(`footer_notify_invalid_stream` & `footer_notify_invalid_emails`), they have somewhat increased code reusability. The main logic, I think, could not be simplified further if I am to proceed with the current implementation.
- The tests are very much related to that of `CYCLE_COMPOSE_FOCUS` as it is derived from there, but there were few places with unnecessary conditions and checks that I have removed.
- I am not quite sure if there is any other way to trigger on 'loosing focus'(without all those conditions). If you or @preetmishra have some other implementation in mind, I would love to learn :). Re mouse event, I can try extending this for mouse events as well, if this implementation looks fine.

zee-bit:

zulipterminal/ui_tools/boxes.py
```diff
@@ -491,6 +504,48 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+                if (self.focus_position == self.FOCUS_CONTAINER_HEADER
+                        and header.focus_col == self.FOCUS_HEADER_BOX_STREAM):
+                    cursor_pos = header[self.FOCUS_HEADER_BOX_STREAM].edit_pos
+                    stream_name = (header[self.FOCUS_HEADER_BOX_STREAM]
+                                   .edit_text)
+                    stream_name_len = len(stream_name)
+                    if stream_name_len == cursor_pos:
+                        if not self.model.is_valid_stream(stream_name):
+                            self.footer_notify_invalid_stream()
+                            return key
```
neiljp, line 524: This likely needs to handle the same setting of recipient ids and stream id as in the other block of very similar code - which is a good reason why this would be handled well with a refactor to handle the different keys together.

Similarly the similar block below, and another for PMs?

neiljp (commented):
@zee-bit This needs rebasing to work around conflicts and potentially update for the latest improvements in the PM UI/data, but the implementation also demonstrates both the duplication and what look like issues due to not duplicating extra code. This could be refactored variously, but we could also "up" from the message content, which might affect how the refactoring could go.

neiljp:
I should say that it does seem to work OK :+1: I've not found a signal emitted within urwid that would solve this for us as it stands.

zulipterminal/ui_tools/boxes.py
```diff
@@ -491,6 +504,48 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                     self.model.save_draft(message)
                 elif message != saved_draft:
                     self.view.controller.save_draft_confirmation_popup(message)
+
+        # Since, there are two keys mapped to GO_RIGHT
+        # i.e. ['right', 'l'], we dont need 'l' key to satisfy
+        # the condition: is_command_key("GO_RIGHT", key).
+        # FIXME: Refactor this condition?
+        elif key == 'right':
```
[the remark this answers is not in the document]
zee-bit: Yes.
The `l` key would otherwise, instead of getting typed as a letter in `header_box`, behave exactly the same as `right` keypress(i.e. cycle focus).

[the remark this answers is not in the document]
neiljp: Great; can we make this behave correctly if other keys are used instead of those currently assigned? I'd prefer not to hard-code the assumption of `['right', 'l']` in the code.

zee-bit:

zulipterminal/ui_tools/boxes.py
```diff
@@ -572,6 +586,35 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+                            return key
+            elif key == 'down':
+                if self.focus_position == self.FOCUS_HEADER_BOX_RECIPIENT:
+                    self.update_recipient_emails(self.to_write_box)
+                    invalid_emails = self.model.get_invalid_recipient_emails(
+                        self.recipient_emails
+                    )
+                    if invalid_emails:
+                        self.footer_notify_invalid_emails(invalid_emails)
+                        return key
```
prah23, line 611: This part of the code seems the same as the section validating on cycling out using `tab`, can you perhaps add a helper within `WriteBox` to do this?

zulipterminal/ui_tools/boxes.py
```diff
@@ -490,6 +491,19 @@ def autocomplete_emojis(self, text: str, prefix_string: str
 
         return emoji_typeahead, emojis
 
+    def footer_notify_invalid_stream(self) -> None:
+        invalid_stream_error = (
+            'Invalid stream name. Use {} or {} to autocomplete.'.format(
+                primary_key_for_command('AUTOCOMPLETE'),
+                primary_key_for_command('AUTOCOMPLETE_REVERSE'))
```
prah23, line 498: It might be more vivid to style the commands in `'code'`, similar to the way we show other commands in hints in the footer.

prah23 (changes requested):
@zee-bit Thanks for this!
This works as expected and restricts the user from bypassing the validation quite well. 👍 
I've dropped a couple of comments inline, as some of these changes would also benefit #937.

zulipterminal/ui_tools/boxes.py
zee-bit: Sure, I was thinking of adding a helper but then I hesitated, I don't know why. :)

zee-bit: It will be difficult to return the `key`(incase of invalid email) from inside a helper function?

prah23: You could move the part above the `invalid_emails` to a helper (would be a prep for), but I guess it wouldn't be very necessary here since it's just a direct call to the model's method. I'll add the helper in #937 directly then.

neiljp (commented):
@zee-bit Extracting the helper(s) for the footer simplifies things to some degree, but the overlap between tab and right/down is still very similar. I'm currently fairly open to this PR since you have a test. However, the code is highly duplicated and I'd much prefer a solution which combined the two branches if we are able, to avoid that. We can always refactor later, but then we accumulate "technical debt".

zee-bit:

zulipterminal/ui_tools/boxes.py
```diff
@@ -490,6 +490,16 @@ def autocomplete_emojis(self, text: str, prefix_string: str
...
 
+    def footer_notify_invalid_recipient(self, error_msg: str) -> None:
+        footer_message = [
+            error_msg + ' Use ',
+            ('code', primary_key_for_command('AUTOCOMPLETE')),
+            ' or ',
+            ('code', primary_key_for_command('AUTOCOMPLETE_REVERSE')),
+            ' to autocomplete.'
+        ]
+        self.view.set_footer_text(footer_message, 3)
```
preetmishra, line 502: The method is calling an existing method, `set_footer_text`, with a particular error message. This could lead us to a state where we have multiple methods that essentially do the same thing with different arguments for different _error_ types.

Another way could be to just have an `ERROR_TEMPLATES` data structure that could be used to pass an appropriate error message to the existing `set_footer_text` method.

zulipterminal/ui_tools/boxes.py
```diff
@@ -500,6 +501,16 @@ def footer_notify_invalid_recipient(self, error_msg: str) -> None:
         ]
         self.view.set_footer_text(footer_message, 3)
 
+    def loosing_focus_from_stream_box(self, key: str) -> bool:
+        stream_header = self.header_write_box[self.FOCUS_HEADER_BOX_STREAM]
+        stream_name_len = len(stream_header.edit_text)
+        cursor_pos = stream_header.edit_pos
+        if (is_command_key('CYCLE_COMPOSE_FOCUS', key)
+                or key in ['right', 'down']):
```
preetmishra, line 509: As these command keys are configurable from `keys.py`, we shouldn't hardcode them here.

zulipterminal/ui_tools/boxes.py
```diff
@@ -500,6 +501,16 @@ def footer_notify_invalid_recipient(self, error_msg: str) -> None:
         ]
         self.view.set_footer_text(footer_message, 3)
 
+    def loosing_focus_from_stream_box(self, key: str) -> bool:
```
preetmishra, line 504: There a typo in the method name and in the commit description.

zulipterminal/ui_tools/boxes.py
```diff
@@ -500,6 +501,16 @@ def footer_notify_invalid_recipient(self, error_msg: str) -> None:
...
+        stream_header = self.header_write_box[self.FOCUS_HEADER_BOX_STREAM]
+        stream_name_len = len(stream_header.edit_text)
+        cursor_pos = stream_header.edit_pos
+        if (is_command_key('CYCLE_COMPOSE_FOCUS', key)
+                or key in ['right', 'down']):
+            if key != 'right' or stream_name_len == cursor_pos:
+                return True
```
preetmishra, line 511: The nested conditional could be simplified with what follows.
```python3
        if (is_command_key('CYCLE_COMPOSE_FOCUS', key) or key == "down"
              or (key == "right" and stream_name_len == cursor_pos)):
            return True
 ```
 
 While writing conditionals, expressing the intent clearly is vital. For instance, in the snippet above it is comparatively apparent that we're checking whether the key is `TAB` or `DOWN` or `RIGHT` (given that the cursor is at the end) for blur ("focus loss") event. Both approaches work, but the latter is more comprehensible.

File: zulipterminal/ui_tools/boxes.py [hunk 9]
```diff
@@ -592,32 +604,37 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+                                and self.loosing_focus_from_stream_box(key)):
                             self.footer_notify_invalid_recipient(
                                 'Invalid stream name.')
                             return key
-                        user_ids = self.model.get_other_subscribers_in_stream(
-                                                    stream_name=stream_name)
-                        self.recipient_user_ids = user_ids
-                        self.stream_id = self.model.stream_id_from_name(
-                                                                stream_name)
+                        if not is_primary_direction_key(key):
```
preetmishra, line 612: In place of recurring conditionals, we could just check for it once and return early? That would remove all the following code changes.

tests/ui_tools/test_boxes.py
```diff
@@ -709,7 +737,8 @@ def test_keypress_CYCLE_COMPOSE_FOCUS(self, write_box, tab_key,
                 write_box.stream_box_view(stream_id)
         else:
             write_box.private_box_view()
-        size = widget_size(write_box)
+        # FIXME: Use widget_size(), instead of hard-coded size.
```
preetmishra, line 740: Why aren't we using `widget_size`?

zulipterminal/ui_tools/boxes.py
```diff
@@ -582,7 +593,8 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                     self.view.controller.save_draft_confirmation_popup(
                         this_draft,
                     )
-        elif is_command_key('CYCLE_COMPOSE_FOCUS', key):
+        elif (is_command_key('CYCLE_COMPOSE_FOCUS', key)
+              or is_primary_direction_key(key)):
```
preetmishra, line 597: Why are capturing all the direction keys with the method? Wouldn't `GO_RIGHT` and `GO_DOWN` suffice?

zulipterminal/ui_tools/boxes.py
preetmishra, line 502: The method is calling an existing method, `set_footer_text`, with a particular error message. This could lead us to a state where we have multiple methods that essentially do the same thing with different arguments for different _error_ types.

Another way could be to just have an `ERROR_TEMPLATES` data structure that could be used to pass an appropriate error message to the existing `set_footer_text` method.

preetmishra (commented):
@zee-bit Thanks for your work. I pulled code locally and it seems to work well! :+1: 

While the PR is functional, we could simplify the code a bit. I have left a few in-line comments. Let me know if there's anything that needs clarification.

zulipterminal/ui_tools/boxes.py
zee-bit: Yep, this is definitely more clear and comprehensible than the earlier conditional. Guess I'll have to look from different angles if the conditionals are getting complex. Thanks. :+1:

zulipterminal/ui_tools/boxes.py
zee-bit: Actually, the command that binds these keys(`right` and `down`) offers a secondary letter-key(i.e. `i` and `j` respectively). Hence, using the commands instead of keys would prevent the letter keys from getting typed in the write-box.

zulipterminal/ui_tools/boxes.py
zee-bit: I can see a typo in the commit description - "diables" in the first line?
But, I can't figure out the typo in the function name. Could you please clarify?

zulipterminal/ui_tools/boxes.py
zee-bit: All of these conditionals have their corresponding `else`/`elif` clauses, so returning early might not give the expected behavior. Let me know if I misunderstood you. :)

tests/ui_tools/test_boxes.py
zee-bit: Well, I tried using `widget_size`, but for some cases, the `write-box` changes to box widget instead of flow and returns a `(maxcol, maxrow)` size instead of just `(maxcol, )`. This made the tests fail for such cases. Let me try to dig more into this and see if I can get a fix for this.

zulipterminal/ui_tools/boxes.py
preetmishra: Yeah, re the latter, it's lose vs loose.

zulipterminal/ui_tools/boxes.py
preetmishra: I understand why you've used them. However, you can see that if the keys change from `keys.py`, it is not going to where. For example, we could have another method for figuring out whether the command keys have letters and take it further from there. In any case, hardcoding shouldn't be our last resort.

zulipterminal/ui_tools/boxes.py
preetmishra: Right, we would have to return early from three places.

I think a cleaner way would be to provision another `elif` and check the stream validity separately.

zulipterminal/ui_tools/boxes.py
zee-bit: Actually, yes they should.
I may have misinterpreted [one of Neil's point from above](https://github.com/zulip/zulip-terminal/pull/864#pullrequestreview-591884868), and thought we needed to handle cycling for the "up" key as well. I'll amend the commit to use only `GO_RIGHT` and `GO_DOWN`. Then, I guess we won't be needing the second commit at all? I'll remove that too and keep a backup of this branch just in case. :)

tests/ui_tools/test_boxes.py
zee-bit: Sorry for the late reply. I got busy with other PR's since finding the root+fix of this particular issue was taking forever. :(
I have tried to dig more on this, but it seems like it's an internal `urwid` issue that I am unable to resolve. While looking at this, I realized that the whole `write_box` is sometimes recognized as "flow widget" for all tests (that's when all the tests pass), but sometimes the `write_box` is determined as "box widget" for all tests (2 tests fail in this case).

I also found that in the latter case, the tests only fail for "DOWN" keypresses, because it can't be handled properly by `super()` i.e. urwid's `container.py`. I can fix the whole issue by not letting the "DOWN" keypress being handled by urwid and returning early after manually handling, but it seems to have other side-effects and does not look clean.

zulipterminal/ui_tools/boxes.py
zee-bit: Would it make sense to define the `ERROR_TEMPLATES` in a separate module?

PR #867, opened 2021-01-06
Handle topic links in message info view

### LINKED ISSUE #709: Handle topic links
Currently, we don't style or linkify [topic links](https://zulipchat.com/api/get-messages). We could potentially style the topic links as we style other in-line message links and extend the linkifiers.

Note: The concerned endpoint was renamed from `subject_links` to `topic_links` in Zulip 3.0 (feature level 1).

Ezio-Sarthak:
This PR is a headstart in handling Topic links in a stream topic.

**Commit flow:**
- Generate link buttons in message info view via a static method (Refactor).
- Add API return types for new and old topic links' responses.
- Convert message responses to support modern formats (in general) via a model helper (Refactor).
- Topic links stored by parsing into `OrderedDict` (like message links).
- Topic links appended in the message info view.

Fixes #709.

View:
Topic links

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -600,6 +600,34 @@ class TestModel:
             assert model.index['pointer'][repr(model.narrow)] == anchor
         assert model._have_last_message[repr(model.narrow)] is True
 
+    @pytest.mark.parametrize('messages, expected_messages_response', [(
+        {'topic_links':   [{'url': 'www.foo.com', 'text': 'Bar'}]},
+        {'topic_links':   [{'url': 'www.foo.com', 'text': 'Bar'}]},
+    ), (
+        {'topic_links':   ['www.foo1.com']},
+        {'topic_links':   [{'url': 'www.foo1.com', 'text': ''}]},
+    ), (
+        {'topic_links': []},
+        {'topic_links': []},
+    ), (
+        {'subject_links':   ['www.foo2.com']},
+        {'topic_links':     [{'url': 'www.foo2.com', 'text': ''}]},
+    ), (
+        {'subject_links': []},
+        {'topic_links':   []},
+    )], ids=[
+        'Zulip_4.0+_ZFL46_response_with_topic_links',
+        'Zulip_3.0+_ZFL1_response_with_topic_links',
+        'Zulip_3.0+_ZFL1_response_empty_topic_links',
+        'Zulip_2.1+_response_with_subject_links',
+        'Zulip_2.1+_response_empty_subject_links',
+    ])
+    def test_modernize_message_response(self, model, messages,
+                                        expected_messages_response):
+
+        assert (model.modernize_message_response(messages)
+                == expected_messages_response)
+
     def test_get_message_false_first_anchor(
             self, mocker, messages_successful_response, index_all_messages,
             initial_data, num_before=30, num_after=10
--- a/zulipterminal/api_types.py
+++ b/zulipterminal/api_types.py
@@ -30,7 +30,12 @@ class Message(TypedDict, total=False):
     timestamp: int
     client: str
     subject: str  # Only for stream msgs.
-    topic_links: List[str]
+    # NOTE: new in Zulip 3.0 / ZFL 1
+    # NOTE: API response format of `topic_links` new in Zulip 4.0 / ZFL 46
+    topic_links: List[Any]
+    # NOTE: `subject_links` new in Zulip 2.1;
+    # Deprecated from Zulip 3.0 / ZFL 1
+    subject_links: List[str]
     is_me_message: bool
     reactions: List[Dict[str, Any]]
     submessages: List[Dict[str, Any]]
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -509,6 +509,11 @@ class Model:
         }
         response = self.client.get_messages(message_filters=request)
         if response['result'] == 'success':
+            response['messages'] = [
+                self.modernize_message_response(msg)
+                for msg in response['messages']
+            ]
+
             self.index = index_messages(response['messages'], self, self.index)
             narrow_str = repr(self.narrow)
             if first_anchor and response['anchor'] != 10000000000000000:
@@ -529,6 +534,35 @@ class Model:
             return ""
         display_error_if_present(response, self.controller)
         return response['msg']
+
+    @staticmethod
+    def modernize_message_response(message: Message
+                                   ) -> Message:
+        """
+        Converts received message into the modern message response format.
+
+        This provides a good single place to handle support for older server
+        releases params, and making them compatible with recent releases.
+
+        TODO: This could be extended for other message params potentially
+        in future.
+        """
+        # (1) `subject_links` param is changed to `topic_links` from
+        # feature level 1, server version 3.0
+        if 'subject_links' in message.keys():
+            message['topic_links'] = message.pop('subject_links')
+
+        # (2) Modernize `topic_links` old response (List[str]) to new response
+        # (List[Dict[str, str]])
+        if 'topic_links' in message.keys():
+            topic_links = [
+                {'url': link, 'text': ''} for link in message['topic_links']
+                if type(link) == str
+            ]
+            if topic_links:
+                message['topic_links'] = topic_links
+
+        return message
 
     def fetch_message_history(self, message_id: int,
                               ) -> List[Dict[str, Union[int, str]]]:
@@ -986,7 +1020,7 @@ class Model:
         Handle new messages (eg. add message to the end of the view)
         """
         assert event['type'] == "message"
-        message = event['message']
+        message = self.modernize_message_response(event['message'])
         # sometimes `flags` are missing in `event` so initialize
         # an empty list of flags in that case.
         message['flags'] = event.get('flags', [])
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -661,6 +661,9 @@ class MessageBox(urwid.Pile):
         self.message_links: 'OrderedDict[str, Tuple[str, int, bool]]' = (
             OrderedDict()
         )
+        self.topic_links: 'OrderedDict[str, Tuple[str, int, bool]]' = (
+            OrderedDict()
+        )
         self.time_mentions: List[Tuple[str, str]] = list()
         self.last_message = last_message
         # if this is the first message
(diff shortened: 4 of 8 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1215,34 +1216,38 @@ def __init__(self, controller: Any, msg: Message, title: str,
...
+            )
 
-        super().__init__(controller, widgets, 'MSG_INFO', popup_width, title)
+            display_attr = None if index % 2 else 'popup_contrast'
+            link_widgets.append(
+                MessageLinkButton(self.controller, caption, link, display_attr)
+            )
+        self.widgets = (self.widgets[:self.slice_index] + link_widgets
+                        + self.widgets[self.slice_index:])
+        self.popup_width = max(self.popup_width, link_width)
```
neiljp, line 1258: You're doing a lot of setting of attributes here, but setting the values before calling - that suggests it may be clearer passing things in as parameters and/or back as return values. In this case can we leave these two lines in `__init__` and simplify this?

tests/conftest.py
```diff
@@ -280,7 +280,7 @@ def unicode_emojis():
     'reactions': [],
     'type': 'private',
     'avatar_url': 'dummy_avatar_url',
-    'subject_links': [],
+    'topic_links': [],
```
neiljp, line 283: We should check this in any case, as I'm not sure why a group-PM should have a topic? Same above.

tests/conftest.py
```diff
@@ -222,7 +222,7 @@ def unicode_emojis():
     'stream_id': 205,
     'subject': 'Test',
     'reactions': [],
-    'subject_links': [],
+    'topic_links': [],
```
neiljp, line 225: Note that we aim to support Zulip 2.1+, which is prior to feature level 1, so we need to support subject_links as well as topic_links.

Given the text below, you might consider whether we can refactor more tests to use the same message templates - so we don't need to make as many changes in multiple places.

zulipterminal/ui_tools/boxes.py
```diff
@@ -578,6 +582,14 @@ def need_recipient_header(self) -> bool:
         else:
             raise RuntimeError("Invalid message type")
 
+    def _set_topic_links(self) -> None:
+        if 'topic_links' not in self.message.keys():
+            return
+        for link in self.message['topic_links']:
+            self.topic_links[link] = (
+                'Link', len(self.topic_links) + 1, True
```
neiljp, line 590: Not having the link text available is a little frustrating :frowning_face:

Perhaps the backend could be improved in this regard, as at least for linkifiers like we regularly use, we know there is "link text", like `#T867` here.

For now could we maybe special-case the view for empty link text and put it on one line?

neiljp (commented):
@Ezio-Sarthak Thanks for looking into this - when we have the open-in-browser feature this will be really helpful :+1:

This works OK for me on czo but I think it needs generalizing for Zulip 2.1+, and could be simpler both in code and possibly in commit structure too. Let me know if you have any questions.

tests/conftest.py
Ezio-Sarthak: Yeah, I ran the tests without subject_links in pm_template and group_pm_template, and as expected, there was no error.

Maybe I'll mention this in the commit text as well. Thanks for pointing out :)

tests/conftest.py
Ezio-Sarthak: For now, I've added `topic_links` wherever we have `subject_links` so that there is no version support issue. This is keeping in mind my reply to the comment below.

neiljp (commented):
@Ezio-Sarthak I just realized I had an outstanding set of responses to this PR so sending them now - let me know if they still make sense :+1:

zulipterminal/ui_tools/views.py
```diff
@@ -1219,31 +1220,42 @@ def __init__(self, controller: Any, msg: Message, title: str,
...
+        if message_links:
+            msg_link_widgets, msg_link_width = self.set_links(message_links)
+            widgets = (widgets[:slice_index] + msg_link_widgets
                        + widgets[slice_index:])
-            popup_width = max(popup_width, message_link_width)
+            popup_width = max(popup_width, msg_link_width)
 
         super().__init__(controller, widgets, 'MSG_INFO', popup_width, title)
 
+    def set_links(
```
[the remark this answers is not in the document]
Ezio-Sarthak: Actually, it uses `self.controller`, but that could be passed as a parameter too. (Since we're not modifying the class instance anyways :+1:) I'll update this in the coming iteration. Thanks for observing it :sweat_smile: 

Re the name, yeah I'll be naming it something like `get_link_buttons()`.

tests/conftest.py
```diff
@@ -246,7 +247,6 @@ def unicode_emojis():
     'reactions': [],
     'type': 'private',
     'avatar_url': 'dummy_avatar_url',
-    'subject_links': [],
```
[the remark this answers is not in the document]
Ezio-Sarthak: Ideally, this shouldn't be the case :thinking:, as there's no such 'topic' in PMs (or group PMs) as you mentioned earlier. This could be potentially filed in the main repo, it not before, in any case.

tests/conftest.py
Ezio-Sarthak: Makes sense. My suggestion would be that we let `subject_links` stay untouched for now in the tests and fixtures. We're validating for both `*_links` parameter in the helper of later commit, in such a way that only one could possibly exist (I don't know why I didn't consider that for this commit :laughing:)

zulipterminal/api_types.py
```diff
@@ -14,6 +14,7 @@ class Message(TypedDict, total=False):
     timestamp: int
     client: str
     subject: str  # Only for stream msgs.
+    subject_links: List[str]  # To support Zulip 2.1+
```
Ezio-Sarthak, line 17: I let here both types stay. Since this is a response dict template, it should contain all the supported fetchable parameters IMO :)

Ezio-Sarthak:
@neiljp Thanks for the review! Your outstanding comments made us discuss client + server related aspects of the PR in detail :)

The motive of this PR would get complete potentially once @sumanthvrao's PR [zulip/zulip#17124](https://github.com/zulip/zulip/pull/17124) gets merged. Till then, we could resolve some related refactor/code related aspects :sweat_smile:

neiljp (commented):
@Ezio-Sarthak Thanks for continuing to work on this :+1:

This looks to work fine, though I do think that having a "message modernization" method when receiving messages will probably be useful in general - for example with 'subject' and 'topic' themselves too! Having an accessor method is generally good over just accessing the data, but the UI already has the Message right now so get_topic_links isn't really an accessor so much as asking the model to simplify the data on-demand.

zulipterminal/model.py
```diff
@@ -438,6 +438,18 @@ def fetch_custom_emojis(self) -> NamedEmojiData:
         display_error_if_present(response, self.controller)
         return custom_emojis
 
+    def get_topic_links(self, message: Message) -> List[str]:
```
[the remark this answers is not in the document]
Ezio-Sarthak: Makes sense. That could be used potentially for modernizing more than one params in general! Thanks for the suggestion :)

zulipterminal/model.py
```diff
@@ -481,6 +481,11 @@ def get_messages(self, *,
         }
         response = self.client.get_messages(message_filters=request)
         if response['result'] == 'success':
+            # Convert message reponse to the modern format
+            response['messages'] = self.modernize_message_response(
+                response['messages']
+            )
```
neiljp, line 487: I'm not sure the comment is necessary - it duplicates the function name? :)

Note that we need to call this for all new messages, both on explicit message fetch and for message events.

tests/model/test_model.py
```diff
@@ -612,6 +612,21 @@ def test_success_get_messages(self, mocker, messages_successful_response,
...
         assert model._have_last_message[repr(model.narrow)] is True
 
+    @pytest.mark.parametrize('messages, expected_messages_response', [(
+        [{'subject_links': ['www.foo1.com', 'www.bar1.com']}],
+        [{'topic_links':   ['www.foo1.com', 'www.bar1.com']}]), (
+        [{'topic_links':   ['www.foo2.com', 'www.bar2.com']}],
+        [{'topic_links':   ['www.foo2.com', 'www.bar2.com']}]),
+    ], ids=[
+        'param_old_server_release',
+        'param_new_server_release',
```
neiljp, line 622: It may be worth adding some empty cases too? The code should already handle it, and the test will then specify that it should.

zulipterminal/ui_tools/boxes.py
```diff
@@ -728,6 +729,19 @@ def need_recipient_header(self) -> bool:
         else:
             raise RuntimeError("Invalid message type")
 
+    def _set_topic_links(self) -> None:
```
neiljp, line 732: Do you expect to expand upon this in future or call it in more than one place? If not then right now this appears to be a dict comprehension, so I think we can have it inline at the one call site?

zulipterminal/ui_tools/views.py
```diff
@@ -1267,7 +1269,18 @@ def __init__(self, controller: Any, msg: Message, title: str,
...
+                self.create_link_buttons(controller, topic_links)
+            )
+
+            widgets = (widgets[:slice_index] + topic_link_widgets
+                       + widgets[slice_index:])
+            popup_width = max(popup_width, topic_link_width)
+
         if message_links:
+            if topic_links:
+                slice_index += (len(topic_links) + 2)
```
neiljp, line 1283: Is this simplified if the slice_index is in the conditional in the earlier commit?

zulipterminal/model.py
```diff
@@ -502,6 +507,23 @@ def get_messages(self, *,
         display_error_if_present(response, self.controller)
         return response['msg']
 
+    @staticmethod
+    def modernize_message_response(response: List[Message]) -> List[Message]:
+        """
+        Cleans recieved messages into the modern format
```
neiljp, line 513: This docstring could be more helpful - why is it useful/helpful?

Also "receive".

neiljp (commented):
@Ezio-Sarthak The modernization pass makes things cleaner :+1:

tests/model/test_model.py
Ezio-Sarthak: By empty do you mean `list()` or `None`? We don't have a `None` value response for `topic_links` from API call anyways. That said I'll add some empty list tests. Thanks for pointing out :)

Ezio-Sarthak:

Ezio-Sarthak:

neiljp:
@Ezio-Sarthak Thanks for this! I just merged with the remaining few changes left over after the last back and forth as discussed :tada:
(mainly removing the need for `.keys()` on a dict and simplifying some text/comments)

PR #868, opened 2021-01-06
Time format setting

### LINKED ISSUE #770: Introduce application-wide time format setting
Currently, we default to the 24-hour time format in Zulip Terminal. It would be reasonable to introduce an application-wide setting to select between the 12-hour and the 24-hour time format. 

For v1, we might want to introduce it as a custom `zuliprc` setting. (In the web app, it is shown under `Settings > Display settings`.)

Lastly, we would want to discuss which time format should be the _new_ default.

mkp6781:
Initial time format settings are fetched during initial register call.
Event handler enables update of this setting as events are recieved
from the server.

Fixes #770

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1,4 +1,5 @@
 import datetime
+import itertools
 from collections import OrderedDict, defaultdict
 from typing import Any, Dict
 
@@ -1778,10 +1779,6 @@ class TestMessageBox:
             'timestamp': 1532103879,
         }
     ])
-    @pytest.mark.parametrize('current_year', [2018, 2019, 2050],
-                             ids=['now_2018', 'now_2019', 'now_2050'])
-    @pytest.mark.parametrize('starred_msg', ['this', 'last', 'neither'],
-                             ids=['this_starred', 'last_starred', 'no_stars'])
     @pytest.mark.parametrize('expected_header, to_vary_in_last_message', [
         (['alice', ' ', 'DAYDATETIME'], {'sender_full_name': 'bob'}),
         ([' ', ' ', 'DAYDATETIME'], {'timestamp': 1532103779}),
@@ -1789,6 +1786,10 @@ class TestMessageBox:
     ], ids=['show_author_as_authors_different',
             'merge_messages_as_only_slightly_earlier_message',
             'dont_merge_messages_as_much_earlier_message'])
+    @pytest.mark.parametrize('current_year', [2018, 2019, 2050],
+                             ids=['now_2018', 'now_2019', 'now_2050'])
+    @pytest.mark.parametrize('starred_msg', ['this', 'last', 'neither'],
+                             ids=['this_starred', 'last_starred', 'no_stars'])
     def test_main_view_content_header_without_header(self, mocker, message,
                                                      expected_header,
                                                      current_year,
@@ -1804,7 +1805,9 @@ class TestMessageBox:
         all_to_vary = dict(to_vary_in_last_message, **stars['last'])
         last_msg = dict(message, **all_to_vary)
         msg_box = MessageBox(this_msg, self.model, last_msg)
-        expected_header[1] = msg_box._time_for_message(message)
+        self.model.formatted_local_time.side_effect = itertools.cycle(
+                                ["Fri Jul 20 21:54", " ", "Thu Jan 01 05:30"])
+        expected_header[1] = "Fri Jul 20 21:54"
         if current_year > 2018:
             expected_header[1] = '2018 - ' + expected_header[1]
         expected_header[2] = '*' if starred_msg == 'this' else ' '
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -1,5 +1,4 @@
 import threading
-import time
 from collections import OrderedDict
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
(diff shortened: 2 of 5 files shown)


mkp6781:

Ezio-Sarthak (commented):
@mkp6781 Thanks for working on this! I recently started reviewing PRs, so pardon me if I run into some confusion :sweat_smile: 

I left some comments, primarily concerning reformatting and visual diffs.

The update of display settings via a new method seems cleaner, and would potentially help in extending for other visual updates as well :+1: 

A follow-up to this might be the addition of tests for the time format function added :)

Ezio-Sarthak:
@zulipbot remove "PR needs review"
@zulipbot add "PR needs update"

neiljp (commented):
@mkp6781 Thanks for giving this a try :+1: This looks like a good start, but it's tricky to review due to the structure, though I can see some good elements in there.

You currently have two commits, which aren't very specific in what they do. The second adds some features, but also moves some code around? While it can seem unnecessary at first, naming commits more specifically gives one an idea of what each is supposed to achieve and narrows down what you include in it. You've extracted the time-formatting method into its own function in the second commit, and a "refactor" commit first which only moves that into the model would be a good first commit. Then changing the behavior of that function based on the initial_data and events means you only have to update the model. For the second commit you could just update that behavior based on the on-loading (initial_data) format preference, and leave event-handling update of that value to a third commit. You could even split the message-list update, which I understand you currently have added in your second commit too, into a fourth commit. These are all improvements of the current situation, and should have value - and be working - after each commit.

It helps to think what your levels of features (or fixes!) are, and in what way you can add them independently to best present them for review and merging, as well as whether you can refactor code first (no externally visible changes) to make it easier to add those features or fixes. While you're working on the code, you can of course have whatever commits you want, but for explaining to others (and so getting reviewed and ultimately merged) it helps to present them cleanly as a flow, like chapters in a book. git interactive rebase is super useful for tidying things up :)

tests/ui/test_ui_tools.py
```diff
@@ -1804,7 +1811,9 @@ def test_main_view_content_header_without_header(self, mocker, message,
         all_to_vary = dict(to_vary_in_last_message, **stars['last'])
         last_msg = dict(message, **all_to_vary)
         msg_box = MessageBox(this_msg, self.model, last_msg)
-        expected_header[1] = msg_box._time_for_message(message)
+        self.model.formatted_local_time.side_effect = self.formatted_local_time
+        expected_header[1] = self.model.formatted_local_time(
```
mkp6781, line 1815: Initially a few tests were failing when I replaced the `msg_box._time_for_message(message)` with the mocked function call to `self.model.formatted_local_time`. The `self.model` in the class `TestMessageBox` is a MagicMock object and it only returns a mock object when `self.model.formatted_local_time` is called. As a result, we do not get the required return value of type string.
To rectify this, I have set the `.side_effect` of `self.model.formatted_local_time` to another function `formatted_local_time` within the same class `TestMessageBox`. On doing this, the `formatted_local_time` method within the class is accessed whenever a call to the mocked function `self.model.formatted_local_time` is made. The passed arguments are processed to return the required string in this case.

I have repeated the same changes in `test_popups.py` where a similar issue was occuring.

mkp6781:

tests/ui/test_ui_tools.py
neiljp: We shouldn't be including these duplicates in the tests in any case; that's not a good solution - though I can see your problem.

Previously the test simply used `ctime` internally, however with the big model mock it cannot access the original method. There are two alternatives, in addition to the 'reimplement part of model' approach you have tried already:
* don't mock all of model, only the parts that are required; that could involve quite a refactor before doing this
* set a `return_value` instead of a `side_effect`, custom to each test parametrization; rather than working with timestamps, we then just work with the resulting text we expect to see, and would get from the new function

I'd recommend trying the last option first.

neiljp (commented):
@mkp6781 This commit is getting there for the refactor :+1:

Knowing what to do with tests is tricky, particularly when mocks are involved; I hope my response helps point you in the right direction!

zulipterminal/ui_tools/views.py
```diff
@@ -1177,9 +1176,11 @@ def __init__(self, controller: Any, msg: Message, title: str,
         self.msg = msg
         self.message_links = message_links
         self.time_mentions = time_mentions
+        date_and_time = controller.model.formatted_local_time(
+                                                    msg['timestamp'], True)
 
         msg_info = [
-            ('', [('Date & Time', time.ctime(msg['timestamp'])[:-5]),
+            ('', [('Date & Time', str(date_and_time)),
```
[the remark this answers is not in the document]
mkp6781: I have removed them now.

mkp6781:

neiljp:
@mkp6781 I pushed a small update to this at https://github.com/neiljp/zulip-terminal/commit/25d61a0b51f98308c28e3cf133b7437b3043e7c8 - let me know what you think. I finished tidying some code, which needed a few test updates too.

neiljp:
@mkp6781 I just merged the altered form of this commit manually as indicated on #**zulip-terminal**, which included some minor test and string simplifications :tada:

Let's treat the setting in a different PR now?

@jalajcodes This should open up that other issue now if you're interested.

jalajcodes:
> @mkp6781 I just merged the altered form of this commit manually as indicated on #**zulip-terminal**, which included some minor test and string simplifications 🎉
> 
> Let's treat the setting in a different PR now?
> 
> @jalajcodes This should open up that other issue now if you're interested.

Sorry for replying late, you're talking about this one ryt? #879

PR #869, opened 2021-01-08
Codecov improvements

neiljp:
@sumanthvrao This is a follow-up to the discussion we had in #831 - it doesn't make much difference to the runtime, but this now just runs codecov when running the tests under one version of python - 3.9 (we could pick some fixed element in the python version array instead if we could determine the appropriate syntax).

I also noticed that we can filter the default view to the master branch on the badge and get an improved UI using the `app.` website for Codecov (going back further than 6 months, for example).

@rht Thanks for pointing out the github action for codecov in the API repo PR :)

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -65,5 +65,7 @@ jobs:
         run: pip install .[testing]
       - name: Run tests with pytest
         run: pytest 
-      - name: Run codecov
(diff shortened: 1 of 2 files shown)


sumanthvrao:
@neiljp Looks good, thanks for spinning up a PR for this!

neiljp:
@sumanthvrao Thanks for the original note regarding this! :+1:

PR #870, opened 2021-01-09
boxes: Disable stream change during edit.

### LINKED ISSUE #774: Disable cycling to stream and recipients while editing
It is not possible to edit stream name for stream messages, not the recipient emails for private messages. Thus, we should disable cycling to these boxes.

Abhirup-99:
Partially fixes #774

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -82,6 +82,8 @@ class WriteBox(urwid.Pile):
         # Private message recipient text entry, None if stream-box
         # or not initialized
         self.to_write_box: Optional[ReadlineEdit] = None
+        self.in_edit_message_box = False
+        self.edit_caption: Optional[str] = None
 
         # For tracking sending typing status updates
         self.send_next_typing_update = datetime.now()
@@ -208,8 +210,8 @@ class WriteBox(urwid.Pile):
             write_box.edit_text
         )
 
-    def stream_box_view(self, stream_id: int, caption: str='', title: str='',
-                        ) -> None:
+    def _stream_view(self, stream_id: int, caption: str='', title: str='',
+                     ) -> None:
         self.set_editor_mode()
         self.stream_id = stream_id
         self.recipient_user_ids = self.model.get_other_subscribers_in_stream(
@@ -221,13 +223,6 @@ class WriteBox(urwid.Pile):
             key=primary_key_for_command('AUTOCOMPLETE'),
             key_reverse=primary_key_for_command('AUTOCOMPLETE_REVERSE')
         )
-        self.stream_write_box = ReadlineEdit(edit_text=caption)
-        self.stream_write_box.enable_autocomplete(
-            func=self._stream_box_autocomplete,
-            key=primary_key_for_command('AUTOCOMPLETE'),
-            key_reverse=primary_key_for_command('AUTOCOMPLETE_REVERSE')
-        )
-        self.stream_write_box.set_completer_delims("")
 
         self.title_write_box = ReadlineEdit(edit_text=title)
         self.title_write_box.enable_autocomplete(
@@ -259,6 +254,18 @@ class WriteBox(urwid.Pile):
         ]
         self.contents = write_box
 
+    def stream_box_view(self, stream_id: int, caption: str='', title: str='',
+                        ) -> None:
+        self.stream_write_box = ReadlineEdit(
+            edit_text=caption
+        )
+        self.stream_write_box.enable_autocomplete(
+            func=self._stream_box_autocomplete,
+            key=primary_key_for_command('AUTOCOMPLETE'),
+            key_reverse=primary_key_for_command('AUTOCOMPLETE_REVERSE')
+        )
+        self.stream_write_box.set_completer_delims("")
+        self._stream_view(stream_id, caption, title)
         # Use and set a callback to set the stream marker
         self._set_stream_write_box_style(None, caption)
         urwid.connect_signal(self.stream_write_box, 'change',
@@ -266,7 +273,10 @@ class WriteBox(urwid.Pile):
 
     def stream_box_edit_view(self, stream_id: int, caption: str='',
                              title: str='') -> None:
-        self.stream_box_view(stream_id, caption, title)
+        self.in_edit_message_box = True
+        self.edit_caption = caption
+        self.stream_write_box = urwid.Text(caption)
+        self._stream_view(stream_id, caption, title)
         self.edit_mode_button = EditModeButton(self.model.controller, 20)
 
         self.header_write_box.widget_list.append(self.edit_mode_button)
@@ -539,6 +549,8 @@ class WriteBox(urwid.Pile):
                     self.keypress(size, 'esc')
         elif is_command_key('GO_BACK', key):
             self.msg_edit_id = None
+            self.in_edit_message_box = False
+            self.edit_caption = None
             self.msg_body_edit_enabled = True
             self.send_stop_typing_status()
             self.view.controller.exit_editor_mode()
@@ -576,8 +588,11 @@ class WriteBox(urwid.Pile):
             if self.focus_position == self.FOCUS_CONTAINER_HEADER:
                 if self.to_write_box is None:
                     if header.focus_col == self.FOCUS_HEADER_BOX_STREAM:
-                        stream_name = (header[self.FOCUS_HEADER_BOX_STREAM]
-                                       .edit_text)
+                        if not self.in_edit_message_box:
+                            stream_name = (header[self.FOCUS_HEADER_BOX_STREAM]
+                                           .edit_text)
+                        else:
+                            stream_name = self.edit_caption
                         if not self.model.is_valid_stream(stream_name):
                             invalid_stream_error = (
                                 'Invalid stream name.'
@@ -635,7 +650,10 @@ class WriteBox(urwid.Pile):
             else:
                 self.focus_position = self.FOCUS_CONTAINER_HEADER
             if self.to_write_box is None:
-                header.focus_col = self.FOCUS_HEADER_BOX_STREAM
+                if self.in_edit_message_box:
+                    header.focus_col = self.FOCUS_HEADER_BOX_TOPIC
+                else:
+                    header.focus_col = self.FOCUS_HEADER_BOX_STREAM
             else:
                 header.focus_col = self.FOCUS_HEADER_BOX_RECIPIENT
 
@@ -1593,6 +1611,8 @@ class PanelSearchBox(urwid.Edit):
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         if ((is_command_key('ENTER', key) and self.get_edit_text() == '')
                 or is_command_key('GO_BACK', key)):
+            self.in_edit_message_box = False
+            self.edit_caption = None
             self.panel_view.view.controller.exit_editor_mode()
             self.reset_search_text()
             self.panel_view.set_focus("body")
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -133,9 +138,10 @@ def stream_box_view(self, stream_id: int, caption: str='', title: str='',
 
     def stream_box_edit_view(self, stream_id: int, caption: str='',
                              title: str='') -> None:
-        self.stream_box_view(stream_id, caption, title)
+        self.stream_box_view(stream_id, caption, title, True)
```
prah23, line 141: The boolean being passed by itself does not provide much context on why we're using it, you could perhaps use `*,` to require the parameter's name to be specified along with the value. This should provide context to understand the need for the boolean.

zulipterminal/ui_tools/boxes.py
```diff
@@ -96,17 +98,20 @@ def stream_box_view(self, stream_id: int, caption: str='', title: str='',
...
+            )
+            self.stream_write_box.enable_autocomplete(
+                func=self._stream_box_autocomplete,
+                key=primary_key_for_command('AUTOCOMPLETE'),
+                key_reverse=primary_key_for_command('AUTOCOMPLETE_REVERSE')
+            )
+            self.stream_write_box.set_completer_delims("")
+        else:
+            self.stream_write_box = urwid.AttrMap(urwid.SelectableIcon(
+                                    "Stream: " + caption), None, 'selected')
```
prah23, line 114: Is there any particular reason why you've used `SelectableIcon`?

zulipterminal/ui_tools/boxes.py
```diff
@@ -40,6 +40,8 @@ def __init__(self, view: Any) -> None:
         self.stream_id = None  # type: Optional[int]
         self.recipient_user_ids = []  # type: List[int]
         self.msg_body_edit_enabled = True
+        self.in_edit_message_box = False
```
prah23, line 43: This extra boolean can be avoided by making use of `stream_box_edit_view`, because it seems to indicate a message being edited.

prah23 (commented):
@Abhirup-99 Thanks for working on this 👍

I would suggest restructuring to use `stream_box_edit_view` solely to make the changes instead of the `is_in_edit` state, `self.stream_write_box`'s nature can be overwritten inside `stream_box_edit_view` since the latter is extended from `stream_box_view`. You might have to mention `self.header_write_box` to include the overwritten stream box in the header box, but it would reduce the extra boolean you currently need, considering that we have a separate edit view for such specifications.

zulipterminal/ui_tools/boxes.py
Abhirup-99: Yes that seems to be a better idea, I would you keyword arguments for this purpose. Thanks for pointing this out.

zulipterminal/ui_tools/boxes.py
Abhirup-99: I missed this one, thanks for pointing this out.

Abhirup-99:
> @Abhirup-99 Thanks for working on this 
> 
> I would suggest restructuring to use `stream_box_edit_view` solely to make the changes instead of the `is_in_edit` state, `self.stream_write_box`'s nature can be overwritten inside `stream_box_edit_view` since the latter is extended from `stream_box_view`. You might have to mention `self.header_write_box` to include the overwritten stream box in the header box, but it would reduce the extra boolean you currently need, considering that we have a separate edit view for such specifications.

Thanks for the review, I agree it's a valid concern that this should completely be done in the edit box but if I use `self.header_write_box`instead of the current approach, the corresponding syntax becomes
`python
self.header_write_box.widget_list[0] = .....
`
this would seem a bit more confusing than the current approach to anyone extending it, at least in my view because the entire first element gets overwritten. The current approach grants just seem better to me for the sake of understanding. I guess we can wait for Neil to suggest this part.

prah23:
@Abhirup-99 I meant rewriting the `stream_write_box` to be a `urwid.Text` or so (as appropriate), and then use it in the same `header_write_box` as before. We wouldn't be changing the orientation or structure of the `header_write_box.` I believe the `header_write_box`'s syntax or behavior shouldn't change.
Here's a snippet that might help make my suggestion clear for extending it in the `stream_box_edit_view`.
```python
        self.stream_write_box = urwid.Text(u"")
        self.header_write_box = urwid.Columns(
            [self.stream_write_box, self.title_write_box],
            dividechars=1
        )
```

Abhirup-99 (commented):
> @Abhirup-99 I meant rewriting the `stream_write_box` to be a `urwid.Text` or so (as appropriate), and then use it in the same `header_write_box` as before. We wouldn't be changing the orientation or structure of the `header_write_box.` I believe the `header_write_box`'s syntax or behavior shouldn't change.
> Here's a snippet that might help make my suggestion clear for extending it in the `stream_box_edit_view`.
> 
> ```python
>         self.stream_write_box = urwid.Text(u"")
>         self.header_write_box = urwid.Columns(
>             [self.stream_write_box, self.title_write_box],
>             dividechars=1
>         )
> ```

I got your point on this part, let me try adding a helper function for this change then because even with this approach, there would be code duplication.

Abhirup-99:
@prah23 take a look now

zulipterminal/ui_tools/boxes.py
prah23: I think that's because the type of edit shows a pop-up with options to choose from. It might be better to have the stream non-selectable.

prah23:
@Abhirup-99 It looks well segregated 👍. It seems a little more complicated now though, but I got your point about prioritizing non-redundant code. I'm probably still biased towards the `stream_box_edit_view` extension approach because though it needed registering the stream box again, it seemed simpler to understand for me. We can definitely wait for Neil's opinion, I just wanted to drop in some suggestions, in case they helped.

Also, in your current approach, we wouldn't be needing the `is_in_edit` boolean, would we? Considering that you have separate views now.

prah23:
@Abhirup-99 It looks well segregated 👍. It seems a little more complicated now though, but I got your point about prioritizing non-redundant code. I'm probably still biased towards the stream_box_edit_view extension approach because though it needed registering the stream box again, it seemed simpler to understand for me. We can definitely wait for Neil's opinion, I just wanted to drop in some suggestions, in case they helped.

Also, in your current approach, we wouldn't be needing the is_in_edit boolean, would we? Considering that you have separate views now.

Abhirup-99:
> @Abhirup-99 It looks well segregated . It seems a little more complicated now though, but I got your point about prioritizing non-redundant code. I'm probably still biased towards the stream_box_edit_view extension approach because though it needed registering the stream box again, it seemed simpler to understand for me. We can definitely wait for Neil's opinion, I just wanted to drop in some suggestions, in case they helped.
> 
> Also, in your current approach, we wouldn't be needing the is_in_edit boolean, would we? Considering that you have separate views now.

Yeah it's required while cycling through the tabs.

zulipterminal/ui_tools/boxes.py
```diff
@@ -393,8 +401,11 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             if self.focus_position == self.FOCUS_CONTAINER_HEADER:
                 if self.to_write_box is None:
                     if header.focus_col == self.FOCUS_HEADER_BOX_STREAM:
-                        stream_name = (header[self.FOCUS_HEADER_BOX_STREAM]
-                                       .edit_text)
+                        if not self.in_edit_message_box:
```
Abhirup-99, line 404: @prah23 see this portion.

prah23: @Abhirup-99 I was referring to the `is_in_edit` boolean.

Ezio-Sarthak: We could know that a user is in edit mode on keypress (i.e., `stream_box_edit_view` called). That said, the extra boolean operator could be an unwanted variable (unless you aim to use it for a different purpose) :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -131,11 +120,28 @@ def stream_box_view(self, stream_id: int, caption: str='', title: str='',
...
+        self._stream_view(stream_id=stream_id, title=title)
+
     def stream_box_edit_view(self, stream_id: int, caption: str='',
                              title: str='') -> None:
-        self.stream_box_view(stream_id, caption, title)
+        self.in_edit_message_box = True
+        self.edit_caption = caption
+        self.stream_write_box = urwid.AttrMap(urwid.SelectableIcon(
+                                "Stream: " + caption), None, 'selected')
+        self._stream_view(stream_id=stream_id, title=title)
```
Ezio-Sarthak, line 143: Since the wrapper function's sole purpose lies in setting the box in edit mode, we could redeclare `self.stream_write_box` in the same :)
Also, it would be cleaner if you add the new method `self._stream_view` below the old `self.stream_box_view` instead of renaming it, so as to get rid of the shifted diffs having the same code :)

Ezio-Sarthak (commented):
Thanks for working on this! The UI seems to be on the right path! 

However, there is a scope for improving/simplifying the code logic. Leaving some quick comments below :)

zulipterminal/ui_tools/boxes.py
Abhirup-99: @Ezio-Sarthak The extra boolean value is for autocompletion parts, which is completely detached from this place. That's why it was necessary to have that.

zulipterminal/ui_tools/boxes.py
Abhirup-99: We basically have to modify the widget list in that case, but that would lead to more duplication as we have to edit the `header_write_box ` and then finally `write_box`, which basically would mean half the `stream_write_box` function itself. I preferred the current approach as it leads to the least duplication.
The f string one I definitely agree with.

Ezio-Sarthak: Ah, I misled my above comment. Actually, I was talking about setting the `self.header_write_box` widget list element (stream box) as `Urwid.Text()`, in edit mode. I think this should be straightforward? (i.e., setting this in `self.stream_box_edit_view()`)

zulipterminal/ui_tools/boxes.py
```diff
@@ -131,11 +120,28 @@ def stream_box_view(self, stream_id: int, caption: str='', title: str='',
...
         self.edit_mode_button = EditModeButton(self.model.controller, 20)
-
```
Ezio-Sarthak, line 138: Unwanted diff :)

zulipterminal/ui_tools/boxes.py
Ezio-Sarthak: I agree on a declaration of boolean :+1:. Another approach could be that the user never cycles to the stream box, if in edit mode. Through this, we could skip this conditional check setting `stream_name`. One may think then, that this would throw an error when user tries to send the message, but our `self.stream_write_box` never changes in the whole process (cycling + editing + sending), so it would not be erroneous :)

Abhirup-99: > I agree on a declaration of boolean . Another approach could be that the user never cycles to the stream box, if in edit mode. Through this, we could skip this conditional check setting `stream_name`. One may think then, that this would throw an error when user tries to send the message, but our `self.stream_write_box` never changes in the whole process (cycling + editing + sending), so it would not be erroneous :)
Since the cycling logic is completely detached, it might create more constraints in future implementations. Already the entire logic is a serious of if else, pushing one more on that list doesn't seem to me a good idea.

zulipterminal/ui_tools/boxes.py
Abhirup-99: > Ah, I misled my above comment. Actually, I was talking about setting the `self.header_write_box` widget list element (stream box) as `Urwid.Text()`, in edit mode. I think this should be straightforward? (i.e., setting this in `self.stream_box_edit_view()`)

Any update on header_write_box would trigger the same changes as above.

neiljp:
I see there has been discussion about this - how ready is this?

Abhirup-99:
@neiljp this pr is ready for review. I have removed all conflicts.

neiljp:
This code is merged after simplification as #1161, which was itself merged manually.

PR #872, opened 2021-01-11
Github Actions: fix warnings and dependencies conflicts 

Abhirup-99:


.github/workflows/codeql-analysis.yml
```diff
@@ -10,22 +10,13 @@ jobs:
...
-    - run: git checkout HEAD^2
-      if: ${{ github.event_name == 'pull_request' }}
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v1
       
-      # Override language selection by uncommenting this and choosing your languages
-      # with:
-      #   languages: go, javascript, csharp, python, cpp, java
```
neiljp, line 29: This belongs with the previous section, so we shouldn't change the indent level.

neiljp (commented):
@Abhirup-99 Thanks for looking into this :) I was considering merging just the top of your change to the codeql action (hence my comment inline), but noted the other issue you referenced. So other than the warnings, this has potential to also affect how effectively the codeql action works.

As you pointed out we can fix the Pipfile to resolve the QL installation issue. However, we need to do a lot more to properly fix the Pipfile - and we don't use it elsewhere in CI right now (pip is much faster - though admittedly due to repeated locking with pipenv). Given that, I'd prefer that we override what looks like QL autodetecting the Pipfile, and use the standard approach we use in the other CI action, at least for now.

More generally we can consider retaining support for Pipfile and ensuring it still works reliably, though there are more options out there - we can discuss that in the stream.

.github/workflows/codeql-analysis.yml
Abhirup-99: Removed :).

Abhirup-99:
zulipbot add "PR needs Review".

neiljp:
@Abhirup-99 Your codeql commit still included the pipfile change in .gitignore, so I've removed that and just added the checkout change for now.

CodeQL gives an alternative setting where we can specify the dependency installation commands - could you try that approach in this PR perhaps? We can fix the pipfile in another PR after we've discussed it further as I mentioned above.

Abhirup-99:
Closing as has been superseded by #875 and #876.

PR #874, opened 2021-01-15
Scrollable large single msg view

### LINKED ISSUE #543: Allow scrolling of large single messages
Depending on the size of the terminal window, which in turn limits the size of the message list area, certain large messages cannot be viewed in full.

If only a part of a message is being displayed, then we could probably respond to up/down as scrolling through the message only, or perhaps have slightly different keys to do so? We might not want to jump to the next message by 'accident', which could happen if we use regular keys, and different in-message navigation would solve that. A scrollable 'pop-up' for the full message could be another option here.

The maximum message size should be available in the server repo.

Ezio-Sarthak:
**What does the PR do?!** :

 - Allows scrolling for large single messages 
   **Note: currently true for every message**
 - Provides raw and rendered message view to users.

**Commit flow:**
- refactor: views: Use urwid.Frame for PopUpView.
- boxes: Add members to MessageBox to fetch header, content, footer.
- hotkeys/keys: Add f hotkey for full rendered message popup.
- views: Structure FullRenderedMsgView class.
- core: Add controller helper to show full rendered message on toggle.
- views: Add keypress events for showing/hiding full rendered message.
- model: Add support for fetching a raw message.
- refactor: boxes: Use new model method to fetch a raw message.
- hotkeys/keys: Add r hotkey for full raw message popup view.
- views: Structure FullRawMsgView class.
- core: Add controller helper to show full raw message on toggle.
- views: Add keypress events for showing/hiding full raw message.

Further discussions [here](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Full.20message.20source.20scrollable.20view.20.23T543) on czo :)

Fixes #543.

**Screenshots/GIFs:**
*Full Rendered Message View*
rendered_view
*Full Raw Message View*
raw_message_view

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -853,6 +853,46 @@ class TestModel:
             model = Model(self.controller)
 
     @pytest.mark.parametrize(
+        "response, expected_raw_content, display_error_called",
+        [
+            (
+                {
+                    "result": "success",
+                    "msg": "",
+                    "raw_content": "Feels **great** to be back!",
+                },
+                "Feels **great** to be back!",
+                False,
+            ),
+            (
+                {
+                    "result": "error",
+                    "msg": "Invalid message(s)",
+                    "code": "BAD_REQUEST",
+                },
+                None,
+                True,
+            ),
+        ],
+    )
+    def test_fetch_raw_message_content(
+        self,
+        mocker,
+        model,
+        expected_raw_content,
+        response,
+        display_error_called,
+        message_id=1,
+    ):
+        self.client.get_raw_message.return_value = response
+
+        return_value = model.fetch_raw_message_content(message_id)
+
+        self.client.get_raw_message.assert_called_once_with(message_id)
+        assert self.display_error_if_present.called == display_error_called
+        assert return_value == expected_raw_content
+
+    @pytest.mark.parametrize(
         "initial_muted_streams, value",
         [
             ({315}, True),
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -899,6 +899,27 @@ class TestMsgInfoView:
             time_mentions=list(),
         )
 
+    @pytest.mark.parametrize("key", keys_for_command("FULL_RAW_MESSAGE"))
+    def test_keypress_full_raw_message(self, message_fixture, key, widget_size):
+        msg_info_view = MsgInfoView(
+            self.controller,
+            message_fixture,
+            title="Message Information",
+            topic_links=OrderedDict(),
+            message_links=OrderedDict(),
+            time_mentions=list(),
+        )
+        size = widget_size(msg_info_view)
+
+        msg_info_view.keypress(size, key)
+
+        self.controller.show_full_raw_message.assert_called_once_with(
+            message=message_fixture,
+            topic_links=OrderedDict(),
+            message_links=OrderedDict(),
+            time_mentions=list(),
+        )
+
     @pytest.mark.parametrize(
         "key", {*keys_for_command("GO_BACK"), *keys_for_command("MSG_INFO")}
     )
@@ -918,10 +939,11 @@ class TestMsgInfoView:
         assert self.controller.open_in_browser.called
 
     def test_height_noreactions(self):
-        expected_height = 5
-        # 5 = 1 (date & time) +1 (sender's name) +1 (sender's email)
+        expected_height = 6
+        # 6 = 1 (date & time) +1 (sender's name) +1 (sender's email)
         # +1 (view message in browser)
         # +1 (full rendered message)
+        # +1 (full raw message)
         assert self.msg_info_view.height == expected_height
 
     # FIXME This is the same parametrize as MessageBox:test_reactions_view
@@ -984,9 +1006,9 @@ class TestMsgInfoView:
             OrderedDict(),
             list(),
         )
-        # 11 = 5 labels + 1 blank line + 1 'Reactions' (category)
+        # 12 = 6 labels + 1 blank line + 1 'Reactions' (category)
         # + 4 reactions (excluding 'Message Links').
-        expected_height = 11
+        expected_height = 12
         assert self.msg_info_view.height == expected_height
 
     @pytest.mark.parametrize(
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -10,6 +10,7 @@ from zulipterminal.ui_tools.views import (
     AboutView,
     EditHistoryView,
     EditModeView,
+    FullRawMsgView,
     FullRenderedMsgView,
     HelpView,
     MsgInfoView,
@@ -420,6 +421,87 @@ class TestFullRenderedMsgView:
         super_keypress = mocker.patch(MODULE + ".urwid.Frame.keypress")
 
         self.full_rendered_message.keypress(size, key)
+
+        super_keypress.assert_called_once_with(size, expected_key)
+
+
+class TestFullRawMsgView:
+    @pytest.fixture(autouse=True)
+    def mock_external_classes(self, mocker, msg_box, initial_index):
+        self.controller = mocker.Mock()
+        mocker.patch.object(
+            self.controller, "maximum_popup_dimensions", return_value=(64, 64)
+        )
+        self.controller.model.fetch_raw_message_content = mocker.Mock(
+            return_value="This is a `raw` message content :+1:"
+        )
+        mocker.patch(MODULE + ".MessageBox", return_value=msg_box)
+        # NOTE: Given that the FullRawMsgView just uses the message ID from
+        # the message data currently, message_fixture is not used to avoid
+        # adding extra test runs unnecessarily.
+        self.message = {"id": 1}
+        self.full_raw_message = FullRawMsgView(
+            controller=self.controller,
+            message=self.message,
+            topic_links=OrderedDict(),
+            message_links=OrderedDict(),
+            time_mentions=list(),
+            title="Full Raw Message",
+        )
+
+    def test_init(self, mocker, msg_box):
+        assert self.full_raw_message.title == "Full Raw Message"
+        assert self.full_raw_message.controller == self.controller
+        assert self.full_raw_message.message == self.message
+        assert self.full_raw_message.topic_links == OrderedDict()
+        assert self.full_raw_message.message_links == OrderedDict()
+        assert self.full_raw_message.time_mentions == list()
+        assert self.full_raw_message.header.widget_list == msg_box.header
+        assert self.full_raw_message.footer.widget_list == msg_box.footer
+
+    @pytest.mark.parametrize("key", keys_for_command("MSG_INFO"))
+    def test_keypress_exit_popup(self, key, widget_size):
+        size = widget_size(self.full_raw_message)
+
+        self.full_raw_message.keypress(size, key)
+
+        assert self.controller.exit_popup.called
+
+    def test_keypress_exit_popup_invalid_key(self, widget_size):
+        size = widget_size(self.full_raw_message)
+        key = "a"
+
+        self.full_raw_message.keypress(size, key)
+
+        assert not self.controller.exit_popup.called
+
+    @pytest.mark.parametrize(
+        "key",
+        {
+            *keys_for_command("FULL_RAW_MESSAGE"),
+            *keys_for_command("GO_BACK"),
+        },
+    )
+    def test_keypress_show_msg_info(self, key, widget_size):
+        size = widget_size(self.full_raw_message)
+
+        self.full_raw_message.keypress(size, key)
+
+        self.controller.show_msg_info.assert_called_once_with(
+            msg=self.message,
+            topic_links=OrderedDict(),
+            message_links=OrderedDict(),
+            time_mentions=list(),
+        )
+
+    def test_keypress_navigation(
+        self, mocker, widget_size, navigation_key_expected_key_pair
+    ):
+        size = widget_size(self.full_raw_message)
+        key, expected_key = navigation_key_expected_key_pair
+        super_keypress = mocker.patch(MODULE + ".urwid.Frame.keypress")
+
+        self.full_raw_message.keypress(size, key)
 
         super_keypress.assert_called_once_with(size, expected_key)
 
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -349,6 +349,11 @@ KEY_BINDINGS: 'OrderedDict[str, KeyBinding]' = OrderedDict([
         'help_text': 'Show/hide full rendered message (from message information)',
         'key_category': 'msg_actions',
     }),
+    ('FULL_RAW_MESSAGE', {
+        'keys': ['r'],
+        'help_text': 'Show/hide full raw message (from message information)',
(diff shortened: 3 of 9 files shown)


neiljp:
@Ezio-Sarthak This has been discussed on the stream? To clarify, I think it would be fine to include the "full view" and "raw view" options in the same popup for now, if we could make it clear how to toggle between them.

Ezio-Sarthak:

zee-bit (changes requested):
Thanks for working on this @Ezio-Sarthak. :+1: 

It works pretty well functionally. I am sure after some improvements the `Tab` to switch context in popups would be extended to other popups as well in the future. The title is pretty huge currently and offsets the 'Full message view' text to one side of the header, is there any way to fix this?
Also, the message header inside the popup breaks on to another line, if the topic name is large, it's better to take care of this case as well. for e.g, see my message here: [#test here](https://chat.zulip.org/#narrow/stream/7-test-here/topic/a.20really.2C.20really.2C.20really.2C.20really.20long.20topic.20title.20here.2E.2E.2E.2E.2E.2E/near/1129610)

I really like the style of your PR message above. :+1: 

I've dropped in a few comments in your PR, feel free to discuss further. :)

zulipterminal/ui_tools/views.py
```diff
@@ -1441,3 +1441,97 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        elif is_command_key('GO_UP', key):
+            key = 'up'
+        elif is_command_key('GO_DOWN', key):
+            key = 'down'
+        elif is_command_key('SCROLL_UP', key):
+            key = 'page up'
+        elif is_command_key('SCROLL_DOWN', key):
+            key = 'page down'
+        elif is_command_key('GO_TO_BOTTOM', key):
+            key = 'end'
```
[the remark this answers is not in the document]
Ezio-Sarthak: Nope. I realized I need simply `PopUpView` as base class for content. Thanks :)

zulipterminal/ui_tools/views.py
```diff
@@ -1441,3 +1441,90 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        self.tab_focus = 0  # Initial tab
+
+        self._render_body()  # Responsible render the body
+
+        super().__init__(self.contents['body'][0], (urwid.Pile(header, None)),
+                         focus_part='body')
+
+        max_cols, max_rows = controller.maximum_popup_dimensions()
+
+        self.width = min(max_cols, self.contents['body'][0].width)
```
neiljp, line 1503: This seems to generate narrow popups, even when the base window is quite wide?
For large messages we want to display as much as we can :)

zulipterminal/ui_tools/views.py
```diff
@@ -1441,3 +1441,90 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        """
+        Returns popup height. The popup height is calculated using urwid's
+        .rows method on every widget.
+        """
+        return sum(widget.rows((popup_width, )) for widget in widgets)
+
+    def _render_body(self) -> None:
+        content = PopUpView(self.controller,
+                            [self.tab_contents[self.tab_focus]],
+                            'FULL_MSG_VIEW', 64, self.title)
```
neiljp, line 1522: Could you explain the logic here? What makes 64 special? This explains the narrow popup.

tests/ui_tools/test_popups.py
```diff
@@ -185,6 +186,127 @@ def test_feature_level_content(self, mocker, zulip_version):
         )
 
 
+class TestFullMsgView:
+    @pytest.fixture(autouse=True)
+    def mock_external_classes(self, mocker, msg_box, initial_index):
+        self.controller = mocker.Mock()
+        mocker.patch.object(self.controller, 'maximum_popup_dimensions',
+                            return_value=(64, 64))
```
neiljp, line 194: This seems conveniently similar to the hardcoded "magic" value in the actual code; maybe we should parametrize over things somehow so a test notices, or at least fix that width issue!

tests/ui_tools/test_popups.py
```diff
@@ -185,6 +186,127 @@ def test_feature_level_content(self, mocker, zulip_version):
...
+            (0, 1),
+            (1, 0),
+        ],
+        ids=[
+            'rendered_msg_to_raw_msg',
+            'raw_msg_to_rendered_msg',
+        ]
+    )
+    @pytest.mark.parametrize("tab_key",
+                             keys_for_command("CYCLE_COMPOSE_FOCUS"))
```
neiljp, line 297: This is confusing to read, and in the code - it's not compose related.

neiljp (commented):
@Ezio-Sarthak This works well but I'm a little wary of the hooks into the "tabs" which aren't rendered and use of `CYCLE_COMPOSE_FOCUS`. I'm open to leaving that code in there since it does the switching, but it feels a bit strange, as if you've ripped out some code, which you sort of have?

Perhaps minor, but we might add some kind of check to ensure that we're not fetching the raw message multiple times - which I know the code currently satisfies.

It seems that footlinks don't show in the rendered version?

The rendered text is also indented to the left and right, which looks a little strange perhaps.

It would be good to have `i` exit the view completely, like we do for edit history, if we're to be consistent.

zulipterminal/ui_tools/views.py
Ezio-Sarthak: I initially thought that 64 is the maximum width :sweat_smile:. I'll refactor that using `max_cols` from controller :+1:

zulipterminal/ui_tools/views.py
```diff
@@ -1441,3 +1441,97 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        max_cols, max_rows = controller.maximum_popup_dimensions()
+
+        self.width = min(max_cols, 64)
+
+        height = self.calculate_popup_height(
+            (self.contents['body'][0].body
+             + self.contents['header'][0].widget_list),
+            self.width
+        )
+        self.height = min(max_rows, height)
```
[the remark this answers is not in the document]
Ezio-Sarthak: True. It would be then good to first show the raw content then? (Since mostly the length of raw content is more than the formatted one).

On a side note, I'm a bit skeptical on whether to change the width/height of the popup upon switching. Is this something we want to achieve?

tests/ui_tools/test_popups.py
Ezio-Sarthak: (See my other comment)

zulipterminal/core.py
```diff
@@ -236,6 +237,19 @@ def show_about(self) -> None:
             'area:help'
         )
 
+    def show_full_msg(
```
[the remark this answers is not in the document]
Ezio-Sarthak: Thanks for pointing that out. Initially, it was `show_full_msg_view`, but `view` seemed obvious, so I skipped it. I'll change it to `show_full_message` :+1:

zulipterminal/ui_tools/views.py
zee-bit: `PopUpView` already has both `max_rows` and `max_cols` present in its constructor, we don't need to pass those again, right?

zulipterminal/ui_tools/views.py
```diff
@@ -1441,3 +1441,97 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+
+    @staticmethod
+    def calculate_popup_height(widgets: List[Any],
+                               popup_width: int) -> int:
+        """
+        Returns popup height. The popup height is calculated using urwid's
+        .rows method on every widget.
+        """
+        return sum(widget.rows((popup_width, )) for widget in widgets)
```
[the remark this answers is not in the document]
Ezio-Sarthak: Right. I forgot we could utilize the static method from `PopUpView` in this. Thanks for pointing out :)

zulipterminal/ui_tools/views.py
Ezio-Sarthak: What I'm passing is the `requested_width` (required parameter), as we want to display as much info on that view as we can. Hope I'm able to clarify :sweat_smile:

Ezio-Sarthak:

zulipterminal/model.py
```diff
@@ -603,6 +603,17 @@ def fetch_message_history(self, message_id: int,
...
+    def fetch_raw_message(self, message_id: int) -> str:
+        """
+        Fetches raw message content of a message using its ID.
+        """
+        response = self.client.get_raw_message(message_id)
+        if response['result'] == 'success':
+            return response['raw_content']
+        display_error_if_present(response, self.controller)
+
+        return "Raw content unavailable."
```
preetmishra, line 615: I believe it would be reasonable to return an empty string so that you can check elsewhere for a failure easily.

tests/model/test_model.py
```diff
@@ -723,6 +723,38 @@ def test_fail_get_messages(self, mocker, error_response,
         with pytest.raises(ServerConnectionFailure):
             model = Model(self.controller)
 
+    @pytest.mark.parametrize('response, expected_raw_content', [({
```
preetmishra, line 726: Two things about the commit text:
* We could remove the `refactor:` label as this is not a refactor commit rather it adds a method.
* It's better to explicate what the commit would add/do in the commit title. In this case, we could mention that it would add support for fetching a raw message.

tests/ui_tools/test_popups.py
```diff
@@ -617,7 +635,7 @@ def test_keypress_exit_popup(self, key, widget_size):
         assert self.controller.exit_popup.called
 
     def test_height_noreactions(self):
-        expected_height = 3
+        expected_height = 4           # +1 due to full msg view key
```
preetmishra, line 638: We should add why it is four than saying +1.

zulipterminal/ui_tools/views.py
```diff
@@ -1519,3 +1519,92 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+
+        header = msg_w[:2] + [urwid.Divider()]
+        _content_formatted = urwid.Pile([msg_w[2].original_widget[-1][1]])
+
+        # Get raw message content
+        _content_raw = controller.model.fetch_raw_message(message['id'])
+        _content_raw_w = urwid.Pile([urwid.Text(_content_raw)])
+
+        tab_list = [
+            {'label': 'Raw content',       'content': _content_raw_w},
```
preetmishra, line 1562: The label in the Message Information view says "Full message" but this renders "Raw content" first. I think we should show the "Raw content" later if we end up going the _tabs_ way.

preetmishra: Also, it is not evident whether the user is being presented with "Full message" or "Raw content" in the UI.

preetmishra (commented):
@Ezio-Sarthak Thanks for working on this!: I tested it locally and it seems to function well. :+1: 

However, it is not evident which mode (between "Full message" and "Raw content") is currently active. Also, let's discuss the tab-switching further on CZO.

I have not given it a thorough review for now as the implementation might change based upon our discussion on CZO. That said, I have left comments for things that are likely to stay anyway.

zulipterminal/ui_tools/views.py
```diff
@@ -1687,3 +1687,49 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        ):
+            self.controller.show_msg_info(
+                msg=self.message,
+                topic_links=self.topic_links,
+                message_links=self.message_links,
+                time_mentions=self.time_mentions,
+            )
+            return key
+        elif is_command_key("MSG_INFO", key):
+            self.controller.exit_popup()
```
neiljp, line 1764: I can see why this is necessary, but I'm unclear why we don't use it for the edit history popup, since that seems a close equivalent of this one.

There's no test for this?

neiljp (commented):
@Ezio-Sarthak This is a big PR that's gone through various iterations now - phew!

zulipterminal/ui_tools/boxes.py
```diff
@@ -1407,6 +1410,7 @@ def main_view(self) -> List[Any]:
         content, self.message_links, self.time_mentions = self.transform_content(
             self.message["content"], self.model.server_url
         )
+        self.content = urwid.Text(content)
```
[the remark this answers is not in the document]
Ezio-Sarthak: You mean the local variable `content` only or the class variable `self.content` too?

zulipterminal/ui_tools/views.py
Ezio-Sarthak: Ah. This didn't grab my attention before! I didn't expect this to be the case :sweat_smile:. Thanks for pointing this out! I am actually expecting the opposite. What I expect is exiting of popup using either `FULL_RENDERED_MESSAGE` (say) or `GO_BACK` **only**

zulipterminal/config/keys.py
```diff
@@ -112,6 +112,11 @@ class KeyBinding(TypedDict, total=False):
         'help_text': 'Cycle through recipient and content boxes',
         'key_category': 'msg_compose',
     }),
+    ('SWITCH_TABS', {
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yes. This shouldn't be present in the current iteration, right? :sweat_smile:

zulipterminal/ui_tools/views.py
neiljp: Good discussion! Thanks @preetmishra :)

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
neiljp: We clarified this in PMs - it looks much clearer now :+1:

zulipterminal/config/keys.py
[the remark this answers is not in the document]
neiljp: I think was left from a previous review draft!

zulipterminal/ui_tools/views.py
```diff
@@ -1749,3 +1749,53 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        self.time_mentions = time_mentions
+        max_cols, max_rows = controller.maximum_popup_dimensions()
+
+        # Get rendered message header and footer
+        msg_box = MessageBox(message, controller.model, None)
+
+        # Get raw message content widget list
+        response = controller.model.fetch_raw_message_content(message["id"])
+        if response is None:
+            controller.report_error("Error: Couldn't fetch raw message content")
```
neiljp, line 1792: This is going to be a duplicate of the error handling in the model method?

neiljp:
@Ezio-Sarthak I've merged the first feature, but given the report_error issue wanted to briefly pause here. The secondary reason is that the raw content commit is ripe for refactoring the existing code, though we can do that after this PR of course (but shouldn't forget!)

zulipterminal/ui_tools/views.py
Ezio-Sarthak: Ah yes. I'll simply return in that case :+1:. Thanks for pointing out!

Ezio-Sarthak:
@neiljp Thanks for your patience on this! I just pushed an updated version. :+1:

neiljp:
@Ezio-Sarthak Thanks for the last update with the refactor to use the new model wrapper for the other cases :+1: I'm merging after fixing one small typo in that refactor :tada: 
(However, we should add tests to fix that!)

PR #876, opened 2021-01-15
requirements: Upgrade flake8 from 3.7.9 to 3.8.3.

Abhirup-99:
This allows the use of pipenv without any errors.


Abhirup-99:
Closed in favor of #936.

PR #878, opened 2021-01-16
bugfix: views: Fixed Exit Popup from Stream Members view.

zee-bit:
Follow-up bugfix for merged PR #T856.


neiljp:
@zee-bit Thanks for the prompt update to your own PR :+1: I would have merged as-is, but note that the 'T' prefix in commit text is chat.zulip.org specific, and on GitHub there is no prefix, it is just #856.

PR #880, opened 2021-01-19
Feature: Improve Stream Info Popup

zee-bit:
- [x] Add Stream Members hotkey in README. [merged commit 8ce2edb]
- [x] Add Stream creation date in Stream Info popup [New in Server v4.0, ZFL 30]
- [x] Add Stream email
- [x] bugfix: Fix case when Weekly message count is None
- [x] Add Message retention days [New in server v3.0, ZFL 17]
- [ ] Notification settings [issue #887]

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -217,6 +217,7 @@ def general_stream() -> Dict[str, Any]:
         "stream_weekly_traffic": 0,
         "push_notifications": False,
         "email_address": "general@example.comm",
+        "message_retention_days": 10,
         "subscribers": [1001, 11, 12],
     }
 
@@ -240,6 +241,7 @@ def secret_stream() -> Dict[str, Any]:
         "is_old_stream": True,
         "desktop_notifications": False,
         "stream_weekly_traffic": 0,
+        "message_retention_days": -1,
         "push_notifications": False,
         "subscribers": [1001, 11],
     }
@@ -267,6 +269,7 @@ def streams_fixture(
                 "desktop_notifications": False,
                 "stream_weekly_traffic": 0,
                 "push_notifications": False,
+                "message_retention_days": i + 30,
                 "email_address": f"stream{i}@example.com",
                 "subscribers": [1001, 11, 12],
             }
@@ -742,6 +745,7 @@ def initial_data(
         },
         "twenty_four_hour_time": True,
         "realm_emoji": realm_emojis,
+        "realm_message_retention_days": 74,
         "last_event_id": -1,
         "muted_topics": [],
         "realm_user_groups": [],
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -220,6 +220,73 @@ class TestModel:
             include_subscribers=True,
         )
 
+    @pytest.mark.parametrize(
+        [
+            "to_vary_in_stream_dict",
+            "realm_msg_retention_days",
+            "feature_level",
+            "expect_msg_retention_text",
+        ],
+        [
+            case(
+                {1: {}},
+                None,
+                10,
+                {1: "Indefinite [Organization default]"},
+                id="ZFL=None_no_stream_retention_realm_retention=None",
+            ),
+            case(
+                {1: {}, 2: {}},
+                -1,
+                16,
+                {
+                    1: "Indefinite [Organization default]",
+                    2: "Indefinite [Organization default]",
+                },
+                id="ZFL=16_no_stream_retention_realm_retention=-1",
+            ),
+            case(
+                {2: {"message_retention_days": 30}},
+                60,
+                17,
+                {2: "30"},
+                id="ZFL=17_stream_retention_days=30",
+            ),
+            case(
+                {
+                    1: {"message_retention_days": None},
+                    2: {"message_retention_days": -1},
+                },
+                72,
+                18,
+                {1: "72 [Organization default]", 2: "Indefinite"},
+                id="ZFL=18_stream_retention_days=[None, -1]",
+            ),
+        ],
+    )
+    def test_normalize_and_cache_message_retention_text(
+        self,
+        model,
+        stream_dict,
+        to_vary_in_stream_dict,
+        realm_msg_retention_days,
+        feature_level,
+        expect_msg_retention_text,
+    ):
+        model.stream_dict = stream_dict
+        model.server_feature_level = feature_level
+        model.initial_data["realm_message_retention_days"] = realm_msg_retention_days
+        for stream_id in to_vary_in_stream_dict:
+            model.stream_dict[stream_id].update(to_vary_in_stream_dict[stream_id])
+
+        model.normalize_and_cache_message_retention_text()
+
+        for stream_id in to_vary_in_stream_dict:
+            assert (
+                model.cached_retention_text[stream_id]
+                == expect_msg_retention_text[stream_id]
+            )
+
     @pytest.mark.parametrize("msg_id", [1, 5, set()])
     @pytest.mark.parametrize(
         "narrow",
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -1127,6 +1127,7 @@ class TestStreamInfoView:
         mocker.patch(LISTWALKER, return_value=[])
         self.stream_id = general_stream["stream_id"]
         self.controller.model.server_feature_level = 40
+        self.controller.model.cached_retention_text = {self.stream_id: "10"}
         self.controller.model.stream_dict = {self.stream_id: general_stream}
         self.stream_info_view = StreamInfoView(self.controller, self.stream_id)
 
@@ -1145,37 +1146,65 @@ class TestStreamInfoView:
         )
 
     @pytest.mark.parametrize(
-        ["to_vary_in_stream_data", "server_feature_level", "expected_height"],
+        [
+            "to_vary_in_stream_data",
+            "cached_message_retention_text",
+            "server_feature_level",
+            "expected_height",
+        ],
         [
             case(
                 {"date_created": None},
+                "74 [Organization default]",
                 None,
-                11,
-                id="ZFL=None__no_date_created",
+                12,
+                id="ZFL=None_no_date_created__no_retention_days",
             ),
             case(
                 {"date_created": None},
+                "74 [Organization default]",
+                16,
+                12,
+                id="ZFL<30_no_date_created__ZFL<17_no_retention_days",
+            ),
+            case(
+                {"date_created": None, "message_retention_days": 200},
+                "200",
+                17,
+                12,
+                id="ZFL<30_no_date_created__ZFL=17_custom_finite_retention_days",
+            ),
+            case(
+                {"date_created": None, "message_retention_days": None},
+                "Indefinite [Organization default]",
                 29,
-                11,
-                id="ZFL<30__no_date_created",
+                12,
+                id="ZFL<30_no_date_created__ZFL>17_default_indefinite_retention_days",
             ),
             case(
-                {"date_created": 1472091253},
+                {"date_created": 1472091253, "message_retention_days": 31},
+                "31",
                 30,
-                12,
-                id="ZFL=30__with_date_created",
+                13,
+                id="ZFL=30_with_date_created__ZFL>17_custom_finite_retention_days",
             ),
             case(
-                {"date_created": 1472047124},
+                {"date_created": 1472047124, "message_retention_days": None},
+                "72 [Organization default]",
                 40,
-                12,
-                id="ZFL>30__with_date_created",
+                13,
+                id="ZFL>30_with_date_created__ZFL>17_default_finite_retention_days",
             ),
             case(
-                {"date_created": 1472046489, "stream_weekly_traffic": None},
+                {
+                    "date_created": 1472046489,
+                    "stream_weekly_traffic": None,
+                    "message_retention_days": 60,
+                },
+                "60",
                 50,
-                12,
-                id="ZFL>30__new_stream_with_date_created",
+                13,
+                id="ZFL>30_new_stream_with_date_created__ZFL>17_finite_retention_days",
             ),
         ],
     )
@@ -1183,18 +1212,21 @@ class TestStreamInfoView:
         self,
         general_stream,
         to_vary_in_stream_data,
+        cached_message_retention_text,
         server_feature_level,
         expected_height,
     ):
+        model = self.controller.model
         stream_id = general_stream["stream_id"]
-        self.controller.model.stream_dict = {stream_id: general_stream}
-        self.controller.model.stream_dict[stream_id].update(to_vary_in_stream_data)
-        self.controller.model.server_feature_level = server_feature_level
+        model.stream_dict = {stream_id: general_stream}
+        model.stream_dict[stream_id].update(to_vary_in_stream_data)
+        model.cached_retention_text = {stream_id: cached_message_retention_text}
+        model.server_feature_level = server_feature_level
 
         stream_info_view = StreamInfoView(self.controller, stream_id)
 
         # height = 1(description) + 2(blank lines) + 2(category)
-        # + 3(checkboxes) + [2-4](fields, depending upon server_feature_level)
+        # + 3(checkboxes) + [2-5](fields, depending upon server_feature_level)
         assert stream_info_view.height == expected_height
 
     @pytest.mark.parametrize("key", keys_for_command("COPY_STREAM_EMAIL"))
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -162,6 +162,8 @@ class Model:
             for stream in self.stream_dict.values():
                 stream["date_created"] = None
 
+        self.normalize_and_cache_message_retention_text()
+
         # NOTE: The expected response has been upgraded from
         # [stream_name, topic] to [stream_name, topic, date_muted] in
         # feature level 1, server version 3.0.
@@ -191,6 +193,38 @@ class Model:
         self.twenty_four_hr_format = self.initial_data["twenty_four_hour_time"]
         self.new_user_input = True
         self._start_presence_updates()
+
+    def message_retention_days_response(self, days: int, org_default: bool) -> str:
+        suffix = " [Organization default]" if org_default else ""
+        return ("Indefinite" if (days == -1 or days is None) else str(days)) + suffix
+
+    def normalize_and_cache_message_retention_text(self) -> None:
+        # NOTE: The "message_retention_days" field was added in server v3.0, ZFL 17.
+        # For consistency, we add this field on server iterations even before this
+        # assigning it the value of "realm_message_retention_days" from /register.
+        # The server defines two special values for this field:
+        # • None: Inherits organization-level setting i.e. realm_message_retention_days
+        # • -1: Messages in this stream are stored indefinitely
+        # We store the abstracted message retention text for each stream mapped to its
+        # sream_id in model.cached_retention_text. This will be displayed in the UI.
+        self.cached_retention_text: Dict[int, str] = {}
+        realm_message_retention_days = self.initial_data["realm_message_retention_days"]
+        if self.server_feature_level is None or self.server_feature_level < 17:
+            for stream in self.stream_dict.values():
+                stream["message_retention_days"] = None
+
+        for stream in self.stream_dict.values():
+            message_retention_days = stream["message_retention_days"]
+            is_organization_default = message_retention_days is None
+            final_msg_retention_days = (
+                realm_message_retention_days
+                if is_organization_default
+                else message_retention_days
+            )
+            message_retention_response = self.message_retention_days_response(
+                final_msg_retention_days, is_organization_default
+            )
+            self.cached_retention_text[stream["stream_id"]] = message_retention_response
 
     def get_focus_in_current_narrow(self) -> Union[int, Set[None]]:
         """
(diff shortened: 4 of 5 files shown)


neiljp:
@zee-bit We should have included the README in the previous iteration, so it's good to see it here :+1:

For the stream creation date, it's good to see the check vs server feature level here, but note the type of this property, and we'll want to test against a range of values.

Also note that we may want to adjust the model data of the subscriptions to be uniform across different levels (abbreviated to ZFL in places), and then have the UI just use the data that's guaranteed to be present - ie. place the conditional(s) in the model. Ultimately this could be part of a move to store/expose only `StreamData` rather than the underlying data from the server.

zee-bit:

tests/ui_tools/test_popups.py
```diff
@@ -590,6 +594,7 @@ def test_keypress_navigation(self, mocker, widget_size,
             'subscribers': [],
             'date_created': 1472047124,
             'stream_weekly_traffic': 140,
+            'email_address': 'zt@chat.zulip.org',
```
neiljp, line 601: In practice these email addresses can be very long. A separate test with that kind of example would be useful. For the reason of long emails, I would suggest we consider allowing/forcing the URL onto the next line. For example, try applying your PR to the #announce stream on czo.

zulipterminal/ui_tools/views.py
```diff
@@ -1098,7 +1098,8 @@ def timestamp_to_local_datetime(timestamp: int) -> str:
         member_keys = ', '.join(map(repr, keys_for_command('STREAM_MEMBERS')))
         stream_email = stream['email_address']
 
-        weekly_msg_count = stream['stream_weekly_traffic']
+        weekly_msg_count = ('N/A' if stream['stream_weekly_traffic'] is None
+                            else str(stream['stream_weekly_traffic']))
```
neiljp, line 1107: You're referring to a commit, not a PR, in this commit text.

If `None` is always due to it being just-created, it would be useful to indicate that, and expand N/A to something clearer in any case.

I'd be happy merging this commit sooner, but the tests in the first commit are dependencies of all the others.

zulipterminal/model.py
```diff
@@ -131,6 +131,26 @@ def __init__(self, controller: Any) -> None:
...
+            if feature_level is None or feature_level < 17:
+                stream['message_retention_days'] = None
+            else:
+                stream['message_retention_days'] = (
+                    message_retention_days_response(
+                        *(self.initial_data['realm_message_retention_days'],
+                          True) if stream['message_retention_days'] is None
+                        else (stream['message_retention_days'], False)
+                    )
+                )
```
neiljp, line 152: This is rather confusing to read. Do we need that nested function? The one-liner if/else is not easy to read.

neiljp (commented):
@zee-bit This generally looks good :+1: Comments inline.

tests/ui_tools/test_popups.py
```diff
@@ -550,24 +550,62 @@ def test_keypress_navigation(self, mocker, widget_size,
         super_keypress.assert_called_once_with(size, expected_key)
 
 
+@pytest.mark.parametrize('stream, stream_id, server_feature_level', [
```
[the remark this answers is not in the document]
zee-bit: Yes. I thought it will increase our scope of testing, and potentially detect/eliminate any bug that may get unnoticed, if we tried to localize the parameters inside the class?

README.md
```diff
@@ -196,6 +196,7 @@ source ~/.zshenv
 | Toggle topics in a stream                             | <kbd>t</kbd>                                  |
 | Mute/unmute Streams                                   | <kbd>m</kbd>                                  |
 | Show stream information & modify settings             | <kbd>i</kbd>                                  |
+| Show stream members from stream information box       | <kbd>m</kbd>                                  |
```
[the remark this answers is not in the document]
zee-bit: So, I'll drop this commit from the PR?

zulipterminal/model.py
zee-bit: I'll fix this.
Just to clarify, do you want me to remove the whole function altogether, or just simplify the conditions for calling that function with those parameters?

zulipterminal/ui_tools/views.py
```diff
@@ -1079,6 +1081,19 @@ def __init__(self, controller: Any, stream_id: int) -> None:
         self.controller = controller
         stream = controller.model.stream_dict[stream_id]
 
+        def timestamp_to_local_datetime(timestamp: int) -> str:
+            local_timezone = tzlocal.get_localzone()
+            local_time = datetime.fromtimestamp(timestamp, local_timezone)
+
+            formatted_datetime = local_time.strftime("%B %d, %Y")
+            return formatted_datetime
```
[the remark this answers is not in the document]
zee-bit: Hmm. Seems like the class parametrization was not a very good idea afterall. But, I am not sure how else to add tests to this commit without making the subsequent commits dependent on it.

If you want to centralize this datetime conversion first, I can put it up on a separate PR inside maybe `helper.py` and then refactor this commit to use that function. If you have any other idea in mind, it would help me to know. :)

zulipterminal/ui_tools/views.py
zee-bit: Adding this commit in the same PR has led to this. I asked in #zulip-terminal about adding this fix in this PR or another PR, and we decided to go with the former. If you want, I can add this to the PR I mentioned few comments above. This should segregate this fix from the class parameters I declared in the prior commit?

tests/ui_tools/test_popups.py
zee-bit: I understand your concern. I'll fix this too.

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
neiljp: The centralized preferred formatting is merged now as a Model method.

zee-bit:
@neiljp PR has been updated.
Using `Overlay` for pyperclip error was looking ugly(and, not in place with our current UI), so I have decided to use the already implemented `NoticeView`, prompting those specific users to install the required utility package for `pyperclip`. Other previous reviews have been addressed as well(except maybe class parameterization, since I wasn't sure if you wanted me to remove those?).

zee-bit:

zee-bit:
On second look, I dont think we use Class parameterization anywhere else in our codebase. So, I have removed it, and moved the parameters to another function that tests `stream_info`'s popup height.

tests/ui_tools/test_popups.py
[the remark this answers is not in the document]
zee-bit: I have removed the class parameterization and moved the fixtures to another test function(`test_popup_height`), since the popup height will change depending on the server version running - which corresponds to the numbers of fields supported in that version. :)

zulipterminal/ui_tools/views.py
```diff
@@ -1180,6 +1180,13 @@ def __init__(self, controller: Any, stream_id: int) -> None:
             else []
         )
 
+        # New in feature level 17, server verison 3.0
+        msg_retention_days = (
+            [("Message retention days", stream["message_retention_days"])]
```
preetmishra, line 1185: See my other comment regarding extracting another variable.

zulipterminal/model.py
```diff
@@ -156,6 +156,32 @@ def __init__(self, controller: Any) -> None:
...
+                        "realm_message_retention_days"
+                    ]
+                    stream["message_retention_days"] = message_retention_days_response(
+                        organization_default, True
+                    )
+                else:
+                    stream["message_retention_days"] = message_retention_days_response(
+                        stream["message_retention_days"], False
+                    )
+
```
preetmishra, line 184: I think we can simplify the logic and conditionals a bit if we structure this differently.

preetmishra (commented):
@zee-bit Thanks for working on this! This seems to work well locally. :+1:

The changes look good overall. Though, since the PR is a bit dated, it would benefit from the skills that you've acquired over the past weeks. :) I have left comments in-line.

tests/ui_tools/test_popups.py
```diff
@@ -755,7 +755,7 @@ def test_keypress_stream_members(self, mocker, key, widget_size):
                     "rendered_description": "hey",
                     "subscribers": [],
                     "date_created": None,
-                    "stream_weekly_traffic": 123,
```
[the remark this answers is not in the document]
zee-bit: As per the commit description, if a stream is newly created (within a week), the `stream_weekly_traffic` would be `None` as per the API docs [here](https://zulip.com/api/get-subscriptions#response).

I was not sure if I should add new tests with this field as `None` or just change the pre-existing tests (to prevent adding unnecessary tests and keep the commit smaller). Let me know if adding new tests is better. :)

The latter did not break the tests because the tests check for popup height, which remains the same even if the field is `None` cause we're displaying an appropriate message in that case.

zulipterminal/model.py
zee-bit: I have taken the `realm_message_retention_days` field out of the conditional which makes it a bit cleaner, but I don't get how it can be simplified further. It already looks pretty straightforward?

Could you please explain what structure you had in mind?

zee-bit:
@preetmishra PR updated. :)

tests/ui_tools/test_popups.py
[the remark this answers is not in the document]
zee-bit: I have added a new fixture with `stream_weekly_traffic` as `None`, without modifying the existing ones.

zulipterminal/core.py
```diff
@@ -338,6 +339,29 @@ def stream_muting_confirmation_popup(self, button: Any) -> None:
         )
         self.loop.widget = PopUpConfirmationView(self, question, mute_this_stream)
 
+    def copy_stream_email(self, stream_email: str) -> None:
+        try:
+            pyperclip.copy(stream_email)
+            clipboard_text = pyperclip.paste()
+            assert clipboard_text == stream_email
```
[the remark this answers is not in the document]
zee-bit: It's difficult to test this since we have an external module involved which makes use of system packages (`xclip`/`xsel` on Linux, and others on different platforms). We can't know the platform on which our tests are running, and if those packages are installed in the system, nor if the text was successfully copied to clipboard.

We can test for the exception case using `side_effects` though, but I don't think that'd be any useful?

neiljp (commented):
@zee-bit This looks useful and functionally appears to work well :+1: (the copy seems to work fine for me!)

I think if we can simplify the tests then this will be a lot easier to work with - and the tests you add ea rly on block the later commits, which I think we discussed previously. So, once they are clean a lot should be mergeable, pending my other feedback.

zulipterminal/core.py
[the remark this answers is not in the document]
preetmishra: I agree with Neil. That is what I was talking about in the first statement.

Re testing, we don't have to test the external module. We could just test whether we're calling the external module correctly.

zulipterminal/model.py
preetmishra: Yeah, that makes it better. Additionally, you could do something like:
```python3
                    stream["message_retention_days"] = message_retention_days_response(
                        realm_message_retention_days, stream["message_retention_days"] is None
                    )
```
That would simplify the `else` block.

zulipterminal/core.py
```diff
@@ -436,6 +437,31 @@ def stream_muting_confirmation_popup(
...
+                (None, ", for copying texts to your clipboard, which could not"),
+                (None, " find a copy/paste mechanism for your system."),
+                (None, " This error should only appear on Linux."),
+                (
+                    None,
+                    "\nYou can fix this by installing any ONE of the "
+                    "copy/paste mechanisms using:\n\n",
+                ),
+                ("msg_code", "sudo apt-get install xclip [Recommended]\n"),
+                ("msg_code", "sudo apt-get install xsel"),
```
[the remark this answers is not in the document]
zee-bit: The first 2 (xclip, xsel) are system packages whereas the other 2 (gtk, PyQt4) are python modules, which still needs the system packages to run so I'm not sure why they are listed in the website. The copying also seems to work fine without these python modules installed.

preetmishra:
@zee-bit Thanks for the update on the feedback. This looks pretty good already! I have left replies to the respective in-line comments. :+1:

Let's work on simplifying the tests as Neil mentioned before moving ahead with the next review iteration.

zulipterminal/model.py
zee-bit: I believe you are talking about the first else block?
The conditional within that have `stream['message_retention_days']` and `realm_message_retention_days` as args to the function call. I don't think they will have the same value at any point of time?

zee-bit:
@neiljp @preetmishra I have updated this PR as per both of your reviews.
Re the `stream_data_to_vary`, I have defined and used the `general_stream` fixture from conftest, of which I'm not sure if it is a bad decision. Let me know if I should instead define a stream data dict in the main test and then build up from fixtures.

neiljp (commented):
@zee-bit This addresses a lot of my structural points, probably most left are the details :+1:

The use of the fixture looks good to me, certainly for now. We may want to follow-up with private variations and similar later, but we can build on this.

zulipterminal/core.py
```diff
@@ -441,6 +442,29 @@ def stream_muting_confirmation_popup(
...
         self.loop.widget = PopUpConfirmationView(self, question, mute_this_stream)
 
+    def copy_to_clipboard(self, text: str, text_category: str) -> None:
+        try:
+            pyperclip.copy(text)
+            clipboard_text = pyperclip.paste()
+            if clipboard_text == text:
+                self.report_success(f"{text_category} copied successfully!")
+            else:
+                self.report_error(f"Could not copy {text_category}. Please try again.")
```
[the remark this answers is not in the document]
zee-bit: I was not sure what to mention as a possible *try*/*hint*, since failing to copy would mean it's an internal `pyperclip` issue. Re-thinking now, I believe we could mention something along the lines of: "Could not copy Test. Error in the dependency library" or just remove the second sentence altogether?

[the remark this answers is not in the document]
neiljp: Possibly worth rewording and avoiding the 'try again' sentence - after all, we *have* copied, we can just warn (strongly) that a pasted value may be incorrect, or something like that? That's more transparent to the user - conceptually they may have some special clipboard system which could upset this, for example? Then they'd know it was copied but a verification didn't work.

zulipterminal/model.py
zee-bit: @neiljp All of the above makes sense. Just had one small doubt; for server versions < 17 do we need to include this field in the `Subscription` data?

zee-bit: If we include this field with a value of `None`, as before, then it'll get mixed with the special `None` value (denoting that it inherits the organization-level setting) since we are not over-writing this field.

preetmishra (commented):
@zee-bit Thanks for the updates! :+1: This continues to work well. I've left some in-line comments.

zulipterminal/core.py
```diff
@@ -441,6 +442,31 @@ def stream_muting_confirmation_popup(
...
 
+    def copy_to_clipboard(self, text: str, text_category: str) -> None:
+        try:
+            pyperclip.copy(text)
+            clipboard_text = pyperclip.paste()
+            if clipboard_text == text:
+                self.report_success(f"{text_category} copied successfully")
+            else:
+                self.report_warning(
+                    f"{text_category} copied, but the clipboard text does not match"
```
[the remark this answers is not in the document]
zee-bit: This might happen if there is an internal issue with the `pyperclip` module. Neil also explained other possible trigger conditions in [this](https://github.com/zulip/zulip-terminal/pull/880#discussion_r682727216) comment.

zulipterminal/core.py
```diff
@@ -441,6 +442,31 @@ def stream_muting_confirmation_popup(
         mute_this_stream = partial(self.model.toggle_stream_muted_status, stream_id)
         self.loop.widget = PopUpConfirmationView(self, question, mute_this_stream)
 
+    def copy_to_clipboard(self, text: str, text_category: str) -> None:
+        try:
+            pyperclip.copy(text)
+            clipboard_text = pyperclip.paste()
+            if clipboard_text == text:
```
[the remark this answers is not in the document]
zee-bit: This was done to address your previous review [here](https://github.com/zulip/zulip-terminal/pull/880#discussion_r679679485), where we decided to not use `assert` statements but simple conditional to check the correctness of the copied text.

tests/conftest.py
```diff
@@ -192,45 +192,52 @@ def logged_on_user():
...
-    "audible_notifications": False,
-    "description": "General Stream",
-    "is_old_stream": True,
-    "desktop_notifications": False,
-    "stream_weekly_traffic": 0,
-    "push_notifications": False,
-    "email_address": "general@example.comm",
-    "subscribers": [1001, 11, 12],
-}
+@pytest.fixture
```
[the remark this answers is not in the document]
neiljp: Some appear to have just shifted to the next commit; let's address these in a different PR?

zulipterminal/core.py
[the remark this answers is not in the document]
preetmishra: Ah, I understand now. I was under the impression that the module would copy correctly or not throw an exception.

neiljp:
@zee-bit For the record, I just merged the first 7 commits as 0e3d91e6bf77986c8fccb7d520412cfd0de777c2 and those prior :tada:

This should allow focus on the last commit by itself :+1:

tests/model/test_model.py
```diff
@@ -220,6 +220,70 @@ def test_register_initial_desired_events(self, mocker, initial_data):
...
+        model.stream_dict = stream_dict
+        model.server_feature_level = feature_level
+        model.initial_data["realm_message_retention_days"] = realm_msg_retention_days
+        for key in to_vary_in_stream_dict:
+            model.stream_dict[key].update(to_vary_in_stream_dict[key])
+
+        model.normalize_and_cache_message_retention_text()
+
+        for key in to_vary_in_stream_dict:
+            assert model.cached_retention_text[key] == expect_msg_retention_text[key]
```
neiljp, line 285: Minor: "key" is very broad; these are stream ids, right?

tests/model/test_model.py
```diff
@@ -220,6 +220,70 @@ def test_register_initial_desired_events(self, mocker, initial_data):
...
+            "feature_level",
+            "expect_msg_retention_text",
+        ],
+        [
+            case(
+                {1: {}},
+                None,
+                10,
+                {1: "Indefinite [Organization default]"},
+                id="ZFL=None_no_stream_retention_realm_retention=None",
```
neiljp, line 236: Ah, this is the `None` case that I queried elsewhere, where the organization value slips through to the stream level.

Was `None` a realm value, though? I assume it followed the original rule, ie. -1 for indefinite and +n for days, so I'm not sure about None?

The #api-design topic doesn't seem to have clarified this exactly, so perhaps Mateusz could answer in a follow-up or you could refer him here?

neiljp (commented):
@zee-bit This last commit is mainly minor points since we've discussed a lot, but the early Zulip version behavior still seems unclear. Other than the existing commit being quite large, having the text extraction in a separate commit will help with focusing anyone else on how we're handling it, without having display code in the way.

tests/model/test_model.py
zee-bit: Right, this is why we were handling the `days == None` case earlier.
To clarify, `None` was a realm value earlier (between server v2-3, as per my testing on the dev server) and it had the same meaning as the current `-1` for indefinite which was changed by @mateuszmandera's [commit](https://github.com/zulip/zulip/commit/7a03e2a7fe259a3410cadbb1a9a4955aa1fae3c7). So to summarize, I believe this was how the server's representation changed over time:
- Server version 2-3: realm value could have `None` for indefinite and +n for days
- Server version 3.0: added stream message retention days, with `None` for org default, `-1` for indefinite and +n for days
- Server version 3+: realm's `None` value was changed to `-1` by Mateusz's above commit.

(I hope Mateusz should get notification from the above mention, still I'll refer him here from the stream to confirm my hypothesis)

mateuszmandera: Ah, that was an interesting dive through old commits :grinning: This overview sounds correct to me based on that. To be a bit more precise, `message_retention_days` was a thing for streams before 3.0 but only in the sense of the backend logic - but it indeed wasn't exposed via the API in any way (the value had to be set manually via `manage.py shell` or db query). I presume that `zulip-terminal` only cares about the API history here though, where indeed this is only a thing since 3.0 - https://github.com/zulip/zulip/commit/c488a35f10ac7b1e5e20fc46302c85ce2364e825

zulipterminal/model.py
zee-bit: Yes, this is resolved with the `cached_retention_text` and the idea to fallback to the realm value on older server versions.

zulipterminal/model.py
```diff
@@ -192,6 +194,37 @@ def __init__(self, controller: Any) -> None:
         self.new_user_input = True
         self._start_presence_updates()
 
+    def message_retention_days_response(self, days: int, org_default: bool) -> str:
+        suffix = " [Organization default]" if org_default else ""
+        return ("Indefinite" if (days == -1 or days is None) else str(days)) + suffix
```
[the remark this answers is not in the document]
zee-bit: I think this should be already answered by the below comment of mine and Mateusz?

neiljp:
@zee-bit Thanks for getting this final commit work done :+1: This now passes CI after the pytz fixup. I'm going to merge this now :tada: (we can follow-up with other features in separate PRs - could you file issues for what we discussed?)

@mateuszmandera Thanks for confirmation of the findings - do we want to improve the API docs for this?

PR #882, opened 2021-01-20
refactor: views: Add year in message info view.

### LINKED ISSUE #879: Message Info Popup: Add Year to date of message
Messages (in the list of messages) show the year of posting (if it is a prior year), but currently we don't seem to be showing any year for the date in the message info popup. In the popup we can always show the year.

jalajcodes:
Added Year in the message info popup.

Fixes #879

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -438,10 +438,13 @@ class TestModel:
         ({'result': 'success'}, True),
         ({'result': 'some_failure'}, False),
     ])
+    @pytest.mark.parametrize('recipients', [
+        ['iago@zulip.com'],
+        ['iago@zulip.com', 'hamlet@zulip.com']
+    ])
     def test_send_private_message(self, mocker, model,
-                                  response, return_value,
-                                  content="hi!",
-                                  recipients="notification-bot@zulip.com"):
+                                  recipients, response, return_value,
+                                  content="hi!"):
         self.client.send_message = mocker.Mock(return_value=response)
 
         result = model.send_private_message(recipients, content)
@@ -452,6 +455,12 @@ class TestModel:
         assert result == return_value
         self.display_error_if_present.assert_called_once_with(response,
                                                               self.controller)
+
+    def test_send_private_message_with_no_recipients(self, model,
+                                                     content="hi!",
+                                                     recipients=[]):
+        with pytest.raises(RuntimeError):
+            model.send_private_message(recipients, content)
 
     @pytest.mark.parametrize('response, return_value', [
         ({'result': 'success'}, True),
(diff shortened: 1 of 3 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1179,7 +1180,10 @@ def __init__(self, controller: Any, msg: Message, title: str,
         self.time_mentions = time_mentions
 
         msg_info = [
-            ('', [('Date & Time', time.ctime(msg['timestamp'])[:-5]),
+            ('', [('Date & Time', datetime.datetime.fromtimestamp(
```
Ezio-Sarthak, line 1183: This shouldn't need an extra module to import. The `time.ctime` method provides a full string containing the year too.

On a side note, it is good to mention any added module to the dependency files (Pipfile, requirements, setup), if the module isn't listed there :)

Ezio-Sarthak (commented):
Thanks for taking this up :+1: 

As mentioned in the comment below, it could be more simpler code and support wise :)

zulipterminal/ui_tools/views.py
jalajcodes: `datetime` is part of the Python standard library just like `time`. I don't think we need to mention it in the dependency files.

This is what `ctime` gives us, `Wed Jan 26 07:06:55 2021` whereas `datetime` allows us to call `strftime` which let's us format it however we want... `Jan 20 2021, 07:06:55` . 

Latter one is better as Year at the end doesn't feel intuitive. (atleast to me)

Also in the code I noticed a  `TODO` mentioning that maybe in the future we may add 12-hour format support with application wide support for different formats.
 `datetime` is the perfect module for that.

If you want I can change it back to `time.ctime` OR as `time.ctime` is used only two times in this file, I can change the second instance to use `datetime` and then `time` module import can be removed?

Ezio-Sarthak: Ah, yes. Thanks for correcting. `datetime` doesn't need mentioning. (You may take that comment for a module not in standard python library :sweat_smile:)

Good catch. I didn't take that TODO into consideration. Now I get your point.

You may update the PR accordingly :+1: (mentioning the `time` to `datetime` in commit text)

I merely dabble iin reviewing currently. Apologies for that :sweat_smile:

neiljp: Re dependencies: as you've agreed, this isn't an issue.

Note that the message format has the year at the left, to allow it to look aligned when the year is optional, so we use an adjusted form of `ctime` already, in some sense.

There is an open PR I've not got to yet which approaches the 12/24-hour issue, so let's look at that for comparison. Given the dependency of formatting on user setting, it may make sense to place a standard formatter in the model, moved from the MessageBox.

jalajcodes: @neiljp should I change the code back to use ctime?

PR #883, opened 2021-01-21
Handle internal PM links

### LINKED ISSUE #764: TRACKING: Handle message links

Ezio-Sarthak:
This is a follow-up PR from #708. It adds support for handling internal PM links.

I highly appreciate the work of @preetmishra in #708, by which it became simpler and cleaner to work with the helper functions in general :)

Partially fixes #764.

### DIFF
--- a/tests/ui_tools/test_buttons.py
+++ b/tests/ui_tools/test_buttons.py
@@ -439,36 +439,40 @@ class TestMessageLinkButton:
     @pytest.mark.parametrize(
         "link, expected_parsed_link",
         [
-            (
+            case(
                 SERVER_URL + "/#narrow/stream/1-Stream-1",
                 ParsedNarrowLink(
                     narrow="stream", stream=DecodedStream(stream_id=1, stream_name=None)
                 ),
+                id="modern_stream_narrow_link",
             ),
-            (
+            case(
                 SERVER_URL + "/#narrow/stream/Stream.201",
                 ParsedNarrowLink(
                     narrow="stream",
                     stream=DecodedStream(stream_id=None, stream_name="Stream 1"),
                 ),
+                id="deprecated_stream_narrow_link",
             ),
-            (
+            case(
                 SERVER_URL + "/#narrow/stream/1-Stream-1/topic/foo.20bar",
                 ParsedNarrowLink(
                     narrow="stream:topic",
                     topic_name="foo bar",
                     stream=DecodedStream(stream_id=1, stream_name=None),
                 ),
+                id="topic_narrow_link",
             ),
-            (
+            case(
                 SERVER_URL + "/#narrow/stream/1-Stream-1/near/1",
                 ParsedNarrowLink(
                     narrow="stream:near",
                     message_id=1,
                     stream=DecodedStream(stream_id=1, stream_name=None),
                 ),
+                id="stream_near_narrow_link",
             ),
-            (
+            case(
                 SERVER_URL + "/#narrow/stream/1-Stream-1/topic/foo/near/1",
                 ParsedNarrowLink(
                     narrow="stream:topic:near",
@@ -476,27 +480,33 @@ class TestMessageLinkButton:
                     message_id=1,
                     stream=DecodedStream(stream_id=1, stream_name=None),
                 ),
+                id="topic_near_narrow_link",
             ),
-            (SERVER_URL + "/#narrow/foo", ParsedNarrowLink()),
-            (SERVER_URL + "/#narrow/stream/", ParsedNarrowLink()),
-            (SERVER_URL + "/#narrow/stream/1-Stream-1/topic/", ParsedNarrowLink()),
-            (SERVER_URL + "/#narrow/stream/1-Stream-1//near/", ParsedNarrowLink()),
-            (
+            case(
+                SERVER_URL + "/#narrow/foo",
+                ParsedNarrowLink(),
+                id="invalid_narrow_link_1",
+            ),
+            case(
+                SERVER_URL + "/#narrow/stream/",
+                ParsedNarrowLink(),
+                id="invalid_narrow_link_2",
+            ),
+            case(
+                SERVER_URL + "/#narrow/stream/1-Stream-1/topic/",
+                ParsedNarrowLink(),
+                id="invalid_narrow_link_3",
+            ),
+            case(
+                SERVER_URL + "/#narrow/stream/1-Stream-1//near/",
+                ParsedNarrowLink(),
+                id="invalid_narrow_link_4",
+            ),
+            case(
                 SERVER_URL + "/#narrow/stream/1-Stream-1/topic/foo/near/",
                 ParsedNarrowLink(),
+                id="invalid_narrow_link_5",
             ),
-        ],
-        ids=[
-            "modern_stream_narrow_link",
-            "deprecated_stream_narrow_link",
-            "topic_narrow_link",
-            "stream_near_narrow_link",
-            "topic_near_narrow_link",
-            "invalid_narrow_link_1",
-            "invalid_narrow_link_2",
-            "invalid_narrow_link_3",
-            "invalid_narrow_link_4",
-            "invalid_narrow_link_5",
         ],
     )
     def test__parse_narrow_link(
@@ -755,15 +765,16 @@ class TestMessageLinkButton:
     @pytest.mark.parametrize(
         "parsed_link, narrow_to_stream_called, narrow_to_topic_called",
         [
-            (
+            case(
                 ParsedNarrowLink(
                     narrow="stream",
                     stream=DecodedStream(stream_id=1, stream_name="Stream 1"),
                 ),
                 True,
                 False,
+                id="stream_narrow",
             ),
-            (
+            case(
                 ParsedNarrowLink(
                     narrow="stream:topic",
                     topic_name="Foo",
@@ -771,8 +782,9 @@ class TestMessageLinkButton:
                 ),
                 False,
                 True,
+                id="topic_narrow",
             ),
-            (
+            case(
                 ParsedNarrowLink(
                     narrow="stream:near",
                     message_id=1,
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/buttons.py
```diff
@@ -471,6 +494,13 @@ def _switch_narrow_to(self, parsed_link: ParsedNarrowLink) -> None:
             self.topic_name = parsed_link['topic_name']
             self.message = dict(id=parsed_link['message_id'])
             self.controller.narrow_to_topic(self)
+        elif 'pm-with:near' == narrow:
```
neiljp, line 497: Is there such a thing as `pm-with` without a `near`?

zulipterminal/ui_tools/buttons.py
```diff
@@ -273,11 +273,17 @@ def __init__(self, user_id: int, email: str) -> None:
     'stream_name': Optional[str],
 })
 
+DecodedPM = TypedDict('DecodedPM', {
+    'type': str,
```
neiljp, line 277: Can we describe this using `Literal`?

tests/ui_tools/test_buttons.py
```diff
@@ -96,6 +96,20 @@ def test__decode_stream_data(self, stream_data, expected_response):
 
         assert return_value == expected_response
 
+    @pytest.mark.parametrize('pm_data, expected_response', [
+            ('1,2-pm', dict(recipient_ids=[1, 2], type='pm')),
+            ('1,2-group', dict(recipient_ids=[1, 2], type='group')),
```
neiljp, line 101: Is a 1-size group OK? Is pm a synonym for group?

neiljp (commented):
@Ezio-Sarthak I've given this a quick look-over, initially since it was button-related and I moved some tests around. I've not run it yet.

The 'link to this message' code does seem to generate the code you have, but narrow URLs are more flexible than this in general, and other formats may be present in messages due to that. In the webapp itself there are more URL types generated; we may want to check with the server what values to expect. A few others I've come across:
* `<id>-<bot-name>` (eg. 2414-shiny-bot)
* `<id>-user<id>` (eg. 2406-user2406)

The latter seems like a strange choice (ending with email prefix?), but it's definitely widespread; it's unclear to me why these formats exist in addition to a simple eg. 2406/2414-pm/group. Other than the server source we might search czo for reasons.

tests/ui_tools/test_buttons.py
Ezio-Sarthak: Oops, that needs correction. Thanks for pointing out.

neiljp: In principle that sounds likely - but pm and group appear to work in both cases on czo!

zulipterminal/ui_tools/buttons.py
```diff
@@ -277,11 +277,17 @@ def __init__(self, user_id: int, email: str) -> None:
     'stream_name': Optional[str],
 })
 
+DecodedPM = TypedDict('DecodedPM', {
```
neiljp, line 280: We're using the updated form for TypedDict now.

zulipterminal/ui_tools/buttons.py
```diff
@@ -475,6 +499,13 @@ def _switch_narrow_to(self, parsed_link: ParsedNarrowLink) -> None:
             self.topic_name = parsed_link['topic_name']
             self.message = dict(id=parsed_link['message_id'])
             self.controller.narrow_to_topic(self)
+        elif 'pm-with' == narrow:
```
neiljp, line 502: What I meant about my last comment is that I'm fairly sure there are valid pm-with and pm-with:near (equivalent in URLs), so we need to likely handle both cases.

neiljp (commented):
@Ezio-Sarthak There are some conflicts to be resolved and the TypedDict issue, but otherwise this needs to be broadened to include both non-near and near forms.

With the basic structure in place, we can consider the other forms as a follow-up since we haven't had feedback about this.

Ezio-Sarthak:

zulipterminal/ui_tools/buttons.py
```diff
@@ -359,6 +365,22 @@ def _decode_stream_data(encoded_stream_data: str) -> DecodedStream:
...
+        """
+        Returns a dict with PM type and IDs of PM recipients.
+        """
+        # Recipient ids (seperated by `,`) are of prime interest.
+        recipient_ids, *_ = encoded_pm_data.split('-')
+        recipient_ids_list = list(map(int, recipient_ids.split(',')))
+
+        # Currently webapp uses `pm` and `group` suffix interchangeably.
+        # Treat more-than-2 pms to group pms to avoid confusion.
+        pm_type = 'pm' if len(recipient_ids_list) < 3 else 'group'
```
zee-bit, line 379: Are we defining `pm` and `group` properly here?
The URL doesn't include the *current-user's* `user_id`. So, if the length of `recipient_ids_list` is `1` the number of persons involved would be `2`(i.e. *current-user* and another user). This should be called `pm`, right?
Similarly, if the length of `recipient_ids_list` is `>= 2`, then number of persons involved would be `1 + len(recipient_ids_list)`, and this should be called `group`?

Correct me if I am wrong, please. :)

Ezio-Sarthak: You're right! To clarify, the currently exposed formats of url _may or may not_ contain the current user's _user_id_ (manually tested this locally).

I'll add a conditional check to confirm if recipient_ids contain user id, and then calculate length accordingly. That should clear that edge case.

Thanks for poiting that out :)

zee-bit: I think the URL will contain the current user's `user_id` only if the user is conversing with himself?
Or, were you able to discover some other edge-case for this while testing?

Ezio-Sarthak: The Zuilp server has exposed many URL narrow formats as of now (maybe a potential caveat :) ). Try narrowing to a PM by removing your id from the list, you'd still end up narrowing. I discovered that and some more easter eggs during manual testing ;)

neiljp:
@Ezio-Sarthak This seems like it would be a fairly straightforward PR to get in, though:
* this needs rebasing over the black transition?
* this doesn't always narrow to the right message - this may be due to this PR being at a state prior to a fix that we merged later?

preetmishra (commented):
@Ezio-Sarthak Thanks for following up on the internal links work. This works well locally and the changes look good too. :+1: I have left a few minor in-line comments.

zulipterminal/ui_tools/buttons.py
```diff
@@ -375,6 +381,27 @@ def _decode_stream_data(encoded_stream_data: str) -> DecodedStream:
...
+        Returns a dict with PM type and IDs of PM recipients.
+        """
+        # Recipient ids (seperated by `,`) are of prime interest.
+        recipient_ids, *_ = encoded_pm_data.split("-")
+        recipient_ids_list = list(map(int, recipient_ids.split(",")))
+
+        no_of_recipients = len(recipient_ids_list)
+        # Bump no. of recipients tp include current user_id if not already
+        # present
+        if user_id not in recipient_ids_list:
```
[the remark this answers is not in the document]
Ezio-Sarthak: Well, the server has some (strange exposed) URL formats, that includes only the id(s) of the recipient(s) other than the current user, e.g.:
- narrow/pm-with/1-user1
- narrow/pm-with/1-bot-name

Those are the ones I'm aware of as of now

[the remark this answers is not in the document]
Ezio-Sarthak: The above ones are those that don't have `user_id` included. Below are ones that have `user_id` included:

- narrow/pm-with/1,2-pm
- narrow/pm-with/1,2-group
- narrow/pm-with/1,2,3-pm
- narrow/pm-with/1,2,3-group

tests/ui_tools/test_buttons.py
```diff
@@ -464,7 +521,7 @@ def test__decode_message_id(self, message_id, expected_return_value):
         ],
     )
     def test__parse_narrow_link(self, link, expected_parsed_link):
-        return_value = MessageLinkButton._parse_narrow_link(link)
+        return_value = MessageLinkButton._parse_narrow_link(link, 1)
```
preetmishra, line 524: It is not apparent what 1 is. Perhaps we should mandate keyword args with `*`.

zulipterminal/ui_tools/buttons.py
```diff
@@ -561,13 +574,29 @@ def _switch_narrow_to(self, parsed_link: ParsedNarrowLink) -> None:
                 topic_name=parsed_link["topic_name"],
                 contextual_message_id=parsed_link["message_id"],
             )
+        elif "pm-with:near" == narrow:
```
preetmishra, line 577: We should validate whether the PM links are correct like we do for streams.

zulipterminal/ui_tools/buttons.py
```diff
@@ -561,13 +574,29 @@ def _switch_narrow_to(self, parsed_link: ParsedNarrowLink) -> None:
...
+            emails = list(
```
preetmishra, line 578: It would be cleaner to only keep the narrowing part in the method like other conditionals.

preetmishra (commented):
@Ezio-Sarthak I am not sure whether your commits were pushed properly. You seem to have reacted positively to my previous comments but the changes aren't present. That said, this continues to work well. I have left some more in-line comments. :+1:

zulipterminal/ui_tools/buttons.py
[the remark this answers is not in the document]
Ezio-Sarthak: I'll add a note here to refer to the URL formats below in any case :+1:

tests/ui_tools/test_buttons.py
```diff
@@ -422,6 +431,58 @@ def test__decode_stream_data(
 
         assert return_value == expected_response
 
+    @pytest.mark.parametrize(
+        "pm_data, expected_response",
+        [
+            case(
+                "1001,12-pm",
+                dict(type=None, recipient_ids=[1001, 12], recipient_emails=None),
```
prah23, line 439: Since you type these `expected_response` cases as `DecodedPM` in the test parameters below, it would be more appropriate to have these as `DecodedPM` instances rather than `dict` instances (similar to how we have `ParsedNarrowLink` instances as parameter cases in the other tests in this file).

PR #886, opened 2021-01-22
ci: Add cancel workflow.

Abhirup-99:
This cancels any pending workflows from
any previous push.


neiljp:
@Abhirup-99 Thanks for persevering with this :+1: As discussed in the stream, this is now merged as 223ad81818f954ce68348faea49c5af9c72105b9 with an updated commit text :tada:

PR #889, opened 2021-01-25
bugfix: model: Fix incorrect condition in update_message_event.

zee-bit:
Fixed an inaccurate condition in `handle_update_message_event` that assumed `subject` will only be present in event list if the message had a topic/subject edit.


zee-bit:

zee-bit:
From the discussions at: [#zulipterminal>subject always present in event](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/subject.20always.20present.20in.20event).
This PR is incorrect, so there's no need to keep this open, IMO?

I am closing this now, FYI @neiljp. :)

PR #893, opened 2021-01-27
Refactor run.py with DEFAULT_SETTINGS & simple error exits

neiljp:
This is the first part of #841 with some extra work and a few other commits first to help application of that refactor, and to extract DEFAULT_SETTINGS into a central place.

@Abhirup-99 Happy to take feedback :)


neiljp:
@Abhirup-99 Thanks for your initial work on that commit - good motivation to get this file more organized :+1:

PR #902, opened 2021-01-31
refactor:Extract function for message being in a narrow

neiljp:
This could do with a unit test, though it's just a refactor right now and the tests would potentially cover a large state space.

I'm thinking this could be useful in many ways right now already as an available method that could be used in various places.

@zee-bit Could this be useful for #824 if you constructed a fake message dict to pass in?

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -270,6 +270,47 @@ class Model:
             ids = index['mentioned_msg_ids']
         return ids.copy()
 
+    def current_narrow_contains_message(self, message: Message) -> bool:
+        """
+        Determine if a message conceptually belongs to a narrow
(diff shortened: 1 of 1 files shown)


zee-bit:
@neiljp Thanks for opening this PR. This could potentially be useful for #824, considering I was currently constructing a fake `narrow` dict there, instead of `message`. I can definitely see how we can construct a message dict(from `request` and `narrow`) as an argument for this function as well.

However, I still see both the functions to be particularly similar, considering the only difference would be that the function in #824 has its use-case limited to just notifying the user if the message is not from the same narrow, whereas this function could be used in various places that need this particular logic. Is there any other difference that I may have missed?

Issue #428, opened 2019-07-10
Limit rate of actions associated with keys that 'toggle' and send to server


neiljp:
Some keys, such as `+` and `*` (or `ctrl s`) send a message on pressing, and are currently only limited by the rate the keys can be pressed. This likely causes the server to ratelimit ZT, so we should track how frequently this is sent and probably artificially limit the rate that messages can be sent to the server.

Work on this might be related to #427 and general work on improving the resilience in case the server stops responding for any reason (network issues, server issues, ratelimiting).

neiljp:
For the record, this should include repeatedly scrolling beyond the end of known messages, which also hits ratelimits.

zulipbot:
Hello @kaustubh-nair, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

PR #903, opened 2021-01-31
bugfixes: Improve first/last message handling

neiljp:
This is the result of exploring seeing intermittent cases where the `_have_last_message` dict had a `KeyError` in event handling via the new Notice popup.

This PR:
* improves the handling of this dict so that the logic is more consistent and expected
* uses the first/last message tracking to avoid hitting the API on trying to scroll beyond the start/end of the message view

This is a partial fix of #428, in particular https://github.com/zulip/zulip-terminal/issues/428#issuecomment-627187932.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -176,6 +176,9 @@ class Model:
         self.new_user_input = True
         self._start_presence_updates()
 
+    def have_last_message_in_current_narrow(self) -> bool:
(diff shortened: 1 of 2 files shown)


neiljp:
These tests may be fixed by rebasing onto something like #904.

I've run with this branch a little and am slightly concerned that this has or exposes an outstanding issue with movement between eg. stream and topic narrows, but makes it worse since it may limit the scrolling.

Further thoughts welcome!

PR #908, opened 2021-02-04
refactor: run: Moved opening print statements outside main.

Abhirup-99:


neiljp (commented):
@Abhirup-99 Thanks for following up on the points I raised :+1:

The function extraction is definitely a useful refactor. However, this would be clearer if the method extraction was a one-step process with as little modification as possible during that particular commit.

I'd suggest some early preparation refactor commits which set the `zterm` elements during the processing of them, rather than afterwards, so that `print_welcome_message` ultimately can deal with just output of some header and then a value/origin for each settings, with that latter origin being optionally extended - eg. if it came from a different setting (like footlinks vs maximum-footlinks on current new line 315 vs 320). In principle each of those values could look up an indexed description for each setting, which could also be used elsewhere, and we may be able to reduce the setting output to a loop.

Abhirup-99:
Thank you for the review, I would update it based on the suggestions.

Abhirup-99:
@zulipbot add "PR needs review."

Abhirup-99:

Abhirup-99:

zulipterminal/cli/run.py
```diff
@@ -294,6 +309,21 @@ def list_themes() -> str:
...
+    complete, incomplete = complete_and_incomplete_themes()
+    if zterm['theme'][0] in incomplete:
+        print(in_color('yellow',
+                       "   WARNING: Incomplete theme; "
+                       "results may vary!\n"
+                       "      (you could try: {})".
+                       format(", ".join(complete))))
+    zterm.pop('theme')
+    for key in sorted(zterm):
+        print(WELCOME_MESSAGE_TEXT[key](zterm[key]))
```
neiljp, line 324: This is pretty fancy, but the standard line follows something like:
```
"   autohide setting '{}' specified {}.".format(*autohide),
```
On that basis, I think we could massage the second part of zterm to loop through everything, and for theme perhaps have "extra text" which is otherwise `None`?
(all this in a third commit, based on my main feedback)

zulipterminal/cli/run.py
```diff
@@ -430,7 +433,8 @@ def main(options: Optional[List[str]]=None) -> None:
             theme_data = theme_with_monochrome_added(THEMES[theme_to_use[0]])
         else:
             theme_data = THEMES[theme_to_use[0]]
-
+        zterm['theme'] = theme_to_use
+        zterm.pop('notify')
```
neiljp, line 445: This change seems unrelated to this commit?

neiljp (commented):
@Abhirup-99 This is heading in the right direction :+1:

I'm glad to see the two commits, but you've changed the implementation of the welcome message in the same commit. Moving a chunk of code around (eg. into a function) can benefit from making it just that - partially since if you have your .gitconfig (diff colorMoved ?) right, or perhaps use the right tool, then it shows moved lines differently from just added/removed :)

So, I'd suggest you could have at least an extra commit, to simplify the method afterwards. This then might look like:
* tidy the "inputs" (zterm+theme+another? -> just zterm) - as discussed in last review?
* extract the method, which outputs using zterm using the existing code
* simplify the new method (which might also facilitate the `notify` option being output automatically?)

Other than seeing the extraction is just a move of code via git-diff, smaller changes per commit are easier to review :)

It's not always magically clear what to do in these changes - it's often easier to look back and see the flow that makes sense - and I do that myself when writing commits/PRs. Things get moved around, commits end up representing different changes or groups of changes that make sense together. I'll often do a lot of that before pushing; it's just a case of seeing the patterns.

Abhirup-99:
> It's not always magically clear what to do in these changes - it's often easier to look back and see the flow that makes sense - and I do that myself when writing commits/PRs. Things get moved around, commits end up representing different changes or groups of changes that make sense together. I'll often do a lot of that before pushing; it's just a case of seeing the patterns.

Thanks for the insights. I would update the pr shortly.

zulipterminal/cli/run.py
Abhirup-99: I updated the pr to make it simpler. Yeah, looking back I also think this was really fancy and could have been simply avoided.

Abhirup-99:

Abhirup-99:

PR #909, opened 2021-02-05
Remove space to split recipients by ','.

prah23:
Autocomplete now splits on just `','` instead of `', '`.
The users would only have to use a `,` and continue with typing the next recipient and receive suggestions through autocomplete, without having to worry about adding a space.


neiljp:
@prah23 I held off from merging as I wasn't sure if you planned to add a few test cases?

prah23:
@neiljp Yes, I plan on adding test cases, thank you for waiting. I'll update you once this is ready for your review.

Issue #901, opened 2021-01-30
Move from type comments to inline type annotations


neiljp:
This breaks out the last part of #838, which is enabled by the move to python 3.6+.

This could be automated using `com2ann`, though note that tool requires python 3.8+.

zulipbot:
Hello @zulip/server-refactoring members, this issue was labeled with the "area: refactoring" label, so you may want to check it out!

neiljp:
^ This seems like a bug in zulipbot label-checking?

ahmedabuamra:
@zulipbot claim

zulipbot:
Hello @ahmedabuamra!

Thanks for your interest in Zulip! You have attempted to claim an issue without the labels "help wanted", "good first issue". Since you're a new contributor, you can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/zulip-terminal/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) or [good first issue](https://github.com/zulip/zulip-terminal/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22good+first+issue%22) labels.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

ahmedabuamra:
I'm sorry for working on this issue without assignment.
I wanted to try `com2ann` and save you some time. Thanks.

PR #910, opened 2021-02-05
refactor: Move from type comments to inline type annotations.

ahmedabuamra:
Ran `com2ann` in the project folder to auto re-write the files.

Fixes: #901

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -99,10 +99,10 @@ class Model:
         self.controller = controller
         self.client = controller.client
 
-        self.narrow = []  # type: List[Any]
-        self._have_last_message = {}  # type: Dict[str, bool]
+        self.narrow: List[Any] = []
+        self._have_last_message: Dict[str, bool] = {}
         self.stream_id = -1
-        self.recipients = frozenset()  # type: FrozenSet[Any]
+        self.recipients: FrozenSet[Any] = frozenset()
         self.index = initial_index
 
         self.user_id = -1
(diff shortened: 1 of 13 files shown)


zulipterminal/model.py
```diff
@@ -458,10 +460,13 @@ def update_stream_message(self, topic: str, message_id: int,
...
-                                         'type': 'realm_emoji'}
-                         for emoji_code, emoji in response['emoji'].items()
-                         if not emoji['deactivated']}  # type: NamedEmojiData
+        custom_emojis: NamedEmojiData = {
+            emoji['name']: {
+                'code': emoji_code,
+                'type': 'realm_emoji'
+                }
+            for emoji_code, emoji in response['emoji'].items()
+            if not emoji['deactivated']}
```
neiljp, line 469: Did com2ann generate this new layout? We're moving towards a style where large bracketed code have the closing one on the starting line, though the original version isn't ideal in this respect either.

zulipterminal/ui.py
```diff
@@ -94,8 +94,8 @@ def set_typeahead_footer(self, suggestions: List[str],
                              is_truncated: bool) -> None:
         if suggestions:
             # Wrap by space.
-            footer_text = [' ' + s + ' '
-                           for s in suggestions]  # type: List[Any]
+            footer_text: List[Any] = [' ' + s + ' '
+                                      for s in suggestions]
```
neiljp, line 98: Fits on one line?

neiljp (commented):
@ahmedabuamra Welcome to the project and thanks for looking into this :+1:

The com2ann tool is automated so I've not looked in fine detail, but noticed a few cases we may want to look into in terms of formatting, checking the state of lister/run-mypy compared to zulip/zulip, and OrderedDict (as discussed on chat.zulip.org).

This does point to various cases where we could improve the typing - while equivalent I'd prefer `Optional[X]` to `Union[None, X]`, for example - but we can address those in another PR.

zulipterminal/model.py
ahmedabuamra: Meaning to be like that, right?

```
        custom_emojis: NamedEmojiData = {
            emoji['name']: {
                'code': emoji_code,
                'type': 'realm_emoji'
            }
            for emoji_code, emoji in response['emoji'].items()
            if not emoji['deactivated']
        }
```

tools/lister.py
```diff
@@ -69,8 +69,8 @@ def list_files(targets=[], ftypes=[], use_shebang=True, modified_only=False,
     # throw away empty lines and non-files (like symlinks)
     files = list(filter(os.path.isfile, files_gen))
 
-    result_dict = defaultdict(list)  # type: Dict[str, List[str]]
-    result_list = []  # type: List[str]
+    result_dict: Dict[str, List[str]] = defaultdict(list)
+    result_list: List[str] = []
```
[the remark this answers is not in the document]
ahmedabuamra: I understand. `tools/lister.py` is the only changed file in `tools`.

ahmedabuamra:
@zulipbot remote "PR needs update"

ahmedabuamra:

andersk:
To deal with types like `OrderedDict[K, V]` that parse in mypy but don’t evaluate in Python at runtime, you can quote them:

```python
KEY_BINDINGS: "OrderedDict[str, KeyBinding]" = OrderedDict([
    …
])
```

zulipterminal/model.py
neiljp: Right :+1: Also trailing commas, but that's simpler to quickly fix :)

ahmedabuamra:
Thanks @andersk, I updated the PR as you suggested and waiting for @neiljp 's confirmation too.

zulipterminal/model.py
ahmedabuamra: @neiljp Done :)

neiljp:
@ahmedabuamra Thanks for getting this done - I'm merging this now :tada:

PR #915, opened 2021-02-08
Minor typing adjustments to reach Zulip's client expectations

### LINKED ISSUE #885: Make various minor adjustments to match Zulip typing notifications protocol
We now do a reasonable job of handling typing events as documented here:
https://zulip.com/api/set-typing-status

However, there are some points which are not quite handled right, so far including:
* [ ] We don't handle the TYPING_STARTED_EXPIRY_PERIOD case; that means the footer can get stuck if a PM user doesn't send a stop event, which happens if the other user disappears (closes client, falls offline)
* [ ] We potentially send a 'stop' event even if the user hasn't sent a 'start' event, such as when using GO_BACK and not in the msg_write_box

For the first case, while a timeout should handle this, I would expect that closing a ZT client mid-type, which should close/unregister the queue, doesn't trigger sending stop-typing events.

@prah23 You worked on #845 (#884) so you might be interested in this?

prah23:
- The first commit adds a 'state' boolean that indicates whether a typing `start` event has been sent, so as to avoid sending unnecessary `stop` events.
- The second commit adds assertions within the `_handle_typing_event()` tests, to assert on the message being passed to the footer call based on the `op`.
- The third commit adds a duration of 15 seconds to the `set_footer_text()` function call so that the footer doesn't get stuck displaying the 'is typing...' status when it doesn't receive a `stop` event.
 
Fixes #885.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -52,6 +52,7 @@ class TestWriteBox:
         assert write_box.model == self.view.model
         assert write_box.view == self.view
         assert write_box.msg_edit_state is None
+        assert not write_box.sent_start_typing_status
 
     def test_not_calling_typing_method_without_recipients(self, mocker, write_box):
         write_box.model.send_typing_status_by_user_ids = mocker.Mock()
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -1001,6 +1001,9 @@ class Model:
         Handle typing notifications (in private messages)
         """
         assert event["type"] == "typing"
+
+        TYPING_STARTED_EXPIRY_PERIOD = 15
+
         if hasattr(self.controller, "view"):
             # If the user is in pm narrow with the person typing
             narrow = self.narrow
@@ -1012,7 +1015,11 @@ class Model:
                 if event["op"] == "start":
                     user = self.user_dict[event["sender"]["email"]]
                     self.controller.view.set_footer_text(
-                        [" ", ("footer_contrast", user["full_name"]), " is typing..."]
+                        [" ", ("footer_contrast", user["full_name"]), " is typing..."],
+                        # Restore the footer if there hasn't been a corresponding
+                        # typing 'stop' event after a 'start' event for greater
+                        # than 15 seconds.
+                        TYPING_STARTED_EXPIRY_PERIOD,
                     )
                 elif event["op"] == "stop":
                     self.controller.view.set_footer_text()
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -88,6 +88,7 @@ class WriteBox(urwid.Pile):
         self.send_next_typing_update = datetime.now()
         self.last_key_update = datetime.now()
         self.idle_status_tracking = False
+        self.sent_start_typing_status = False
 
         # Constant indices into self.contents
         # (CONTAINER=vertical, HEADER/MESSAGE=horizontal)
@@ -113,13 +114,19 @@ class WriteBox(urwid.Pile):
         self.view.controller.enter_editor_mode_with(self)
 
     def send_stop_typing_status(self) -> None:
-        # Send 'stop' updates only for PM narrows.
-        if self.to_write_box and self.recipient_user_ids:
+        # Send 'stop' updates only for PM narrows, when there are recipients
+        # to send to and a prior 'start' status has already been sent.
+        if (
+            self.to_write_box
+            and self.recipient_user_ids
+            and self.sent_start_typing_status
+        ):
             self.model.send_typing_status_by_user_ids(
                 self.recipient_user_ids, status="stop"
             )
             self.send_next_typing_update = datetime.now()
             self.idle_status_tracking = False
+            self.sent_start_typing_status = False
 
     def private_box_view(
         self,
@@ -201,6 +208,7 @@ class WriteBox(urwid.Pile):
                         self.recipient_user_ids, status="start"
                     )
                     self.send_next_typing_update += start_period_delta
+                    self.sent_start_typing_status = True
                     # Initiate tracker function only if it isn't already
                     # initiated.
                     if not self.idle_status_tracking:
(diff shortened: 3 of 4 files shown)


prah23:
@neiljp I'll also change the tutorial in the documentation for `_handle_typing_event()` if the change seems okay to you.

prah23:
@neiljp I'll also change the tutorial in the documentation for _handle_typing_event() if the change seems okay to you

prah23:

zulipterminal/ui_tools/boxes.py
```diff
@@ -82,6 +82,7 @@ def __init__(self, view: Any) -> None:
         self.send_next_typing_update = datetime.now()
         self.last_key_update = datetime.now()
         self.idle_status_tracking = False
+        self.sent_start_typing_status = False
```
Ezio-Sarthak, line 85: Could we improve this naming? Maybe "start_typing_status_sent"?

Ezio-Sarthak, line 85: Could we improve this naming? Maybe "start_typing_status_sent"?

Ezio-Sarthak (commented):
Thanks for looking into this. This looks good overall :+1:
Note that the 'fixes' text in the first commit will close the issue, so you'll need to rephrase it slightly or just remove it.

zulipterminal/ui_tools/boxes.py
zee-bit: I think the current name is fine.
It all depends on how you break that name while reading. e.g. currently with `sent_start_typing_status` you would break it something like: `sent` - `start_typing_status`, whereas with Sarthak's version of name, something like: `start_typing_status` - `sent`.

Plus, the current name is consistent with a similar naming above: `send_next_typing_update`. :)

prah23: @zee-bit thanks for clarifying, that is what I had in mind.
I've also named it in a way that's easy to read, `sent_typing_start_status` meaning that we've "sent the `start` status". If it still seems like it can be improved, I'm happy to make it more readable. :)

Rohitth007:
@prah23 Hi, could you clarify what causes the error in the first commit?
Because the async function would never start in the first place if the `start` op was never sent ryt?

prah23:
@Rohitth007 The problem the first commit is trying to solve is sending unnecessary `stop` events without prior `start` events.
Currently, `stop` events are sent when
- The user has been idle for more than `TYPING_STOPPED_WAIT_PERIOD` (5) seconds.
- The user sends a message.
- The user exits the compose box.

The async function I presume you're referring to is the `track_idleness_and_update_status` method, which handles only the first point above by waiting for the specified idle time. 

The second point can send a `stop` event when say the user types a message and then remains idle for a while (let's just say 20 seconds) and then sends the message he typed without typing anything more at that point. An unnecessary `stop` event is sent because the server has already been notified of the typing being stopped after 5 seconds of inactivity.

For the third point, for example, the user opens the compose box, does nothing, and quits the compose box, which again is another way in which we hit the API unnecessarily.

To avoid calling the API this way, I've used a state boolean, `sent_start_typing_status` so we know whether or not to send a `stop` event.

Rohitth007:
@prah23 ooh yes, got it makes sense, thanks for the detailed explanation, the code looks good :+1:

zulipterminal/ui_tools/boxes.py
```diff
@@ -108,13 +109,15 @@ def set_editor_mode(self) -> None:
 
     def send_stop_typing_status(self) -> None:
         # Send 'stop' updates only for PM narrows.
```
Ezio-Sarthak, line 111: Maybe we could extend this note to include the conditional check below?

zulipterminal/model.py
```diff
@@ -858,6 +858,11 @@ def _handle_typing_event(self, event: Event) -> None:
         Handle typing notifications (in private messages)
         """
         assert event['type'] == "typing"
+
+        # Restore the footer if there hasn't been a corresponding typing
```
preetmishra, line 862: This comment doesn't seem to belong here. You could add it prior to the next change.

zulipterminal/ui_tools/boxes.py
preetmishra: Let's amend the test associated with it as well in the corresponding `test_init`.

zulipterminal/model.py
```diff
@@ -869,7 +874,7 @@ def _handle_typing_event(self, event: Event) -> None:
                         ' ',
                         ('code', user['full_name']),
                         ' is typing...'
-                    ])
+                    ], TYPING_STARTED_EXPIRY_PERIOD)
```
preetmishra, line 877: We could amend the existing test to check whether the method was called with the required args.

preetmishra (commented):
@prah23 Thanks for working on this. This looks good mostly. :+1: Let's rebase and resolve minor in-line comments.

prah23:
Thanks for the review @preetmishra! I've accommodated all your suggestions. I've added an extra commit to assert on the footer notification to be displayed, and then modified it to also include the expiry period in the last commit.

tests/ui_tools/test_boxes.py
```diff
@@ -52,6 +52,7 @@ def test_init(self, write_box):
         assert write_box.model == self.view.model
         assert write_box.view == self.view
         assert write_box.msg_edit_state is None
+        assert not write_box.sent_start_typing_status
```
prah23, line 55: This was my understanding of your suggestion to amend `test_init`, is this what you had in mind? If yes, should I add another commit asserting on other attributes initiated in the `write_box`?

Rohitth007:
@prah23 While working on the typing animation PR, I had realized that the edge case of animation being stuck needed the defining of TYPING_STARTED_EXPIRY_PERIOD which I had borrowed from your PR but it had to be treated differently than this PR because it is asynchronous. Hence we would have to discuss how to go about these changes (commit 2 and 3)

prah23:
@Rohitth007 sure, what do you suggest?

Rohitth007:
I think there is no point adding them and redoing them in the other PR. They are small changes anyway right?

neiljp (commented):
@prah23 I've merged the first commit as-is, which was the bugfix unique to this PR now? If so, is this PR good to close?

You mentioned about updating the docs for this, which could come in a separate PR?

prah23:
@neiljp thanks!

The docs update was related to the second and the third commit, because the tutorial we have speaks about `_handle_typing_event()`, should I let @Rohitth007 take that up instead since we're picking the update to that method off of his PR?

neiljp:
@prah23 That sounds relevant - ok with you @Rohitth007 ?

PR #916, opened 2021-02-12
autocomplete: Reposition cursor before autocomplete in message box.

### LINKED ISSUE #899: Reposition cursor before autocompleting
While autocompleting, the position of the cursor is where the autocompleted text is added, appending the text after the cursor to the autocompleted text, resulting in inconsistent text. 
For example, `#**zulip-ter|min` is autocompleted to `#**zulip-terminal**min` due to the cursor being placed after `r`.

We should reposition the cursor before we autocomplete on the text.

ahmedabuamra:
Setting the cursor position to the next white space or newline if exists,
otherwise move it to the end of the message body.

Fixes: #899

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -631,6 +631,22 @@ class WriteBox(urwid.Pile):
                 header.focus_col = self.FOCUS_HEADER_BOX_STREAM
             else:
                 header.focus_col = self.FOCUS_HEADER_BOX_RECIPIENT
+        elif is_command_key('AUTOCOMPLETE', key) and hasattr(
+                            self, 'msg_write_box'):
+            if self.focus_position == self.FOCUS_CONTAINER_MESSAGE:
+                text_box = self.msg_write_box if hasattr(
+                    self, 'focus_position') else None
+            elif self.focus_position == self.FOCUS_CONTAINER_HEADER:
+                header = self.header_write_box
+                text_box = header[header.focus_col]
+            if text_box:
+                text_len = len(text_box.edit_text)
+                new_edit_pos = text_len
+                for i in range(text_box.edit_pos, text_len):
+                    if text_box.edit_text[i] in [' ', '\n']:
+                        new_edit_pos = i
+                        break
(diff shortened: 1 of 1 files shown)


ahmedabuamra:

zulipterminal/ui_tools/boxes.py
```diff
@@ -631,6 +631,22 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                 header.focus_col = self.FOCUS_HEADER_BOX_STREAM
             else:
                 header.focus_col = self.FOCUS_HEADER_BOX_RECIPIENT
+        elif is_command_key('AUTOCOMPLETE', key) and hasattr(
+                            self, 'msg_write_box'):
```
neiljp, line 635: Is this to handle the case of the compose box not being used yet?

neiljp (commented):
@ahmedabuamra There's not too much code, though this does present an improvement it seems :+1: In general I'd like to see some tests to validate the behavior here, ie. given some text in various positions with the cursor at different locations, what would we expect to see after `ctrl f`?

I've tried some limited manual tests as follows: (`|` is the cursor position)
* This helps with PM header autocomplete - I'm not sure if that was expected!
* This also helps with stream autocomplete in the header
* This doesn't resolve the spaces issue, which I think we expected?
* #**zulip-term|inal** -> moves cursor to end, but shows `[No matches found]` in the footer; I'm not sure of the behavior, but that seems not quite right
* #**zulip-terminal*| -> won't autocomplete (this is maybe a pre-existing issue)

Generally it would be useful to add some tests to define the failing behavior - you could even mark them as xfail for now - whether in this case or not. For example, tests with the cursor in different locations should expose and document the (bad right now!) behavior with the cursor around spaces.

zulipterminal/ui_tools/boxes.py
ahmedabuamra: Yes, because some test cases test `autocomplete` without initiating `msg_write_box`, so I think that could happen in some scenarios based on these tests. That's why I made this check.

PR #917, opened 2021-02-12
enhancement: Introduces application-wide time format setting.

### LINKED ISSUE #770: Introduce application-wide time format setting

mkp6781:
Initial time format settings are fetched during initial register call.
Event handler enables update of this setting as events are recieved
from the server.

Fixes #770

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -179,6 +179,7 @@ class TestModel:
             'subscription',
             'typing',
             'update_message_flags',
+            'update_display_settings',
         ]
         fetch_event_types = [
             'realm',
@@ -189,6 +190,7 @@ class TestModel:
             'muted_topics',
             'realm_user',
             'realm_user_groups',
+            'update_display_settings',
             'zulip_version',
         ]
         model.client.register.assert_called_once_with(
@@ -1939,6 +1941,34 @@ class TestModel:
         for stream_id in stream_ids:
             new_subscribers = model.stream_dict[stream_id]['subscribers']
             assert new_subscribers == expected_subscribers
+
+    @pytest.mark.parametrize(['setting', 'twenty_four_hr_format'], [
+        (True, False), (False, True)
+    ])
+    def test_update_twenty_four_hour_format(self, mocker, model, setting,
+                                            twenty_four_hr_format):
+        event = {
+            'type': 'update_display_settings',
+            'setting_name': 'twenty_four_hour_time',
+        }
+        event['setting'] = setting
+
+        first_msg_w = mocker.Mock()
+        second_msg_w = mocker.Mock()
+        first_msg_w.original_widget.message = {'id': 1}
+        second_msg_w.original_widget.message = {'id': 2}
+        self.controller.view.message_view = mocker.Mock(
+                                               log=[first_msg_w, second_msg_w])
+        create_msg_box_list = mocker.patch('zulipterminal.model.'
+                                           'create_msg_box_list')
+        # Test for change in time format
+        model.twenty_four_hr_format = twenty_four_hr_format
+        assert model.twenty_four_hr_format != event['setting']
+        model._handle_update_display_settings(event)
+        assert model.twenty_four_hr_format == event['setting']
+        assert create_msg_box_list.call_count == len(
+                                    self.controller.view.message_view.log)
+        assert model.controller.update_screen.called
 
     @pytest.mark.parametrize('muted_streams, stream_id, is_muted', [
         ({1},   1, True),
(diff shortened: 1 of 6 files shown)


tests/model/test_model.py
```diff
@@ -1916,6 +1918,23 @@ def test__handle_subscription_event_subscribers_one_user_multiple_streams(
...
+        ({'setting_name': 'twenty_four_hour_time',
+          'setting': True}, False),
+    ], ids=['twenty_four_hour_time'])
+    def test__handle_update_display_settings(self, mocker, model,
+                                             twenty_four_hr_format,
+                                             event):
+        event['type'] = 'update_display_settings'
+
+        # Test for change in time format
+        if event['setting_name'] == 'twenty_four_hour_time':
```
neiljp, line 1931: This is a test for only the 24-hour update, and it's likely best to keep it at that - name it that, and if we add other settings event updates then we can write separate tests as we need them, or adapt this one. Those extra parameters are just one set, which you could put as a dict in the body of the test.

You're only testing False -> True? A lot of those parameters are constants once you focus on just this part of the event, and you can focus on the values of the settings.

neiljp (commented):
@mkp6781 Thanks for following up on this after the refactor in the other PR :+1:

This looks broadly good in practice, though I did note that with the year included the am/pm drops onto another line? (eg. look at messages before/after new-year 2020/2021)

zulipterminal/ui_tools/boxes.py
```diff
@@ -689,7 +689,6 @@ def __init__(self, message: Message, model: 'Model',
 
         # mouse_event helper variable
         self.displaying_selection_hint = False
-
```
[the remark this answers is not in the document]
mkp6781: Sorry these were left unintentionally.

tests/model/test_model.py
mkp6781: > You're only testing False -> True? A lot of those parameters are constants once you focus on just this part of the event, and you can focus on the values of the settings.

Should I add tests for checking whether `create_msg_box_list` is being called and a separate test for `formatted_local_time`?

neiljp: Both in the same test seems OK for now. I'll look again in the next iteration.

mkp6781:

mkp6781:
> This looks broadly good in practice, though I did note that with the year included the am/pm drops onto another line? (eg. look at messages before/after new-year 2020/2021)

I missed looking into this part. I'll make the change along with next iteration.

tests/model/test_model.py
```diff
@@ -1940,6 +1942,32 @@ def test__handle_subscription_event_subscribers_one_user_multiple_streams(
...
+        first_msg_w = mocker.Mock()
+        second_msg_w = mocker.Mock()
+        first_msg_w.original_widget.message = {'id': 1}
+        second_msg_w.original_widget.message = {'id': 2}
+        self.controller.view.message_view = mocker.Mock(
+                                               log=[first_msg_w, second_msg_w])
+        create_msg_box_list = mocker.patch('zulipterminal.model.'
+                                           'create_msg_box_list')
+        # Test for change in time format
+        if event['setting_name'] == 'twenty_four_hour_time':
```
neiljp, line 1962: This is now guaranteed with the specific area of this test.

tests/model/test_model.py
```diff
@@ -1940,6 +1942,32 @@ def test__handle_subscription_event_subscribers_one_user_multiple_streams(
             new_subscribers = model.stream_dict[stream_id]['subscribers']
             assert new_subscribers == expected_subscribers
 
+    def test_update_twenty_four_hour_format(self, mocker, model):
+        event = {
+            'type': 'update_display_settings',
+            'setting_name': 'twenty_four_hour_time',
+            'setting': True
+        }
+        twenty_four_hr_format = False
```
neiljp, line 1951: We can extract the True/False values into parametrized values here, to ensure we test different initial and expected values of the setting after the event.

neiljp (commented):
@mkp6781 Just to let you know that the conflict is related to the year-related code I just merged from #930, which should also help clarify the comment from the last review re strftime. I left some comments, but this looks fairly ready to go if we can resolve these points well and get the missed points from the last review.

tests/model/test_model.py
mkp6781: Yes 👍🏼

neiljp:
@mkp6781 Thanks for working on the server-related part of this after the refactoring PR :+1: I just manually merged this with some minor test and naming changes for consistency as 06082092ad994b5c2e23d277920a95a9ddf0585b :tada:

For reference, this would also have been fine as a "use initial data" commit followed by an "update based on events" commit - this PR should be simple enough that we've not overlooked anything that may have been clearer in that form :+1:

PR #918, opened 2021-02-12
workflows: Add github action checking emoji data matching with server.

### LINKED ISSUE #912: Add GitHub Action to regularly check unicode emoji are up to date with those in server
These change infrequently, but it would be useful to be alerted when these need to be updated.

We have a script which fetches and converts the emoji right now, which should allow this to be automated fairly easily.

Ezio-Sarthak:
The workflow uses CRON time syntax to schedule the check daily at a specific time.
In case of a mismatch in fetched data and existing data, the check would fail with exit status: Emoji data not up-to-date.

[unrelated]: Corrects a couple of diffs :)

Fixes #912

### DIFF
--- a/.github/workflows/emoji-server-check-workflow.yml
+++ b/.github/workflows/emoji-server-check-workflow.yml
@@ -1,0 +1,65 @@
+name: "Emoji sync check from Zulip server"
+
+on:
+  # What if the workflow doesn't run as scheduled?
+  # GitHub actions work on request-response mechanism. So, there may be
+  # tons of scheduled jobs queued, and so there may be a delayed response.
+  #
+  # For more information, checkout a blog post:
+  # https://upptime.js.org/blog/2021/01/22/github-actions-schedule-not-working/
+  schedule:
+    # The CRON time format is as follows:
+    # ---------------------------------------
+    # minutes | hours |  day  | month | day
+    #                  (month)         (week)
+    # ---------------------------------------
+    # For more information on CRON time formats, see https://crontab.guru/
+    #
+    # The below line suggests that the event would be triggered everyday
+    # at 17:00 UTC times.
+    - cron:  '0 17 * * *'
+
+jobs:
+  check:
+    name: Emoji sync check from Zulip server
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      - name: Install dependencies
+        run: pip install -e .
+
+      - name: Fetch emojis from server
+        run: ./tools/fetch-unicode-emoji-data
+      - name: Convert emojis dict to list
+        run: ./tools/convert-unicode-emoji-data
+
+      - name: Check if emojis are added/deleted/modified in server
+        run: |
+          git diff --exit-code zulipterminal/unicode_emojis.py >> $GITHUB_ENV
+
+          if [ $? -eq 0 ] ; then
+            echo "Emojis are up-to-date"
+          fi
+
+      - name: Notify in stream if emojis are not up-to-date
+
+        # Ignore this step if emojis are up-to-date.
+        if: '${{ github.GITHUB_ENV != 0 }}'
+
+        uses: zulip/github-actions-zulip@v0.1.0
+        with:
+          username: ${{ secrets.ZT_BOT_EMAIL }}
+          api-key: ${{ secrets.ZT_BOT_API_KEY }}
+          organization-url: 'https://chat.zulip.org'
+          to: 'zulip-terminal'
+          type: 'stream'
+          topic: 'CI action feedback'
+          content: 'Heads up :caution:! Emoji sync with the server failed!
+                    This originated from an update in server emoji.
+                    Kindly re-generate `unicode_emojis.py` in main.'
+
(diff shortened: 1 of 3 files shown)


Abhirup-99:
Wouldn't it be better if it's checked in every push and PR? If something breaks in a certain pr, it would be very difficult to identify it. Also, the main branch is supposed to be stable and we are all using it during development, so this goes against that.

prah23:
I second @Ezio-Sarthak's approach here. A CRON job might be better suited as these change infrequently as mentioned in the issue. We wouldn't have to run this as frequently as on every push and PR, also considering the frequency of that is highly variable. I presume the check is meant to be more general. Since the emoji list is auto-generated and not meant to be modified manually, the chance of something breaking in a PR should be significantly low.

prah23:
I second @Ezio-Sarthak's approach here. A CRON job might be better suited as these change infrequently as mentioned in the issue. We wouldn't have to run this as frequently as on every push and PR, also considering the frequency of that is highly variable. I presume the check is meant to be more general. Since the emoji list is auto-generated and not meant to be modified manually, the chance of something breaking in a PR should be significantly low.

prah23:
I second @Ezio-Sarthak's approach here. A CRON job might be better suited as these change infrequently as mentioned in the issue. We wouldn't have to run this as frequently as on every push and PR, also considering the frequency of that is highly variable. I presume the check is meant to be more general. Since the emoji list is auto-generated and not meant to be modified manually, the chance of something breaking in a PR should be significantly low.

Ezio-Sarthak:
@Abhirup-99 I initially had a thought of doing so, but as mentioned by Neil in the issue, this should be infrequent, and so a regular check could be enough. Also, since we have some workflows triggering already (specially codeql), having one more with every push would be a bloat to the repo.

@prah23 Thanks for the clarification 👍 

PS - The PR is open for reviews  :)

neiljp (commented):
@Ezio-Sarthak Thanks for looking into this :+1:

The guts of this check seem good, but I do think we need to think about how the triggering and notification is going to occur.

I'm not sure how feasible it is, but sending the results to a Zulip topic may be the tidiest result here. If I understand the parallels, this is being worked on for zulip/zulip; I just posted in #**automated testing>Unicode emoji sync** to query this.

Given how minimal this is, we could limit this to trigger only on pushes to main on zulip/zulip-terminal. I'm not sure if we want this to fail CI due to this, but I'm not sure what other notification options we have that are straightforward.

If this runs on a timed schedule, I'm not sure where the result appears?

.github/workflows/emoji-server-check-workflow.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: python --version
+      - name: Fetch emojis from server
+        run: ./tools/fetch-unicode-emoji-data
+      - name: Convert emojis dict to list
+        run: ./tools/convert-unicode-emoji-data
+      - name: Check if emojis are up-to-date
+        run: |
+          export DIFF="$(git diff zulipterminal/unicode_emojis.py)"
+          export DIFF_LEN=${#DIFF}
+          if [[ ${DIFF_LEN} == 0 ]] ; then
```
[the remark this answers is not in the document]
Ezio-Sarthak: Thanks. I didn't know it before :sweat_smile:

Ezio-Sarthak:

The notification of check on czo topic is a pretty awesome idea!
I'll see if I could make a script to send a message to czo **#zulip-terminal > Unicode emoji sync with server** if that's possible :)
Also, do we have to notify in czo only if there's a change in data?

Ezio-Sarthak:

.github/workflows/emoji-server-check-workflow.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Convert emojis dict to list
+        run: ./tools/convert-unicode-emoji-data
+      - name: Notify if emojis are not up-to-date
+        run: |
+          git diff --exit-code zulipterminal/unicode_emojis.py
+          if [ $? -eq 0 ] ; then
+            echo "Emojis up-to-date"
+          else
+            curl -X POST https://chat.zulip.org/api/v1/messages \
+              -u ${{ secrets.ZT_BOT_EMAIL }}:${{ secrets.ZT_BOT_API_KEY }} \
```
neiljp, line 39: From a skim of the secrets documents, we would likely want this to be in an environment variable rather than exposed in clear text.

.github/workflows/emoji-server-check-workflow.yml
```diff
@@ -0,0 +1,45 @@
...
+              --data-urlencode type=stream \
+              --data-urlencode 'to=zulip-terminal'\
+              --data-urlencode 'subject=Emoji sync with server' \
+              --data-urlencode 'content=Heads up :caution:!
+                                        Emoji data needs to be updated.'
```
neiljp, line 44: We can easily change these later, but:
* I expect it'd be better to have a more general single topic for CI feedback, since we have everything in one stream.
* Re the content, we may wish to specify more that this originates from a server emoji update.

.github/workflows/emoji-server-check-workflow.yml
Ezio-Sarthak: As of my knowledge, `secrets` are default environment variable itself in GitHub actions.
However, It's good storing them as `env` variables in any case :+1:. This will tidy-up the relevant code wherever they are used. (currently double brackets seem visually messy).

.github/workflows/emoji-server-check-workflow.yml
Ezio-Sarthak: - Re topic, maybe something like: "CI action feedback"?
- Re the content, the below message could be a potential description?:
          "Heads up! Emoji sync with the server failed!
           This originated from an update in server emoji. Kindly re-generate `unicode_emojis.py`"

PR #921, opened 2021-02-13
docstring: Lint and fix docstrings.

Abhirup-99:


tools/lint-docstring
```diff
@@ -0,0 +1,119 @@
...
+
+import argparse
+import importlib
+import sys
+from os import getcwd, listdir
+from os.path import isfile, join
+from typing import Dict, List, Tuple, cast
+
+
+developer_doc = join(getcwd(), 'docs', 'developer-file-overview.md')
```
neiljp, line 11: `getcwd` won't always work for this, eg. if the script is run from in `tools/`. Consider using operations derived from `__file__`, perhaps?

tools/lint-docstring
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python3
+
+import argparse
+import importlib
+import sys
+from os import getcwd, listdir
+from os.path import isfile, join
```
neiljp, line 7: You may find `pathlib` more practical now that we're in late python3.

tools/lint-docstring
```diff
@@ -0,0 +1,119 @@
...
+
+
+def get_docstrings() -> Tuple[Dict[str, str], Dict[str, List[str]]]:
+    root = join(getcwd(), 'zulipterminal')
+    folder_paths = {
+        'zulipterminal': root,
+        'zulipterminal/cli': join(root, 'cli'),
+        'zulipterminal/config': join(root, 'config'),
+        'zulipterminal/ui_tools': join(root, 'ui_tools')
+    }
```
neiljp, line 21: This does work, but we should be able to at minimum automate this dictionary generation, and at most we can likely look inside all of `zulipterminal` with a `glob`?

I'd also expect the `zulipterminal` to be able to be a parameter we pass in, as long as we scan through sub-directories automatically without needing this dict.

tools/lint-docstring
```diff
@@ -0,0 +1,119 @@
...
+        total_files[folder] = [file for file in listdir(path)
+                               if(isfile(join(path, file))
+                               and file != '__init__.py')]
+    all_zt_files_dict = {}
+    for folder, files in total_files.items():
+        for file in files:
+            imported_file = importlib.import_module(
+                            f'{folder.replace("/",".")}.{file[:-3]}')
+            docstring = cast(str, imported_file.__doc__)
+            all_zt_files_dict[file] = docstring.strip().replace('\n', ' ')
```
neiljp, line 33: This seems like a post-processing step we can handle elsewhere?

tools/lint-docstring
```diff
@@ -0,0 +1,119 @@
...
+import sys
+from os import getcwd, listdir
+from os.path import isfile, join
+from typing import Dict, List, Tuple, cast
+
+
+developer_doc = join(getcwd(), 'docs', 'developer-file-overview.md')
+
+
+def get_docstrings() -> Tuple[Dict[str, str], Dict[str, List[str]]]:
```
neiljp, line 14: This is specifically *module* docstrings, so it would be clearer to name it as such?

neiljp (commented):
@Abhirup-99 I couldn't get this to run properly - I get a traceback.

I've made a few inline comments, as I think the folder+file listing could be simpler and easier using `pathlib`.

It seems that you're editing the markdown file? I'd suggest generating it using a template each time, which should be a lot simpler.

We don't need to include a lint mode in a first version, though if you do then we could generate the file to a `NamedTemporaryFile` and compare.

tools/lint-docstring
```diff
@@ -0,0 +1,119 @@
...
+    return (all_zt_files_dict, total_files)
+
+
+def get_file_overview_doc_docstrings() -> Dict[str, str]:
+    with open(developer_doc, 'r') as file:
+        doc_data = file.readlines()
+
+    # table is present from line 6 to (length - 1)
+    computed_doc_dict = {}
+    for i in range(6, len(doc_data)):
```
Ezio-Sarthak, line 43: We may refactor the starting line index as well (so as to easily handle any changes in future). Something like
```python
TABLE_START_LINE = 6
```
would be great :)

neiljp:
@Abhirup-99 Thanks for working on this! This was used as a good foundation for #1208.
Since that was merged earlier this year, I'm closing this.

PR #926, opened 2021-02-17
Generate hotkey using script

### LINKED ISSUE #126: Generate hotkeys section/doc using script and check with CI

mkp6781:
Fixes #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Zulip Terminal - [Zulip](https://zulip.com)'s official terminal client
 
-[Recent changes](https://github.com/zulip/zulip-terminal/blob/main/CHANGELOG.md) | [Configuration](#Configuration) | [Hot Keys](#hot-keys) | [FAQs](https://github.com/zulip/zulip-terminal/blob/main/docs/FAQ.md) | [Development](#contributor-guidelines)
+[Recent changes](https://github.com/zulip/zulip-terminal/blob/main/CHANGELOG.md) | [Configuration](#Configuration) | [Hot Keys](https://github.com/zulip/zulip-terminal/blob/main/docs/hotkey.md) | [FAQs](https://github.com/zulip/zulip-terminal/blob/main/docs/FAQ.md) | [Development](#contributor-guidelines)
 
 [![Zulip chat](https://img.shields.io/badge/zulip-join_chat-brightgreen.svg)](https://chat.zulip.org/#narrow/stream/206-zulip-terminal)
 [![PyPI](https://img.shields.io/pypi/v/zulip-term.svg)](https://pypi.python.org/pypi/zulip-term)
@@ -17,7 +17,7 @@ Zulip Terminal is the official terminal client for Zulip, providing a [text-base
 
 Specific aims include:
 * Providing a broadly similar user experience to the Zulip web client, ultimately supporting all of its features
-* Enabling all actions to be achieved through the keyboard (see [Hot keys](#hot-keys))
+* Enabling all actions to be achieved through the keyboard (see [Hot keys](https://github.com/zulip/zulip-terminal/blob/main/docs/hotkey.md))
 * Exploring alternative user interface designs suited to the display and input constraints
 * Supporting a wide range of platforms and terminal emulators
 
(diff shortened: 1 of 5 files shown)


tools/generate_hotkey.py
```diff
@@ -0,0 +1,52 @@
...
+    'navigation': navigation,
+    'searching': searching,
+    'stream_list': stream_list,
+}
+
+for key, item in KEY_BINDINGS.items():
+    CATEGORY[item['key_category']].append((item['help_text'], item['keys']))
+
+OUTPUT_FILE = "docs/hotkey.md"
+mdFile = open(OUTPUT_FILE, "w")
```
Abhirup-99, line 26: Thank you for the pr, but it would be better if you use the `with open` syntax to do operations on the file. This would allow all closing and garbage collection to take place automatically when you move out. Else, you would have to handle it manually.

mkp6781: Thank you for mentioning that. I'll make the change 👍🏼

tools/generate_hotkey.py
```diff
@@ -0,0 +1,52 @@
...
+stream_list = []
+
+CATEGORY = {
+    'general': general,
+    'msg_actions': msg_actions,
+    'msg_compose': msg_compose,
+    'navigation': navigation,
+    'searching': searching,
+    'stream_list': stream_list,
+}
```
neiljp, line 20: Can we just generate these from what is in `HELP_CATEGORIES`? Otherwise we'll need to update this script if we change the categories.

tools/generate_hotkey.py
```diff
@@ -0,0 +1,52 @@
...
+
+for key, item in KEY_BINDINGS.items():
+    CATEGORY[item['key_category']].append((item['help_text'], item['keys']))
```
neiljp, line 23: Would this be clearer as a dict comprehension?

I'll need to reread again later, but is CATEGORY a constant? If not, it doesn't need to be capitalized.

tools/generate_hotkey.py
```diff
@@ -0,0 +1,52 @@
...
+                new_key_list[i] = "<kbd>"+new_key_list[i]+"</kbd>"
+            new_key_list = " + ".join(new_key_list)
+            various_key_combinations.append(new_key_list)
+        various_key_combinations = " / ".join(reversed(various_key_combinations))
+        mdFile.write("|{}|{}|\n".format(help_text, various_key_combinations))
+    mdFile.write("\n\n")
+
+mdFile.write("**Note:** You can use `arrows`, `home`, `end`,"
+                        "`Page up` and `Page down` keys to move around"
+                        "in Zulip-Terminal.")
```
neiljp, line 50: This may be unnecessary given the section with navigation?

neiljp (commented):
@mkp6781 This looks like a good start - I left a few comments :+1:

tools/generate_hotkey.py
mkp6781: I left it there as it was there in the README file.

tools/generate_hotkey.py
mkp6781: I have tried using dict comprehension in two different ways. I have left one commented in the code. Please let me know which one would be preferred.

tools/generate_hotkey.py
```diff
@@ -0,0 +1,36 @@
...
+
+from zulipterminal.config.keys import HELP_CATEGORIES, KEY_BINDINGS
+
+
+# category = {}
+# {category.setdefault(item['key_category'],[]).append((item['help_text'], item['keys'])) for (key, item) in KEY_BINDINGS.items()}
+
+# OR
+category = defaultdict(list)
+{category[item['key_category']].append((item['help_text'], item['keys'])) for (key, item) in KEY_BINDINGS.items()}
```
neiljp, line 12: Comprehensions generally build an object itself and don't use side-effects (walrus operator aside), so I'd expect this dict to be assigned to `category` or some other name.

You may need a nested comprehension, or another data-structure may be better.

tools/generate_hotkey.py
```diff
@@ -0,0 +1,36 @@
...
+OUTPUT_FILE = "docs/hotkey.md"
+
+with open(OUTPUT_FILE,"w") as mdFile:
+    mdFile.write("<!--- Generated automatically by tools/generate_hotkey.py -->\n")
+    mdFile.write("<!--- Do not modify -->\n\n")
+    mdFile.write("# HOT KEYS\n")
```
neiljp, line 18: These blocks can be one call with a large string.

neiljp (commented):
@mkp6781 Just had another quick look - the order of sections appears to have changed? This may change as you finalize the data structures from keys.py in any case.

tools/generate_hotkey.py
mkp6781: > You may need a nested comprehension, or another data-structure may be better.

Sorry for the delay. 
I have used a nested comprehension this time. Also, I would like to know which alternate data structure you were referring to here.

mkp6781:
To generate the hotkey section exactly as before, a few minor adjustments may be needed in the keys.py file.

tools/generate_hotkey.py
neiljp: What you have now looks fine :+1:

tools/generate_hotkey.py
neiljp: I see you removed it - that's great, though let's ensure that's noted in the commit text.

neiljp:
@mkp6781 I'm excited to have this done! :+1: 

We should perhaps adjust the generated file to be plural (...keys.md), and I'm not sure why we have the title capitalized; I'm not sure we do in other documentation?

Regarding your point about the difference between the files, I'd suggest adding an intermediate commit with the extra/adjusted entries.

tools/generate_hotkey.py
```diff
@@ -0,0 +1,33 @@
...
+    for action in categories.keys():
+        mdFile.write("## {}\n".format(HELP_CATEGORIES[action]))
+        mdFile.write("|Command|Key Combination|\n")
+        mdFile.write("| :--- | :---: |\n")
+        for help_text, key_combinations_list in categories[action].items():
+            various_key_combinations = []
+            for key_combination in key_combinations_list:
+                new_key_list = key_combination.split()
+                for i in range(len(new_key_list)):
+                    new_key_list[i] = "<kbd>"+new_key_list[i]+"</kbd>"
```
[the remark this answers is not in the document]
mkp6781: I should have noticed this earlier while making the change for `categories`. :)

tools/generate_hotkey.py
```diff
@@ -0,0 +1,39 @@
...
+        item['help_text']: item['keys'] 
+        for item in KEY_BINDINGS.values()
+        if item['key_category'] == category
+    }
+    for category in HELP_CATEGORIES.keys()
+}
+OUTPUT_FILE = Path(__file__).parent.parent.joinpath('docs', 'hotkeys.md')
+
+with open(OUTPUT_FILE,"w") as mdFile:
+    mdFile.write("<!--- Generated automatically by tools/generate_hotkey.py -->\n"
```
neiljp, line 18: fstring?

neiljp (commented):
@mkp6781 I hope you can already see the readability improvements from the last few iterations?

There are a few more updates we need based on the extra commit, and there are some changes to the generation script which could simplify that even further which you may be able to identify from the previous iterations. The plural form looks better, but let's change the script name to match too?

tools/generate_hotkey.py
mkp6781: Since we are not using any variables here, do we need an fstring?

neiljp: The filename itself is a variable, so we can avoid adjusting this text (or part of it) when the script runs.

mkp6781:

tools/generate_hotkey.py
mkp6781: I do not quite follow you here. Isn't the filename a constant with the root directory as zulipterminal. A similar case can be seen in `convert-unicode-emoji-data`.

mkp6781: Noticed the change in `convert-unicode-emoji-data`, I've made the same change 👍🏼

neiljp:
@mkp6781 Thanks for this - this exposes various cases where we weren't keeping the documents synchronized, and will make updating categories and text much easier :+1: I've just merged this with a few adjustments to the first commit due to broken links, chose to exclude/update various key data from the second, and merged the third as it was :tada:

PR #928, opened 2021-02-17
boxes: Use `user_id`s in autocomplete for user mentions.

### LINKED ISSUE #151: Create extended user mention syntax when needed.
Recently, in zulip/zulip#10243 we merged support for the following mention syntax to zulip/zulip:

`@**Full Name|1234(id)**`

This is to allow us to mention users with the same full name, and is optional. The web app intelligently adds the new syntax only when the users have the same full name, and uses the simpler syntax for others.

Example: https://chat.zulip.org/#narrow/stream/7-test-here/topic/user.20mention

We should:

1. Add support for this syntax in the terminal client.
2. Ensure that it is behind a feature flag because it would break mentions in older servers/any non CZO server for now.

prah23:
This PR adds a commit that append `user_id`s to the users' names in the autocomplete typeahead for user mentions. This helps in distinguishing between users having the same name, similar to the pm recipients autocomplete.

Tests updated.

Fixes #151.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -101,6 +101,8 @@ class TestWriteBox:
                     "Human Myself",
                     "Human 1",
                     "Human 2",
+                    "Human Duplicate",
+                    "Human Duplicate",
                     "Group 1",
                     "Group 2",
                     "Group 3",
@@ -108,14 +110,74 @@ class TestWriteBox:
                 ],
             ),
             ("@*", 0, ["Group 1", "Group 2", "Group 3", "Group 4"]),
-            ("@**", 0, ["Human Myself", "Human 1", "Human 2"]),
+            (
+                "@**",
+                0,
+                [
+                    "Human Myself",
+                    "Human 1",
+                    "Human 2",
+                    "Human Duplicate",
+                    "Human Duplicate",
+                ],
+            ),
             # mentions
-            ("@Human", 0, ["Human Myself", "Human 1", "Human 2"]),
-            ("@**Human", 0, ["Human Myself", "Human 1", "Human 2"]),
-            ("@_Human", 0, ["Human Myself", "Human 1", "Human 2"]),
+            (
+                "@Human",
+                0,
+                [
+                    "Human Myself",
+                    "Human 1",
+                    "Human 2",
+                    "Human Duplicate",
+                    "Human Duplicate",
+                ],
+            ),
+            (
+                "@**Human",
+                0,
+                [
+                    "Human Myself",
+                    "Human 1",
+                    "Human 2",
+                    "Human Duplicate",
+                    "Human Duplicate",
+                ],
+            ),
+            (
+                "@_Human",
+                0,
+                [
+                    "Human Myself",
+                    "Human 1",
+                    "Human 2",
+                    "Human Duplicate",
+                    "Human Duplicate",
+                ],
+            ),
             ("@_*Human", None, []),  # NOTE: Optional single star fails
-            ("@_**Human", 0, ["Human Myself", "Human 1", "Human 2"]),
-            ("@Human", None, ["Human Myself", "Human 1", "Human 2"]),
+            (
+                "@_**Human",
+                0,
+                [
+                    "Human Myself",
+                    "Human 1",
+                    "Human 2",
+                    "Human Duplicate",
+                    "Human Duplicate",
+                ],
+            ),
+            (
+                "@Human",
+                None,
+                [
+                    "Human Myself",
+                    "Human 1",
+                    "Human 2",
+                    "Human Duplicate",
+                    "Human Duplicate",
+                ],
+            ),
             ("@NoMatch", None, []),
             # streams
             (
@@ -159,13 +221,19 @@ class TestWriteBox:
             ("@Human", 0, "@**Human Myself**"),
             ("@Human", 1, "@**Human 1**"),
             ("@Human", 2, "@**Human 2**"),
-            ("@Human", -1, "@**Human 2**"),
-            ("@Human", -2, "@**Human 1**"),
-            ("@Human", -3, "@**Human Myself**"),
-            ("@Human", -4, None),
+            ("@Human", 3, "@**Human Duplicate**"),
+            ("@Human", 4, "@**Human Duplicate**"),
+            ("@Human", -1, "@**Human Duplicate**"),
+            ("@Human", -2, "@**Human Duplicate**"),
+            ("@Human", -3, "@**Human 2**"),
+            ("@Human", -4, "@**Human 1**"),
+            ("@Human", -5, "@**Human Myself**"),
+            ("@Human", -6, None),
             ("@_Human", 0, "@_**Human Myself**"),
             ("@_Human", 1, "@_**Human 1**"),
             ("@_Human", 2, "@_**Human 2**"),
+            ("@_Human", 3, "@_**Human Duplicate**"),
+            ("@_Human", 4, "@_**Human Duplicate**"),
             ("@H", 1, "@**Human 1**"),
             ("@Hu", 1, "@**Human 1**"),
             ("@Hum", 1, "@**Human 1**"),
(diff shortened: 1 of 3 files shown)


neiljp (commented):
@prah23 This seems to work, but while you mentioned adding tests later on czo, you note in the commit text that tests are included - did you forget to add them?

Manually this seems to work, but once we have tests then we could ensure we exercise the code a little more. The duplicates check relies on adjacent elements being identical given the sort?

prah23:
@neiljp I haven't removed the "`Tests updated.`" part of the commit text from the previous iteration, where I've appended `user_id`s to all the names in the list. I'm adding tests and will update you once the PR is ready. Also, I maintain a separate list of duplicate names, which I check with before appending `user_id`s to the final list that is passed to the typeahead, so the duplicates check does not rely on adjacency, or any particular order for that matter.

The duplicates list is formed by checking if the current name in the loop exists in the sub-list of elements leading to the current name. I've run this loop on the `user_names` list.

prah23:
@neiljp I haven't removed the "`Tests updated.`" part of the commit text from the previous iteration, where I've appended `user_id`s to all the names in the list. I'm currently adding tests and will update you once the PR is ready. Also, I maintain a separate list of duplicate names, which I check with before appending `user_id`s to the final list that is passed to the typeahead, so the duplicates check does not rely on adjacency, or any particular order for that matter.

The duplicates list is formed by checking if the current name in the loop exists in the sub-list of elements leading to the current name. I've run this loop on the `user_names` list.

prah23:
@neiljp I haven't removed the "`Tests updated.`" part of the commit text from the previous iteration, where I've appended `user_id`s to all the names in the list. I'm currently adding tests and will update you once the PR is ready. Also, I maintain a separate list of duplicate names, which I check with before appending `user_id`s to the final list that is passed to the typeahead, so the duplicates check does not rely on adjacency or any particular order for that matter.

The duplicates list is formed by checking if the current name in the loop exists in the sub-list of elements leading to the current name. I've run this loop on the `user_names` list.

neiljp:
Re adjacency, I was referring to the sub-list check you indicated. If the list is sorted, can we rely upon the previous entry only? Alternatively, it might be clearer to use Counter. In any case, once we have tests then this will be easier to explore.

prah23:
@zulipbot remove "PR needs review"

prah23:

prah23:

tests/ui_tools/test_boxes.py
```diff
@@ -242,6 +242,49 @@ def test_generic_autocomplete_mentions_subscribers(self, write_box, text,
...
+    @pytest.mark.parametrize('text', [
+        '@H', '@Hu', '@Hum', '@Huma', '@Human',
+        '@**', '@**H', '@**Hu', '@**Hum', '@**Huma', '@**Human'
+    ])
+    @pytest.mark.parametrize('matching_users, distinct_matching_users', [
+        (['Human Myself', 'Human 1', 'Human 2', 'Human Duplicate',
+          'Human Duplicate'],
+         ['@**Human Myself**', '@**Human 1**', '@**Human 2**',
+          '@**Human Duplicate|13**', '@**Human Duplicate|14**'])
+    ])
```
neiljp, line 254: This is just one pair, right? If so, let's place it in the body of the test.

tests/ui_tools/test_boxes.py
```diff
@@ -242,6 +242,49 @@ def test_generic_autocomplete_mentions_subscribers(self, write_box, text,
         typeahead_string = write_box.generic_autocomplete(text, state)
         assert typeahead_string == required_typeahead
 
+    @pytest.mark.parametrize('text', [
+        '@H', '@Hu', '@Hum', '@Huma', '@Human',
+        '@**', '@**H', '@**Hu', '@**Hum', '@**Huma', '@**Human'
```
neiljp, line 247: This looks like something which we could generate.

tests/ui_tools/test_boxes.py
```diff
@@ -242,6 +242,49 @@ def test_generic_autocomplete_mentions_subscribers(self, write_box, text,
...
+        _process_typeaheads = mocker.patch(BOXES
+                                           + '.WriteBox._process_typeaheads')
+
+        write_box.generic_autocomplete(text, state)
+
+        _process_typeaheads.assert_called_once_with(distinct_matching_users,
+                                                    state, matching_users)
+
+    @pytest.mark.parametrize('text', [
+        '@_', '@_H', '@_Hu', '@_Hum', '@_Huma', '@_Human'
```
neiljp, line 267: Can we combine with the previous test?

zulipterminal/ui_tools/boxes.py
```diff
@@ -445,11 +445,23 @@ def autocomplete_users(self, text: str, prefix_string: str
                                        reverse=True)
 
         user_names = [user['full_name'] for user in sorted_matching_users]
+
+        user_name_duplicates = [user_name
+                                for count, user_name in enumerate(user_names)
+                                if user_name in user_names[:count]]
```
neiljp, line 451: This remains my main concern with this implementation. It's not a common case, but for example I get a pretty bad response trying a mention of `@user`<kbd>ctrl-f</kbd> on czo.

Instinctively I'd suggest using `Counter`, but anything which doesn't make a slice in every loop would likely be sufficient.

You have tests to ensure you get good results, so refactoring should be easier :+1:

zulipterminal/ui_tools/boxes.py
```diff
@@ -445,11 +445,23 @@ def autocomplete_users(self, text: str, prefix_string: str
...
+
+        user_names_with_distinct_duplicates = [
+            f"{user['full_name']}|{user['user_id']}"
+            if user['full_name'] in user_name_duplicates
+            else user['full_name']
```
neiljp, line 456: Minor: I know you need user as the base object to get the user-id in the loop, but it seems strange that we're generating a list with names and then only use it once.

neiljp (commented):
@prah23 This seems close to being merged :+1:

tests/ui_tools/test_boxes.py
prah23: The `distinct_matching_users` list differs for these autocomplete `text`'s, which is why I moved these to another test.

prah23: I meant that the typeahead would need `@_**Human 1**` instead of `@**Human 1**` because this test is for silent mentions, which makes the `distinct_matching_users` list different.

prah23 (commented):
@neiljp I've addressed all your comments from the previous review. I haven't combined the tests due to the typeahead list being different (normal and silent mentions) and I've also had to explicitly add a test case, which I've explained inline. Apart from that, it's ready for a review.

prah23:

prah23:
@neiljp this has been `black` rebased, is up-to-date, and ready for another review.

zulipterminal/ui_tools/boxes.py
```diff
@@ -457,12 +457,23 @@ def autocomplete_users(
         )
 
         user_names = [user["full_name"] for user in sorted_matching_users]
+
+        user_names_counter = Counter(user_names)
```
neiljp, line 461: It's worth adding a note/todo/fixme here as to why we use this - this definitely seems faster and more efficient than the previous implementation, but it's worth noting why, and how/why we could improve this in future (we generate this counter each time we do an autocomplete?)

tests/ui_tools/test_boxes.py
```diff
@@ -340,6 +340,68 @@ def test_generic_autocomplete_mentions_subscribers(
...
+            "Human Duplicate",
+            "Human Duplicate",
+        ]
+        distinct_matching_users = [
+            "@**Human Myself**",
+            "@**Human 1**",
+            "@**Human 2**",
+            "@**Human Duplicate|13**",
+            "@**Human Duplicate|14**",
+        ]
```
neiljp, line 368: If these are only specific to the assert part then they can sit with it, below the act[ion] part. That makes it clear they're not part of the setting-up.

tests/ui_tools/test_boxes.py
```diff
@@ -340,6 +340,68 @@ def test_generic_autocomplete_mentions_subscribers(
...
+            "Human Duplicate",
+            "Human Duplicate",
+        ]
+        distinct_matching_users = [
+            "@_**Human Myself**",
+            "@_**Human 1**",
+            "@_**Human 2**",
+            "@_**Human Duplicate|13**",
+            "@_**Human Duplicate|14**",
+        ]
```
neiljp, line 397: This test seems the same as the previous but with a different prefix; can we not just add that to the parametrize?

neiljp (commented):
@prah23 As discussed this looks close to merging and really clean :+1:

I just left a few comments inline.

neiljp:
@prah23 Thanks for working on this - I'm going to merge this now :tada:

We could debate the best test structure, but having the feature there and tested is good and we can always iterate from there :+1:

PR #930, opened 2021-02-18
refactor: views: Add year in message info view.

### LINKED ISSUE #879: Message Info Popup: Add Year to date of message

jalajcodes:
Added Year in the message info popup box.
image

Fixes #879

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -1164,15 +1164,16 @@ class Model:
                       self.controller, -1)
 
     def formatted_local_time(self, timestamp: int,
-                             *, show_seconds: bool = False,
+                             *, show_seconds: bool,
                              show_year: bool = False) -> str:
         local_time = datetime.datetime.fromtimestamp(timestamp)
-        if show_seconds:
-            return local_time.strftime('%a %b %d %H:%M:%S')
-        if show_year:
-            return local_time.strftime('%a %b %-d %Y %-H:%M:%S')
-        else:
-            return local_time.strftime('%a %b %d %H:%M')
+        format_codes = (
+            "%a %b %d"
+            f"{' %Y ' if show_year else ''}"
+            "%H:%M"
+            f"{':%S' if show_seconds else ''}"
+        )
+        return local_time.strftime(format_codes)
 
     def _update_rendered_view(self, msg_id: int) -> None:
         """
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -1164,10 +1164,13 @@ class Model:
                       self.controller, -1)
 
     def formatted_local_time(self, timestamp: int,
-                             *, show_seconds: bool) -> str:
+                             *, show_seconds: bool = False,
+                             show_year: bool = False) -> str:
         local_time = datetime.datetime.fromtimestamp(timestamp)
         if show_seconds:
             return local_time.strftime('%a %b %d %H:%M:%S')
+        if show_year:
+            return local_time.strftime('%a %b %-d %Y %-H:%M:%S')
         else:
             return local_time.strftime('%a %b %d %H:%M')
 
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -1216,7 +1216,7 @@ class MsgInfoView(PopUpView):
         self.message_links = message_links
         self.time_mentions = time_mentions
         date_and_time = controller.model.formatted_local_time(
-                                        msg['timestamp'], show_seconds=True)
+                                        msg['timestamp'], show_year=True)
 
         msg_info = [
             ('', [('Date & Time', date_and_time),
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -1216,7 +1216,8 @@ class MsgInfoView(PopUpView):
         self.message_links = message_links
         self.time_mentions = time_mentions
         date_and_time = controller.model.formatted_local_time(
-                                        msg['timestamp'], show_year=True)
+            msg['timestamp'], show_seconds=True, show_year=True
+            )
 
         msg_info = [
             ('', [('Date & Time', date_and_time),


zulipterminal/model.py
```diff
@@ -1164,10 +1164,13 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
...
 
     def formatted_local_time(self, timestamp: int,
-                             *, show_seconds: bool) -> str:
+                             *, show_seconds: bool = False,
+                             show_year: bool = False) -> str:
         local_time = datetime.datetime.fromtimestamp(timestamp)
         if show_seconds:
             return local_time.strftime('%a %b %d %H:%M:%S')
+        if show_year:
+            return local_time.strftime('%a %b %-d %Y %-H:%M:%S')
```
Ezio-Sarthak, line 1173: This makes `show_seconds` True by default?
We may have a `format_date` like string variable, in which we could append supported formats as per query?

jalajcodes: That's a good idea... I'll refactor it😅

zulipterminal/model.py
```diff
@@ -1164,10 +1164,13 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
                       self.controller, -1)
 
     def formatted_local_time(self, timestamp: int,
-                             *, show_seconds: bool) -> str:
+                             *, show_seconds: bool = False,
+                             show_year: bool = False) -> str:
```
neiljp, line 1168: You can leave show_seconds without a default here.

zulipterminal/model.py
neiljp: I suggested a similar approach in #917, which we'll need to integrate with this at some point in any case.

zulipterminal/model.py
```diff
@@ -1164,15 +1164,16 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
                       self.controller, -1)
 
     def formatted_local_time(self, timestamp: int,
-                             *, show_seconds: bool = False,
+                             *, show_seconds: bool,
```
Ezio-Sarthak, line 1167: This change could be done in the previous commit, as changing show_seconds code isn't the aspect of this commit :)

zulipterminal/model.py
```diff
@@ -1164,15 +1164,16 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
...
                              show_year: bool = False) -> str:
         local_time = datetime.datetime.fromtimestamp(timestamp)
-        if show_seconds:
-            return local_time.strftime('%a %b %d %H:%M:%S')
-        if show_year:
-            return local_time.strftime('%a %b %-d %Y %-H:%M:%S')
-        else:
-            return local_time.strftime('%a %b %d %H:%M')
+        format_codes = (
```
Ezio-Sarthak, line 1170: It's a good practice to make intermediate variable names context dependent, in general. Something like "format_date" could work?

zulipterminal/ui_tools/views.py
```diff
@@ -1216,7 +1216,8 @@ def __init__(self, controller: Any, msg: Message, title: str,
         self.message_links = message_links
         self.time_mentions = time_mentions
         date_and_time = controller.model.formatted_local_time(
-                                        msg['timestamp'], show_year=True)
+            msg['timestamp'], show_seconds=True, show_year=True
```
Ezio-Sarthak, line 1219: You might have observed, that you had to change the parameter you added, in both the commits, when you refactor in a later commit. See my review comment for reference :)

Ezio-Sarthak (commented):
This does seem to work OK, but there are some general aspects in general you might wanna know 👍 

It is generally preferred to keep a refactor commit in the beginning. One way this helps is to avoid changing related code in multiple commits (referring to one of my comments below). If you think this is something we could add in the docs, do let us know :)

The show_year param might also not need a default value (coz we're validating it anyway).

jalajcodes:
> It is generally preferred to keep a refactor commit in the beginning
Can you please elaborate?

> The show_year param might also not need a default value (coz we're validating it anyway).
I don't think so, as removing the default value will make the `show_year` parameter required and this will cause an error where we don't pass it. 

@Ezio-Sarthak

zulipterminal/ui_tools/views.py
jalajcodes: Sorry, I didn't quite catch what you're trying to say. What do you mean by refactor in a later commit? I didn't have to make any changes in this file in the second commit.

Ezio-Sarthak:
> Can you please elaborate?

Sure, For instance, I have a commit that adds a key-value pair in a dictionary. As a refactor, I want that dict to convert into an OrderedDict. So, ideally, what I wish to do is to convert that dict to OrderedDict prior to adding key-value pairs in it. To compare with your code, you've added the `show_year` parameter first and declared a general `format_date` string in the later commit. Hope this helps :)

> I don't think so, as removing the default value will make the `show_year` parameter required and this will cause an error where we don't pass it.

Ah I see, I thought we're adding `show_year` param for each call. Apologies for confusion :sweat_smile:

zulipterminal/model.py
```diff
@@ -1164,10 +1164,13 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
                       self.controller, -1)
 
     def formatted_local_time(self, timestamp: int,
-                             *, show_seconds: bool) -> str:
+                             *, show_seconds: bool = False,
+                             show_year: bool = False) -> str:
         local_time = datetime.datetime.fromtimestamp(timestamp)
         if show_seconds:
             return local_time.strftime('%a %b %d %H:%M:%S')
+        if show_year:
```
Ezio-Sarthak: I was referring this addition. Notice that you're indirectly modifying this condition in the later commit (where you're declaring intermediate variable). However, since code-wise this is an insignificant change, I guess we shouldn't bother about discussing it more :)

zulipterminal/ui_tools/views.py
Ezio-Sarthak: Ah, that is surely a confusion. See my comment [here](https://github.com/zulip/zulip-terminal/pull/930/commits/45fda745a561e50f8dd4d3d42c85e323d10e14eb#r579353017) Apologies :sweat_smile:

neiljp:
@jalajcodes Thanks for working on this! :+1: I just squashed these commits together, made a little reformatting and merged as 74a78beb91e37c768890c7f320dcd059152e1ccb :tada:

I squashed since the commits were very small, though refactoring to put the f-string conditional in-line might have been one option. There was a minor bug in the output with regard to spaces which I fixed - though we should probably add a test for this method to prevent introducing this kind of regression :)

PR #932, opened 2021-02-20
feature: Capture screenshots in ZT

Ezio-Sarthak:
This PR adds a feature to capture the screen of ZT using the hotkey `meta c`.
The screenshots are saved locally in `screenshots/` (format: `png`)

This PR could also potentially be an initial step in automating screenshots update in docs :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ zulip_term.egg-info
 *.log
 zt_venv/*
 zulip-terminal-venv/*
+screenshots/*
 
 ## Config files
 zuliprc
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -13,6 +13,7 @@
 |Redraw screen|<kbd>ctrl</kbd> + <kbd>l</kbd>|
 |Quit|<kbd>ctrl</kbd> + <kbd>c</kbd>|
 |View user information (From Users list)|<kbd>i</kbd>|
+|Capture screenshot of the window|<kbd>meta</kbd> + <kbd>c</kbd>|
 
 ## Navigation
 |Command|Key Combination|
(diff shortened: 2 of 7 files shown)


Ezio-Sarthak:

neiljp (commented):
@Ezio-Sarthak I've just given this a try, and provisionally this works well, at least in development.

There are a few factors that we'd need to improve here, including at least:
* pillow isn't that large a dependency (2.2M), but compared to all but lxml (5.5M) and cryptography (3.2M) it's quite large for one that we only use for screenshots, compared to all others being <650k each (.whl files)
* In a basic non-developer install, `tools/` doesn't exist, so this will fail
* `os.system` is not as clean as other operations - though we're calling a python script, so why not treat it as a module instead? :)

We could maybe make it a development dependency for now, though of course that'd need some kind of check or way to distinguish between a development and simple install in the application itself.

neiljp:
Some other notes since the prototype by @Ezio-Sarthak.

Note that this was discussed at [#zulip-terminal>Updating screenshot](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Updating.20screenshot/near/1096770)

For a first version, this would be great as a dev tool:
- we may not need more than pillow to achieve screenshots now?
- pillow can be added to the appropriate 'dev' extra list
- code would check for the presence of pillow on startup, and do nothing if absent
- we have `platform_code`, which seems like a good place for this
- keep the key out of the help (in-app and markdown file) for now
- add some notes to the dev section to document the shortcut key exists :)
- as a follow-up, the PR template can be updated to note about this tool for the visual changes section :)

For a followup v2, more for use by users, with more polish:
- it'd be preferable to keep install size minimal, so add something like a `screenshot` install extra, with additional installation dependencies for that (which could be included in `dev` too)
- this can likely still fail on non-GUI terminal platforms, which needs handling
- now include the key in the help popup, but this could be dynamic to avoid confusion, or show an error message when attempted, if the key is not present (possibly with a custom message in the markdown?)

criticic:
id like to take this up

PR #935, opened 2021-02-23
bugfix: boxes: Call typing status method only when recipients exist.

prah23:
This PR adds a commit that prevents calling `model`'s `send_typing_status_by_user_ids` method without recipients specified.
I've also added a test for the same.


neiljp:
@prah23 Thanks for the bugfix :+1: Merged with extra line and commit text tweak as 0e9be4c20 :tada:

PR #937, opened 2021-02-27
Enhance validation and check for bad/no recipients in PMs

prah23:
This PR adds a commit that refactors the `get_invalid_recipients_emails()` method
of the model to `check_recipient_validity()` which checks if the
recipient's name and email match.

This commit follows this strategy:
* Ensures that the email and the name are of the same user.
* Removes extra text after the email.

The recipients' information is also tidied for all the valid recipients,
by formatting the information to fit the `name <email>` format.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -79,6 +79,107 @@ class TestWriteBox:
         assert not write_box.model.send_private_message.called
 
     @pytest.mark.parametrize(
+        ["raw_recipients", "tidied_recipients"],
+        [
+            ("Human 1 person1@example.com", "Human 1 <person1@example.com>"),
+            (
+                "Human 2 person2@example.com random text",
+                "Human 2 <person2@example.com>",
+            ),
+            (
+                "Human Myself FOOBOO@gmail.com random, Human 1 <person1@example.com>",
+                "Human Myself <FOOBOO@gmail.com>, Human 1 <person1@example.com>",
+            ),
+            (
+                "Human Myself <FOOBOO@gmail.com>, Human 1 person1@example.com random",
+                "Human Myself <FOOBOO@gmail.com>, Human 1 <person1@example.com>",
+            ),
+            (
+                "Human Myself FOOBOO@gmail.com random,"
+                "Human 1 person1@example.com random",
+                "Human Myself <FOOBOO@gmail.com>, Human 1 <person1@example.com>",
+            ),
+            (
+                "Human Myself FOOBOO@gmail.com random, Human 1 person1@example.com "
+                "random, Human 2 person2@example.com random",
+                "Human Myself <FOOBOO@gmail.com>, Human 1 <person1@example.com>, "
+                "Human 2 <person2@example.com>",
+            ),
+            (
+                "Human Myself FOOBOO@gmail.com, Human 1 person1@example.com random, "
+                "Human 2 person2@example.com",
+                "Human Myself <FOOBOO@gmail.com>, Human 1 <person1@example.com>, "
+                "Human 2 <person2@example.com>",
+            ),
+        ],
+        ids=[
+            "untidy_with_improper_formatting",
+            "untidy_with_extra_text",
+            "untidy_first_recipient_out_of_two",
+            "untidy_second_recipient_out_of_two",
+            "two_untidy_recipients",
+            "three_untidy_recipients",
+            "untidy_middle_recipient_out_of_three",
+        ],
+    )
+    @pytest.mark.parametrize("key", keys_for_command("CYCLE_COMPOSE_FOCUS"))
+    def test_tidying_recipients_on_cycling_out(
+        self, mocker, write_box, widget_size, key, raw_recipients, tidied_recipients
+    ):
+        write_box.model.is_valid_private_recipient = mocker.Mock(return_value=True)
+        write_box.private_box_view()
+        write_box.focus_position = write_box.FOCUS_CONTAINER_HEADER
+        write_box.header_write_box.focus_col = write_box.FOCUS_HEADER_BOX_RECIPIENT
+
+        write_box.to_write_box.set_edit_text(raw_recipients)
+        write_box.to_write_box.set_edit_pos(len(raw_recipients))
+
+        size = widget_size(write_box)
+        write_box.keypress(size, key)
+
+        assert write_box.to_write_box.edit_text == tidied_recipients
+
+    @pytest.mark.parametrize(
+        ["raw_recipients", "invalid_recipients"],
+        [
+            ("Human 1 <person2@example.com>", "Human 1 <person2@example.com>"),
+            ("person1@example.com", "person1@example.com"),
+            ("Human 1", "Human 1"),
+        ],
+        ids=["name_email_mismatch", "no_name_specified", "no_email_specified"],
+    )
+    @pytest.mark.parametrize("key", keys_for_command("CYCLE_COMPOSE_FOCUS"))
+    def test_footer_notification_on_invalid_recipients(
+        self,
+        write_box,
+        key,
+        mocker,
+        widget_size,
+        raw_recipients,
+        invalid_recipients,
+    ):
+
+        write_box.model.is_valid_private_recipient = mocker.Mock(return_value=False)
+        write_box.private_box_view()
+        write_box.focus_position = write_box.FOCUS_CONTAINER_HEADER
+        write_box.header_write_box.focus_col = write_box.FOCUS_HEADER_BOX_RECIPIENT
+
+        write_box.to_write_box.edit_text = raw_recipients
+        write_box.to_write_box.set_edit_pos(len(raw_recipients))
+        expected_lines = "Invalid recipient(s) - " + invalid_recipients
+
+        size = widget_size(write_box)
+        write_box.keypress(size, key)
+
+        self.view.controller.report_error.assert_called_once_with(expected_lines)
+        # If there are invalid recipients, we expect the focus
+        # to remain in the to_write_box.
+        assert write_box.focus_position == write_box.FOCUS_CONTAINER_HEADER
+        assert (
+            write_box.header_write_box.focus_col == write_box.FOCUS_HEADER_BOX_RECIPIENT
+        )
+
+    @pytest.mark.parametrize(
         "text, state",
         [
             ("Plain Text", 0),
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -122,8 +122,13 @@ class TestWriteBox:
             "untidy_middle_recipient_out_of_three",
         ],
     )
-    @pytest.mark.parametrize("key", keys_for_command("CYCLE_COMPOSE_FOCUS"))
-    def test_tidying_recipients_on_cycling_out(
+    @pytest.mark.parametrize(
+        "key",
+        keys_for_command("SEND_MESSAGE")
+        + keys_for_command("SAVE_AS_DRAFT")
+        + keys_for_command("CYCLE_COMPOSE_FOCUS"),
+    )
+    def test_tidying_recipients_on_keypresses(
         self, mocker, write_box, widget_size, key, raw_recipients, tidied_recipients
     ):
         write_box.model.is_valid_private_recipient = mocker.Mock(return_value=True)
@@ -148,7 +153,12 @@ class TestWriteBox:
         ],
         ids=["name_email_mismatch", "no_name_specified", "no_email_specified"],
     )
-    @pytest.mark.parametrize("key", keys_for_command("CYCLE_COMPOSE_FOCUS"))
+    @pytest.mark.parametrize(
+        "key",
+        keys_for_command("SEND_MESSAGE")
+        + keys_for_command("SAVE_AS_DRAFT")
+        + keys_for_command("CYCLE_COMPOSE_FOCUS"),
+    )
     def test_footer_notification_on_invalid_recipients(
         self,
         write_box,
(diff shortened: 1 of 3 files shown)


zulipterminal/model.py
```diff
@@ -906,11 +907,35 @@ def _handle_typing_event(self, event: Event) -> None:
...
+                                                      .split(recipient_email))
+
+                        # Ensure that there is no text after the email
+                        # (extra_text) and that the current name matches
+                        # the actual name of the recipient.
+                        if not extra_text.strip() and (recipient_name.strip()
+                           == self.user_dict[recipient_email]['full_name']):
+                            continue
+
+                invalid_recipients.append(recipient)
```
neiljp, line 936: The code design here feels strange - 4 nested if's, with a continue, or else it appends. Could you use a more complex regex to avoid all the manual cleaning?

Also the function of the code is different now - it is connected to the structure of the UI display. Ideally the model would simply validate name/email pairs passed in, with the caller doing any parsing into that format.

zulipterminal/model.py
```diff
@@ -906,11 +907,35 @@ def _handle_typing_event(self, event: Event) -> None:
                 else:
                     raise RuntimeError("Unknown typing event operation")
 
-    def get_invalid_recipient_emails(self, recipient_emails: List[str]
-                                     ) -> List[str]:
-
-        return [email for email in recipient_emails
-                if email not in self.user_dict]
+    def get_invalid_recipients(self, recipient_text: str) -> List[str]:
```
neiljp, line 910: You said this commit is WIP so I'm guessing we'll have a test in a later iteration :)

neiljp (commented):
@prah23 Thanks for the bugfix and looking into the PM validation :+1:

Some comments inline, but also feel free to chat in the stream.

zulipterminal/ui_tools/boxes.py
```diff
@@ -521,17 +521,21 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                         content=self.msg_write_box.edit_text
                     )
             else:
+                self.update_recipient_emails(self.to_write_box)
```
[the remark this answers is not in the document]
prah23: Alright, I'll separate it and nest the condition.

zulipterminal/model.py
prah23: Absolutely, once we decide on how we want to handle this, I'll add relevant tests.

zulipterminal/model.py
prah23: I'll try using a more complex regex, I was a little uncertain about the design, and hence I pushed the commit as WIP.

tests/ui_tools/test_boxes.py
```diff
@@ -718,7 +718,7 @@ def focus_val(x: str) -> int:
         if write_box.focus_position == write_box.FOCUS_CONTAINER_HEADER:
             write_box.header_write_box.focus_col = (focus_val(
                                                     initial_focus_col_name))
-        write_box.model.get_invalid_recipient_emails.return_value = []
+        write_box.model.check_recipient_validity.return_value = []
```
zee-bit, line 721: return type specified here doesn't match with that of the functions. :)

zee-bit (changes requested):
Thanks for working on it. 
This looks good and works as expected. :+1: 

I suppose tests would be added after some more review and feedback?

tests/ui_tools/test_boxes.py
prah23: Right, this commit is just functionality at the moment, because it hasn't been decided on completely. I needed a little feedback on the validation before I proceeded with the tests. Let me know if you have suggestions and I'll modify it.

prah23:

zulipterminal/ui_tools/boxes.py
```diff
@@ -619,14 +619,49 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+                            recipient
+                        )
+                        recipient_name, recipient_email, invalid_text = (
+                            cleaned_recipient_list[0]
+                        )
+                        valid = self.model.check_recipient_validity(
+                            recipient_name, recipient_email
+                        )
+                        tidy_recipient_info += (", " if tidy_recipient_info
+                                                else "")
```
neiljp, line 641: This behavior suggests we should use `join` for this too, ie. build a list during this loop.

prah23:

prah23:

zulipterminal/model.py
```diff
@@ -914,11 +914,14 @@ def _handle_typing_event(self, event: Event) -> None:
...
-        return [email for email in recipient_emails
-                if email not in self.user_dict]
+    def check_recipient_validity(self, recipient_name: str,
+                                 recipient_email: str,
+                                 ) -> bool:
+        if recipient_email and recipient_email in self.user_dict:
+            verified_name = self.user_dict[recipient_email]['full_name']
+            if recipient_name == verified_name:
+                return True
+        return False
```
Ezio-Sarthak, line 924: Currently, it's fine, though I feel it'd be nice to split this into a prior commit, where you could add relevant tests for this method as well.

zulipterminal/ui_tools/boxes.py
```diff
@@ -208,6 +208,37 @@ def update_recipient_emails(self, write_box: ReadlineEdit) -> None:
...
+        recipients = [recipient.strip()
+                      for recipient in write_box.edit_text.split(',')
+                      if recipient.strip()]
+
+        for recipient in recipients:
+            cleaned_recipient_list = re.findall(
+                r"^(.*?)(?:\s*?<?([\w\.-]+@[\w\.-]+)>?(.*))?$",
+                recipient
+            )
+            recipient_name, recipient_email, invalid_text = (
```
Ezio-Sarthak, line 224: minor: If we aren't interested in `invalid_text`, we could replace it with `_` :)

Ezio-Sarthak (changes requested):
This looks good overall :+1:. Added few comments. Hope it helps :)

zulipterminal/model.py
prah23: I considered doing that, but a change here in a previous commit would need me to change the function call and the calling parameters, which needs extraction of the user's name as well, which is basically what the current standalone commit is doing. :) 

For tests, I've covered the functionality in the more general tests for validation, with extensive cases that test the functioning of this method, I don't think we'd need more explicit tests for the method.

zulipterminal/ui_tools/boxes.py
prah23: We're not interested in `invalid_text` but I'm inclined to let the name stay so that it's somewhat understandable from reading the code why we're discarding it, but that also sounds like a need for some comments within, I'll add them.

neiljp (commented):
@prah23 This does continue to work fine, the code looks better organized and there are now tests of course :+1: However, while not always meaningful, this is looking quite large for one commit now - though of course now including tests does affect that!

What is noticeable, partly from the different *sets* of tests, is that there do seem to be distinct changes all in the one commit. If these are separate then it makes it easier not just to review (or read later in the git log) but also see the progressive refinement of the code.

It seems like this could be split into changes (commits) to:
* Extend the invalid-recipients/email code to mention autocomplete (that wasn't in the previous iteration I think? is that from another PR?) - with one of those new tests?
* Validate & tidy - possibly even split into two commits, one building on the other, perhaps the first including the model change and providing more information to the footer feedback? The footer test(s) would then developed as you add that feature, and that change in the test would document that change

I'm not completely averse to this large commit, partly since a lot is new code, but if you're ok with giving the splitting a try, it's a good skill to develop :)

The advantage of multiple commits is that you get multiple chances to describe what you're adding/changing and why, "attached" to only each change, rather than needing to give a summary of potentially a large change with bullets that relate to... some location?

**(corrected mention)**

zulipterminal/model.py
neiljp: I just noticed this comment. This method is really sort of just a helper used in one place - ideally we would have a test, but it's like an external (but encapsulated in another class) implementation function at least for now. I suspect this code will change again in the near future.

As per my review below, ideally this change could be accompanied by only a small change elsewhere. That is, in some commit we need this different function in the model to make a feature work elsewhere, but if we try and avoid doing too much in the commit at once, it'll be clearer that this change is also made just to support that change/feature.

zulipterminal/ui_tools/boxes.py
```diff
@@ -216,6 +216,50 @@ def track_idleness_and_update_status() -> None:
...
+            # Discard invalid_text as part of tidying up the recipient.
+
+            if recipient_email:
+                valid = self.model.check_recipient_validity(
+                    recipient_name, recipient_email
+                )
+            else:
+                valid = False
+
+            if valid:
```
neiljp, line 245: This may be simplified and clearer with the `is_*` form I mention above?

zulipterminal/ui_tools/boxes.py
```diff
@@ -605,6 +605,11 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                         msg_id=self.msg_edit_id,
                     )
                 else:
+                    valid = self.tidy_valid_recipients_and_notify_invalid_ones(
```
neiljp, line 616: Would it be more accurate to name these as `all_valid` or similar?

tests/ui_tools/test_boxes.py
```diff
@@ -808,6 +808,58 @@ def test_tidying_recipients_on_keypresses(
...
+        write_box.private_box_view()
+        write_box.focus_position = write_box.FOCUS_CONTAINER_HEADER
+        write_box.header_write_box.focus_col = write_box.FOCUS_HEADER_BOX_RECIPIENT
+
+        write_box.to_write_box.edit_text = raw_recipients
+        write_box.to_write_box.set_edit_pos(len(raw_recipients))
+        expected_lines = [
+            "Invalid recipient(s) - " + invalid_recipients,
+            " - Use ",
+            ("code", primary_key_for_command("AUTOCOMPLETE")),
```
neiljp, line 846: "code" should likely be "footer_contrast"?

neiljp (commented):
@prah23 Splitting into smaller commits is definitely an improvement so far and I like the suggestions in the last commit :+1:

I've left some comments inline on specific code.

Generally I'd like you to think about smaller commit structuring a little more in future: things happening in adjacent commits are broadly OK (adding, migrating to new code, removing - in first 3 commits here) as it's possible to *remember* the flow of code to some extent across commits. However, adjusting the code so that each commit is a migration (refactor) makes it possible to directly look at a change within a single commit, in the context of the overall flow in a PR.

prah23:
Thanks for the review, @neiljp!
I've restructured the commits as suggested, and accommodated the inline changes suggested.

zulipterminal/ui_tools/boxes.py
```diff
@@ -587,6 +636,11 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         elif is_command_key("SAVE_AS_DRAFT", key):
             if not self.msg_edit_id:
                 if self.to_write_box:
+                    all_valid = self.tidy_valid_recipients_and_notify_invalid_ones(
+                        self.to_write_box
+                    )
+                    if not all_valid:
+                        return key
```
neiljp, line 651: These 2 extra checks (draft/send)  are still part of this commit; you don't mention this change in the commit text either, but it'd probably come across as an "also applies checking to x and y", which indicates it could go into another commit.

zulipterminal/ui_tools/boxes.py
```diff
@@ -650,21 +704,12 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
-                        invalid_emails_error = (
-                            f"Invalid recipient(s) - {', '.join(invalid_emails)}"
-                        )
-                        self.view.controller.report_error(invalid_emails_error)
+                    all_valid = self.tidy_valid_recipients_and_notify_invalid_ones(
+                        self.to_write_box
+                    )
+                    if not all_valid:
                         return key
+                    self.update_recipient_emails(self.to_write_box)
```
neiljp, line 720: Is there some change in behavior by having this after instead of earlier? Does it warrant a mention?

zulipterminal/ui_tools/boxes.py
```diff
@@ -650,21 +704,12 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
-                        if not self.model.is_valid_private_recipient(
-                            recipient_email
-                        )
-                    ]
-
-                    if invalid_emails:
-                        invalid_emails_error = (
-                            f"Invalid recipient(s) - {', '.join(invalid_emails)}"
-                        )
-                        self.view.controller.report_error(invalid_emails_error)
```
neiljp: Other than adding the feature for the other keypresses, this commit also extracts the code into a function *and* changes it - those could easily be a separate refactor followed by the extra checks.

neiljp (commented):
@prah23 Thanks for responding to the minor points, and the restructure is a good first step - see if you can identify the others I pointed at :+1:

zulipterminal/ui_tools/boxes.py
prah23: It ensures that the emails extracted are valid for sure, I've added a comment for the same.

prah23:
Thanks for your patience on this, @neiljp!
I've added another commit to extract validation logic to a helper, and the thorough validation is a commit after that, which changes the behavior inside the helper. The extension of the validation to `SEND_MESSAGE` and `SAVE_AS_DRAFT` is pushed to another commit as well.

neiljp:
@prah23 Thanks for the restructure - this has cleaner smaller commits now :+1:

I just removed a duplicate assignment in the last commit and pushed to the PR before merging :tada:

prah23:
Thanks for your patience!

prah23:
Thanks for your patience!

PR #942, opened 2021-03-05
bugfix: boxes: Fix quote text rendering for xfails.

zee-bit:


tests/ui/test_ui_tools.py
```diff
@@ -1936,25 +1936,20 @@ def test_keypress_EDIT_MESSAGE(self, mocker, message_fixture, widget_size,
...
-                            <p>B</p>
-                            <blockquote>
-                                <p>C</p>
-                            </blockquote>
-                        </blockquote>
-        """, "{} {} A\n{} B\n{} {} C",
-                     marks=pytest.mark.xfail(reason="rendered_bug")),
+        ("""<blockquote>
+                <blockquote>
+                    <p>A<br/>B</p>
```
neiljp, line 1941: A quick test indicated the server actually generates a `<br>` (not xml) and after both A and B too?

There are also extra newlines in each blockquote associated with ::before and ::after - I've not looked at this in a while but are we relying on those to be present for spacing purposes?

zee-bit: There's a similar test just above this, that had `<br/>` as well, so I didn't cross-check with the server.

Yes, I presume they are there because the quoting expects some text after the quote, hence the `newline`?

neiljp (commented):
@zee-bit Thanks for tackling this - This looks like a great improvement already :+1: Our quoting is OK right now, but more complex quotes certainly need some work!

As you indicated, the next step is adding more test cases; these need not be xfails if you can fix them at the same time :)
Some areas that stand out:
* blank lines after multi-quotes, eg. no response to a quote and quoted again -> extra blank line per quote level
* text above the quoted text, which is much more common with the "<user> said..." format we have now, for quote-level>=2
* handling "split" or multiple quotes in the same message; this is particularly evident with the simple quotes enabled by starting lines with '>', though this isn't used in auto-quoting

Some of that may be part of #402. More work on this would be great, but if not then we should likely add more test cases with xfails, or at least a tracking issue which could point to #402 or perhaps replace it.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1310,25 +1310,28 @@ def indent_quoted_content(soup: Any, padding_char: str) -> int:
...
+                if len(child_block) == 0:
+                    child_iterator = child_list
+                else:
+                    child_iterator = child_list[1:]
             for child in child_iterator:
                 new_tag = soup.new_tag('p')
                 new_tag.string = actual_padding
                 # If the quoted message is multi-line message
                 # we deconstruct it and pad it at break-points (<br/>)
-                if child.findAll('br'):
```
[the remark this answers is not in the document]
zee-bit: Yes, this seemed unnecessary. The fact that if `findAll` returns an empty list, the for loop won't run anyways, was enough?

tests/ui/test_ui_tools.py
```diff
@@ -1930,31 +1930,26 @@ def test_keypress_EDIT_MESSAGE(self, mocker, message_fixture, widget_size,
...
             </blockquote>""", "{} czo [1]\n"),
-        pytest.param("""<blockquote>
-                            <blockquote>
-                                <p>A<br>
-                                B</p>
-                            </blockquote>
-                        </blockquote>
-            """, "{} {} A\n{} {} B",
-                     marks=pytest.mark.xfail(reason="rendered_bug")),
-        pytest.param("""<blockquote>
```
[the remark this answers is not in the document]
zee-bit: So, do I need to use `pytest.param`, with inline ids, for all the test or just xfails?
I am guessing the former?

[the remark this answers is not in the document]
zee-bit: Currently, we use them just for `xfails`.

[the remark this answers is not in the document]
neiljp: By having it for all, this allows the ids to be near to each parametrization - and also makes it easier to add/remove things like xfails if we wish.

zee-bit:
@neiljp PR has been updated keeping in mind your review above. 
- I have managed to clear out most of the extra newlines for multilevel quotes.
- The `user said ...` lines for multilevel quotes have also been fixed, though, for quote-level >= 3, the message formatting seems out-of-place(probably some bug in `soup2mark()`?)
- I tried fixing handling "splits", but it had some adverse side-effects in other simpler quotes, that were previously fixed.
- Tests have also been added(and amended) with a few xfails.

zee-bit:

tests/ui/test_ui_tools.py
```diff
@@ -1908,61 +1909,76 @@ def test_keypress_EDIT_MESSAGE(self, mocker, message_fixture, widget_size,
...
+                    <p>C</p>
+                </blockquote>
+                <p>D</p>""",
+             ("{} {} {} A\n\n"
+              "{} {} B\n\n"
+              "{} C\n\n"
+              "D"),
+             id="quoted level 3"),
+        case("""<blockquote>
+                    <p>A<br>B</p>
```
neiljp, line 1946: Does the server send this over two lines?

tests/ui/test_ui_tools.py
```diff
@@ -1937,17 +1937,17 @@ def test_keypress_EDIT_MESSAGE(self, mocker, message_fixture, widget_size,
                     <p>C</p>
                 </blockquote>
                 <p>D</p>""",
-             ("{} {} {} A\n\n"
-              "{} {} B\n\n"
-              "{} C\n\n"
```
neiljp: Could you explain why the newlines are removed?

tests/ui/test_ui_tools.py
```diff
@@ -1976,8 +1975,86 @@ def test_keypress_EDIT_MESSAGE(self, mocker, message_fixture, widget_size,
...
              ("{} {} A\n"
               "{} B\n"
-              "{} {} C"),
-             id="quoted level 2-1-2",
+              "{} \n"
+              "{} {} C\n"),
+             id="quoted level 2-1-2"),
+        case("""<p><a href='https://chat.zulip.org/1'</a>czo</p>
+                <blockquote>
+                    <p><a href='https://chat.zulip.org/2'</a>czo</p>
```
neiljp, line 1983: This applies to the original links too, but these are actually invalid links - how do these pass?

zulipterminal/ui_tools/boxes.py
```diff
@@ -1136,6 +1136,13 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
                 metadata['time_mentions'].append((time_string, source_text))
             else:
                 markup.extend(cls.soup2markup(element, metadata)[0])
+            # Remove extra newline from the end of blockquotes
+            if(markup[-1] == '\n'
+               and isinstance(markup[-2], tuple)
+               and markup[-2][0] == 'msg_quote'
+               and isinstance(markup[-2][1], list)
+               and markup[-2][1][-1] == '\n'):
+                markup = markup[:-1]
```
neiljp, line 1145: This looks rather messy. Can this be included in the quote handler? If so, would it be clearer there?

neiljp (commented):
@zee-bit The refactor definitely improves the readability, though it does expose the broken links!
This is definitely an improvement over the previous situation and works for higher levels of quoting including text above it, but we still end up with some strange behavior and style corruption at level 3-4+?

This appears an improvement over the previous iteration, though I'm not sure precisely what you've changed - and the newline changes are somewhat unclear.

This doesn't seem to address #402 directly (mixing of quote styles), though given the improvement already I'm open to merging a simplified version and iterating, if we can clarify the open points.

tests/ui/test_ui_tools.py
zee-bit: Ahh, yes. Sorry, forgot to change this!

tests/ui/test_ui_tools.py
zee-bit: > blank lines after multi-quotes, eg. no response to a quote and quoted again -> extra blank line per quote level

This is for addressing the first point of your review above.

Are those extra newlines important?

zulipterminal/ui_tools/boxes.py
zee-bit: This piece of code checks for extra newlines after quotes and removes them. But, I agree, this looks messy and complex, mainly due to the following reason:
- Our `soup2mark` function is recursive, hence it becomes difficult to check for **unnecessary** blank lines after quotes.
Currently, I am relying on `markup` to check for any appended `\n` at the end of each recursion-step, then comparing it with the previous element if it was `blockquote` and had a `\n` at the end already. We then remove this extra newline from the end of markup.

Let me see if I can simplify this!

tests/ui/test_ui_tools.py
zee-bit: I don't understand what do you mean by "invalid links". Are you pointing out the incorrect syntax used for `a` tag?

zee-bit:

neiljp:
@zee-bit I've just merged the first two commits, and replied on the stream about this.

neiljp:
As per my comment in the stream, let's take the remainder of this regarding improving the newline handling into a separate PR? I'm more convinced about the remaining commit now, but in any case there seem to be remaining issues, possibly related to the html returned by the server.

zee-bit:
The later commit from this PR, related to removing extra newlines from the end of quoted messages, is branched off to #1003, as requested. Further improvements will be pushed to that PR. Closing this now.

PR #943, opened 2021-03-06
bugfix: starred messages: Update messages in starred narrow.

### LINKED ISSUE #940: Remove unstarred messages from starred messages narrow
Steps to reproduce:
1) Star a message. You can find it added to the starred messages narrow.
2) Now unstar it. The star in the top right corner is no longer visible but the message can still be seen in the starred messages narrow. 

This is the case only within the same ZT session. As per the behaviour in the webapp, unstarred messages should be removed from starred messages narrow immediately.

Abhirup-99:
The commit updates ZT behavior in starred message section to match the web app. The current behavior involves removing the message when the user moves out of the starred section.
Fixes #940 .

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -1153,9 +1153,14 @@ class Model:
             if operation == 'add':
                 if flag_to_change not in msg['flags']:
                     msg['flags'].append(flag_to_change)
+                if flag_to_change == 'starred':
+                    self.index['starred_msg_ids'].add(message_id)
             elif operation == 'remove':
                 if flag_to_change in msg['flags']:
                     msg['flags'].remove(flag_to_change)
+                if(message_id in self.index['starred_msg_ids']
+                   and flag_to_change == 'starred'):
+                    self.index['starred_msg_ids'].remove(message_id)
             else:
                 raise RuntimeError(event, msg['flags'])
 
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -113,7 +113,8 @@ class MessageView(urwid.ListBox):
         ids_to_keep = self.model.get_message_ids_in_current_narrow()
         if self.log:
             top_message_id = self.log[0].original_widget.message['id']
-            ids_to_keep.remove(top_message_id)  # update this id
+            if top_message_id in ids_to_keep:
+                ids_to_keep.remove(top_message_id)  # update this id
             no_update_baseline = {top_message_id}
         else:
             no_update_baseline = set()
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1531,7 +1531,10 @@ class TestModel:
         operation, model.server_feature_level = update_message_flags_operation
 
         model.index = dict(messages={msg_id: {'flags': flags_before}
-                                     for msg_id in indexed_ids})
+                                     for msg_id in indexed_ids},
+                           starred_msg_ids=set([msg_id
+                                                for msg_id in indexed_ids
+                                                if 'starred' in flags_before]))
         event = {
             'type': 'update_message_flags',
             'messages': event_message_ids,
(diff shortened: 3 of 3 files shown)


zulipterminal/model.py
```diff
@@ -355,6 +355,8 @@ def toggle_message_star_status(self, message: Message) -> None:
         base_request = dict(flag='starred', messages=[message['id']])
         if 'starred' in message['flags']:
             request = dict(base_request, op='remove')
+            if message['id'] in self.index['starred_msg_ids']:
+                self.index['starred_msg_ids'].remove(message['id'])
```
neiljp, line 359: This would be better occurring in the event handler, as if you have multiple clients running, you want the update in both clients. For example, if you have the webapp and zulip-terminal both open.

Abhirup-99:

Abhirup-99:
@zulipbot remove "PR awaiting feedback"

Abhirup-99:

zulipterminal/ui_tools/views.py
```diff
@@ -113,7 +113,8 @@ def load_old_messages(self, anchor: int) -> None:
         ids_to_keep = self.model.get_message_ids_in_current_narrow()
         if self.log:
             top_message_id = self.log[0].original_widget.message['id']
-            ids_to_keep.remove(top_message_id)  # update this id
+            if top_message_id in ids_to_keep:
+                ids_to_keep.remove(top_message_id)  # update this id
```
neiljp, line 117: Is this a separate bugfix?

zulipterminal/model.py
```diff
@@ -1156,6 +1156,9 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
             elif operation == 'remove':
                 if flag_to_change in msg['flags']:
                     msg['flags'].remove(flag_to_change)
+                if(message_id in self.index['starred_msg_ids']
+                   and flag_to_change == 'starred'):
+                    self.index['starred_msg_ids'].remove(message_id)
```
neiljp, line 1161: This handles the remove case, but it seems we also are missing the add case?

neiljp (commented):
@Abhirup-99 I just tested this manually and the removal appears to work but I think we need to handle addition too?

Have you adjusted the tests to just pass so far? It'd be useful to have test cases for events triggering these change(s).

It's quicker to merge if you can set the commit title to follow the style we have in the rest of our git log (also see the README)

zulipterminal/ui_tools/views.py
Abhirup-99: This is to fix a race round condition when inside the current starred narrow, we unstar and remove the first message from the top, but when loading older messages, we try to remove this message_id again which is technically not present in the set anymore throwing errors.

zulipterminal/model.py
Abhirup-99: Updating shortly.

zulipterminal/ui_tools/views.py
Abhirup-99: This is more of a hypothesis, rather than an exact cause. I tested unstaring and moving the cursor up with and without the if block. Without the if block, I go an exception twice for trying to remove the message_id, which is not present. This although didn't occur with the if block. That's why added the if block.

Abhirup-99: Update: I tested again without the if block, and found the error. 
Steps to check it out:
- Remove the if block.
- Move to the topmost loaded message inside the starred_messages narrow. Ensure that new messages don't get loaded.
- Unstar it.
- Try moving up(loading older messages).

Abhirup-99: Screenshot from 2021-03-08 17-52-13

Abhirup-99:

Abhirup-99:

Abhirup-99:
@neiljp please take a look when you are free. I have updated the commit message to make it more descriptive with all the individual changes I have done in this pr. Thanks :).

neiljp:
@Abhirup-99 Thanks for working on this, this is a great bugfix to have :+1: As per my message in the stream, I've merged this manually with a few reformatting changes only :tada: I excluded the view change for now since I think we should address this separately.

PR #946, opened 2021-03-12
boxes: Add support for renaming (no topic) messages.

### LINKED ISSUE #837: Message editing: Add support for renaming of "(no topic)" messages
We currently support editing message topics of one or many messages, but only via one's own messages in those topics.

If a topic is "topicless", ie. shown as "(no topic)", then depending on server settings it should be possible to edit it independently of contribution to the topic:
https://zulip.com/help/configure-message-editing-and-deletion

zee-bit:
This PR adds support for renaming messages that had no topic, by anyone as long as `realm_allow_message_editing` is `True` for the organization. This is the behavior that is followed by webapp for message editing events.

Fixes #837.


tests/ui/test_ui_tools.py
```diff
@@ -1863,20 +1863,56 @@ def test_main_view_generates_EDITED_label(self, mocker,
...
+        ({
+            'sender_id': 1,
+            'timestamp': 1,
+            'subject': '(no topic)'
+         }, False, True, False),
     ], ids=['msg_sent_by_other_user',
             'topic_edit_only_after_time_limit',
             'editing_not_allowed',
-            'all_conditions_met'])
+            'all_conditions_met',
```
Ezio-Sarthak, line 1904: minor: Let's keep this in the end to provide more clarity.

Ezio-Sarthak (changes requested):
Thanks for taking this up :+1:  This provisionally works well.
There are some aspects that I'd suggest worth taking a look at, other than minor code changes in my comments:
 - One can't edit a topic multiple times (in a session) for messages sent from the same author?
 - [unrelated] In web app, changing topic name (other user's message) shows an edit marker for all following message sent by that user, which seems not to be the case if seen in ZT? (to clarify: editing in web app, while running a ZT session)

zee-bit:
@Ezio-Sarthak Thanks for the review. :relaxed: :
Re "unrelated", you are probably talking about community topic editing, here: #793.
(I should probably query `Kaustubh Nair` about the status of that PR, and if he's busy, maybe I can take that forward. :) )

zee-bit:
@Ezio-Sarthak Thanks for the review. :relaxed:
Re "unrelated", you are probably talking about community topic editing, here: #793.
(I should probably query @kaustubh-nair about the status of that PR, and if he's busy, maybe I can take that forward. :) )

tests/ui/test_ui_tools.py
```diff
@@ -1896,20 +1896,56 @@ def test_keypress_STREAM_MESSAGE(self, mocker, msg_box, widget_size,
...
+            'editing_not_allowed_for_no_topic',
+            'all_conditions_met', ])
     def test_keypress_EDIT_MESSAGE(self, mocker, message_fixture, widget_size,
                                    expect_editing_to_succeed,
                                    to_vary_in_each_message,
                                    realm_editing_allowed,
                                    msg_body_edit_enabled,
                                    key):
+        if message_fixture['type'] == 'private':
+            to_vary_in_each_message['subject'] = ''
```
neiljp, line 1948: Is this what we expect in PMs, or is the key absent?

neiljp (commented):
@zee-bit I just merged a quick bugfix for #1034 (#1035) today, and wanted to follow-up on this PR since it'll need a small rebase. Most of the behavior of this I'm interested in seeing checked in the tests rather than manually - since it depends on server settings.

#793 does seem to be a good follow-up to rebase/amend, and in terms of testing I realize that the settings we have won't be updated from server->ZT, so another good follow-up would be to add event handling for those changes - otherwise to test this manually we have to change the setting and restart ZT each time?

tests/ui/test_ui_tools.py
zee-bit: The API docs do say that we should expect an empty string here. Quote:
> subject: string The topic of the message. Currently always "" for private messages, though this could change if Zulip adds support for topics in private message conversations.

I also did a manual test just to be extra sure. :)

zee-bit:

zulipterminal/ui_tools/boxes.py
```diff
@@ -1637,37 +1637,63 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
-                )
+            if (
+                self.message["sender_id"] != self.model.user_id
+                and self.message["subject"] != "(no topic)"
+            ):
+                if self.message["type"] == "stream":
+                    self.model.controller.report_error(
+                        " You can't edit messages sent by other users"
+                        " with non-empty topic."
+                    )
```
zee-bit, line 1648: This should be the accurate definition of "(no topic)" messages i.e. empty topic. But, I am a bit hesitant to print this in the footer (the user might not know how we represent empty topics?).
I also had `You can't edit messages sent by other users which do not have "(no topic)"`

zee-bit: After a short discussion with @prah23 via PM's we've decided to go with "You can't edit messages sent by other users that already have a topic." :)

tests/ui/test_ui_tools.py
```diff
@@ -2251,13 +2251,40 @@ def test_keypress_STREAM_MESSAGE(
...
+                60,
+                False,
+                " Editing sent message is disabled.",
+                " Editing sent message is disabled.",
+            ),
+            ({"sender_id": 1, "timestamp": 45}, True, True, 60, True, None, None),
+            ({"sender_id": 1, "timestamp": 1}, True, True, 0, True, None, None),
         ],
         ids=[
             "msg_sent_by_other_user",
```
neiljp, line 2344: With more parametrize fields and the next commit adding more cases, it seems good to inline the ids first before the current 2nd commit.

neiljp (commented):
@zee-bit The tests have more information, but putting the case ids inline will make it easier to determine that the right output is happening :+1:

For the code itself, as we get closer to including all the range of edit possibilities, and given your "do we need this" in the code, I'm a little concerned it's tricky to follow the branches. We could put together a document which translates the "who can edit what" table in the zulip help pages into a clear rejection (can't edit because x) flow and mapping of those to text, given the range of inputs - which is sort of what the tests will illustrate, and be clearer with those inline ids :)

It might be useful to think about where community topic editing fits into the flow - I think we're sort of enabling this by default when we edit from our own messages but rename more, and used to seeing it work on czo.

zee-bit:
Thanks for the review Neil,
Re the "do we need this" in code, I'm 98% sure that we don't need that condition, since we should have already handled that at the top? To make the conditional branches easier to comprehend, I've put up small comments at the top that should provide much more clarity to our handling of message-editing, for now.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1676,10 +1676,20 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
             self.model.controller.view.middle_column.set_focus("footer")
         elif is_command_key("EDIT_MESSAGE", key):
-            if self.message["sender_id"] != self.model.user_id:
-                self.model.controller.report_error(
-                    " You can't edit messages sent by other users."
-                )
+            # User can't edit messages of others that already have a topic
+            if (
+                self.message["sender_id"] != self.model.user_id
+                and self.message["subject"] != "(no topic)"
```
neiljp, line 1682: We mentioned this in the test, but might be worth noting here that a private subject is ""?
That said, can you move the subject check into the stream conditional? That seems cleaner.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1690,23 +1700,45 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+                    # Allow editing topic if the message has "(no topic)"
+                    if self.message["subject"] == "(no topic)":
                         self.model.controller.report_warning(
-                            " Only topic editing allowed."
-                            " Time Limit for editing the message body has"
-                            " been exceeded.",
+                            " Only topic editing is allowed."
                         )
                         msg_body_edit_enabled = False
+                    # FIXME: Is this else necessary, we already handle this at the top?
```
neiljp, line 1729: I think this will be easier to reason about if you move the first conditional around as I indicate above.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1690,23 +1700,45 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+                            )
+                            msg_body_edit_enabled = False
+                elif self.message["type"] == "stream":
+                    # Allow editing topic if the message has "(no topic)"
+                    if self.message["subject"] == "(no topic)":
                         self.model.controller.report_warning(
-                            " Only topic editing allowed."
-                            " Time Limit for editing the message body has"
-                            " been exceeded.",
+                            " Only topic editing is allowed."
```
neiljp, line 1726: I'm inclined to think that we should indicate why it's allowed - this is someone else's message but with no-topic?

zulipterminal/ui_tools/boxes.py
```diff
@@ -1690,23 +1700,45 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+                    else:
+                        self.model.controller.report_error(
+                            " You can't edit messages sent by other users that"
+                            " already have a topic."
+                        )
+                        return key
+                else:
+                    # The remaining case is of a private message not belonging to user.
+                    self.model.controller.report_error(
+                        " You can't edit private messages sent by other users."
```
neiljp, line 1739: You're not testing this - is it already covered?

neiljp (commented):
@zee-bit The inlining of ids definitely helps - but there are a few other changes which may help simplify the tests that I've noted too. Reading the code again, it seems easier to follow now - particularly if one of the conditionals are moved around a little?

I'm also wondering if now or later we may want to reorder checks in order to give better error messages - but that's different than getting the new logic working :)

zulipterminal/ui_tools/boxes.py
zee-bit: I think we can remove this whole block from here?
They should already be handled fine with the below conditionals.

zulipterminal/ui_tools/boxes.py
zee-bit: Actually, because of the equivalent conditional at the top, this case was already handled by that and returned so we never reached here.
Removing the top block as I suggested in one of the above comments should make this feasible.

tests/ui/test_ui_tools.py
```diff
@@ -2282,27 +2282,70 @@ def test_keypress_STREAM_MESSAGE(
...
+            case(
+                {"sender_id": 1, "timestamp": 1},
+                True,
+                False,
+                60,
+                True,
+                " Only topic editing allowed."
+                " Time Limit for editing the message body has been exceeded.",
+                " Time Limit for editing the message has been exceeded.",
+                id="topic_edit_only_after_time_limit",
```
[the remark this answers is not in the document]
zee-bit: > This isn't related to this feature, but the test (at first glance!) says editing should succeed here, yet in manual testing (and per the warnings/errors you added) editing (partially) succeeds with streams and fails with PMs!

This seems to also true for `msg_body_edit_enabled`?
Anyways, I've expanded both to contain values for both types of messages.

zulipterminal/ui_tools/boxes.py
zee-bit: We can't remove this block. It handles the "anytime" message-edit case - almost forgot about that earlier. :sweat_smile: 

I don't think we can move the subject check into the stream conditional below too. We don't enter the next block since `realm_message_content_edit_limit_seconds` will be `0` for "anytime" message-edit, so we need to handle that case outside either in the beginning or end?

zulipterminal/ui_tools/boxes.py
zee-bit: I have amended the equivalent error message at the top to print the same thing as here.

neiljp (commented):
@zee-bit The test refactoring are good clean changes making the feature commit much easier to follow :+1:

This is complicated logic, so the tests are understandably complicated - but the more cleanly we can express the behavior the better, and this looking great!

tests/ui/test_ui_tools.py
```diff
@@ -2380,6 +2397,16 @@ def test_keypress_EDIT_MESSAGE(
...
+        if message_type == "stream" and expected_footer_text_stream:
+            if expected_editing_to_succeed[message_type]:
+                report_warning.assert_called_once_with(expected_footer_text_stream)
+            else:
+                report_error.assert_called_once_with(expected_footer_text_stream)
+        elif message_type == "private" and expected_footer_text_private:
+            if expected_editing_to_succeed[message_type]:
+                report_warning.assert_called_once_with(expected_footer_text_private)
+            else:
+                report_error.assert_called_once_with(expected_footer_text_private)
```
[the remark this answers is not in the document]
zee-bit: For each dict there could be different `report_*` functions for different message types. I think that will make the test even more complex if I understand your statement correctly?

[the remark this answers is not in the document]
neiljp: This seems to be resolved with the change to a dict for the footer_text?

neiljp:
@zee-bit While I made some comments, I'm just giving this a final look and may merge soon as-is.

neiljp:
@zee-bit Thanks for getting this done! This has nicely tidied up these tests too :+1: I just pushed here with some very minor commit text changes and will merge shortly :tada:

PR #947, opened 2021-03-13
ui/boxes: Improve behaviour of non reply compose keys.

### LINKED ISSUE #871: Improve compose hotkey behavior (as compared to replies)
Most of the compose-message actions behave fairly well, and many are reply-based, often with extra features.

Since our autocomplete of streams/topics has been (and for users still is) less useful, the best approach for ZT users has been to rely upon ways to reply, rather than do straight compose actions. For example, one might go to a stream or topic and reply but editing the stream/topic, or reply to a user in PMs, or by using the user list.

Now that we have autocomplete for stream/topic in the compose header, this is already a lot better, and hopefully we will have this for PMs soon too #587. **Before we do, we should ensure that non-reply compose works well.**

* [ ] Currently r/enter and other keys work fine for replying, but x/c seem not quite right - in particular `c` while in a PM narrow.
* [ ] We can expand non-replying compose (x/c) to many other contexts - they should work almost everywhere?

mkp6781:
Brings in consistent behaviour for use of non reply compose keys `x`/`c`.
Now, keypresses `x` and `c` always opens up `private_box_view` and
`stream_box_view` respectively.

Earlier, non reply compose hotkeys were configured for use only when focus
was on the `MiddleColumnView`. This commit expands its scope to enable use
of `x`/`c` even when focus is on users list, stream list, topics list, or
top buttons.

Tests added.

Fixes #871

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -243,6 +243,18 @@ class TestView:
         view.model.controller.show_all_mentions.assert_called_once_with(view)
         assert view.body.focus_col == 1
 
+    @pytest.mark.parametrize('key', keys_for_command('STREAM_MESSAGE'))
+    def test_keypress_STREAM_MESSAGE(self, view, mocker, key, widget_size):
+        view.middle_column = mocker.Mock()
+        view.body = mocker.Mock()
+        view.controller.is_in_editor_mode = lambda: False
+        size = widget_size(view)
+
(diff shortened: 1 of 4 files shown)


neiljp:
@mkp6781 Thanks for fixing this :tada: I've merged this with updates to the commit text and to include an expected result in the parametrized test, rather than conditional code, which makes the tests clearer.

PR #948, opened 2021-03-14
bugfix: core: Improve consistency while toggling from topic to stream.

### LINKED ISSUE #934: Topic -> Stream zoom-out fails for old topic after viewing recent up-to-date stream
To reproduce:
1. Narrow to a stream you're up to date on (select stream in stream list) -> interleaved stream view
2. Move cursor left, toggle into topics view, `q` and search for eg. `388`, hit enter -> topic view for that topic
3. Hit `z`

I'd expect this to zoom out to the context of the message that was selected, in the stream view, eg. around April 2020.

Instead this goes back to the first stream view.

Note that this does not occur if you go straight to the particular topic, ie. skip step (1).

Abhirup-99:
Fixes #934.

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -138,6 +138,8 @@ class TestController:
         assert controller.model.stream_id == msg_box.stream_id
         assert controller.model.narrow == expected_narrow
         controller.view.message_view.log.clear.assert_called_once_with()
+        assert (controller.model
+                .index['stream_msg_ids_by_stream_id'][205] == set())
 
         widget = (controller.view.message_view.log
                   .extend.call_args_list[0][0][0][0])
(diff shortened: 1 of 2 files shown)


Abhirup-99:

Ezio-Sarthak (commented):
Thanks for taking this up! The toggle behavior seems OK now :+1: 

As you rightly pointed out about the gap in 2 sets of indexed messages, I believe fixing that should fix this too?

I also found another fix locally for the stream toggle bug. Discussion [link](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Stream.20toggle.20bugfix.20(.23T934)) :)

neiljp:
@Abhirup-99 I think we should avoid blowing away the message cache if we can - though I appreciate that this does look to be a solution - as we could end up fetching messages multiple times if nothing else. If we did take this approach, we would likely want to isolate the changes in the model in any case, as the change digs into the model from the controller; code like that still exists in ZT, but it's a lot easier to reason about otherwise.

See the topic @Ezio-Sarthak linked for other discussion.

PR #949, opened 2021-03-15
refactor: Migrate to f-strings in tests wherever reasonable.

Ezio-Sarthak:
Some expressions are tidy enough in the current format and are therefore excluded.


neiljp:
@Ezio-Sarthak Thanks for the follow-up - I just merged with changes in a couple of lines to simplify further :tada:

PR #950, opened 2021-03-16
tests: boxes: Add `SAVE_AS_DRAFT` keypress test for write box.

mkp6781:


tests/ui_tools/test_boxes.py
```diff
@@ -68,6 +68,51 @@ def test_not_calling_send_private_message_without_recipients(self, key,
...
+                'content': 'Random message',
+                'subject': 'Topic'
+            }
+        else:
+            write_box.to_write_box = mocker.Mock()
+            draft = {
+                'type': 'private',
+                'to': ['emailgateway@zulip.com'],
+                'content': 'Random message'
+            }
```
Ezio-Sarthak, line 104: Maybe a `parametrize` could be replaced with this conditional check?

tests/ui_tools/test_boxes.py
```diff
@@ -68,6 +68,51 @@ def test_not_calling_send_private_message_without_recipients(self, key,
 
         assert not write_box.model.send_private_message.called
 
+    @pytest.mark.parametrize('key', keys_for_command('SAVE_AS_DRAFT'))
+    @pytest.mark.parametrize('stream_message', [True, False])
+    @pytest.mark.parametrize('draft_message_in_current_session', [
+        True, None,
```
neiljp, line 74: You're using a descriptive parameter as a setting here, which is a little confusing.

neiljp (commented):
@mkp6781 I support @Ezio-Sarthak 's feedback, but other than my point below this looks close to ready :+1:

Note that the 'fixes' text will close the issue, so you'll need to rephrase it slightly or just remove it.

tests/ui_tools/test_boxes.py
mkp6781: Thanks for the feedback @Ezio-Sarthak . I have parametrized the `draft_message` now.

mkp6781:

tests/ui_tools/test_boxes.py
```diff
@@ -68,6 +68,53 @@ def test_not_calling_send_private_message_without_recipients(self, key,
...
+            'type': 'stream',
+            'to': 'Current Stream',
+            'content': 'Random message',
+            'subject': 'Topic'
+        },
+        {
+            'type': 'private',
+            'to': ['emailgateway@zulip.com'],
+            'content': 'Random message'
+        },
```
zee-bit, line 83: Maybe add one test fixture for PM huddle too?

tests/ui_tools/test_boxes.py
```diff
@@ -68,6 +68,53 @@ def test_not_calling_send_private_message_without_recipients(self, key,
...
+    def test_keypress_SAVE_DRAFT_MESSAGE(self, key, mocker, draft_composition,
+                                         draft_saved_in_current_session,
+                                         write_box, widget_size):
+        mocker.patch(BOXES + '.WriteBox.update_recipient_emails')
+        (write_box.model.session_draft_message.
+            return_value) = draft_saved_in_current_session
+        write_box.stream_write_box = mocker.Mock(edit_text="Current Stream")
+        write_box.title_write_box = mocker.Mock(edit_text="Topic")
+        write_box.msg_write_box = mocker.Mock(edit_text="Random message")
+        write_box.recipient_emails = ['emailgateway@zulip.com']
```
zee-bit, line 100: Looks like you aren't using the data from test fixture, and still providing a hard-coded value matching with the fixtures - that's why the tests aren't failing. This shouldn't happen; otherwise, there's no reason to use `parametrize`, right :)

Plus, the above three `write_box` objects are only for stream message and the last one for PM's, so probably better to have separate cases; but it's minor.

tests/ui_tools/test_boxes.py
```diff
@@ -68,6 +68,53 @@ def test_not_calling_send_private_message_without_recipients(self, key,
...
+
+        if draft_composition['type'] == "stream":
+            write_box.stream_id = 1
+        else:
+            write_box.to_write_box = mocker.Mock()
```
zee-bit, line 105: The second point in the above review could be shifted here.

zee-bit (changes requested):
Thanks for noticing the missing tests. This should be almost ready to merge. :+1:

You still haven't addressed Neil's concern about removing/rephrasing the fixes text, and the `parametrize` is not used properly as we'd expect. You can take a look at other similar tests to figure out how `parametrize` works, or you can consult the [`pytest`](https://docs.pytest.org/en/stable/parametrize.html) docs. :)

tests/ui_tools/test_boxes.py
mkp6781: > Looks like you aren't using the data from test fixture, and still providing a hard-coded value matching with the fixtures - that's why the tests aren't failing. This shouldn't happen; otherwise, there's no reason to use parametrize, right 

Yeah my bad. :)

tests/ui_tools/test_boxes.py
mkp6781: We have only two data structures `PrivateComposition` and `StreamComposition`. Testing already covers both these cases. For PM huddle, its just a case of appending more emails to the existing list. Is there a need for a separate parameter?

neiljp (commented):
.

mkp6781:
I have rebased this branch. Used Black to reformat the code so that it passes linting tests.

neiljp:
Marking this as a completion candidate, but the underlying code has changed somewhat, so in addition to a rebase and improving the tests, this will need some adjustment to fit the new code - this will require digging for eg. renames and similar.

criticic:
I'd like to work on this.

PR #951, opened 2021-03-16
ui: Add starred message count in UI.

klfoulk16:
Commit 1:
- starred_messages are added to the list of initial_data_to_fetch in model.py.

Commit 2:
- StarredButton now stores the count of starred_messages. 
- The displayed starred_message count is now updated during appropriate message
    flag events.
- Styles were addd to theme.py for the displayed starred_message count. To do this I:
    a. Added optional count_style arg for TopButton.
    b. Set this arg to 'unread_count' or 'starred_count' for the appropriate
    buttons.
    c. Pass count_style to TopButton.update_count function.


zee-bit:
@klfoulk16 Thanks for working on this. :+1: 
This seems to work pretty well functionally.(though I have not tested it extensively)

Seems like you haven't rebased on top of the recent changes in upstream, you may want to refer [the docs here](https://zulip.readthedocs.io/en/latest/git/using.html#keep-your-fork-up-to-date).
Apart from that, I tried running tests on your PR and it seems to fail in some cases. Did you ran the tests locally before pushing?
[ZT-development guide](https://github.com/zulip/zulip-terminal#development-tasks) can be quite helpful in figuring out how to run the tests locally.

klfoulk16:
Good points! My bad, this was my first code PR and clearly I forgot some things in my excitement. I'll do the rebase, check out the tests and also write some tests to make sure everything I added is working properly (so far I've been manually checking things - oopsies). Thanks for looking over this

zulipterminal/config/themes.py
```diff
@@ -160,6 +160,8 @@
          None,             DEF['light_blue:bold'],    DEF['black']),
         ('unread_count',   'yellow',                  'black',
          None,             DEF['yellow'],             DEF['black']),
+        ('starred_count',  'black',                  'yellow',
```
Ezio-Sarthak, line 163: We'd want to extend this to other themes as well :)

zulipterminal/ui_tools/views.py
```diff
@@ -762,10 +762,11 @@ def menu_view(self) -> Any:
             width=self.width,
             count=self.model.unread_counts['all_mentions'])
 
-        # Starred messages are by definition read already
+        # # Starred messages are by definition read already
```
Ezio-Sarthak, line 765: Unwanted diff :)

Ezio-Sarthak (changes requested):
Thanks for taking this up :+1: With some refining, this would be quite useful :)

I understand getting started with code/debugging, push/pull is quite messy 😛. It's always cleaner to remove any diffs and commented code that isn't necessary for the PR. This makes it more convenient to view actual changes in the PR :)

Also, I'd encourage always to split UI and event handling elements into their separate commits.

minor: It isn't required to mention PR number in the title itself :)

klfoulk16:
Screen Shot 2021-03-17 at 3 32 02 PM
Screen Shot 2021-03-17 at 3 33 09 PM
Screen Shot 2021-03-17 at 3 33 34 PM
Screen Shot 2021-03-17 at 3 33 54 PM
Images of the UI. Personally I feel the star count should mimic the text styling so as to not be confused with the "unread_counts." In zt_dark I made the count grey instead of white because the white looked to similar to yellow in my opinion

klfoulk16:
The tests currently aren't passing because I'm attempting to modify test_model.test_update_star_status to also test the recent changes; however, I'm running into lots of issues figuring out how this test and the function it's testing should be working. I've added some comments in the code to explain the issues I'm finding...I'd love if someone could give me a helping hand explaining why things work the way they work

klfoulk16:
Ready for review! *Finally* finished figuring out some tests for the updating the starred message count during a message flags event.

zulipterminal/ui_tools/buttons.py
```diff
@@ -61,7 +63,11 @@ def update_count(self, count: int, text_color: Optional[str]=None) -> None:
             count_text = ''
         else:
             count_text = str(count)
-        self.update_widget(('unread_count', count_text), new_color)
+
+        if self.starred is True:
+            self.update_widget(('starred_count', count_text), new_color)
+        else:
+            self.update_widget(('unread_count', count_text), new_color)
```
Rohitth007, line 70: I think it's better not to change `TopButton` to fit the needs of `StarredButton`. The parent has to be agnostic of its children. Instead you can create an _optional theme arg_ which defaults to 'unread_count' in `__init__` and `update_count` or create `modified_update_count` in child which calls `update_count`.

Rohitth007 (commented):
It's a nice addition to ZT. :+1: 
Left a few comments, didn't go through everything as I haven't properly explored the `message_flags` section of the code base.

zulipterminal/model.py
```diff
@@ -1298,6 +1312,8 @@ def _register_desired_events(self, *, fetch_data: bool=False) -> str:
         if response['result'] == 'success':
             if fetch_data:
                 self.initial_data.update(response)
+                with open("alternative_initial_data.txt", "w") as file:
+                    file.write(str(self.initial_data))
```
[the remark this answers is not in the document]
klfoulk16: oopsies!! thanks for catching that

zulipterminal/ui_tools/buttons.py
klfoulk16: interesting! I'll look into doing that. Mind if a reach out if I have questions?

Rohitth007: Sure, no problem :+1:

klfoulk16: Just updated the PR! Everything is now handled in the child StarredButton element

klfoulk16:
Updated and ready for another review!

zulipterminal/ui_tools/buttons.py
Rohitth007: It's nice that's its now handled by `StarredButton` but its just repetition of code. There are 2 ways you could do this (by creating a variable that stores text you wan't to display which can default to `unread count` when not specified):
* **Simpler Way**: Pass `'starred count'` to `__init__`, `__init__` to `update_count` , to `update_widget`
* or from the Starred button's `update_count` call `super().update_count` and pass `'starred count'` to it.

klfoulk16: I'm not sure I quite follow what you mean...I added the ability to pass the count_type to the Top Button (with a default of unread_count) which the update_count uses to pass the correct words to update_widget....but that feels similar to what I had in the first place?

Rohitth007: It looks fine now, initially you defined `starred` in the TopButton itself and said if True do this else do that. Now it's independent of the text. A future button may use this to write `'something else'` instead of `'starred count'`.
(I meant _optional text arg_ actually, I mistyped :))

Rohitth007: So naming it `count_theme` makes more sense than `count_type` ?

klfoulk16: Oh true! That makes a lot more sense, let me change the variable name real quick

klfoulk16:
Updated once again and all comments have been resolved 👍

zulipterminal/ui_tools/buttons.py
```diff
@@ -31,6 +31,7 @@ def __init__(
         prefix_character: Union[str, Tuple[Any, str]]='\N{BULLET}',
         text_color: Optional[str]=None,
         count: int=0,
+        count_theme: str='unread_count',
```
zee-bit, line 34: Would it be better to use `count_color`, to have a name that is consistent with `text_color`?
Also, you might want to consider having a `TypeAnnotation` similar to `text_color`.

zee-bit (commented):
@klfoulk16 Thanks for continuing your work on this. This definitely seems to be on the right path. :+1: 

I like the slightly grayed-out color for `star_count`, in the default theme. It is very different from `unread_counts` and is less distracting to look at, which is important, since the star count is not as critical to the user as unread count.
Apart from the general UI, this seems to mostly work fine, although I did encounter a `KeyError` while testing. I am not sure if this is particularly related to this PR, but you should be able to reproduce this yourself if you unstar a message from the webapp with ZT still open and focused in the `starred_message` narrow.
Oh, and as already pointed out by @Ezio-Sarthak, its better if you could split out the event handling and UI part into separate commits. Below is detailed review, hope it helps. :)

zulipterminal/ui_tools/buttons.py
klfoulk16: @Rohitth007 and I discussed naming it count_theme because 'starred_count'/'unread_count' are styles from the theme palette so what's being stored in the arg is the item's style name versus the actual 'color'. Does that make sense? I'd be happy to discuss further and change it if we decide count_color would be better. Good catch with the TypeAnnotation!

klfoulk16:
> @klfoulk16 Thanks for continuing your work on this. This definitely seems to be on the right path. 👍
> 
> I like the slightly grayed-out color for `star_count`, in the default theme. It is very different from `unread_counts` and is less distracting to look at, which is important, since the star count is not as critical to the user as unread count.
> Apart from the general UI, this seems to mostly work fine, although I did encounter a `KeyError` while testing. I am not sure if this is particularly related to this PR, but you should be able to reproduce this yourself if you unstar a message from the webapp while being in the `starred_message` narrow in ZT.
> Oh, and as already pointed out by @Ezio-Sarthak, it's better if you could split out the event handling and UI part into separate commits. Below is a more detailed review, hope it helps. :)

So for separating the event handling and UI part, would that just involve making the changes to the theme file and the addition of the 'count_theme' arg to the TopButton a different commit?

zulipterminal/ui_tools/buttons.py
Rohitth007: Actually even `text_color` is a style in the palette. Eg: The user button uses `user_idle`, `user_active` etc.
Actually now that you brought this up I feel both the names feel wrong. 
* `text_color` technically can be used not just to change the text color (foreground) but also the background along with styles like _bold_, _italics_, etc. This can be used when we select a different background color from the default.
* `count_theme` feels wrong as theme is more like a collection of styles?

`text_style` and `count_style` seem apt here. This has to be done in a different commit if going forward with it.
Thoughts?

zee-bit: > Now that you brought this up I feel both the names feel wrong.

Right. I was about to say this.
`*_style` here seems fine. `urwid`, however, likes to call those `display_attributes`, but that seems a bit too long IMO?

Re `TypeAnnotation`, Note that there's also a `None` display_attr in `themes`, which could be passed here, so its better to generalize the type as `Optional[str]`.

zee-bit:
> So for separating the event handling and UI part, would that just involve making the changes to the theme file and the addition of the 'count_theme' arg to the TopButton a different commit?

Yep, that should be all.
Generally, the `model.py` file handles all the events so it's better to split that change into a single commit at the beginning.
You can also add the changes in the `themes` file in an intermediate commit after the event handling.
Then work on the UI part in the later commit(s). :)

tests/ui_tools/test_buttons.py
```diff
@@ -92,6 +94,32 @@ def test_text_content(self, mocker,
         assert len(text[0]) == len(expected_text) == (width - 1)
         assert text[0] == expected_text
 
+    def test_update_count(self, mocker):
+        """Make sure update_count is called with correct parameters"""
+        show_function = mocker.Mock()
+        top_button = TopButton(controller=mocker.Mock(),
```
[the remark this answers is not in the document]
klfoulk16: The goal of the test I added on lines 112-121 was to make sure that when the update_count function was called on the starred_button that it used the correct call args...I guess it would be more logical to test the StarredButton.__init__ function itself since that's where the actual change to the count_theme is taking place?

zulipterminal/config/themes.py
```diff
@@ -255,6 +257,8 @@
          None,             LIGHTBLUE,         BLACK),
         ('unread_count',   'yellow',          'black',
          None,             YELLOW,            BLACK),
+        ('starred_count',   'light gray',      'black',
+         None,             GRAY,              BLACK),
```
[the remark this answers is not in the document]
klfoulk16: What do you mean by "above change"?

zulipterminal/ui_tools/buttons.py
klfoulk16: So shall I update the name of count_theme in the original commit? Seems logical to name it correct from the start. I can create a third commit to fix the `text_color` name. I agree with the name change I think it makes a lot more sense

klfoulk16:
@zee-bit I can't seem to reproduce the `KeyError` you speak of? I've tried unstarring and starring multiple messages from within the webapp while keeping ZT open to the `starred_messages` narrow and everything seems to work as expected?

zulipterminal/ui_tools/buttons.py
klfoulk16: If I change `count_style` to `Optional[str]=None` it causes an error with the `TopButton.update_widget` function...Technically the UserButton doesn't have a `count_style` but nor does it use the `update_widget` function...Not sure how to handle this. Would you change the types that `update_widget` accepts? Or would you change the `TypeAnnotation` for the `count_style` arg?

This was the error message:
`zulipterminal/ui_tools/buttons.py:67: error: Argument 1 to "update_widget" of "TopButton" has incompatible type "Tuple[Optional[str], str]"; expected "Tuple[str, str]"`

tests/model/test_model.py
```diff
@@ -1575,6 +1578,70 @@ def test_update_star_status(self, mocker, model, event_op,
...
+    @pytest.mark.parametrize('event_message_ids, indexed_ids', [
+        ([1], [1]),
+        ([1, 2], [1]),
+        ([1, 2], [1, 2]),
+        ([1], [1, 2]),
+    ])
+    def test_update_star_count_in_view_remove(self, mocker, model,
+                                              event_message_ids,
+                                              indexed_ids,
+                                              update_message_flags_operation):
```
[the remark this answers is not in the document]
klfoulk16: I've created an alternative test and added it to the most recent version of the pull request, do you mind looking at it and seeing if it looks like a better version? This is the new test: https://github.com/zulip/zulip-terminal/pull/951/commits/7051886ad2865ba0b285093a5628075777da4b8e#diff-6b425a54c6e29c8e98d58bc743f98ffcc82a83e8ceb39e063cf31e3381e439e5R1581-R1626

zulipterminal/ui_tools/buttons.py
Rohitth007: Fixing the Type of `update widget` would do as previously `unread_count` was hard coded.
Re UserButton, it too displays count in yellow ryt? 
Also I just saw that UserButton uses update_widget in a weird way, it uses `(you)` as count text. Maybe it can be changed in another PR.

klfoulk16: I can't seem to find where UserButton uses update_widget, do you mind sharing the file/line number where you found that?

klfoulk16: I fixed the `TopButton.update_widget`'s `TypeAnnotation` to be `Tuple[Optional[str], str]`. I also declared `count_style='unread_count'` in all of the non-starred TopButton children's' `super.__init__` functions....except for the user button because I couldn't find where it was used. Shall I go in and add `count_style='unread_count',` to the `UserButton`'s init function as well?

klfoulk16:
Changes in the previous commit:
1. Condensed the relatively similar functions for testing `add`, `remove star_count` into one function.
2. Removed unnecessary `StarredButton.update_widget` test and changed it to tests checking the StarredButton's init functions to check that the new count_style arg is correctly assigned.
3. Renamed TopButton's text_color arg to text_style per our discussions above

Thanks @zee-bit for the clarification about the bug you found; however, I still am unable to reproduce it. After the final part where I unstar the messages from the web app and then scroll through those messages in ZT (I assume I am scrolling through them in the `starred_messages` narrow?) I do not get a `TypeError`. Instead it appears as if I had unstarred the messages from ZT, with the messages still displayed in the `starred_messages` narrow, but they no longer have a star nor are they included in the displayed Starred Messages count. Scrolling through them causes no issues.

neiljp (commented):
@klfoulk16 Just done a first pass over this having tested it - there are a few things to fix/improve, but this is looking good :+1: I've mainly focused on functionality and commit structure this time, as smaller commits are that much easier to review. If there are additional parts I've not indicated that are independent separate improvements - maybe to tests? - then feel free to break into a separate commit too if that.

zulipterminal/model.py
```diff
@@ -1237,14 +1237,20 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
             if operation == 'add':
                 if flag_to_change not in msg['flags']:
                     msg['flags'].append(flag_to_change)
-                if flag_to_change == 'starred':
+                    if flag_to_change == 'starred':
+                        self.controller.view.starred_button.update_count(
+                            self.controller.view.starred_button.count + 1)
```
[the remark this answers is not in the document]
klfoulk16: Resolved and updated tests!

zulipterminal/ui_tools/buttons.py
```diff
@@ -31,6 +31,7 @@ def __init__(
         prefix_character: Union[str, Tuple[Any, str]]='\N{BULLET}',
         text_color: Optional[str]=None,
         count: int=0,
+        count_style: Optional[str]=None,
```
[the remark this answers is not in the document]
neiljp: This doesn't seem to have been done yet? Did you understand the parts I was referring to?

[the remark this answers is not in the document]
klfoulk16: Sorry about that, I pulled apart the commit in a new branch (in case I royally screwed things up) and it seems I accidentally pushed the old branch instead of the new one. The correct version is now up.

I felt it made the most sense to break it into three separate commits:
1. Refactor StarredButton to actually display the count (it uses the existing unread count styling).
2. Add the optional count_style arg to TopButton, refactor the update_count function and add define the style for the various buttons.
3. Add a new style - called starred_count - to themes and change the StarredButton's count_style to it. 

I'd be happy to reorganize or squash two of those together if you think I should. Sorry again about not having the correct version up

neiljp (commented):
@klfoulk16 Thanks for the reworking - the split of commits makes this much cleaner to review :+1: As before this seems to work fine, with the possible exception of monochrome mode (1-color bit depth), having now read the theme changes in their own commit.

I didn't focus as much on tests last time, but with smaller commits it's easier to read and so that's the bulk of my (mostly minor) suggestions this time around.

Some of the edits do change multiple files which aren't reflected in the commit areas (eg `model: title` vs `model/buttons/views: title`?)

We could keep the color/style naming refactor (you can add that as a separate commit prefix, if you see in git log) here or in another PR if you like, since there seem to be a cascade of other names which might be affected?

In any case I'd like to get the core of this merged as it'll be useful! :+1:
(a good follow-up would be to use the new server feature where this is enabled/disabled via a setting, but we can save that for later in an issue if you'd prefer)

tests/ui_tools/test_buttons.py
```diff
@@ -92,6 +93,29 @@ def test_text_content(self, mocker,
...
+                               caption='caption',
+                               show_function=show_function,
+                               width=12,
+                               count=10,
+                               count_style='unread_count')
+        top_button.update_widget = mocker.patch(TOPBUTTON + ".update_widget")
+        top_button.update_count(11)
+        assert (((top_button.count_style, '11'), None)
+                == top_button.update_widget.call_args[0])
```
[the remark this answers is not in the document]
klfoulk16: I tried using that previously but for some reason it wasn't passing despite everything being correct. I will look into it again

zulipterminal/config/themes.py
```diff
@@ -180,6 +180,8 @@
          None,             DEF['light_blue:bold'],    DEF['black']),
         ('unread_count',   'yellow',                  'black',
          None,             DEF['yellow'],             DEF['black']),
+        ('starred_count',   'light gray',             'black',
```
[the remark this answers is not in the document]
klfoulk16: How do you run in monochrome theme? I ran with`zulip-term --color-depth 1` (which I believe is monochrome?) and it seems to be different...The starred_count is not bolded while the rest are:
Screen Shot 2021-05-05 at 10 15 01 PM

zulipterminal/ui_tools/buttons.py
```diff
@@ -258,7 +258,7 @@ def __init__(self, user: Dict[str, Any], controller: Any,
             show_function=self._narrow_with_compose,
             width=width,
             prefix_character=(color, state_marker),
-            text_color=color,
+            text_style=color,
```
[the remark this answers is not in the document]
klfoulk16: I added some comments in the topic about this...I went down a bit of a rabbit hole wondering whether we should actually change `text_color` to `text_style`

zulipterminal/ui_tools/buttons.py
```diff
@@ -59,9 +61,10 @@ def update_count(self, count: int, text_color: Optional[str]=None) -> None:
             count_text = ''
         else:
             count_text = str(count)
-        self.update_widget(('unread_count', count_text), new_color)
 
-    def update_widget(self, count_text: Tuple[str, str],
+        self.update_widget((self.count_style, count_text), new_color)
+
+    def update_widget(self, count_text: Tuple[Optional[str], str],
```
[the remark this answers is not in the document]
klfoulk16: No I believe this is the right commit. This commit adds `count_style: Optional[str]=None,` to the `TopButton.__init__` function in line 22. Prior to this commit update_widget was called with `unread_count` therefore the type was `Tuple[str, str]` but due to the fact that in this commit `count_style` is now passed in instead of `unread_count` the type had to be changed

tests/ui_tools/test_buttons.py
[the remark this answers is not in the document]
klfoulk16: I agree that version would be cleaner, but for some reason I keep getting the following error when I use that:
`>       assert top_button.update_widget.assert_called_once_with(('unread_count', '11'), None)
E       AssertionError: assert None
E        +  where None = <bound method wrap_assert_called_once_with of <MagicMock name='update_widget' id='4393243264'>>(('unread_count', '11'), None)
E        +    where <bound method wrap_assert_called_once_with of <MagicMock name='update_widget' id='4393243264'>> = <MagicMock name='update_widget' id='4393243264'>.assert_called_once_with
E        +      where <MagicMock name='update_widget' id='4393243264'> = <TopButton selectable flow widget ''>.update_widget

tests/ui_tools/test_buttons.py:107: AssertionError`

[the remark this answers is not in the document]
klfoulk16: Yet if I get rid of 'None' then the test fails and send this:
```>       assert top_button.update_widget.assert_called_once_with(top_button.count_style, '11')
E       AssertionError: expected call not found.
E       Expected: update_widget('unread_count', '11')
E       Actual: update_widget(('unread_count', '11'), None)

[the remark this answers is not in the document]
Comment (klfoulk16) [reply tests/ui_tools/test_buttons.py:108]: I found that `top_button.update_widget.call_args[0])` was a workaround that had the expected behavior. Any ideas why the `update_widget.assert_called_once_with` was acting funky?

File: tests/ui_tools/test_buttons.py [hunk 15]
```diff
@@ -92,6 +93,29 @@ def test_text_content(self, mocker,
...
 
+    def test_update_count(self, mocker):
+        """Make sure update_count is called with correct parameters"""
+        show_function = mocker.Mock()
+        top_button = TopButton(controller=mocker.Mock(),
+                               caption='caption',
+                               show_function=show_function,
+                               width=12,
+                               count=10,
+                               count_style='unread_count')
```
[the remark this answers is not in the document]
klfoulk16: Good suggestions, thank you!

zulipterminal/config/themes.py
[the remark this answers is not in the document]
klfoulk16: I added `'starred_count': '',` to make sure it remains unbolded while `unread_count` is bolded

neiljp:
@klfoulk16 Thanks! As mentioned in the stream I pushed some minor changes, left CI to run and have just merged :tada:!

You may like to review the changes to follow what may not have worked for you earlier or what you could explore in future - eg. I adjusted some tests to push constants into test function parameters, got that alternative assert style working and made a few minor renames.

PR #952, opened 2021-03-17
UI: Emphasize column titles to differentiate it from regular texts.

zee-bit:
This is a minor UI enhancement, that renders the column-titles in `bold` to remain consistent with the `LineBox` borders while differentiating them from the rest of the text.

**BEFORE:**
*monochrome*
monochrome-before
*default*
default-before
*gruvbox*
gruvbox-before
*zt_light*
light-before
*zt_blue*
blue-before

**AFTER:**
*monochrome*
monochrome-after
*default*
default-after
*gruvbox*
gruvbox-after
*zt_light*
light-after
*zt_blue*
blue-after


neiljp:
@zee-bit Thanks for this UI adjustment - it's quite small but does make a difference :+1: I've just merged it with only small adjustments to the commit text for clarity :tada: (generally we've ended up referring to each entry in a theme as a style)

PR #955, opened 2021-03-20
FAQ: Mention Windows Terminal text rendering issues.

prah23:
This PR adds a commit updating the FAQ with the observed Windows Terminal issue regarding rendering bold and normal text the same way.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -20,12 +20,12 @@ We have reports of success on the following terminal emulators:
 * Termite (https://github.com/thestinger/termite)
 * Terminator (https://github.com/gnome-terminator/terminator)
 * iterm2 (https://iterm2.com/) **Mac only**
-* Microsoft/Windows Terminal (https://github.com/Microsoft/Terminal) **Windows only**
 * mosh (https://mosh.org/)
 
 Issues have been reported with the following:
 
-* terminal app **Mac only** - **Issues with some default keypresses, including for sending messages** [zulip-terminal#680](https://github.com/zulip/zulip-terminal/issues/680)
+* (**major**) terminal app **Mac only** - **Issues with some default keypresses, including for sending messages** [zulip-terminal#680](https://github.com/zulip/zulip-terminal/issues/680)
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -23,6 +23,9 @@ We have reports of success on the following terminal emulators:
 * Microsoft/Windows Terminal (https://github.com/Microsoft/Terminal) **Windows only**
 * mosh (https://mosh.org/)
 
+Among the above, minor non-breaking issues have been noticed with the following:
+* Microsoft/Windows Terminal **Windows Only** - Bold text isn't actually bold, it either renders the same as normal text or renders in a different colour (https://github.com/microsoft/terminal/issues/109)
```
neiljp, line 27: Does this deserve its own section separate from the issues section below it?

prah23: The reason I didn't fit it in the issues section below is because that section only has emulators that we consider unusable and not among the successful emulators. Hence putting Windows Terminal in the successful emulators section as well as the issues section seemed inappropriate. This issue also seemed too minor to take it off of the successful emulators list.

prah23:

prah23:
@neiljp I think we'd benefit from a minor issues section apart from the issues section we already have, because the successful emulators section and the current issues section seem to be disjoint sets. We're bound to notice more minor issues on other emulators which aren't necessarily serious enough that we consider them unsuccessful. What are your thoughts on this?

prah23:
I've updated it to mention issues as a sub-point. Would this be a better pattern to follow?

neiljp:
The sub-point seems to distract from the list flow?

Right now two ideas come to mind:
* place all "perfect" emulators in the top list, clarify that the bottom list have issues but in each one state major/minor 
* place all emulators in a "tested" list, with listed caveats against some below (ie. an "issues" list underneath)

I'm inclined towards the first of these since it keeps it cleaner and requires only an update to the text of the lower list.

ZT is not unusable on Mac Terminal, it can be worked around - we could note how - so I'm not sure this is so different.

prah23:
With the Mac Terminal, I meant the issue was more severe comparatively because it needed the user to change his keybindings, or change `keys.py`. Your suggestion was clearer than the previous iteration nonetheless, and I've updated it.

neiljp:
@prah23 Thanks for the documentation update - merged with minor update as fbe6973646fec6125fe026355a1e00760f612377 for consistency :tada:

PR #960, opened 2021-03-26
boxes: Prefill stream name when user presses `c` in stream/topic narrow.

mkp6781:
An earlier commit 2b8a3050 brought about consistency in using `x` and `c`.
This commit aims to build on top of that by prefilling the steam name in
`stream_box_view` when the user presses `c` from stream/topic narrow.

Minor test amendment.


neiljp:
@mkp6781 Thanks for this adjustment to the previous behavior from the other PR - I'm merging this now :tada:

PR #964, opened 2021-03-27
core/run/views: Add `--notify` and `--no-notify` flags as a command-line argument for ZT

zee-bit:
This PR adds new command-line configurations for ZT - the `--notify` and `--no-notify` flags. These should specify whether the notifications will be enabled in the current session or not respectively. If these flags are used as a command-line argument then the corresponding value in the `zuliprc` file is overridden by its value for the current session.

The attribute of this config is also shown in the terminal while ZT is loading, and in the About menu when it has loaded.

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -227,10 +227,13 @@ class Controller:
 
     def show_about(self) -> None:
         self.show_pop_up(
-            AboutView(self, 'About', zt_version=ZT_VERSION,
+            AboutView(self, 'About',
+                      zt_version=ZT_VERSION,
                       server_version=self.model.server_version,
                       server_feature_level=self.model.server_feature_level,
-                      theme_name=self.theme_name, color_depth=self.color_depth,
+                      theme_name=self.theme_name,
+                      color_depth=self.color_depth,
+                      notify_enabled=self.notify_enabled,
                       autohide_enabled=self.autohide,
                       maximum_footlinks=self.maximum_footlinks),
             'area:help'
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -1017,7 +1017,8 @@ class AboutView(PopUpView):
                  server_version: str,
                  server_feature_level: Optional[int],
                  theme_name: str, color_depth: int,
-                 autohide_enabled: bool, maximum_footlinks: int) -> None:
+                 autohide_enabled: bool, maximum_footlinks: int,
+                 notify_enabled: bool) -> None:
         self.feature_level_content = (
             [('Feature level', str(server_feature_level))]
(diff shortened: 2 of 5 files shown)


neiljp (commented):
@zee-bit From the brief discussion in the stream I'm open to this option on the command-line :+1:

However, I'm wondering if we want to have something along the lines of `--notify` and `no-notify` like we do for `autohide`? I know that currently doesn't parallel the valid values, but requiring `--key=value` when there are only two values makes it more verbose and possibly extra understanding of how there are two parts (key + value), rather than just an option.

Re autohide, I think it would be reasonable to add additional values there so we also can have `autohide=enabled|disabled` in the config?

The commits are quite small, but I appreciate the clarity of keeping each change isolated :+1:

zee-bit:
@neiljp PR has been updated.
Re autohide, I wasn't sure if you wanted me to implement those changes in this PR itself, or if you wanted to have a similar setting for `notify` as well?

zulipterminal/core.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, config_file: str, maximum_footlinks: int,
         self.color_depth = color_depth
         self.in_explore_mode = in_explore_mode
         self.autohide = autohide
-        self.notify_enabled = notify
+        self.notify = notify
```
neiljp, line 52: Let's leave the controller attribute descriptive too.

zulipterminal/ui_tools/views.py
```diff
@@ -1030,7 +1031,9 @@ def __init__(self, controller: Any, title: str, *, zt_version: str,
                 ('Theme', theme_name),
                 ('Autohide', 'enabled' if autohide_enabled else 'disabled'),
                 ('Maximum footlinks', str(maximum_footlinks)),
-                ('Color depth', str(color_depth))])
+                ('Color depth', str(color_depth)),
+                ('Notifications',
+                 'enabled' if notify_enabled else 'disabled')])
```
neiljp, line 1036: You added some good trailing commas; it'd be good if we could have one here too.

zulipterminal/core.py
neiljp: (I noticed this since this commit now modifies more files than indicated in the commit)

zulipterminal/ui_tools/views.py
zee-bit: I have added the trailing comma there as requested, but note that it won't make the diff smaller on adding extra fields(?), since linter wanted me to add extra whitespace after the comma. :)

neiljp:
@zee-bit Thanks for the update - I merged manually with a few minor textual changes to the commit text, and I just bumped the brackets onto the following line to solve the whitespace issue you mentioned :tada:

PR #967, opened 2021-03-29
bugfix: core: Fix cases where message was not focused on narrowing with anchor.

### LINKED ISSUE #954: Narrowing with a contextual message id should select that message
If using a link such as generated in a quoted message, then the UI doesn't select that message - which seems like the behavior to be expected.

To reproduce:
* quote a message and send it
* move cursor onto message with quote
* open message info popup, move cursor onto link to original message & press enter

I'd expect the original message to be selected, but we don't do this.

Current behavior is:
* if narrow is loaded/open, do nothing?
* otherwise, load messages around message id; unclear which message is selected?

zee-bit:
This bugfix fixes selecting messages if a contextual msg-id (anchor) is provided to the `_narrow_to()` function for some edge cases. Earlier, if the user was within the same narrow then narrowing to a quoted message from `MsgLinkButton` did not select that message. Also, sometimes pressing `a` on a particular message did not focus on it when narrowed to the `all_messages` narrow.

Tests added.
Fixes #954.

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -126,13 +126,31 @@ class TestController:
         id_list = index_stream["stream_msg_ids_by_stream_id"][stream_id]
         assert {widget.original_widget.message["id"]} == id_list
 
-    def test_narrow_to_topic(self, mocker, controller, msg_box, index_topic):
+    @pytest.mark.parametrize(
+        ["initial_narrow", "anchor", "expected_final_focus"],
+        [
+            ([], None, 537289),
+        ],
+        ids=[
+            "all-messages_to_topic_narrow_no_anchor",
+        ],
+    )
+    def test_narrow_to_topic(
+        self,
+        mocker,
+        controller,
+        msg_box,
+        index_multiple_topic_msg,
+        initial_narrow,
+        anchor,
+        expected_final_focus,
+    ):
         expected_narrow = [
             ["stream", msg_box.stream_name],
             ["topic", msg_box.topic_name],
         ]
-        controller.model.narrow = []
-        controller.model.index = index_topic
+        controller.model.narrow = initial_narrow
+        controller.model.index = index_multiple_topic_msg
         controller.view.message_view = mocker.patch("urwid.ListBox")
         controller.model.stream_dict = {
             205: {
@@ -143,16 +161,23 @@ class TestController:
         controller.model.muted_streams = []
         controller.model.is_muted_topic = mocker.Mock(return_value=False)
 
-        controller.narrow_to_topic(stream_name="PTEST", topic_name=msg_box.topic_name)
+        controller.narrow_to_topic(
+            stream_name="PTEST",
+            topic_name=msg_box.topic_name,
+            contextual_message_id=anchor,
+        )
 
         assert controller.model.stream_id == msg_box.stream_id
         assert controller.model.narrow == expected_narrow
         controller.view.message_view.log.clear.assert_called_once_with()
 
-        widget = controller.view.message_view.log.extend.call_args_list[0][0][0][0]
+        widgets, focus = controller.view.message_view.log.extend.call_args_list[0][0]
         stream_id, topic_name = msg_box.stream_id, msg_box.topic_name
-        id_list = index_topic["topic_msg_ids"][stream_id][topic_name]
-        assert {widget.original_widget.message["id"]} == id_list
+        id_list = index_multiple_topic_msg["topic_msg_ids"][stream_id][topic_name]
+        msg_ids = {widget.original_widget.message["id"] for widget in widgets}
+        final_focus_msg_id = widgets[focus].original_widget.message["id"]
+        assert msg_ids == id_list
+        assert final_focus_msg_id == expected_final_focus
 
     def test_narrow_to_user(self, mocker, controller, user_button, index_user):
         controller.model.narrow = []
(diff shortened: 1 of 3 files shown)


neiljp (commented):
@zee-bit Thanks for fixing the issue - all our stream narrowing from links should work after this - though I'm looking forward to seeing tests :)

We don't need to refer to the issue in the comment, and I'd only include a comment if you can't express the logic clearly. Given `None` has particular meaning here, I'd suggest using it explicitly.

zee-bit:

neiljp (commented):
@zee-bit Thanks for working on the tests :+1:

As per the comment, let's keep the bugfix commit itself as small as possible for clarity. You've extended the test and also altered some other test lines in the same commit, without touching on why. Feel free to add a refactor commit before or after with a rationale for that and we can explore those or improve these tests - these tests are by no means fixed or optimal I'm sure.

Regarding extending the test state space, we definitely could benefit from this in the longer term. There is likely a case where we don't jump to the message remaining due to having pre-existing messages but not the one in question - but this is a more general bug! (can you see it?)

tests/core/test_core.py
```diff
@@ -116,35 +116,65 @@ def test_narrow_to_stream(self, mocker, controller,
         id_list = index_stream['stream_msg_ids_by_stream_id'][stream_id]
         assert {widget.original_widget.message['id']} == id_list
 
-    def test_narrow_to_topic(self, mocker, controller,
-                             msg_box, index_topic):
-        expected_narrow = [['stream', msg_box.stream_name],
-                           ['topic', msg_box.topic_name]]
-        controller.model.narrow = []
+    @pytest.mark.parametrize('initial_narrow, final_narrow, anchor', [
+        ([], [['stream', 'BOO'], ['topic', 'RDS']], None),
```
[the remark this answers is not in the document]
zee-bit: To clarify, my intention behind keeping the final narrow same was that, the `index_topic` fixture from `conftest` contains just one topic and one message, so having multiple final narrows would need some serious refactoring.

No, we don't gain anything from exploring initial_narrow state space, it was just to assure that narrowing works for all possible combinations of initial narrow.

I'll have to explore more on how I can improve the tests with minimal change. Also, breaking the tests seems very unlikely since the index contains just one message and it would focus on that even if we provide an anchor or don't.

zee-bit:

zee-bit:
There's a prep PR for this that might simplify the multiple-message index generation in `conftest` here : #1022.

tests/core/test_core.py
```diff
@@ -126,13 +126,35 @@ def test_narrow_to_stream(
         id_list = index_stream["stream_msg_ids_by_stream_id"][stream_id]
         assert {widget.original_widget.message["id"]} == id_list
 
-    def test_narrow_to_topic(self, mocker, controller, msg_box, index_topic):
+    @pytest.mark.parametrize(
+        ["initial_narrow", "anchor"],
+        [
+            ([], None),
+            ([["stream", "PTEST"], ["topic", "Test"]], 537286),
```
neiljp, line 133: It's good to know that you've tested that this fails without the code change, but let's now move this line into the fixing commit, possibly with any extra test adjustment necessary to handle this (stream setting below?).

This test then becomes enhanced to handle positioning with multiple messages, and the next commit will extend it with code to pass the extra test case.

zulipterminal/core.py
```diff
@@ -392,8 +392,10 @@ def _narrow_to(self, anchor: Optional[int], **narrow: Any) -> None:
 
         msg_id_list = self.model.get_message_ids_in_current_narrow()
 
-        # if no messages are found get more messages
-        if len(msg_id_list) == 0:
+        # If no messages are found in the current narrow
+        # OR, given anchor is not present in msg_id_list
+        # then, get more messages.
+        if len(msg_id_list) == 0 or (anchor is not None and anchor not in msg_id_list):
```
neiljp, line 398: Worth linking to the discussion in the commit text, perhaps.

neiljp (commented):
@zee-bit This looks close to merging and will fix a pretty important bug :+1:

This may also fix #934 ?

tests/core/test_core.py
```diff
@@ -126,13 +126,35 @@ def test_narrow_to_stream(
...
             ["stream", msg_box.stream_name],
             ["topic", msg_box.topic_name],
         ]
-        controller.model.narrow = []
-        controller.model.index = index_topic
+        expected_final_focus_msg_id = 537289 if anchor is None else anchor
+        controller.model.narrow = initial_narrow
+        controller.model.index = index_multiple_topic_msg
+        if len(initial_narrow) == 2 and initial_narrow[0][0] == "stream":
+            controller.model.stream_id = 205
```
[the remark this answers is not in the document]
zee-bit: We can perhaps have a `initial_stream_id` in parameterize?
That would be `None` when narrow is `[]`?

zulipterminal/core.py
zee-bit: To clarify, do you mean adding a link (or commit-hash?) to this commit?

zee-bit: I've added the commit hash in the comments. I am not sure if I can add a link the commit?

neiljp:
@zee-bit I adjusted some commit text, added test cases for the the anchor being another value in the narrow, and moved the expected value into the parametrize block in the last commit (like in the earlier one) - merging now :tada:

PR #969, opened 2021-03-29
refactor: Rearrange registered events dictionaries.

Rohitth007:
This refactor moves the dictionary `fetch_types` to `__init__`.
This dictionary which was previously in `_register_desired_events`
made it look like those were the events being registered to
while actually they were initial fetch events.

The function `_update_initial_data()` has been given a more
apt name `_fetch_initial_data()`.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -88,6 +88,23 @@ class Model:
 
         self._notified_user_of_notification_failure = False
 
+        # Events fetched once at startup
+        self.initial_desired_events: List[str] = [
+            'realm',
+            'presence',
+            'subscription',
+            'message',
+            'update_message_flags',
+            'muted_topics',
+            'realm_user',  # Enables cross_realm_bots
+            'realm_user_groups',
+            'update_display_settings',
+            # zulip_version and zulip_feature_level are always returned in
+            # POST /register from Feature level 3.
+            'zulip_version',
+        ]
+
+        # Events desired while in runtime with their corresponding callback.
         self.event_actions: 'OrderedDict[str, Callable[[Event], None]]' = (
             OrderedDict([
                 ('message', self._handle_message_event),
@@ -106,7 +123,7 @@ class Model:
 
         # Register to the queue before initializing further so that we don't
         # lose any updates while messages are being fetched.
-        self._update_initial_data()
+        self._fetch_initial_data()
 
         self.server_version = self.initial_data['zulip_version']
         self.server_feature_level = (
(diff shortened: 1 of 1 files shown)


Rohitth007:
I tried to but couldn't satisfy PEP8's line length condition :(
83 > 79

zulipterminal/model.py
```diff
@@ -88,25 +88,42 @@ def __init__(self, controller: Any) -> None:
...
+            'realm_user',  # Enables cross_realm_bots
+            'realm_user_groups',
+            'update_display_settings',
+            # zulip_version and zulip_feature_level are always returned in
+            # POST /register from Feature level 3.
+            'zulip_version',
+        ]
+
+        # Events desired while in runtime.
+        self.desired_events_action: 'OrderedDict[str, Callable[[Event], None]]' = (
```
neiljp, line 108: I don't think we need to modify this variable at all - and that will keep the main movement of the list into `__init__` a lot clearer.

neiljp (commented):
@Rohitth007 Thanks for the refactoring contribution! :+1: I'm glad you figured out the difference between the sets of strings and this should help others too :)

If you plan to contribute further I'd suggest checking the ZT code style guidelines or reviewing recent git log entries for examples.

zulipterminal/model.py
Rohitth007: Oh ok, I'll reset this back to the original name.

zulipterminal/model.py
```diff
@@ -96,17 +113,17 @@ def __init__(self, controller: Any) -> None:
                 ('subscription', self._handle_subscription_event),
                 ('typing', self._handle_typing_event),
                 ('update_message_flags',
-                 self._handle_update_message_flags_event),
+                    self._handle_update_message_flags_event),
                 ('update_display_settings',
-                 self._handle_update_display_settings_event),
+                    self._handle_update_display_settings_event),
```
zee-bit, line 118: I guess these diffs aren't needed. :)

zee-bit (commented):
This is a nice refactoring. :+1:
There are few points to mention, though;
1. You can squash the second commit, or drop it and amend the first commit since it just reverts the changes of your first commit. Also, we probably don't want to have an intermediate phase where the tests were broken.
2. As Neil already mentioned, we follow a specific commit style that you can have a look at in your `git log`, and it'll be convenient if you can adapt your commit to follow those guidelines. :)

zulipterminal/model.py
Rohitth007: Don't remember messing with that, I'll remove them

Rohitth007:
Sorry, I didn't get the 2nd point, was my code style wrong or my commit style? @zee-bit @neiljp

Rohitth007:
I have reverted unnecessary changes and squashed commits as needed. This is ready to be merged, please take a look.Thank you for your suggestions. @zee-bit @neiljp

neiljp:
@Rohitth007 Thanks again for the refactor! I made a few minor textual adjustments and merged it manually as f9d09825896e54d49a1b89c10d6d098c9182ea0b :tada:

PR #971, opened 2021-03-30
Add basic fields in footer notification styling

### LINKED ISSUE #782: Support a variety of footer notification styling

Ezio-Sarthak:
**What does this PR do?!**
 - Adds general styling fields: `task:success`, `task:error`, `task:warning`.
 - Adds contrast-color style for footer hint text.
 - refactor: Add default report helpers for success/error/warning footer notifications.
 - Replace the instances of `set_footer_text()` with added helpers wherever reasonable.

I'd highly appreciate any feedback :)

Fixes #782.


Ezio-Sarthak:

Rohitth007:
@Ezio-Sarthak Really nice addition to ZT.
Why `area:neutral` though? `footer` should make sense as there is also a `header`, for the default color.
Also in terms of readability, I feel someone looking at `themes.py`for the very first time would instantly make sense of what a `header` is and what a `footer` is than `area:neutral`.
Just a suggestion :)

Ezio-Sarthak:
@Rohitth007 thanks for the feedback. Re `area: neutral`, it's true that it wouldn't be readable for footer styling sense, though my intent was that this could be used as a general style elsewhere in the UI in future, if that makes sense?

Rohitth007:
I'm assuming the footer would have an independent styling as acts sort of like a notification bar?
Also, I think applying same setting to different components reduces independent customizability.

zulipterminal/config/themes.py
```diff
@@ -49,6 +48,9 @@
     'area:msg': 'standout',
     'area:stream': 'standout',
     'area:error': 'standout',
+    'area:neutral': 'standout',
+    'task:success': 'standout',
+    'task:failure': 'standout',
```
zee-bit, line 53: I am not sure if we currently have any use-case, but could we also support a `task:warning` perhaps? -for something that's not as severe as a failure.

zulipterminal/config/themes.py
```diff
@@ -189,6 +189,12 @@
          None,             DEF['white'],              DEF['dark_cyan']),
         ('area:error',     'white',                   'dark red',
          None,             DEF['white'],               DEF['dark_red']),
+        ('area:neutral',   'black',                   'light gray',
+         None,             DEF['black'],              DEF['light_gray']),
```
zee-bit, line 193: We should change this to `task`, to be more consistent with other footer styles. We could always have another one for `area:neutral` if the need be.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1516,13 +1520,13 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
                             " Time Limit for editing the message has"
-                            " been exceeded.", 3)
+                            " been exceeded.", 'task:failure', 3)
                     return key
                 elif self.message['type'] == 'stream':
                     self.model.controller.view.set_footer_text(
                             " Only topic editing allowed."
                             " Time Limit for editing the message body has"
-                            " been exceeded.", 3)
+                            " been exceeded.", 'task:failure', 3)
```
zee-bit, line 1529: This seems like a good candidate for `task:warning`?
As in, the user is still allowed to edit, but not the message body.

zee-bit (commented):
This is really nice work. The footer looks much clean and dynamic now. :+1:

I haven't tested it on all themes yet, but the darker ones are looking good.
I am also inclined to Sai's point above that we should have `footer` instead of `area`, but the `task` sounds like a fine replacement too. I would love to gather some more opinions on this, but the PR, in general, should be very close to being merge-able. I have dropped in some points below, but they are mostly minor and may need discussion. :)

zulipterminal/config/themes.py
Ezio-Sarthak: We could, though I'm unsure of what styles to expect for that. Could you suggest some color designs (which could potentially be rendered across all themes)?

zulipterminal/config/themes.py
Ezio-Sarthak: Thanks for pointing that out. The original intent for it was that it might be used (in general) for other areas (in future) where there's no highlighting required, e.g., typing status. However, since I couldn't think of more such examples, currently we could just align it with other footer-dedicated attributes. We could simply add such themes later in any case :+1:

zulipterminal/config/themes.py
zee-bit: how about a `dark_yellow` or `light_orange`(not sure if they feasible, though). We could probably test other variants too.

Ezio-Sarthak: Well, we would tend to support all themes, so this might be a support issue for `gruvbox`. We could style it similar to `area:msg`, since its color scheme kind of aligns with what we call `warning` (?)

zee-bit: Yep, sounds good!

neiljp (commented):
@Ezio-Sarthak Thanks for taking this up - I agree with others that the visual feedback is much better :+1:

I've left some comments inline, but I think we'll likely be able to squash these commits together.

Reading this PR it's more clear that we now call a function in a very similar way in many places! I'd consider a refactor before this to add something like `report_success` and similar methods to the controller which would have the default timeout and know the style to use and where to report it (ie. footer). That would make it less explicit where/how we're reporting it, but remove the need for all the explicit style variables here and the existing time variable (which we could scale up slightly based on the string length?).

This needs a black rebase in any case.

zulipterminal/config/themes.py
```diff
@@ -49,6 +48,10 @@
     'area:msg': 'standout',
     'area:stream': 'standout',
     'area:error': 'standout',
+    'footer:neutral': 'standout',
+    'task:success': 'standout',
+    'task:failure': 'standout',
+    'task:warning': 'standout',
```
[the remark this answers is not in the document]
Ezio-Sarthak: I think having a color (say green) on a result (say success) draws user's attention more than just changing the text with the default footer background. We could surely gain feedback in stream in any case.

zulipterminal/config/themes.py
```diff
@@ -206,6 +207,14 @@
          None,             DEF['white'],              DEF['dark_cyan']),
         ('area:error',     'white',                   'dark red',
          None,             DEF['white'],               DEF['dark_red']),
+        ('footer:neutral', 'black',                   'light gray',
+         None,             DEF['black'],              DEF['light_gray']),
```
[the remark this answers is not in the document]
Ezio-Sarthak: How about yellow-blue (foreground-background)?

zulipterminal/config/themes.py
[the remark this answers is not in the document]
neiljp: Right, my initial thought would be to temporarily do something like invert from the default style; I'm not sure what that translates to. Discussing on the stream would be good, though this would be a small update later too.

zulipterminal/config/themes.py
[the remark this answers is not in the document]
neiljp: That sounds quite noticeable - maybe too much compared to the others! - but maybe put some examples together and see what others think?

neiljp:
@Ezio-Sarthak Thanks for the polish on this; I adapted some of the commit text and have just merged it :tada:

PR #972, opened 2021-03-30
[WIP] refactor?: Linear Time vs Constant Time

Rohitth007:
Changes include:
* Optimization for keypress events
* Optimization for alt-keys
* self.model.controller -> self.controller

Do share your thoughts on this.

### DIFF
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -191,14 +191,17 @@ class View(urwid.WidgetWrap):
             self.middle_column.keypress(size, key)
             return key
         elif is_command_key('ALL_PM', key):
-            self.model.controller.narrow_to_all_pm()
+            self.controller.narrow_to_all_pm()
             self.body.focus_col = 1
+            return key
         elif is_command_key('ALL_STARRED', key):
-            self.model.controller.narrow_to_all_starred()
+            self.controller.narrow_to_all_starred()
             self.body.focus_col = 1
+            return key
         elif is_command_key('ALL_MENTIONS', key):
-            self.model.controller.narrow_to_all_mentions()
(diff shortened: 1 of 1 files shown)


andersk:
This change doesn’t actually reduce the time complexity. Instead of executing a chain of *n* `if` statements, you’re still defining *n* functions and building a dictionary of *n* elements on each key press, and both of these operations still take *O*(*n*) time.

Rohitth007:
Since, a python dictionary is an implementation of a hash table, 
it takes O(1) time as it looks up the value wrt the hash of the key 
as opposed to a sequence of 20 or so if else statements which is linear. 
See https://stackoverflow.com/questions/1963507/time-complexity-of-accessing-a-python-dict#1963514

andersk:
It takes *O*(1) time to look up one value in the dictionary, but it takes *O*(*n*) time to build the dictionary, and you’re doing both on each key press.

(To be clear, this isn’t a concerning amount of work with *n* fixed at around 20—but the original code wasn’t concerning by the same token and was clearer.)

Rohitth007:
@andersk Yes maybe that dictionary can be moved into `__init__` so that it builds only once.
It is similar to making a dictionary for registered event callbacks in `model.py` instead of sequencial if else.

Concerning 2nd point, that is why I asked if it was feasible on the stream, I think on the whole it should be effective and give time for other things to happen. I feel this also looks more concise. Also, it's not bad to make things ideal without compromising on readability ryt.

neiljp:
@Rohitth007 I'm going to close this since aside from the readability/optimization drawbacks, this doesn't respect the multiple-hotkey solution we use. That's not to say that there is not some repetition here that could potentially be refactored, but that's separate to the approach here.

Issue #905, opened 2021-02-02
Minimize length of stream box during compose based on available space & maximum stream length 


neiljp:
Currently the stream box can be quite wide; we spread the boxes between the available width.

While topic length can vary, we should be able to find some way to determine a smaller length using urwid's features based on what we know about the maximum stream length at the time.

Good related features to consider
* if we can/should limit the width of the 'topic edit' box while editing (edit-one, edit-later, edit-all messages in a topic)
* fix the length of the stream box while editing, if the stream cannot be changed (pending message moving feature)

I've labelled this as a good first issue initially as this would be good to help someone new to this to get to know urwid and the core UI, but this may be challenging or trickier than I expect and involve digging a little to find the right combination of spacing, so that this works over a wide range of window widths.

cyberShaw:
@zulipbot claim

zulipbot:
Welcome to Zulip, @cyberShaw! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

zulipbot:
@avijitdas126 You have been unassigned from this issue because you have not made any updates for over 14 days. Please feel free to reclaim the issue if you decide to pick up again. Thanks!

silentC1C4D4:
@zulipbot claim

zulipbot:
Hello @silentC1C4D4, it looks like someone has already claimed this issue! Since we believe multiple assignments to the same issue may cause some confusion, we encourage you to search for other unclaimed issues to work on. However, you can always reclaim this issue if no one is working on it.

We look forward to your valuable contributions!

cyberShaw:
@zulipbot abandon

jalajcodes:
@zulipbot claim

zulipbot:
[...]

PR #973, opened 2021-03-30
adding bug fix for issued 905

suzaram3:
Modified line 27 in zulipterminal/ui_tools/ui.py to change the width of the streams column


neiljp:
This does not address the original issue (#905) - which is not about setting the width of one of the major columns, but improving the width of the area assigned to the area for the stream name while composing. If you have further queries I'd suggest re-reading the issue and asking questions to clarify what is necessary there or in a new topic in the zulip-terminal stream on chat.zulip.org.

PR #978, opened 2021-04-02
view: Fix focus index before and after Panel Search.

### LINKED ISSUE #923: crash while using search
I was using the zulip search box [q] and I went to edit a former stream search. I deleted the entire search and pressed enter. Then the app crashed. I was unable to replicate this incident so I'm not sure if I should be creating an issue for it?

I am running Zulip Terminal 0.6.0+git on mac in a Z shell....The following was in my zulip-terminal-tracebacks.log:

{e}

'<' not supported between instances of 'NoneType' and 'int'
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zulipterminal/cli/run.py", line 434, in main
    Controller(zuliprc_path,
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zulipterminal/core.py", line 438, in main
    self.loop.run()
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/main_loop.py", line 287, in run
    self._run()
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/main_loop.py", line 385, in _run
    self.event_loop.run()
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/main_loop.py", line 790, in run
    self._loop()
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/main_loop.py", line 827, in _loop
    self._watch_files[fd]()
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/raw_display.py", line 416, in <lambda>
    wrapper = lambda: self.parse_input(
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/raw_display.py", line 515, in parse_input
    callback(processed, processed_codes)
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/main_loop.py", line 412, in _update
    self.process_input(keys)
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/main_loop.py", line 513, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zulipterminal/ui.py", line 181, in keypress
    return self.controller.current_editor().keypress((size[1],), key)
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zulipterminal/ui_tools/boxes.py", line 1592, in keypress
    self.panel_view.keypress(size, 'esc')
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zulipterminal/ui_tools/views.py", line 369, in keypress
    self.log.set_focus(self.focus_index_before_search)
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/listbox.py", line 203, in set_focus
    self.focus = position
  File "/Users/kellyfoulk/Documents/code/zulip/zulip-terminal/zt_venv/lib/python3.9/site-packages/urwid/monitored_list.py", line 164, in _set_focus
    if index < 0 or index >= len(self):
TypeError: '<' not supported between instances of 'NoneType' and 'int'

### LINKED ISSUE #975: Focus_Index_Before_Search doesn't work as intended
Suppose one is currently focused on the 10th Stream. After "escaping" search it is expected that the focus is back on the 10th Stream but it rather focuses on the stream it was in focus **during the search** right before escaping.
wrong-focus
This is also related to the crash in #923

Rohitth007:
This PR corrects what the code for `focus_index_before_search` was intended to do. See #975
The focus index was still being updated even in the search results.

This is fixed by moving `get_focus` into the SEARCH_STREAM/TOPIC
conditional.

Fixes #975

| Before Commit: | After Commit: |
| :-------------------: | :-----------------: |
| wrong-focus | right-focus |

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -568,64 +568,38 @@ class TestStreamsView:
     def test_keypress_SEARCH_STREAMS(self, mocker, stream_view, key, widget_size):
         size = widget_size(stream_view)
         mocker.patch.object(stream_view, "set_focus")
+        stream_view.log.extend(["FOO", "foo", "fan", "boo", "BOO"])
+        stream_view.log.set_focus(3)
+
         stream_view.keypress(size, key)
+
+        assert stream_view.focus_index_before_search == 3
         stream_view.set_focus.assert_called_once_with("header")
 
     @pytest.mark.parametrize("key", keys_for_command("GO_BACK"))
     def test_keypress_GO_BACK(self, mocker, stream_view, key, widget_size):
         size = widget_size(stream_view)
         mocker.patch.object(stream_view, "set_focus")
+        mocker.patch(VIEWS + ".urwid.Frame.keypress")
         mocker.patch.object(stream_view.stream_search_box, "reset_search_text")
+        stream_view.streams_btn_list = ["FOO", "foo", "fan", "boo", "BOO"]
+        stream_view.focus_index_before_search = 3
+
+        # Simulate search
+        stream_view.log.clear()
+        stream_view.log.extend(stream_view.streams_btn_list[3])
+        stream_view.log.set_focus(0)
+        stream_view.keypress(size, "down")
+        assert stream_view.log.get_focus()[1] != stream_view.focus_index_before_search
+
+        # Exit search
         stream_view.keypress(size, key)
+
+        # Check state reset after search
         stream_view.set_focus.assert_called_once_with("body")
         assert stream_view.stream_search_box.reset_search_text.called
-        assert stream_view.log == self.streams_btn_list
-
-    @pytest.mark.parametrize("search_streams_key", keys_for_command("SEARCH_STREAMS"))
-    @pytest.mark.parametrize("go_back_key", keys_for_command("GO_BACK"))
-    @pytest.mark.parametrize(
-        "current_focus, stream",
-        [
-            (0, "FOO"),
-            (2, "fan"),
-            (4, "BOO"),
-        ],
-    )
-    def test_return_to_focus_after_search(
-        self,
-        mocker,
-        stream_view,
-        widget_size,
-        current_focus,
-        stream,
-        search_streams_key,
-        go_back_key,
-    ):
-        # Initialize log
-        stream_view.streams_btn_list = [
-            mocker.Mock(stream_name=stream_name)
-            for stream_name in ["FOO", "foo", "fan", "boo", "BOO"]
-        ]
-        stream_view.log.extend(stream_view.streams_btn_list)
-
-        # Set initial stream focus to 'current_focus' and name to 'stream'
-        stream_view.log.set_focus(current_focus)
-        stream_view.focus_index_before_search = current_focus
-        previous_focus = stream_view.log.get_focus()[1]
-        previous_focus_stream_name = stream
-
-        # Toggle Stream Search
-        size = widget_size(stream_view)
-        stream_view.keypress(size, search_streams_key)
-
-        # Exit Stream Search
-        stream_view.keypress(size, go_back_key)
-
-        # Obtain new stream focus
-        new_focus = stream_view.log.get_focus()[1]
-        new_focus_stream_name = stream_view.log[new_focus].stream_name
-        assert new_focus == previous_focus
-        assert previous_focus_stream_name == new_focus_stream_name
+        assert stream_view.log == stream_view.streams_btn_list
+        assert stream_view.log.get_focus()[1] == stream_view.focus_index_before_search
 
 
 class TestTopicsView:
@@ -741,65 +715,39 @@ class TestTopicsView:
     def test_keypress_SEARCH_TOPICS(self, mocker, topic_view, key, widget_size):
         size = widget_size(topic_view)
         mocker.patch(VIEWS + ".TopicsView.set_focus")
+        topic_view.log.extend(["FOO", "foo", "fan", "boo", "BOO"])
+        topic_view.log.set_focus(3)
+
         topic_view.keypress(size, key)
+
         topic_view.set_focus.assert_called_once_with("header")
         topic_view.header_list.set_focus.assert_called_once_with(2)
+        assert topic_view.focus_index_before_search == 3
 
     @pytest.mark.parametrize("key", keys_for_command("GO_BACK"))
     def test_keypress_GO_BACK(self, mocker, topic_view, key, widget_size):
         size = widget_size(topic_view)
         mocker.patch(VIEWS + ".TopicsView.set_focus")
+        mocker.patch(VIEWS + ".urwid.Frame.keypress")
         mocker.patch.object(topic_view.topic_search_box, "reset_search_text")
+        topic_view.topics_btn_list = ["FOO", "foo", "fan", "boo", "BOO"]
+        topic_view.focus_index_before_search = 3
+
+        # Simulate search
+        topic_view.log.clear()
+        topic_view.log.extend(topic_view.topics_btn_list[3])
+        topic_view.log.set_focus(0)
+        topic_view.keypress(size, "down")
+        assert topic_view.log.get_focus()[1] != topic_view.focus_index_before_search
+
+        # Exit search
         topic_view.keypress(size, key)
+
+        # Check state reset after search
         topic_view.set_focus.assert_called_once_with("body")
         assert topic_view.topic_search_box.reset_search_text.called
-        assert topic_view.log == self.topics_btn_list
-
-    @pytest.mark.parametrize("search_topics_key", keys_for_command("SEARCH_TOPICS"))
-    @pytest.mark.parametrize("go_back_key", keys_for_command("GO_BACK"))
-    @pytest.mark.parametrize(
-        "current_focus, topic",
-        [
-            (0, "FOO"),
-            (2, "fan"),
-            (4, "BOO"),
-        ],
-    )
-    def test_return_to_focus_after_search(
-        self,
-        mocker,
-        topic_view,
-        widget_size,
-        current_focus,
-        topic,
-        search_topics_key,
-        go_back_key,
-    ):
-        # Initialize log
-        topic_view.topics_btn_list = [
-            mocker.Mock(topic_name=topic_name)
-            for topic_name in ["FOO", "foo", "fan", "boo", "BOO"]
-        ]
-        topic_view.log.extend(topic_view.topics_btn_list)
-
-        # Set initial focus to 'current_focus' and name to 'topic'
-        topic_view.log.set_focus(current_focus)
-        topic_view.focus_index_before_search = current_focus
-        previous_focus = topic_view.log.get_focus()[1]
-        previous_focus_topic_name = topic
-
-        # Toggle Search
-        size = widget_size(topic_view)
-        topic_view.keypress(size, search_topics_key)
-
-        # Exit Search
-        topic_view.keypress(size, go_back_key)
-
-        # Obtain new focus
-        new_focus = topic_view.log.get_focus()[1]
-        new_focus_topic_name = topic_view.log[new_focus].topic_name
-        assert new_focus == previous_focus
-        assert previous_focus_topic_name == new_focus_topic_name
+        assert topic_view.log == topic_view.topics_btn_list
+        assert topic_view.log.get_focus()[1] == topic_view.focus_index_before_search
 
 
 class TestUsersView:
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -365,6 +365,7 @@ class StreamsView(urwid.Frame):
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         if is_command_key("SEARCH_STREAMS", key):
+            _, self.focus_index_before_search = self.log.get_focus()
             self.set_focus("header")
             return key
         elif is_command_key("GO_BACK", key):
@@ -375,9 +376,7 @@ class StreamsView(urwid.Frame):
             self.log.set_focus(self.focus_index_before_search)
             self.view.controller.update_screen()
             return key
-        return_value = super().keypress(size, key)
-        _, self.focus_index_before_search = self.log.get_focus()
-        return return_value
+        return super().keypress(size, key)
 
 
 class TopicsView(urwid.Frame):
@@ -477,6 +476,7 @@ class TopicsView(urwid.Frame):
             self.view.show_left_panel(visible=False)
             self.view.body.focus_col = 1
         if is_command_key("SEARCH_TOPICS", key):
+            _, self.focus_index_before_search = self.log.get_focus()
             self.set_focus("header")
             self.header_list.set_focus(2)
             return key
@@ -488,9 +488,7 @@ class TopicsView(urwid.Frame):
             self.log.set_focus(self.focus_index_before_search)
             self.view.controller.update_screen()
             return key
-        return_value = super().keypress(size, key)
-        _, self.focus_index_before_search = self.log.get_focus()
-        return return_value
+        return super().keypress(size, key)
 
 
 class UsersView(urwid.ListBox):


Rohitth007:
@zulipbot remove "further discussion needed"

Rohitth007:
@zulipbot remove "further discussion required"

zulipterminal/ui_tools/views.py
```diff
@@ -293,13 +293,17 @@ def __init__(self) -> None:
 class StreamsView(urwid.Frame):
     def __init__(self, streams_btn_list: List[Any], view: Any) -> None:
         self.view = view
-        self.log = urwid.SimpleFocusListWalker(streams_btn_list)
+        # Create Stream List Box
         self.streams_btn_list = streams_btn_list
-        self.focus_index_before_search = 0
+        self.log = urwid.SimpleFocusListWalker(streams_btn_list)
```
preetmishra, line 298: Could you explain your reasoning behind the change? I don't think it is necessary.

zulipterminal/ui_tools/views.py
```diff
@@ -293,13 +293,17 @@ def __init__(self) -> None:
 class StreamsView(urwid.Frame):
     def __init__(self, streams_btn_list: List[Any], view: Any) -> None:
         self.view = view
-        self.log = urwid.SimpleFocusListWalker(streams_btn_list)
+        # Create Stream List Box
```
preetmishra, line 296: While comments are a good thing to include, it is always recommended to write code in a way that is readable, self-explanatory and self-documenting instead.

 When the code already explains what is happening, it is superfluous to include an obvious comment. :)

zulipterminal/ui_tools/views.py
```diff
@@ -293,13 +293,17 @@ def __init__(self) -> None:
...
         self.streams_btn_list = streams_btn_list
-        self.focus_index_before_search = 0
+        self.log = urwid.SimpleFocusListWalker(streams_btn_list)
         list_box = urwid.ListBox(self.log)
+        # Create Stream Search Box
```
preetmishra, line 300: It is evident that a search box is being created in the following statement. Let's remove the comment.

zulipterminal/ui_tools/views.py
```diff
@@ -293,13 +293,17 @@ def __init__(self) -> None:
...
         self.stream_search_box = PanelSearchBox(self,
                                                 'SEARCH_STREAMS',
                                                 self.update_streams)
+        self.focus_index_before_search = 0
+        self.in_search_mode = False
+        # Create Stream View Frame
```
preetmishra, line 306: Let's remove this as well. See my other comment.

zulipterminal/ui_tools/views.py
```diff
@@ -293,13 +293,17 @@ def __init__(self) -> None:
...
+        # Create Stream List Box
         self.streams_btn_list = streams_btn_list
-        self.focus_index_before_search = 0
+        self.log = urwid.SimpleFocusListWalker(streams_btn_list)
         list_box = urwid.ListBox(self.log)
+        # Create Stream Search Box
         self.stream_search_box = PanelSearchBox(self,
                                                 'SEARCH_STREAMS',
                                                 self.update_streams)
+        self.focus_index_before_search = 0
```
preetmishra, line 304: A better default value would be `None`, I gather. That would indicate that we didn't have any searches yet.

zulipterminal/ui_tools/views.py
```diff
@@ -293,13 +293,17 @@ def __init__(self) -> None:
...
+        self.in_search_mode = False
```
preetmishra, line 305: We use `is_in` prefix for booleans that track whether an entity _is_ currently _in_ some state. For instance, see `is_in_typeahed_mode` in our source.

zulipterminal/ui_tools/views.py
```diff
@@ -365,27 +370,33 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             self.set_focus('body')
             self.log.set_focus(self.focus_index_before_search)
             self.view.controller.update_screen()
+            self.in_search_mode = False
             return key
         return_value = super().keypress(size, key)
-        _, self.focus_index_before_search = self.log.get_focus()
+        if not self.in_search_mode:
+            _, self.focus_index_before_search = self.log.get_focus()
```
preetmishra, line 377: We could just set the focus in the above if condition instead? That would eliminate the need for another attribute, i.e., `in_search_mode`.

preetmishra (commented):
@Rohitth007 Thanks for putting the work through! This seems to work well. :+1: 

However, I think we could simplify the approach (see my in-line comment). I would encourage you to add tests so that we can ship it with more confidence.

Also, it is not apparent through the commit title where the search is being done. You might want to explicate it a bit more.

zulipterminal/ui_tools/views.py
Rohitth007: So, all the comments I have done are so that I can group related variables together. It would be confusing to know in a single glance what the variables are for and where they are used. Infact, even `empty_search` and `search_lock` can be brought above.
I think `Create` can be removed from the comments, so that the single line actual creation is not confused with.

zulipterminal/ui_tools/views.py
Rohitth007: The exchange was done to facilitate the grouping

zulipterminal/ui_tools/views.py
Rohitth007: This would cause an error with `self.log.set_focus()` as it accepts only `int`. ZT can crash in that case

zulipterminal/ui_tools/views.py
Rohitth007: :+1: I'll change it to `is_in_search_mode`

zulipterminal/ui_tools/views.py
Rohitth007: Oh yes it could be done inside SEARCH_STREAMS right? It was previously outside so I thought it had to be updated everytime. Now, I see there's no use for that.

Rohitth007: I've made this change and updated the PR

zulipterminal/ui_tools/views.py
Rohitth007: I've made some changes to this part in the first refactor commit. It might be useful doing this?

zulipterminal/ui_tools/views.py
preetmishra: Right, that would require extra logic. The current implementation looks good.

zulipterminal/ui_tools/views.py
preetmishra: Hmm. I am not entirely sure about the refactor. Perhaps @neiljp could pitch in.

preetmishra (commented):
@Rohitth007 Thanks for the update. This seems to work well! :+1: 

It would be good to have it in the User search as well while we're at it. It should be straightforward to implement.

The second commit looks good to me. However, I would strongly encourage you to accompany the changes with tests.

Re the first commit, let's see what Neil has to say about the refactor.

Rohitth007:
We could also have it for the user list but it wasn't there to begin with so I thought it wasn't needed. Also we have to make sure that if a user goes to the bottom of the list and come back to the message box, the list doesn't get stuck theme, as same section of code is used. I haven't checked if this would happen.

Rohitth007:
A more useful feature would be to focus back to where it was before search if it was not in the list. Eg: If we are in the message view and press q, escaping should bring us back to that message. That doesn't happen currently but is also out of the scope of this PR

Rohitth007:
Re tests, I'll update them by tomorrow

preetmishra:
@Rohitth007 I am not sure if I understand your first premise completely. Let's talk more about it on CZO. Re the latter, I had something similar while reviewing. Could you file an issue for the same as it is something that we can take up separately?

Rohitth007:
@preetmishra I have added some tests for this and removed bad tests which gave false positives

Rohitth007:
I've implemented user list refocus in the 3rd commit. It needs some checking.

Rohitth007:
I have updated the tests for user list refocusing as well

tests/ui/test_ui_tools.py
```diff
@@ -568,64 +568,34 @@ def test_mouse_event(self, mocker, stream_view, widget_size):
...
-        # Obtain new stream focus
-        new_focus = stream_view.log.get_focus()[1]
-        new_focus_stream_name = stream_view.log[new_focus].stream_name
-        assert new_focus == previous_focus
-        assert previous_focus_stream_name == new_focus_stream_name
+        # Check state reset after search
+        stream_view.set_focus.assert_called_once_with("body")
+        assert stream_view.stream_search_box.reset_search_text.called
+        assert stream_view.log == stream_view.streams_btn_list
+        assert stream_view.log.get_focus()[1] == stream_view.focus_index_before_search
```
preetmishra, line 598: Let's also check whether `set_focus` was called as well?

tests/ui/test_ui_tools.py
```diff
@@ -1043,6 +1043,7 @@ def test_init(self, right_col_view):
         assert right_col_view.view.user_search == right_col_view.user_search
         self.thread.Lock.assert_called_with()
         assert right_col_view.search_lock == self.thread.Lock()
+        assert right_col_view.focus_index_before_search == 0
```
preetmishra, line 1046: minor: The commit title is not in the imperative tense. :)

preetmishra (commented):
@Rohitth007 This continues to work well functionally. Thanks for the update! :+1: 

We could potentially secure the tests for RightColumnView. I have left a few in-line comments related to tests.

tests/ui/test_ui_tools.py
```diff
@@ -1150,6 +1157,9 @@ def test_keypress_GO_BACK(self, right_col_view, mocker, key, widget_size):
 
         right_col_view.set_body.assert_called_once_with(right_col_view.body)
         right_col_view.set_focus.assert_called_once_with("body")
+        right_col_view.body.log.set_focus.assert_called_once_with(
```
preetmishra, line 1160: This doesn't check whether the focus before and after the search match. Currently, this passes even when I remove the `self.focus_index_before_search = self.body.log.get_focus()[1]` from the other `if` block. (This was supposed to go with my last review.)

zulipterminal/ui_tools/views.py
neiljp: I appreciate the desire to tidy the code, but this might be reasonable as part of a larger cleanup like deduplicating the stream+topic list.

neiljp:
@Rohitth007 Let's focus on the bugfix commit which functionally works well and tune that test :+1:

We could incorporate the initial refactor into a more general clean-up like I mentioned.
Regarding the last commit, I think we were discussing that in the stream and there are challenges with the automatic refresh which make knowing where to return to unclear, so let's put that into another set of work targeted at understanding how to resolve that separately?

zulipterminal/ui_tools/views.py
```diff
@@ -367,6 +367,7 @@ def mouse_event(
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         if is_command_key("SEARCH_STREAMS", key):
+            self.focus_index_before_search = self.log.get_focus()[1]
```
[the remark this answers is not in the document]
Rohitth007: Hmm, ok thought this would be clearer but it doesn't tell how many elements are being unpacked. I'll change it to the other one.

tests/ui/test_ui_tools.py
```diff
@@ -741,65 +711,35 @@ def test_keypress_GO_RIGHT(self, mocker, topic_view, key, widget_size):
...
-    @pytest.mark.parametrize("go_back_key", keys_for_command("GO_BACK"))
-    @pytest.mark.parametrize(
-        "current_focus, topic",
-        [
-            (0, "FOO"),
-            (2, "fan"),
-            (4, "BOO"),
-        ],
-    )
-    def test_return_to_focus_after_search(
```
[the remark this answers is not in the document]
Rohitth007: So the problem is that it only enters the search bar and not the search results, I don't think there is a way to use the 'down' keypress to change focus inside the log, I was getting errors. Hence I assumed I was already inside the search results and was escaping from there which is also shorter and can be done inside GO_BACK

tests/ui/test_ui_tools.py
Rohitth007: Oh sry I'll change that

tests/ui/test_ui_tools.py
```diff
@@ -568,64 +568,34 @@ def test_mouse_event(self, mocker, stream_view, widget_size):
     def test_keypress_SEARCH_STREAMS(self, mocker, stream_view, key, widget_size):
         size = widget_size(stream_view)
         mocker.patch.object(stream_view, "set_focus")
+        stream_view.log.extend(["FOO", "foo", "fan", "boo", "BOO"])
```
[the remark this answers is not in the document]
Rohitth007: :+1:

tests/ui/test_ui_tools.py
```diff
@@ -568,64 +568,34 @@ def test_mouse_event(self, mocker, stream_view, widget_size):
...
+        stream_view.log.set_focus(3)
+
         stream_view.keypress(size, key)
```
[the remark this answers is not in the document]
Rohitth007: Yes makes sense, I'll do that

tests/ui/test_ui_tools.py
Rohitth007: Isn't it same as checking whether focus_index was updated? Doing either tests the same functionality?

tests/ui/test_ui_tools.py
Rohitth007: But that would mean the other test fails right? The current tests for userlist are checking only what each function is doing. I'm not able to test this like the previous 2 for some reason

tests/ui/test_ui_tools.py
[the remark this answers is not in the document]
Rohitth007: I just realized why I couldn't get 'down' to work.
But since all of the code happens inside SEARCH_STREAMS and GO_BACK unlike before. It makes sense to write these tests inside GO_BACK as this is what is responsible for resetting state and also it's much shorter and cleaner and tests the same functionality

Rohitth007:
I have made few changes as suggested and moved the bugfix commit to the top. I'll move the other commits into new PR's to be worked on separately.

tests/ui/test_ui_tools.py
Rohitth007: I think we can come back to this after the Panel refactor when the tests also become simpler?

tests/ui/test_ui_tools.py
preetmishra: Right, it's a bit redundant here.

tests/ui/test_ui_tools.py
preetmishra: Generally, tests shouldn't rely on other tests to fail. Let's revisit it post the refactor. :+1:

Rohitth007:
I have moved the other commits into separate branches. This is ready to be merged

neiljp:
@Rohitth007 Thanks for the test reworking :+1: There's still a lot of manual implementation-dependent setup going on in the tests which I'll be glad to see gone in future! Identifying the bugfix was a good find and this is small and clean to merge now :tada:

PR #979, opened 2021-04-03
Add script to enable users to test symbol rendering on their terminals

prah23:
This PR has 2 commits that do the following:
- Create a tool to be shipped with ZT that allows users to test symbol rendering on their respective terminal emulators. In the current iteration, the command to run the script is `zulip-term-check-symbols`. This script has been placed in a new folder called `scripts/`.
- `FAQ.md` has been updated to point users/contributors to the same, to help increase awareness on possible symbol rendering issues.

Tested on Windows Terminal and GNOME Terminal on Ubuntu, the layout is maintained in both.

**Screenshots:** As observed on Windows Terminal:
Symbols rendering script

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -66,6 +66,18 @@ One session of Zulip Terminal represents a connection of one user to one Zulip s
 
 Since we expect the above to be straightforward for most users and it allows the code to remain dramatically simpler, we are unlikely to support multiple Zulip servers within the same session in at least the short/medium term. However, we are certainly likely to move towards a system to make access of the different servers simpler, which should be made easier through work such as [zulip-terminal#678](https://github.com/zulip/zulip-terminal/issues/678).
 
+## Unable to render symbols
+
+If some symbols don't render properly on your terminal, it could likely be because of the symbols not being supported on your terminal emulator and/or font.
+
+We provide a tool that you can run with the command `zulip-term-check-symbols` to check whether or not the symbols render properly on your terminal emulator and font configuration.
+
+Ideally, you should see something similar to the following screenshot (taken on the GNOME Terminal) as a result of running the tool:
+
+![Render Symbols Screenshot](https://user-images.githubusercontent.com/60441372/115103315-9a5df580-9f6e-11eb-8c90-3b2585817d08.png)
+
+If you are unable to observe a similar result upon running the tool, please take a screenshot and let us know about it along with your terminal and font configuration by opening an issue or at [#zulip-terminal](https://chat.zulip.org/#narrow/stream/206-zulip-terminal).
+
 ## Unable to open links
 
 If you are unable to open links in messages, then try double right-click on the link.
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -66,17 +66,6 @@ One session of Zulip Terminal represents a connection of one user to one Zulip s
 
 Since we expect the above to be straightforward for most users and it allows the code to remain dramatically simpler, we are unlikely to support multiple Zulip servers within the same session in at least the short/medium term. However, we are certainly likely to move towards a system to make access of the different servers simpler, which should be made easier through work such as [zulip-terminal#678](https://github.com/zulip/zulip-terminal/issues/678).
 
-## Unable to render non-ASCII characters
-
-**NOTE** Releases of 0.3.2 onwards should not have this issue, or require this solution.
-
-If you see `?` in place of emojis or Zulip Terminal gives a `UnicodeError` / `CanvasError`, you haven't enabled utf-8
-encoding in your terminal. To enable it by default, add this to the end of you `~/.bashrc`:
-
-```
-export LANG=en_US.utf-8
-```
-
 ## Unable to open links
 
 If you are unable to open links in messages, then try double right-click on the link.
--- a/zulipterminal/scripts/render_symbols.py
+++ b/zulipterminal/scripts/render_symbols.py
@@ -1,0 +1,52 @@
+import sys
+
+import urwid
+from urwid import set_encoding
+
+from zulipterminal.config import symbols
+
+
+set_encoding('utf-8')
+
+palette = [
+    ('header', 'white, bold', 'dark blue'),
+    ('footer', 'black', 'white'),
+]
+
+symbol_dict = {name: symbol for name, symbol in vars(symbols).items()
+               if not name.startswith('__') and not name.endswith('__')}
+max_symbol_name_length = max([len(name) for name in symbol_dict.keys()])
+
+symbol_names_list = [urwid.Text(name, align='center')
+                     for name in symbol_dict.keys()]
+symbols_list = [urwid.Text(symbol) for symbol in symbol_dict.values()]
+
+symbols_display_box = urwid.Columns([
+    # Allot extra width to the symbol names list to create a neat layout.
+    (max_symbol_name_length + 2, urwid.Pile(symbol_names_list)),
+    # Allot 2 characters to the symbols list to accommodate symbols needing
+    # double character-width to render.
+    (2, urwid.Pile(symbols_list))
+], dividechars=5)
+columns_length = max_symbol_name_length + 10
+
+line_box = urwid.LineBox(symbols_display_box,
+                         title="Render Symbols",
+                         title_attr='header')
+info_box = urwid.Text(('footer', u" Exit: ^C "), align='center')
+display_box = urwid.Pile([
+    line_box,
+    info_box
+])
+
+# Allot extra width to the display_box to render a neat layout.
+body = urwid.Filler(urwid.Padding(display_box, width=columns_length + 5,
+                                  align='center'), valign='middle')
+
+
+def main() -> None:
+    try:
+        loop = urwid.MainLoop(body, palette)
+        loop.run()
+    finally:
+        sys.exit(0)
--- a/setup.py
+++ b/setup.py
@@ -101,6 +101,8 @@ setup(
     entry_points={
         'console_scripts': [
             'zulip-term = zulipterminal.cli.run:main',
+            'zulip-term-check-symbols = '
+            'zulipterminal.scripts.render_symbols:main',
         ],
     },
     extras_require={


docs/FAQ.md
```diff
@@ -77,6 +77,11 @@ encoding in your terminal. To enable it by default, add this to the end of you `
 export LANG=en_US.utf-8
 ```
 
+## Unable to render symbols
+If some symbols don't render quite right on your terminal, it could likely be because of the symbols not being supported on the emulator/font.
+
+You can run `zulip-term-check-symbols` to check whether or not the symbols render properly on your terminal emulator.
```
neiljp, line 83: I'd suggest we combine this with the previous section?

zulipterminal/scripts/render_symbols.py
```diff
@@ -0,0 +1,60 @@
...
+
+display_box = urwid.Pile([
+    line_box,
+    info_box
+])
+body = urwid.Filler(display_box, valign='middle')
+
+
+def main() -> None:
+    loop = urwid.MainLoop(urwid.Padding(body, left=30, right=30),
```
Ezio-Sarthak, line 58: I suggest we could try decreasing the padding, to avoid splitting of large symbol names in multiple lines.

zulipterminal/scripts/render_symbols.py
```diff
@@ -0,0 +1,60 @@
...
+                                    tlcorner='│', tline='─', lline='│',
+                                    trcorner='┬', blcorner='└', rline='│',
+                                    bline='─', brcorner='┴')
+symbols_widget = urwid.LineBox(urwid.Pile(symbols_list), title="Symbols",
+                               title_attr='sub_header',
+                               tlcorner='─', tline='─', lline='',
+                               trcorner='│', blcorner='─', rline='│',
+                               bline='─', brcorner='┘')
+
+symbols_display_box = urwid.Columns([
```
Ezio-Sarthak, line 37: This could be simpler if we wrap just Pile elements with this column widget.
(AFA I understand, the names are pretty straightforward, and so explicit titles of these names and symbols could be avoided. Let me know your thoughts on this :) )

Ezio-Sarthak (commented):
This looks great in production 👍

While we aim to check for symbol support, we'd likely want the UI to be simpler, so as to not cause any other rendering issues :)

Leaving some comments inline :)

zulipterminal/scripts/render_symbols.py
neiljp: I've not checked the docs, but can't we leave the padding not specified and still get it centered?

docs/FAQ.md
prah23: Yeah, that makes sense, I'll do that. I put it in a separate section thinking it would add emphasis on the tool we're providing but that makes sense.

zulipterminal/scripts/render_symbols.py
prah23: It would still be centered, but visually not very clean as it renders a large box and most of the space is left unused. I was initially at a padding of 40 on the right and left on Windows Terminal, and then reverted back to 30 after running it on the GNOME terminal on Ubuntu.

zulipterminal/scripts/render_symbols.py
prah23: They are pretty straightforward, agreed, but I was of the opinion that an explicit mention would simplify things for users, but I suppose not using an explicit heading would remove the need for the nested LineBox, thereby simplifying the layout.

zulipterminal/scripts/render_symbols.py
```diff
@@ -0,0 +1,42 @@
...
+                               urwid.Text(symbol_dict[symbol], align='center')
+                               ]) for symbol in symbol_dict]
+symbols_display_box = urwid.Pile(symbols_list)
+
+line_box = urwid.LineBox(symbols_display_box,
+                         title="Render Symbols",
+                         title_attr='header',
+                         tlcorner='┌', tline='─', lline='│',
+                         trcorner='┐', blcorner='└', rline='│', bline='─',
+                         brcorner='┘')
```
zee-bit, line 27: These all are default parameters of `LineBox`, so no need to specify explicitly?

zulipterminal/scripts/render_symbols.py
```diff
@@ -0,0 +1,42 @@
...
+    ('footer', 'black', 'white'),
+]
+
+variables = vars(symbols)
+symbol_dict = {symbol: variables[symbol] for symbol in variables
+               if not symbol.startswith('__') and not symbol.endswith('__')}
+
+symbols_list = [urwid.Columns([urwid.Text(symbol, align='center'),
+                               urwid.Text(symbol_dict[symbol], align='center')
+                               ]) for symbol in symbol_dict]
```
zee-bit, line 19: This is sort of hacky, but we may specify `align='right'` and `'left'` to respective `Text` widgets and provide some `dividechars` to `Column`, if we want to reduce the distance between name and symbol?

zee-bit (commented):
This is good. :+1: 

And, although the code is much simpler and rendering fine, we do get a very wide box(at least on my terminal). The names and their corresponding symbols are also very far apart, and my comment below may help reduce the distance?

Apart from that, it's nice work. :)

zulipterminal/scripts/render_symbols.py
zee-bit: The `LineBox` however would still remain wide - as far as I tested :(

prah23: Yeah, the width seems to vary based on the font size and the width allowed to the terminal. I've tried reducing that using a padding, but Neil suggested to just keep it simple and not worry about the width. The current iteration would use the complete width available since there is no padding.

prah23: With no padding, each symbol name should render in a single line in most terminal font size and width combinations, beyond which it might not be feasible for us to consider.

prah23:
@zeebit
With the align `left` and `right` hack, with a `dividechars` of 10, it would look like this on varying terminal width:
image

image
It might look misaligned to the user this way, so I've not included the hack for now and just left it center aligned, which seems to provide consistent alignment in most cases, do you have any further thoughts here?

prah23:

docs/FAQ.md
```diff
@@ -71,11 +71,14 @@ Since we expect the above to be straightforward for most users and it allows the
 **NOTE** Releases of 0.3.2 onwards should not have this issue, or require this solution.
 
 If you see `?` in place of emojis or Zulip Terminal gives a `UnicodeError` / `CanvasError`, you haven't enabled utf-8
-encoding in your terminal. To enable it by default, add this to the end of you `~/.bashrc`:
+encoding in your terminal. To enable it by default, add this to the end of your `~/.bashrc`:
```
zee-bit, line 74: Should we specify the shell in a more general way? For e.g. Mac uses `ZSH` if I am correct, so they won't have a `bashrc` but a `zshrc`. I personally moved on to ZSH quite recently too, and although I still have a `bashrc` adding the below line there probably won't work for me(?)

prah23: Updated, does this seem more general?

zulipterminal/scripts/render_symbols.py
```diff
@@ -0,0 +1,39 @@
...
+])
+body = urwid.Filler(display_box, valign='middle')
+
+
+def main() -> None:
+    try:
+        loop = urwid.MainLoop(body, palette)
+        loop.run()
+    finally:
+        sys.exit(1)
```
neiljp, line 39: minor: zero is normally a sign of successful exit.

zulipterminal/scripts/render_symbols.py
```diff
@@ -0,0 +1,39 @@
...
+    ('footer', 'black', 'white'),
+]
+
+variables = vars(symbols)
+symbol_dict = {symbol: variables[symbol] for symbol in variables
+               if not symbol.startswith('__') and not symbol.endswith('__')}
+
+symbols_list = [urwid.Columns([urwid.Text(symbol, align='center'),
+                               urwid.Text(symbol_dict[symbol], align='center')
+                               ]) for symbol in symbol_dict]
```
neiljp, line 19: This would be clearer with `symbol_dict.items()`.

neiljp (commented):
@prah23 I'm not sure we need to linger on this PR too long, but the spaced-out layout makes it tricky to determine which element aligns with each text description. I think you may be able to get a more compact form of this by explicitly looking for the maximum lengths of each column (or at least the names) and specifying the width. You may be able to just use one column pair rather than one per row? Also given the Padding seems challenging to get right, perhaps cheat a little and put the linebox with fixed width in 3 columns?

Re the FAQ text, I think we need to simplify and maybe just say 'symbols' in place of non-ASCII. The previous problem should be long-gone (0.3.0 is some time/versions ago), and we should be resolving the problem listed by setting the encoding in ZT in any case. So maybe just mostly replace that paragraph? @zee-bit Did you have any issues with characters?

prah23:
@neiljp 
I've updated the layout. I've used explicit widths to pack the layout together to make it look more compact and concise. I've used a `Padding` and have given the widgets inside it just a little over the exact number of characters they need. I've also added comments in the code as to how I've used the widths, in case we add symbols in the future (with longer names) and might need to update the script. 

I've tested this on Windows Terminal (updated the description of the PR with the latest screenshot) and the GNOME Terminal, the display renders the same way and does not break on varying the terminal window's layout.

neiljp:
@prah23 To clarify my previous review, while this is expected to be a less-used script I'd prefer that we don't hard-code the widths but calculate them - we don't update the symbols very often, but it just takes one long name and the table goes awry from wrapping.

The FAQ entry reads fine, but the user needs a way to validate what the script should produce - eg. v1 maybe suggest taking a screenshot and asking on czo if there are problems, or v2 have a trimmed-down screenshot of the script as it stands in the FAQ itself? (v3 ideally itself generated for when we update it and doing diffs in CI?)

prah23:
@neiljp
I've updated the script to check for the maximum symbol name length and structure the layout accordingly, I've set the widths relative to this maximum length.

For the FAQ update, I've mentioned up to the suggested v2, I've shared a screenshot of the tool (taken on the GNOME Terminal) that the users can compare with and then let us know here or on CZO.

prah23:

neiljp:
@prah23 I just merged with a few minor textual changes - add that ctrl c == ^C, and some minor commit text tweaks :tada:

PR #981, opened 2021-04-05
bugfix: Update stream-setting checkboxes from subscription events.

### LINKED ISSUE #747: Update stream settings pop-up on subscription events
Currently, the stream settings pop-up is not updated after receiving pin/mute stream events. This causes a mismatch between the stream state and the pop-up checkbox state when the settings are updated elsewhere when the pop-up is open in a ZT instance.

This can likely be fixed by using an absolute state instead of just toggling from previous state in the checkbox callbacks.

zee-bit:
This PR fixes the mismatch between stream settings if the `StreamInfo` popup is opened in ZT and subscription events are received from the server. The current approach registers a callback when the `StreamInfoView` is opened and looks for `subscription` events to trigger the callback, which in turn updates the checkbox UI.

Fixes #747.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -98,6 +98,8 @@ class Model:
 
         self._notified_user_of_notification_failure = False
 
+        self.callbacks: Dict[str, Any] = {}
+
         # Events fetched once at startup
         self.initial_data_to_fetch: List[str] = [
             "realm",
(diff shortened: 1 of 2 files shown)


zee-bit:
@zulipbot add "further discussion required"

neiljp:
I'm wary from the description that a polling approach is the way to go with this, but rather that we do something like register a callback when opening the popup, which is triggered on events. That could be the starting point for a shift in how we glue parts of the application together, including loosening the coupling between the model and UI.

zee-bit:
I have already changed my approach from polling to callback-based. This works well as a v1, but I need some feedback on improvement so that I can generalize this for other UI elements that depend on events for updation.
- The first commit just adds an `asynch` function that will be responsible for updating the checkbox UI and will be passed as callback.
- The second commit defines the logic for registering and processing callbacks from `model` that will eventually be used to update the UI.

PR #987, opened 2021-04-07
Show sender's presence status in message content header

### LINKED ISSUE #896: Show sender/recipient presence status in messages
This is to clarify the intent behind PR #501, which isn't quite #406 (related to highlighting particular users eg. yourself vs others by name), but more to translate the presence information into an extra field in messages to avoid scanning through the user list to see the sender's active status. We could also apply this to recipients in PMs.

Ezio-Sarthak:
**What does this PR do?!**
 - Adds sender's presence status marker in a message content header

**Commit flow**
 - Centralize the state icons in `ui_mappings.py` (*refactor*)
 - Add state marker as a prefix to sender names in the message header.
 - Add helper in `MessageBox` to update author field if present.
 - Add a method to render presence markers in messages regularly.

**Screenshots/GIFs**

Screenshot from 2021-06-26 23-02-58

Fixes #896.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -11,6 +11,7 @@ from zulipterminal.config.keys import keys_for_command
 from zulipterminal.config.symbols import (
     QUOTED_TEXT_MARKER,
     STATUS_ACTIVE,
+    STATUS_INACTIVE,
     STREAM_TOPIC_SEPARATOR,
     TIME_MENTION_MARKER,
 )
@@ -2112,9 +2113,12 @@ class TestMessageBox:
     @pytest.mark.parametrize(
         "expected_header, to_vary_in_last_message",
         [
-            (["alice", " ", "DAYDATETIME"], {"sender_full_name": "bob"}),
-            ([" ", " ", "DAYDATETIME"], {"timestamp": 1532103779}),
-            (["alice", " ", "DAYDATETIME"], {"timestamp": 0}),
+            (
+                [STATUS_INACTIVE, "alice", " ", "DAYDATETIME"],
+                {"sender_full_name": "bob"},
+            ),
+            ([" ", " ", " ", "DAYDATETIME"], {"timestamp": 1532103779}),
+            ([STATUS_INACTIVE, "alice", " ", "DAYDATETIME"], {"timestamp": 0}),
         ],
         ids=[
             "show_author_as_authors_different",
@@ -2142,6 +2146,9 @@ class TestMessageBox:
             " ",
         ] * 2  # called once in __init__ and then in main_view explicitly
 
+        # The empty dict is responsible for INACTIVE status of test user.
+        self.model.user_dict = {}  # called once in main_view explicitly
+
         stars = {
             msg: ({"flags": ["starred"]} if msg == starred_msg else {})
             for msg in ("this", "last")
@@ -2152,10 +2159,10 @@ class TestMessageBox:
 
         msg_box = MessageBox(this_msg, self.model, last_msg)
 
-        expected_header[1] = output_date_time
+        expected_header[2] = output_date_time
         if current_year > 2018:
-            expected_header[1] = "2018 - " + expected_header[1]
-        expected_header[2] = "*" if starred_msg == "this" else " "
+            expected_header[2] = "2018 - " + expected_header[2]
+        expected_header[3] = "*" if starred_msg == "this" else " "
 
         view_components = msg_box.main_view()
 
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -2211,6 +2211,30 @@ class TestMessageBox:
             assert label[0].text == "EDITED"
             assert label[1][1] == 7
 
+    @pytest.mark.parametrize(
+        "to_vary_in_last_message, update_required",
+        [
+            ({"sender_full_name": "Unique name (won't be in next message)"}, True),
+            ({}, False),
+        ],
+        ids=[
+            "author_field_present",
+            "author_field_not_present",
+        ],
+    )
+    def test_update_message_author_status(
+        self,
+        message_fixture,
+        update_required,
+        to_vary_in_last_message,
+    ):
+        message = message_fixture
+        last_msg = dict(message, **to_vary_in_last_message)
+
+        msg_box = MessageBox(message, self.model, last_msg)
+
+        assert msg_box.update_message_author_status() == update_required
+
     @pytest.mark.parametrize("key", keys_for_command("STREAM_MESSAGE"))
     @pytest.mark.parametrize(
         "narrow, expect_to_prefill",
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -1429,6 +1429,31 @@ class MessageBox(urwid.Pile):
         ]
         return [part for part, condition in parts if condition]
 
+    def update_message_author_status(self) -> bool:
+        """
+        Update the author status by resetting the entire message box
+        if author field is present.
+        """
+        author_is_present = False
+        author_column = 1  # Index of author field in content header
+        author_field = None
+
+        # Get message author field (if present)
+        if self.need_recipient_header():
+            author_field = self[1][author_column]
+        elif isinstance(self[0][author_column], urwid.Text):
+            author_field = self[0][author_column]
+
+        if author_field is not None:
+            author_is_present = author_field.text != " "
+
+        if author_is_present:
+            # Re initialize the message if update is required.
+            # FIXME: Render specific element (here author field) instead?
+            super().__init__(self.main_view())
+
+        return author_is_present
+
     @classmethod
     def transform_content(
         cls, content: Any, server_url: str
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -31,6 +31,7 @@ from zulipterminal.config.symbols import (
     STREAM_TOPIC_SEPARATOR,
     TIME_MENTION_MARKER,
 )
+from zulipterminal.config.ui_mappings import STATE_ICON
 from zulipterminal.helper import (
     Message,
     asynch,
@@ -1324,11 +1325,23 @@ class MessageBox(urwid.Pile):
 
         if any_differences:  # Construct content_header, if needed
             TextType = Dict[str, Tuple[Optional[str], str]]
-            text_keys = ("author", "star", "time")
+            text_keys = ("author", "star", "time", "status")
             text: TextType = {key: (None, " ") for key in text_keys}
 
             if any(different[key] for key in ("recipients", "author", "24h")):
                 text["author"] = ("name", message["this"]["author"])
+
+                # TODO: Refactor to use user ids for look up instead of emails.
+                email = self.message.get("sender_email", "")
+                user = self.model.user_dict.get(email, None)
+                # TODO: Currently status of bots are shown as `inactive`.
+                # Render bot users' status with bot marker as a follow-up
+                status = user.get("status", "inactive") if user else "inactive"
+
+                # The default text['status'] value is (None, ' ')
+                if status in STATE_ICON:
+                    text["status"] = (f"user_{status}", STATE_ICON[status])
+
             if message["this"]["is_starred"]:
                 text["star"] = ("starred", "*")
             if any(different[key] for key in ("recipients", "author", "timestamp")):
@@ -1341,6 +1354,7 @@ class MessageBox(urwid.Pile):
 
             content_header = urwid.Columns(
                 [
+                    ("pack", urwid.Text(text["status"])),
                     ("weight", 10, urwid.Text(text["author"])),
                     (26, urwid.Text(text["time"], align="right")),
                     (1, urwid.Text(text["star"], align="right")),
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -552,6 +552,14 @@ class MiddleColumnView(urwid.Frame):
             return pm
         return None
 
+    def update_message_list_status_markers(self) -> None:
+        for message_w in self.body.log:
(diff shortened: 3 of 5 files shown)


Ezio-Sarthak:

zulipterminal/helper.py
```diff
@@ -26,6 +26,12 @@
 from typing_extensions import Literal, TypedDict
 
 from zulipterminal.api_types import Composition, Message
+from zulipterminal.config.symbols import (
```
preetmishra, line 29: Let's include the filenames in the commit title.

zulipterminal/helper.py
```diff
@@ -101,6 +107,14 @@ class UnreadCounts(TypedDict):
 }
 
 
+STATE_ICON = {
```
preetmishra, line 110: Should we call it `STATE_ICONS` instead?

zulipterminal/ui_tools/boxes.py
```diff
@@ -1198,11 +1199,21 @@ def main_view(self) -> List[Any]:
...
             if any(different[key] for key in ('recipients', 'author', '24h')):
                 text['author'] = ('name', message['this']['author'])
+
+                email = self.message.get('sender_email', '')
+                user = self.model.user_dict.get(email, None)
+                # TODO: Currently status of bots are shown as `inactive`.
+                # Render bot users' status with bot marker as a follow-up
+                status = user.get('status', 'inactive') if user else 'inactive'
+
+                if status in STATE_ICON:
```
preetmishra, line 1214: Since this doesn't have an else case, line `1232` might raise an exception, right?

preetmishra (commented):
@Ezio-Sarthak Thanks for working on this! I have tested this locally and it seems to work well. :+1:

I have left some quick comments for the first review iteration. Let's squash the last commit as you mention in the PR description.

zulipterminal/helper.py
Ezio-Sarthak: Hmm. Seems more apt :+1:

zulipterminal/ui_tools/boxes.py
Ezio-Sarthak: To clarify, line `1203` sets the default values for all the `text_keys` (see line `1202`). So, the default value of `text['status']` would be `(None, ' ')`. Hope this helps :)

Nonetheless, I think it's good to add a note here in any case :+1:

zulipterminal/helper.py
neiljp: When naming like this, it's good to think of how it's used - `STATE_ICONS['active']` suggests there are more than one active one? Generally I err towards the non-plural for this reason.

Also the values are icons, but the keys aren't - so as always, naming is tricky to get perfect!

This relates application state values to representation, a little like `edit_mode_captions`, so perhaps it belongs in with that, though I'm not keen on `helper.py` for that.

neiljp (commented):
@Ezio-Sarthak I do enjoy this feature, so wanted to give some more focused feedback.

Our (event+actions)->message-view handling is something I can ideally see being refactored soon, but what you have here does seem to work fine for the most part :+1:

tests/ui/test_ui_tools.py
```diff
@@ -1786,9 +1787,11 @@ def test_msg_generates_search_and_header_bar(self, mocker,
...
                              ids=['this_starred', 'last_starred', 'no_stars'])
     @pytest.mark.parametrize('expected_header, to_vary_in_last_message', [
-        (['alice', ' ', 'DAYDATETIME'], {'sender_full_name': 'bob'}),
-        ([' ', ' ', 'DAYDATETIME'], {'timestamp': 1532103779}),
-        (['alice', ' ', 'DAYDATETIME'], {'timestamp': 0}),
+        ([STATUS_INACTIVE, 'alice', ' ', 'DAYDATETIME'],
+         {'sender_full_name': 'bob'}),
+        ([' ', ' ', ' ', 'DAYDATETIME'], {'timestamp': 1532103779}),
+        ([STATUS_INACTIVE, 'alice', ' ', 'DAYDATETIME'],
+         {'timestamp': 0}),
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yeah. Currently, in the PR, I've marked all users (including bots) as per their activity. Thus, a bot would currently be marked `inactive`.

Yes, my intent was to just amend the tests re header content to be shown (currently), not testing the activity marker rendering in particular.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1198,11 +1199,22 @@ def main_view(self) -> List[Any]:
...
             TextType = Dict[str, Tuple[Optional[str], str]]
-            text_keys = ('author', 'star', 'time')
+            text_keys = ('author', 'star', 'time', 'status')
             text: TextType = {key: (None, ' ') for key in text_keys}
 
             if any(different[key] for key in ('recipients', 'author', '24h')):
                 text['author'] = ('name', message['this']['author'])
+
+                email = self.message.get('sender_email', '')
+                user = self.model.user_dict.get(email, None)
```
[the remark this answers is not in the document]
Ezio-Sarthak: Sure. I'll add a TODO note :+1:

Ezio-Sarthak:

zulipterminal/helper.py
Ezio-Sarthak: Just to resolve this discussion, I've migrated the `STATE_ICON` to `ui_mappings.py` as added to main recently via #1013.

zulipterminal/ui_tools/views.py
```diff
@@ -530,6 +530,25 @@ def get_next_unread_pm(self) -> Optional[int]:
...
+            # Content header is not None
+            elif isinstance(msg_box[0][1], urwid.Text):
+                is_sender_header = (msg_box[0][1].text != ' ')
+
+            # Content header is None
+            else:
+                is_sender_header = False
+
+            if is_sender_header:
+                self.model._update_rendered_view(msg_box.message['id'])
```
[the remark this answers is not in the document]
Ezio-Sarthak: Separated message box digging in 3rd commit.
Proposed a better approach for rendering in the last commit

zulipterminal/ui_tools/boxes.py
preetmishra: Thanks for clarifying.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1361,6 +1361,26 @@ def main_view(self) -> List[Any]:
         ]
         return [part for part, condition in parts if condition]
 
+    def update_msg_content_header(self) -> bool:
```
preetmishra, line 1364: As you note in the commit (also in the next one), this is not a class method. Class methods are not bound to objects.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1361,6 +1361,26 @@ def main_view(self) -> List[Any]:
...
+        """
+        Update the content header of a message box (if required).
+        """
+        update_required = False
+
+        if self.need_recipient_header():
+            update_required = self[1][1].text != " "
```
preetmishra, line 1371: Could we store `self[1][1]` in a variable before so that it is more clear what is being checked?

zulipterminal/ui_tools/boxes.py
```diff
@@ -1361,6 +1361,26 @@ def main_view(self) -> List[Any]:
...
+
+        # Content header is not None
+        elif isinstance(self[0][1], urwid.Text):
```
preetmishra, line 1374: See my other comment re storing these values in a variable.

tests/ui/test_ui_tools.py
```diff
@@ -2211,6 +2211,54 @@ def test_main_view_generates_EDITED_label(
...
+        ],
+    )
+    def test_update_msg_content_header(
+        self,
+        msg_box,
+        message,
+        update_required,
+        mocker,
+        to_vary_in_each_message,
+        header_needed,
```
preetmishra, line 2254: `header_needed` and `mocker` are never used in the method body?

zulipterminal/ui_tools/views.py
```diff
@@ -547,6 +547,18 @@ def get_next_unread_pm(self) -> Optional[int]:
             return pm
         return None
 
+    def update_msg_list_presence_markers(self) -> None:
+        for index in range(len(self.body.log)):
```
preetmishra, line 551: Since you're using `self.body.log` more than once, you could extract it in a variable.

preetmishra (commented):
@Ezio-Sarthak Thanks for the update! This continues to work well. :+1: 

I have left some minor in-line comments. Other than which, this looks pretty close.

We could log issues for partial updates and also the bot presence marker depending upon whether you would take them next once the PR is merged.

zulipterminal/ui_tools/views.py
Ezio-Sarthak: Yes, though I think it would be not helpful for the line L558 (where we're updating `self.body.log` itself)?

Or do you mean updating the whole message list outside the loop in the end:
`self.body.log = msg_list` (if `msg_list` is the variable for instance) ?

preetmishra: I see that you're only mutating the attribute inside the loop, so extracting `self.body.log` would work.

preetmishra: (Assuming by helpful you meant it is not going to update the original log.)

preetmishra (commented):
@Ezio-Sarthak Thanks for incorporating the changes. This seems to work well. :+1:

@neiljp This looks good from my end.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1361,6 +1361,30 @@ def main_view(self) -> List[Any]:
...
+        # Update the content header only if it's present and non-empty.
+        if content_header is not None:
+            update_required = content_header.text != " "
+
+        if update_required:
+            # Re initialize the message if update is required.
+            # FIXME: Render specific element (here content header) instead?
+            super().__init__(self.main_view())
+
+        return update_required
```
neiljp, line 1386: Do we need this? Ah, you're using this in the test?

zulipterminal/ui_tools/boxes.py
```diff
@@ -1361,6 +1361,30 @@ def main_view(self) -> List[Any]:
         ]
         return [part for part, condition in parts if condition]
 
+    def update_msg_content_header(self) -> bool:
+        """
+        Update the content header of a message box (if required).
+        """
+        update_required = False
+        content_header = None
```
neiljp, line 1369: This isn't the content_header, but the author field? If so, let's name it that? Maybe set `author_column = 1` and use that in the indexing to make it clearer?

zulipterminal/ui_tools/boxes.py
```diff
@@ -1361,6 +1361,30 @@ def main_view(self) -> List[Any]:
...
+
+        # Get message content header (if present)
+        if self.need_recipient_header():
+            content_header = self[1][1]
+        elif isinstance(self[0][1], urwid.Text):
+            content_header = self[0][1]
+
+        # Update the content header only if it's present and non-empty.
+        if content_header is not None:
+            update_required = content_header.text != " "
```
neiljp, line 1379: It's not obvious where the `" "` test comes from - it's the default author text, when there's no author in the header, right? That's partly unclear due to the content_header variable name, but "non empty" and "content header" just repeat the code, so if we're going to include a comment, let's make it more useful - or set a variable like `author_is_present = ...` to help.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1361,6 +1361,30 @@ def main_view(self) -> List[Any]:
...
+
+        if update_required:
+            # Re initialize the message if update is required.
+            # FIXME: Render specific element (here content header) instead?
```
neiljp, line 1383: This doesn't seem too much of an issue performance-wise right now, but let's open an issue for improving this.

tests/ui/test_ui_tools.py
```diff
@@ -2211,6 +2211,52 @@ def test_main_view_generates_EDITED_label(
             assert label[0].text == "EDITED"
             assert label[1][1] == 7
 
+    @pytest.mark.parametrize(
+        "message",
```
neiljp, line 2215: This is specifically a stream message, though the behavior shouldn't vary across message types. Can we add a note here regarding that we should refactor this? We have other tests with a single parametrize entry, but it's a little misleading.

neiljp (commented):
@Ezio-Sarthak Thanks for the last update - just a few points inline left I hope :)

@preetmishra If you are satisfied with the points you raised in the review, can you resolve those?

preetmishra:
@neiljp I don't have the "resolve" button available on my end.

tests/ui/test_ui_tools.py
Ezio-Sarthak: Yeah, I used basically it as base template, which I'm modifying as per the `to_vary_in_each_message` parameters. Surely I'll add a note here :+1:.

zulipterminal/ui_tools/boxes.py
Ezio-Sarthak: Yes :sweat_smile:

zulipterminal/ui_tools/boxes.py
Ezio-Sarthak: Makes sense. I'll go with `author_is_present` and skip the comment :+1:

neiljp:
@preetmishra No "resolve conversation"? If so, the GitHub UI is even less effective than I thought in this regard :/

neiljp:
@Ezio-Sarthak Thanks for making the updates - this will be great to use! I'm going to merge shortly :tada:

Thoughts for some follow-ups that we should discuss and implement promptly or file as issues: (some we discussed already)
* Show bot markers (this was also suggested in my role-markers PR)
* Handle "unavailable" visibility and explicit changes (are they events?)
* Only update the status rather than the entire box - maybe not that tricky?
* Show user presence in PM compose?
* Show user presence in user mentions?

PR #990, opened 2021-04-09
Typing Indicator Animation

Rohitth007:
This commit animates the typing feedback using an iterable cycle
object. This is done using an asynchronous function
`_show_typing_animation` and a boolean `is_recipient_typing` to
start and end the animation.

The corner case of not recieving a `stop` signal is handled by
setting `TYPING_STARTED_EXPIRY_PERIOD=15` and checking this
condition using `datetime`. 

**This has to be handled differently from #915 as footer keep updating every half second 
so, footer text's `duration` variable can't be used.**

Tests added.

animate_typing

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -2535,6 +2535,8 @@ class TestModel:
         model.controller.is_typing_notification_in_progress = (
             is_notification_in_progress
         )
+        mocked_datetime = mocker.patch(MODULE + ".datetime")
+        mocked_datetime.now.return_value = "UNIXTIME"
         show_typing_notification = mocker.patch(
             CONTROLLER + ".show_typing_notification"
         )
@@ -2545,6 +2547,10 @@ class TestModel:
             assert (
                 model.controller.active_conversation_info["sender_name"]
                 == expected_active_conversation_info["sender_name"]
+            )
+            assert (
+                model.controller.active_conversation_info["typing_start_time"]
+                == "UNIXTIME"
             )
         assert show_typing_notification.called == expected_show_typing_notification
 
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -2552,6 +2552,7 @@ class TestModel:
                 model.controller.active_conversation_info["typing_start_time"]
                 == "UNIXTIME"
             )
+            assert model.controller.active_conversation_info["narrow"] == narrow
         assert show_typing_notification.called == expected_show_typing_notification
 
     @pytest.mark.parametrize(
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -5,6 +5,7 @@ import sys
 import time
 import webbrowser
 from collections import OrderedDict
+from datetime import datetime, timedelta
 from functools import partial
 from platform import platform
 from types import TracebackType
@@ -42,6 +43,8 @@ from zulipterminal.version import ZT_VERSION
 
 
 ExceptionInfo = Tuple[Type[BaseException], BaseException, TracebackType]
+
+TYPING_STARTED_EXPIRY_PERIOD = 15
 
 
 class Controller:
@@ -410,12 +413,27 @@ class Controller:
 
     @asynch
     def show_typing_notification(self) -> None:
+        # Restore the footer if there hasn't been a corresponding typing
+        # 'stop' event after a 'start' event for greater than
+        # TYPING_STARTED_EXPIRY_PERIOD. Ref: https://zulip.com/api/set-typing-status.
+        max_duration = timedelta(seconds=TYPING_STARTED_EXPIRY_PERIOD)
+        time_elapsed = timedelta(seconds=0)
+
         self.is_typing_notification_in_progress = True
+        active_conversation_info = self.active_conversation_info
         dots = itertools.cycle(["", ".", "..", "..."])
 
         # Until conversation becomes "inactive" like when a `stop` event is sent
-        while self.active_conversation_info:
-            sender_name = self.active_conversation_info["sender_name"]
+        while active_conversation_info:
+            sender_name = active_conversation_info["sender_name"]
+
+            # If `stop` is not sent for time `max_duration` after most recent update.
+            time_elapsed = (
+                datetime.now() - active_conversation_info["typing_start_time"]
+            )
+            if time_elapsed > max_duration:
+                self.active_conversation_info = {}
+
             self.view.set_footer_text(
                 [
                     ("footer_contrast", " " + sender_name + " "),
@@ -423,6 +441,7 @@ class Controller:
                 ]
             )
             time.sleep(0.45)
+            active_conversation_info = self.active_conversation_info
 
         self.is_typing_notification_in_progress = False
         self.view.set_footer_text()
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -434,6 +434,10 @@ class Controller:
             if time_elapsed > max_duration:
                 self.active_conversation_info = {}
 
+            # If user changes narrow
+            if self.model.narrow != active_conversation_info["narrow"]:
+                self.active_conversation_info = {}
+
             self.view.set_footer_text(
                 [
                     ("footer_contrast", " " + sender_name + " "),
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -1197,6 +1197,7 @@ class Model:
             if event["op"] == "start":
                 sender_name = self.user_dict[sender_email]["full_name"]
                 active_conversation_info["sender_name"] = sender_name
+                active_conversation_info["typing_start_time"] = datetime.now()
 
                 if not controller.is_typing_notification_in_progress:
                     controller.show_typing_notification()
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -1198,6 +1198,7 @@ class Model:
                 sender_name = self.user_dict[sender_email]["full_name"]
                 active_conversation_info["sender_name"] = sender_name
                 active_conversation_info["typing_start_time"] = datetime.now()
+                active_conversation_info["narrow"] = narrow
 
                 if not controller.is_typing_notification_in_progress:
                     controller.show_typing_notification()
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -1,5 +1,6 @@
 import os
 import webbrowser
+from datetime import datetime, timedelta
 from platform import platform
 from threading import Thread, Timer
 from typing import Any, Dict, List, Optional, Set, Tuple
@@ -548,7 +549,13 @@ class TestController:
     @pytest.mark.parametrize(
         "active_conversation_info",
         [
-            case({"sender_name": "hamlet"}, id="in_pm_narrow_with_sender_typing:start"),
+            case(
+                {
+                    "sender_name": "hamlet",
+                    "typing_start_time": datetime.now() + timedelta(1),
+                },
+                id="in_pm_narrow_with_sender_typing:start",
+            ),
             case({}, id="in_pm_narrow_with_sender_typing:stop"),
         ],
     )
@@ -566,7 +573,9 @@ class TestController:
             controller.active_conversation_info = {}
 
         Timer(0.1, mock_typing).start()
-        Thread(controller.show_typing_notification()).start()
+        thread = Thread(target=controller.show_typing_notification)
+        thread.start()
+        thread.join()
 
         if active_conversation_info:
             set_footer_text.assert_has_calls(
@@ -581,4 +590,19 @@ class TestController:
         else:
             set_footer_text.assert_called_once_with()
         assert controller.is_typing_notification_in_progress is False
+        assert controller.active_conversation_info == {}
+
+    def test_end_typing_notification_on_unresponsive_event(
+        self,
+        mocker: MockerFixture,
+        controller: Controller,
+    ) -> None:
+        set_footer_text = mocker.patch(VIEW + ".set_footer_text")
+        controller.active_conversation_info = {"sender_name": "hamlet"}
+        controller.active_conversation_info[
+            "typing_start_time"
+        ] = datetime.now() - timedelta(seconds=14.5)
+
+        controller.show_typing_notification()
+
         assert controller.active_conversation_info == {}
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -553,6 +553,7 @@ class TestController:
                 {
                     "sender_name": "hamlet",
                     "typing_start_time": datetime.now() + timedelta(1),
+                    "narrow": ["pm"],
                 },
                 id="in_pm_narrow_with_sender_typing:start",
             ),
@@ -567,6 +568,7 @@ class TestController:
     ) -> None:
         set_footer_text = mocker.patch(VIEW + ".set_footer_text")
         sleep = mocker.patch(MODULE + ".time.sleep")
+        controller.model.narrow = ["pm"]
         controller.active_conversation_info = active_conversation_info
 
         def mock_typing() -> None:
@@ -598,10 +600,29 @@ class TestController:
         controller: Controller,
     ) -> None:
         set_footer_text = mocker.patch(VIEW + ".set_footer_text")
-        controller.active_conversation_info = {"sender_name": "hamlet"}
+        controller.model.narrow = ["pm"]
+        controller.active_conversation_info = {
+            "sender_name": "hamlet",
+            "narrow": controller.model.narrow,
+        }
         controller.active_conversation_info[
             "typing_start_time"
         ] = datetime.now() - timedelta(seconds=14.5)
+
+        controller.show_typing_notification()
+
+        assert controller.active_conversation_info == {}
+
+    def test_end_typing_notification_on_narrow_change(
+        self, mocker: MockerFixture, controller: Controller
+    ) -> None:
+        set_footer_text = mocker.patch(VIEW + ".set_footer_text")
+        controller.model.narrow = [["pm_with", "othello@zulip.com"]]
+        controller.active_conversation_info = {
+            "sender_name": "hamlet",
+            "narrow": [["pm_with", "hamlet@zulip.com"]],
+        }
+        controller.active_conversation_info["typing_start_time"] = datetime.now()
 
         controller.show_typing_notification()
 


zulipterminal/model.py
```diff
@@ -957,12 +973,10 @@ def _handle_typing_event(self, event: Event) -> None:
                     and event['sender']['email'] in narrow[0][1].split(',')):
                 if event['op'] == 'start':
                     user = self.user_dict[event['sender']['email']]
-                    self.controller.view.set_footer_text([
-                        ' ',
-                        ('code', user['full_name']),
-                        ' is typing...'
-                    ])
+                    self.is_typing = True
+                    self.animate_typing(user)
```
preetmishra, line 977: I am not sure about the name `user` here. This person is actually a recipient and the ZT user is the _user_.

preetmishra (changes requested):
@Rohitth007 Thanks for putting this forward! While this is a small addition, this does seem to improve the experience.

I have tested this locally and it seems to work well. :+1: 

Other than the in-line comments, there are a few things that be improved:
- We follow a particular convention for the commit messages. You can read the commit guidelines that we have in the README and also `$ git log --oneline`. I would recommend to you install `gitlint`.
- We would appreciate some tests to go along with the feature. Tests give us the ability to ship new features confidently.
- We use `mypy` for static type checking. Please add type annotation wherever needed.

Feel free to drop a message in **#zulip-terminal** if you need any assistance with anything.

zulipterminal/model.py
```diff
@@ -945,6 +945,22 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int
                         for user_id in user_ids:
                             subscribers.remove(user_id)
 
+    @asynch
+    def animate_typing(self, user):
+        def typing_gen():
+            while True:
+                yield from ['', '.', '..', '...', '....']
```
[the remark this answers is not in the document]
Rohitth007: Yes, even I put 3 initially but I added another just to see how it looks. Maybe 3 is good enough.

zulipterminal/model.py
```diff
@@ -945,6 +945,22 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int
                         for user_id in user_ids:
                             subscribers.remove(user_id)
 
+    @asynch
+    def animate_typing(self, user):
+        def typing_gen():
```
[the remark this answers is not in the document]
Rohitth007: I will look into this :+1:

zulipterminal/model.py
Rohitth007: This was how it originally was, though I think `sender` maybe a better word here?

zulipterminal/model.py
```diff
@@ -957,12 +973,10 @@ def _handle_typing_event(self, event: Event) -> None:
...
                     user = self.user_dict[event['sender']['email']]
-                    self.controller.view.set_footer_text([
-                        ' ',
-                        ('code', user['full_name']),
-                        ' is typing...'
-                    ])
+                    self.is_typing = True
+                    self.animate_typing(user)
                 elif event['op'] == 'stop':
+                    self.is_typing = False
```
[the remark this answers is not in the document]
Rohitth007: How about `sender_typing_status`?

Rohitth007:
@preetmishra Yes, I am aware of the things specified that I have not added yet. This is just a functional prototype and since it's small enough to understand, I was planning to write a proper commit structure after making those changes. I am in between my exams right now hence the not so perfect PR. Hope you understand :)

zulipterminal/model.py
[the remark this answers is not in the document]
preetmishra: Suffixes like `status` convey that there are more than _two_ states that a variable can take. For booleans, it is reasonable to have prefixes like `is`, `has`, `should`. Perhaps, `is_recipient_typing`?

zulipterminal/model.py
preetmishra: Hmm, `sender` would be a bit ambiguous as in this case both, the user and the recipient, could be perceived as a sender. I believe `recipient` would be apt.

preetmishra:
@Rohitth007 No worries. Best of luck with the exams!

zulipterminal/model.py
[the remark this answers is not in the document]
Rohitth007: Maybe `show_typing_animation` because there should be some way to indicate that it not just a normal screen update?

Rohitth007:
@preetmishra PR updated for review

zulipterminal/model.py
```diff
@@ -945,24 +947,44 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int
                         for user_id in user_ids:
                             subscribers.remove(user_id)
 
+    @asynch
+    def _show_typing_animation(self, recipient: Dict[str, Any]) -> None:
+        start_time = datetime.now()
+        # Restore the footer if there hasn't been a corresponding typing
+        # 'stop' event after a 'start' event for greater than 15 seconds.
+        max_duration = timedelta(seconds=TYPING_STARTED_EXPIRY_PERIOD)
```
preetmishra, line 955: Has this constant (15 seconds) been borrowed from somewhere?

zulipterminal/model.py
```diff
@@ -945,24 +947,44 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int
...
+
+        for dots in itertools.cycle(['', '.', '..', '...']):
```
preetmishra, line 957: Could we use the `next` function with a `lambda` to iterate over the dots? That'd be a bit cleaner.

zulipterminal/model.py
```diff
@@ -945,24 +947,44 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int
...
+            if not self.is_recipient_typing:
+                break
+            elif datetime.now() - start_time > max_duration:
```
preetmishra, line 960: This means that even if the recipient is still typing post 15 seconds, the user won't get any feedback, right?

preetmishra (commented):
@Rohitth007 Hey, thanks for your work! I have given quick feedback on the approach. Let's discuss it before moving ahead for another review iteration.

zulipterminal/model.py
Rohitth007: Yes this is the same constant used in @prah23's #915 where I have borrowed it from and also the same is recommended in the API docs for typing events given [here](https://zulip.com/api/set-typing-status).

zulipterminal/model.py
Rohitth007: This means if the recipient hasn't sent a `stop` (which the senders client regularly sends if the user is idle or escaped out of compose box) for some reason like client crash etc, ZT keeps showing `typing...` .
If the recipient still is typing its the clients responsibility to tell ZT `start` and it will. So, ZT re-displays `typing...`

Rohitth007: Though I see now that I haven't reset `is_recipient_typing` to `False`. I think this doesn't change the logic but should I reset anyway?

zulipterminal/model.py
Rohitth007: I don't quite get this, next can be used only when a generator object is created right? Is lambda used to create this object? That would leads to more lines of code for getting the same functionality?

zulipterminal/model.py
preetmishra: My concern is that we do not check whether we have a new `op=start` and just break regardlessly post `TYPING_STARTED_EXPIRY_PERIOD`. What would happen if we have another `op=start` event post `TYPING_STARTED_WAIT_PERIOD`?

zulipterminal/model.py
preetmishra: Agreed; it would require more lines. However, it would facilitate is the separation of concerns. @neiljp Thoughts on this?

zulipterminal/model.py
Rohitth007: oooh! I haven't considered the case where a `start` is received while the async function is running resulting in 2 of them running parallely. Maybe a boolean to fix this?
And when a new `start` is received after 15 seconds this process can restart. So I don't see an issue while breaking after 15 sec, did I miss something?

Rohitth007: I have used the boolean `is_typing_animation_in_progress` to handle the case where 2 of them run in parallel.
It is a bit long but I didn't want to put `animation_in_progress` or `is_animating` as it doesn't tell us about what is animating.
Even `typing_in_progress` can seem similar to `is_recipient_typing`?

Rohitth007:
@preetmishra I have made a few more changes to handle corner cases better. Even the tests work better now. (I have moved time forward instead of reducing it)
It all works but PEP8 tells me that there is a trailing whitespace error on a line I haven't even touched.
I even tried using VSCode's `Trim Trailing Whitespace` command.
`zulipterminal/model.py:960:68: W291 trailing whitespace`

Rohitth007:

zee-bit:
> It all works but PEP8 tells me that there is a trailing whitespace error on a line I haven't even touched.

You did change that line. You can take a look at the `Files changed` tab of this commit and notice that line `960` is changed. You may be in a subsequent commit while running the linter and thought that you didn't change that line. :sweat_smile:
(That's why it's important to run tests for each commit so that you find your mistakes in that commit itself.)

**PRO-TIP:** `git blame` can prove really helpful in finding out who changed a particular line in a repository.

Rohitth007:
Oh wait I was checking in a different file :sweat_smile: , I mistook `model.py` for `test_model.py` cause that was the only file I was editing at that time. :facepalm:

Ezio-Sarthak (commented):
Thanks for continuing work on this! Leaving some minor code-wise comments :+1:

The commit title could be improved a bit, something more specific?

zulipterminal/model.py
neiljp: @preetmishra I'd prefer to see the comparison before judging.

zulipterminal/model.py
```diff
@@ -1235,7 +1265,7 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
     def formatted_local_time(
         self, timestamp: int, *, show_seconds: bool, show_year: bool = False
     ) -> str:
-        local_time = datetime.datetime.fromtimestamp(timestamp)
+        local_time = datetime.fromtimestamp(timestamp)
```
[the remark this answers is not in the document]
Rohitth007: Done this :+1:, makes sense to have it as a prep commit.

tests/model/test_model.py
```diff
@@ -1660,83 +1661,139 @@ def test_toggle_stream_pinned_status(self, mocker, model,
...
-                    'email': 'iago@zulip.com'
-                }],
-                'id': 0
-            },
-            True,
-        )
-    ], ids=['not_in_pm_narrow', 'not_in_pm_narrow_with_sender',
-            'start', 'stop'])
+    @pytest.mark.parametrize(
+        'narrow, event, animation_in_progress, expect_recipient_typing', [
```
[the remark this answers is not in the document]
Rohitth007: It should be interpreted as so:
```python
If "animation is in progress" and we "expect that the recipient is typing" then assert that this happens.
```
We don't expect animation status, maybe `is_animation_in_progress` is better?

Re `expected_recipient_typing_status` can be used if `recipient_typing_status` is the variable output of the function. Also from above discussion with Preet, 'status' implied more than 2 values and not a boolean.

I wanted to use `is_recipient_typing` but its the same name as boolean. How about `is_recipent_actually_typing`?
It would then be interpreted as:
```python
If "animation is in progress" and "recipient is actually typing" then assert that animation happens, etc...
```

zulipterminal/model.py
Rohitth007: @preetmishra If you could look at this particular discussion again I would mark it as resolved.

zulipterminal/model.py
Rohitth007: @preetmishra I've re-implemented this part with minor changes. I've not used lambda as I was getting an error but I think this is pretty much what you had in mind. It looks much more better and readable now though its the same number of lines because of the if statement coming out of the loop.

tests/model/test_model.py
```diff
@@ -2033,10 +2034,10 @@ def test_toggle_stream_pinned_status(
         model.client.update_subscription_settings.assert_called_once_with(request)
 
     @pytest.mark.parametrize(
-        "narrow, event, called",
+        "narrow, event, is_animation_in_progress, is_recipient_actually_typing",
```
preetmishra, line 2037: While you're at it, let's use cases (`from pytest import param as case`), so that it's evident that which test case pairs with which test id.

tests/model/test_model.py
```diff
@@ -2079,20 +2098,73 @@ def test_toggle_stream_pinned_status(
...
                 True,
+                False,
             ),
         ],
-        ids=["not_in_pm_narrow", "not_in_pm_narrow_with_sender", "start", "stop"],
+        ids=[
+            "not_in_pm_narrow",
+            "not_in_pm_narrow_with_sender",
+            "start",
+            "start when animating",
```
preetmishra, line 2108: Let's use underscores as delimiters.

tests/model/test_model.py
```diff
@@ -2079,20 +2098,73 @@ def test_toggle_stream_pinned_status(
...
     )
-    def test__handle_typing_event(self, mocker, model, narrow, event, called):
+    def test__handle_typing_event(
+        self,
+        mocker,
+        model,
+        narrow,
+        event,
+        is_animation_in_progress,
+        is_recipient_actually_typing,
```
preetmishra, line 2119: Let's use an `expected_` as prefix for params that are expected from the test result, so `is_recipient_actually_typing` -> `expected_is_recipient_typing`.

tests/model/test_model.py
```diff
@@ -2079,20 +2098,73 @@ def test_toggle_stream_pinned_status(
...
         model.user_dict = {"hamlet@zulip.com": {"full_name": "hamlet"}}
+        model.is_recipient_typing = False
+        model.is_typing_animation_in_progress = is_animation_in_progress
 
         model._handle_typing_event(event)
 
-        assert model.controller.view.set_footer_text.called == called
+        assert model.is_recipient_typing == is_recipient_actually_typing
+        assert model._show_typing_animation.called == (
+            is_recipient_actually_typing and not is_animation_in_progress
```
preetmishra, line 2133: It'd be cleaner to parametrize over `_show_typing_animation_is_called` than derive it using these two variables?

tests/model/test_model.py
```diff
@@ -2079,20 +2098,73 @@ def test_toggle_stream_pinned_status(
...
+                False,
+            ),
+        ],
+        ids=["start", "stop"],
+    )
+    def test__show_typing_animation(self, mocker, model, event, expected_indicator):
+        mocker.patch("zulipterminal.ui.View.set_footer_text")
+        recipient = {"full_name": "hamlet"}
+        model.is_recipient_typing = event["op"] == "start"
+        model.last_update_time = datetime.now() - timedelta(seconds=14.5)
```
preetmishra, line 2156: Could we have a test for `last_updated_time` > 15?

preetmishra (commented):
@Rohitth007 Thanks for the improvements! This seems to work well locally. :+1:

The implementation looks good! I have left a few in-line comments. We could have another test for `last_updated_time` case.

zulipterminal/model.py
```diff
@@ -1,9 +1,9 @@
-import datetime
 import json
 import time
 from collections import OrderedDict, defaultdict
 from concurrent.futures import Future, ThreadPoolExecutor, wait
 from copy import deepcopy
+from datetime import datetime
```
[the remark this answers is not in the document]
Rohitth007: :+1: I made some changes here but is the part where I mention "this is a prep commit" fine?

tests/model/test_model.py
```diff
@@ -1,6 +1,7 @@
 import json
 from collections import OrderedDict
 from copy import deepcopy
+from datetime import datetime, timedelta
```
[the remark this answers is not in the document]
Rohitth007: I think the title is fine right? `ui` is the `area` and `model` and `test_model` are files changed.

zulipterminal/model.py
```diff
@@ -993,11 +997,39 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int) -> StreamData:
...
+    @asynch
+    def _show_typing_animation(self, recipient: Dict[str, Any]) -> None:
+        # Restore the footer if there hasn't been a corresponding typing
+        # 'stop' event after a 'start' event for greater than 15 seconds.
+        max_duration = timedelta(seconds=TYPING_STARTED_EXPIRY_PERIOD)
+
+        self.is_typing_animation_in_progress = True
+        dots = itertools.cycle(["   ", ".  ", ".. ", "..."])
+
+        # Untill `stop` signal is sent or
```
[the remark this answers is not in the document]
Rohitth007: :sweat_smile: I'll fix it

zulipterminal/model.py
```diff
@@ -1007,12 +1039,13 @@ def _handle_typing_event(self, event: Event) -> None:
                 and event["sender"]["email"] in narrow[0][1].split(",")
             ):
                 if event["op"] == "start":
-                    user = self.user_dict[event["sender"]["email"]]
-                    self.controller.view.set_footer_text(
-                        [" ", ("footer_contrast", user["full_name"]), " is typing..."]
-                    )
+                    recipient = self.user_dict[event["sender"]["email"]]
+                    self.is_recipient_typing = True
+                    self.last_update_time: datetime = datetime.now()
```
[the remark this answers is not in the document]
Rohitth007: Is a name change to `typing_event_start_time` more clear instead of commenting?

tests/model/test_model.py
Rohitth007: :+1:

tests/model/test_model.py
Rohitth007: :+1: Might have missed it

tests/model/test_model.py
Rohitth007: :+1:

tests/model/test_model.py
Rohitth007: I think it would become too many variables to read and too many True's and False's. I've tried extracting the inside expression as another boolean variable(`expect_indicator`) inside the function and it looks better.

tests/model/test_model.py
Rohitth007: I think this is already tested as the async function breaks when last_update_time crosses 15

Rohitth007: And also when `is_recipient_typing` is False it never enters, so breaking the loop because of `is_recipient_typing` is also tested.

Rohitth007:
@preetmishra This is ready for another review.

tests/model/test_model.py
```diff
@@ -2079,20 +2095,66 @@ def test_toggle_stream_pinned_status(
...
+    ):
         event["type"] = "typing"
 
-        mocker.patch("zulipterminal.ui.View.set_footer_text")
+        mocker.patch("zulipterminal.model.Model._show_typing_animation")
         model.narrow = narrow
         model.user_dict = {"hamlet@zulip.com": {"full_name": "hamlet"}}
+        model.is_recipient_typing = False
+        model.is_typing_animation_in_progress = is_animation_in_progress
+        expect_indicator = expected_is_recipient_typing and not is_animation_in_progress
```
preetmishra, line 2119: It would be more explicit and readable if we put it as a parameter in the parametrization. Also, let's include `_show_typing_animation` in the name as with this we're essentially testing whether `_show_typing_animation` was called.

tests/model/test_model.py
```diff
@@ -2079,20 +2095,66 @@ def test_toggle_stream_pinned_status(
...
 
         model._handle_typing_event(event)
 
-        assert model.controller.view.set_footer_text.called == called
+        assert model.is_recipient_typing == expected_is_recipient_typing
+        assert model._show_typing_animation.called == expect_indicator
+
+    @pytest.mark.parametrize(
+        "event, expected_indicator",
```
preetmishra, line 2127: `indicator` doesn't provide any context. Let's refer to it with the method name (see my other comment).

tests/model/test_model.py
```diff
@@ -2079,20 +2095,66 @@ def test_toggle_stream_pinned_status(
...
+                {"op": "stop"},
+                False,
+            ),
+        ],
+        ids=["start", "stop"],
+    )
+    def test__show_typing_animation(self, mocker, model, event, expected_indicator):
+        mocker.patch("zulipterminal.ui.View.set_footer_text")
+        recipient = {"full_name": "hamlet"}
+        model.is_recipient_typing = event["op"] == "start"
```
preetmishra, line 2145: This is another candidate for a `param` in the parametrization block.

preetmishra (commented):
@Rohitth007 Thanks for persisting. This continues to work well and the PR looks good overall! :+1: I have left a few minor in-line comments.

You seem to have updated `last_update_time` in the code but it is still there in the second commit's description.

zulipterminal/model.py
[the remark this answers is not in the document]
preetmishra: Yes! We should always improve the name instead of leaving comments in any case.

tests/model/test_model.py
preetmishra: The former shouldn't be an issue. I would avoid writing expressions/evaluations in the test code and focus on just testing the method with a bunch of inputs.

tests/model/test_model.py
preetmishra: Right, I think this is fine for now.

zulipterminal/model.py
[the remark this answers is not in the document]
Rohitth007: The problem with refactoring `test_ui_tools` is that `pytest` and `mypy` would get confused between the mocked `date` and the actual `date`. So its better to keep it as `datetime.date`

Rohitth007:
@preetmishra This is ready for another review. I have made all the suggested changes except the datetime refactor for
test_ui_tools. Hopefully the rest of it is looking good?

zulipterminal/model.py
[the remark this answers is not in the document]
preetmishra: I don't think it should be not an issue if we get the naming correct.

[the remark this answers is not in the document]
preetmishra: It's just that we should be consistent with the imports or other conventions throughout the source. :)

tests/model/test_model.py
```diff
@@ -2033,12 +2034,13 @@ def test_toggle_stream_pinned_status(
         model.client.update_subscription_settings.assert_called_once_with(request)
 
     @pytest.mark.parametrize(
-        "narrow, event, called",
+        (
```
preetmishra, line 2037: Let's update `last_update_time` to `typing_event_start_time` in the commit description as I mentioned in the last review. :)

tests/model/test_model.py
```diff
@@ -2079,20 +2101,60 @@ def test_toggle_stream_pinned_status(
...
+        model.is_typing_animation_in_progress = is_animation_in_progress
 
         model._handle_typing_event(event)
 
-        assert model.controller.view.set_footer_text.called == called
+        assert model.is_recipient_typing == expected_is_recipient_typing
+        assert model._show_typing_animation.called == expected__show_typing_animation
+
+    @pytest.mark.parametrize(
+        "event, expect_set_footer_animation",
```
preetmishra, line 2134: minor: expected*

preetmishra (commented):
@Rohitth007 Thanks for the update! It continues to work well. :+1:

@neiljp Other than the minor changes that I have left in-line, this looks good from my end.

tests/model/test_model.py
Rohitth007: I think expect is a better word here. Expected would we apt when we compare a variable with its expected value, but here every possibility makes a call to set_footer_text so I've used the above which would translate to - `if you expect footer animation to be set then do so and so tests else do something else`

Rohitth007: Maybe `expect_footer_animation` would be better?

zulipterminal/model.py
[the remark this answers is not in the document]
Rohitth007: @preetmishra I was getting errors as it was using the patched `date` as the imported `date`.
I think you meant "don't think it should be an issue" or otherwise?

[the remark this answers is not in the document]
preetmishra: You could change the variable name to differentiate between the two.

Yeah, that's what I meant. :) (I have amended my comment.)

tests/model/test_model.py
preetmishra: Alternatively, you could call it `is_set_footer_animation_called`?

Rohitth007:
@preetmishra I have updated the PR with all the suggestions. Hopefully this is now complete.

zulipterminal/model.py
```diff
@@ -996,11 +1000,39 @@ def get_stream_by_id(streams: List[StreamData], stream_id: int) -> StreamData:
                         for user_id in user_ids:
                             subscribers.remove(user_id)
 
+    @asynch
+    def _show_typing_animation(self, recipient: Dict[str, Any]) -> None:
```
neiljp, line 1004: This isn't just showing a typing animation, it's now the entire typing notification function.

neiljp (commented):
@Rohitth007 This seems to work well :+1:

Some thoughts:
* Does this incidentally resolve one issue which @prah23 was trying to fix in #915? (maybe we can all discuss) 
* Having the typing notification behavior extracted into a function is great - and necessary for the asynch stateful behavior of the animation :+1: (though as noted, this is not just animation handled there, but all the notification(s)!)
* While we have other code calling the controller/view from the model (including the current notification code), I'd prefer to move to a situation where we have that behavior wrapped in eg. the controller (maybe including the view?) - consider eg. the report_* functions.
* Related to that, the extra state you're storing in the model is really UI state - it feels strange having storage of typing animation state being in the model?
* Do we need to provide the full recipient data each time, or just the name?

Other than helping us move to a cleaner structure as we add new features, I'm hoping doing that with this PR will be helpful for handling multi-user typing and stream-typing events. Let me know if you want more specific pointers - we can certainly discuss :+1:

zulipterminal/model.py
Rohitth007: Yes one could say that, but it is mainly updating the animation "frame" by "frame" and deciding when to end it.
If future animations are done then  `create_animation` or similar function could be extracted out of this. For now I think calling it show_typing_notification and moving it to Controller would be better.

Rohitth007:
@neiljp Thanks for the review. To address some of your points:
* Yes this also happens to resolve the same edge case as @prah23's #915 , where the basic idea was borrowed from, as it had to be handled differently as this is asynchronous.
* Moving this whole function into controller makes sense and would simplify work on typing notifications features later.
* The full recipient was not needed, it was used it the same way before hence I kept it. Maybe just sending the name is better.

tests/model/test_model.py
```diff
@@ -2079,20 +2101,60 @@ def test_toggle_stream_pinned_status(
...
+        narrow,
+        event,
+        is_animation_in_progress,
+        expected_is_recipient_typing,
+        expected__show_typing_animation,
+    ):
         event["type"] = "typing"
 
-        mocker.patch("zulipterminal.ui.View.set_footer_text")
+        mocker.patch("zulipterminal.model.Model._show_typing_animation")
```
prah23, line 2122: You can just use `MODEL + "._show_typing_animation"` here. The `MODEL` constant is defined at the top of the file as `zulipterminal.model.Model`. This will keep your patch consistent with the other tests in the file.

tests/model/test_model.py
```diff
@@ -2079,20 +2101,60 @@ def test_toggle_stream_pinned_status(
...
+        "event, is_set_footer_animation_called",
+        [
+            case({"op": "start"}, True, id="in_pm_narrow_with_sender_op_start"),
+            case({"op": "stop"}, False, id="op-stop"),
+        ],
+    )
+    def test__show_typing_animation(
+        self, mocker, model, event, is_set_footer_animation_called
+    ):
+        mocker.patch("zulipterminal.ui.View.set_footer_text")
```
prah23, line 2143: Echoing from the previous constant point, you can add a constant for `zulipterminal.ui.View` at the top of the file and use that here.

tests/model/test_model.py
```diff
@@ -2079,20 +2101,60 @@ def test_toggle_stream_pinned_status(
...
+        recipient = {"full_name": "hamlet"}
+        model.is_recipient_typing = event["op"] == "start"
+        model.typing_event_start_time = datetime.now() - timedelta(seconds=14.5)
+
+        model._show_typing_animation(recipient)
+
+        if is_set_footer_animation_called:
+            assert model.controller.view.set_footer_text.called is True
```
prah23, line 2151: I understand the general benefit of erring towards being verbose, but I think this statement will maintain readability even without the `is True` part of it. We can just
```python
assert that something is called
```
instead of 
```python
assert that something is called is True
```
The other benefit is consistency with the `assert`s in the other tests.

tests/model/test_model.py
```diff
@@ -2079,20 +2101,60 @@ def test_toggle_stream_pinned_status(
...
+            assert model.controller.view.set_footer_text.called_with()
+        else:
+            assert model.controller.view.set_footer_text.called_once_with()
```
prah23, line 2154: This can instead use `assert_called_once_with()`? The assert a couple of lines above can also use `assert_called_with()`. The benefit, again, lies in consistency with the other tests.

tests/model/test_model.py
```diff
@@ -2079,20 +2101,60 @@ def test_toggle_stream_pinned_status(
...
+        mocker.patch("zulipterminal.ui.View.set_footer_text")
+        recipient = {"full_name": "hamlet"}
+        model.is_recipient_typing = event["op"] == "start"
+        model.typing_event_start_time = datetime.now() - timedelta(seconds=14.5)
+
+        model._show_typing_animation(recipient)
+
+        if is_set_footer_animation_called:
+            assert model.controller.view.set_footer_text.called is True
+            assert model.controller.view.set_footer_text.called_with()
```
prah23, line 2152: Can you also assert on what text the function is being called with? I understand that the dots might be tricky to assert on, but you could make use of `assert_has_calls`. If not with the dots, an assert with the base text?

prah23 (changes requested):
Thanks for all the work @Rohitth007!
This PR looks good to me, and the expiry period condition seems to have been met. I have a few suggestions in terms of the tests that I'm listing in-line.

zulipterminal/model.py
neiljp: Your suggestion was along the lines of what I was meaning :+1:

tests/model/test_model.py
Rohitth007: Thanks for this, it worked wonderfully

Rohitth007:
I have updated this PR as per suggestions and also have improved the tests.

Rohitth007:
Updated the PR as per discussions in CZO

neiljp (commented):
@Rohitth007 Thanks for the split - this becomes clearer now :+1:

Some of my inline feedback reflects some points I made previously, including regarding storing state in an appropriate namespace.

zulipterminal/model.py
```diff
@@ -1130,12 +1130,14 @@ def _handle_typing_event(self, event: Event) -> None:
             and sender_id != self.user_id
         ):
             if event["op"] == "start":
-                sender_name = self.user_dict[sender_email]["full_name"]
-                controller.view.set_footer_text(
-                    [" ", ("footer_contrast", sender_name), " is typing..."]
-                )
+                self.sender_name = self.user_dict[sender_email]["full_name"]
+                self.is_sender_typing = True
+                if not controller.is_typing_animation_in_progress:
```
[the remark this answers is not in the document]
Rohitth007: Yes, makes sense I'll change it.

zulipterminal/core.py
```diff
@@ -384,6 +387,25 @@ def open_in_browser(self, url: str) -> None:
...
+    def show_typing_notification(self) -> None:
+        self.is_typing_animation_in_progress = True
+        dots = itertools.cycle(["", ".", "..", "..."])
+
+        # Until `stop` signal is sent
+        while self.model.is_sender_typing:
+            self.view.set_footer_text(
+                [
+                    ("footer_contrast", " " + self.model.sender_name + " "),
+                    " is typing" + next(dots),
```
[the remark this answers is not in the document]
Rohitth007: Making this as a refactor then animation commit makes sense but the problem is I cannot test only animation without a stop mode if the threads execute linearly

tests/model/test_model.py
```diff
@@ -2304,21 +2341,44 @@ def test_toggle_stream_visual_notifications(
...
+        expected_is_sender_typing,
+        expected_show_typing_notification,
+    ):
         event["type"] = "typing"
 
-        mocker.patch("zulipterminal.ui.View.set_footer_text")
         model.narrow = narrow
         model.user_dict = {"hamlet@zulip.com": {"full_name": "hamlet"}}
         model.user_id = 5  # Iago's user_id
+        model.is_sender_typing = False
```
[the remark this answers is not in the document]
Rohitth007: It does because it is outside the if conditional.

zulipterminal/model.py
```diff
@@ -1130,12 +1130,14 @@ def _handle_typing_event(self, event: Event) -> None:
...
+                    controller.show_typing_notification()
+
             elif event["op"] == "stop":
-                controller.view.set_footer_text()
+                self.is_sender_typing = False
```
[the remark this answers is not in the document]
Rohitth007: I've used `active_conversation_info` dict for this which also makes it easier to implement group typing notification

neiljp:
@Rohitth007 Thanks for working on this different architecture - the first two commits of this are now merged :tada: 
(since the tests seem OK that far through)

I split the first commit to make it smaller and easier to read.

PR #698, opened 2020-06-19
[WIP] Add support for viewing a message in the web browser

preetmishra:
Thanks to @punchagan for the initial work in #397. :+1: 

Unlike the original PR, which narrowed every message to All Messages in the web app, this opens a message in `stream/topic/near` if it belongs to a stream or `pm-with/near` narrow if it belongs to a PM/huddle (see [discussion](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/View.20message.20in.20browser/near/908851)).

The original commit has been split to facilitate reviews. Moreover, the test has been improved based on @neiljp's feedback on the original PR.

The logic for generating near message URLs has been borrowed from Zulip's [`zerver/lib/url_encoding.py`](https://github.com/zulip/zulip/blob/master/zerver/lib/url_encoding.py).

Partially fixes #452.

PR #991, opened 2021-04-09
Add support for opening message and external links in browser.

zee-bit:
Thanks to @punchagan and @preetmishra for their major initial work in #397 and #698 respectively. :+1:

This PR extracts some initial commits from both PR's and cleans and refactors them. It also adds support for opening external links in the browser from the `MessageLinkButton`. All the reviews on the previous PR's have also been addressed, facilitating further reviews.

Fixes last point of #452.

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -1,3 +1,5 @@
+import os
+import webbrowser
 from platform import platform
 from typing import Any
 
@@ -264,6 +266,38 @@ class TestController:
         widgets = controller.view.message_view.log.extend.call_args_list[0][0][0]
         msg_ids = {widget.original_widget.message["id"] for widget in widgets}
         assert msg_ids == id_list
+
+    @pytest.mark.parametrize(
+        "url",
+        [
+            "https://chat.zulip.org/#narrow/stream/test",
+            "https://chat.zulip.org/user_uploads/sent/abcd/efg.png",
+            "https://github.com/",
+        ],
+    )
+    def test_open_in_browser_success(self, mocker, controller, url):
+        # Set DISPLAY environ to be able to run test in CI
+        os.environ["DISPLAY"] = ":0"
+        controller.view.set_footer_text = mocker.Mock()
+        mock_get = mocker.patch(CORE + ".webbrowser.get")
+        mock_open = mock_get.return_value.open
+
+        controller.open_in_browser(url)
+
+        mock_open.assert_called_once_with(url)
+        controller.view.set_footer_text.assert_called_once_with(
+            f"The link was successfully opened using {mock_get.return_value.name}", 3
+        )
+
+    def test_open_in_browser_fail__no_browser_controller(self, mocker, controller):
+        os.environ["DISPLAY"] = ":0"
+        error = "No runnable browser found"
+        controller.view.set_footer_text = mocker.Mock()
+        mocker.patch(CORE + ".webbrowser.get").side_effect = webbrowser.Error(error)
+
+        controller.open_in_browser("https://chat.zulip.org/#narrow/stream/test")
+
+        controller.view.set_footer_text.assert_called_once_with(f"ERROR: {error}", 3)
 
     def test_main(self, mocker, controller):
         controller.view.palette = {"default": "theme_properties"}
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -2,6 +2,7 @@ import os
 import signal
 import sys
 import time
+import webbrowser
 from collections import OrderedDict
 from functools import partial
 from platform import platform
@@ -14,7 +15,7 @@ from typing_extensions import Literal
 
 from zulipterminal.api_types import Composition, Message
 from zulipterminal.config.themes import ThemeSpec
-from zulipterminal.helper import asynch
+from zulipterminal.helper import LINUX, asynch, suppress_output
 from zulipterminal.model import Model
 from zulipterminal.ui import Screen, View
 from zulipterminal.ui_tools.utils import create_msg_box_list
@@ -303,6 +304,35 @@ class Controller:
             ),
             "area:msg",
         )
+
+    def open_in_browser(self, url: str) -> None:
+        """
+        Opens any provided URL in a graphical browser, if found, else
+        prints an appropriate error message.
+        """
+        # Don't try to open web browser if running without a GUI
+        # TODO: Explore and eventually support opening links in text-browsers.
+        if LINUX and not os.environ.get("DISPLAY") and os.environ.get("TERM"):
+            self.view.set_footer_text(
+                "No DISPLAY environment variable specified. This could "
+                "likely mean the ZT host is running without a GUI.",
+                3,
+            )
+            return
+        try:
+            # Checks for a runnable browser in the system and returns
+            # its browser controller, if found, else reports an error
+            browser_controller = webbrowser.get()
+            # Suppress stdout and stderr when opening browser
+            with suppress_output():
+                browser_controller.open(url)
+                self.view.set_footer_text(
+                    f"The link was successfully opened using {browser_controller.name}",
+                    3,
+                )
+        except webbrowser.Error as e:
+            # Set a footer text if no runnable browser is located
+            self.view.set_footer_text(f"ERROR: {e}", 3)
 
     def search_messages(self, text: str) -> None:
         # Search for a text in messages
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -25,6 +25,7 @@ from zulipterminal.config.symbols import (
 )
 from zulipterminal.config.ui_mappings import EDIT_MODE_CAPTIONS
 from zulipterminal.helper import Message, asynch, match_stream, match_user
+from zulipterminal.server_url import near_message_url
 from zulipterminal.ui_tools.boxes import MessageBox, PanelSearchBox
 from zulipterminal.ui_tools.buttons import (
     HomeButton,
@@ -1297,9 +1298,11 @@ class MsgInfoView(PopUpView):
         self.topic_links = topic_links
         self.message_links = message_links
         self.time_mentions = time_mentions
+        self.server_url = controller.model.server_url
         date_and_time = controller.model.formatted_local_time(
             msg["timestamp"], show_seconds=True, show_year=True
         )
+        view_in_browser_keys = ", ".join(map(repr, keys_for_command("VIEW_IN_BROWSER")))
 
         msg_info = [
             (
@@ -1308,6 +1311,10 @@ class MsgInfoView(PopUpView):
                     ("Date & Time", date_and_time),
                     ("Sender", msg["sender_full_name"]),
                     ("Sender's Email ID", msg["sender_email"]),
+                    (
+                        "View message in browser",
+                        f"Press {view_in_browser_keys} to view message in browser",
+                    ),
                 ],
             ),
         ]
@@ -1410,6 +1417,9 @@ class MsgInfoView(PopUpView):
                 message_links=self.message_links,
                 time_mentions=self.time_mentions,
             )
+        elif is_command_key("VIEW_IN_BROWSER", key):
+            url = near_message_url(self.server_url[:-1], self.msg)
+            self.controller.open_in_browser(url)
         return super().keypress(size, key)
 
 
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -55,6 +55,7 @@
 |Add/remove star status of the current message|<kbd>ctrl</kbd> + <kbd>s</kbd> / <kbd>*</kbd>|
 |Show/hide message information|<kbd>i</kbd>|
 |Show/hide edit history (from message information)|<kbd>e</kbd>|
+|View current message in browser (from message information)|<kbd>v</kbd>|
 
 ## Stream list actions
 |Command|Key Combination|
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -580,8 +580,18 @@ class TestMsgInfoView:
         self.msg_info_view.keypress(size, key)
         assert self.controller.exit_popup.called
 
+    @pytest.mark.parametrize("key", keys_for_command("VIEW_IN_BROWSER"))
+    def test_keypress_view_in_browser(self, mocker, widget_size, message_fixture, key):
+        size = widget_size(self.msg_info_view)
+        self.msg_info_view.server_url = "https://chat.zulip.org/"
+        mocker.patch(VIEWS + ".near_message_url")
+
+        self.msg_info_view.keypress(size, key)
+
+        assert self.controller.open_in_browser.called
+
     def test_height_noreactions(self):
-        expected_height = 3
+        expected_height = 4
         assert self.msg_info_view.height == expected_height
 
     # FIXME This is the same parametrize as MessageBox:test_reactions_view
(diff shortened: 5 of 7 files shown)


zulipterminal/core.py
```diff
@@ -253,6 +254,23 @@ def show_edit_history(
...
+        # Don't try to open web browser if running without a GUI
+        if (LINUX and not os.environ.get('DISPLAY')
+                and os.environ.get('TERM')):
+            self.view.set_footer_text(
+                'No DISPLAY environment variable specified. This could '
+                'likely mean your system is running without a GUI.', 3)
+            return
+        # Set a footer text if no runnable browser is located
+        try:
+            webbrowser.get()
```
preetmishra, line 267: What's the requirement for this particular statement? We're not using the controller being returned.

zulipterminal/core.py
```diff
@@ -253,6 +254,23 @@ def show_edit_history(
...
+            webbrowser.open(url)
+        except webbrowser.Error as e:
+            self.view.set_footer_text(f"ERROR: {e}", 3)
```
preetmishra, line 272: Just curious about whether you were able to raise any exception manually to test.

zulipterminal/core.py
```diff
@@ -253,6 +254,23 @@ def show_edit_history(
...
 
+    def view_in_browser(self, message_id: int) -> None:
+        url = '{}#narrow/near/{}'.format(self.model.server_url, message_id)
+
+        # Don't try to open web browser if running without a GUI
+        if (LINUX and not os.environ.get('DISPLAY')
+                and os.environ.get('TERM')):
+            self.view.set_footer_text(
+                'No DISPLAY environment variable specified. This could '
+                'likely mean your system is running without a GUI.', 3)
```
preetmishra, line 263: Their system would have a GUI but the ZT host won't, right?

zulipterminal/ui_tools/views.py
```diff
@@ -1230,11 +1230,16 @@ def __init__(self, controller: Any, msg: Message, title: str,
         date_and_time = controller.model.formatted_local_time(
             msg['timestamp'], show_seconds=True, show_year=True
         )
+        view_in_browser_key = ', '.join(map(
```
preetmishra, line 1235: This is rather _keys_?

zulipterminal/core.py
preetmishra: Also, let's specify the error context, i.e., the error came while opening the message.

tests/core/test_core.py
```diff
@@ -272,18 +272,23 @@ def test_narrow_to_all_mentions(
         msg_ids = {widget.original_widget.message['id'] for widget in widgets}
         assert msg_ids == id_list
 
-    def test_view_in_browser(self, mocker, controller):
+    @pytest.mark.parametrize('server_url', [
```
preetmishra: Re the commit title, this doesn't always open a message in _topic_ narrow (e.g. it enables PMs/huddles as well).

zulipterminal/core.py
```diff
@@ -254,8 +255,10 @@ def show_edit_history(
             'area:msg'
         )
 
-    def view_in_browser(self, message_id: int) -> None:
-        url = '{}#narrow/near/{}'.format(self.model.server_url, message_id)
+    def view_in_browser(self, message: Message) -> None:
+        # Truncate extra '/' from the server url.
+        url = near_message_url(self.model.server_url[:-1], message)
```
preetmishra: Echoing Neil's comment from the last PR, how do you feel about `urljoin`?

preetmishra (commented):
@zee-bit Thanks for continuing the work. I have tested this locally with every type of narrow and this seems to work well! :+1:

For the first iteration, I have left a few quick in-line comments (except for the last two commits). Of course, we would need to flesh out the security concerns prior to the merge.

While the last two commits extend the capability in a subtle way, we should perhaps introduce them post "Handle media links" PR as to an end-user it might not be apparent why some links (media links) don't open in the web browser while others do. This won't be an issue once we have the ability to handle media links in ZT itself.

zulipterminal/core.py
zee-bit: This statement is particularly only for Error handling, in case no browser is found in the system.
Yes, we don't have any use for the returned controller, but if the module could not find any browser(GUI/text-based) then this statement will return a `webbrowser.Error` and we can use the `try-except` block to catch it and display an appropriate error message.

zulipterminal/core.py
zee-bit: Yes, I was able to successfully reproduce this error on my AWS EC2 instance over `ssh`.
The error message is already very specific, so I don't think context would be necessary. See [here](https://stackoverflow.com/questions/39371219/python-webbrowser-check-if-browser-is-available-nothing-happens-when-opening) for the type of error message thrown and how to reproduce it.

zulipterminal/ui_tools/views.py
zee-bit: There is only one key for `view_in_browser` i.e. `v`, right?
Is there any particular reason to specify plural?

zulipterminal/core.py
zee-bit: I am sorry, but I don't understand this.
Aren't ***their system*** and ***ZT host*** the same?

zulipterminal/core.py
zee-bit: I am not too sure what the *pre-existing* code was there, but I am guessing it was something similar to our `server_url.py`?
We could use `urljoin`, but that will require completely refactoring our `server_url` code, and I am skeptical since the server uses the same approach as we have currently.

zulipterminal/ui_tools/views.py
preetmishra: Note that we're join-ing an array. While we do only have one key right now, the list might expand later.

zulipterminal/core.py
preetmishra: Right. @neiljp Any thoughts on this?

zulipterminal/core.py
preetmishra: For instance, the ZT host could be an AWS EC2 or a droplet and your laptop could be the system (with a GUI). What would happen in this case?

zulipterminal/core.py
preetmishra: Thanks for the explanation. Agreed; the error message seems to capture the context itself.

zulipterminal/core.py
zee-bit: So, replacing *your system* with *ZT host* in the above footer text?

zulipterminal/core.py
preetmishra: Makes sense!

zulipterminal/core.py
preetmishra: Yeah, that seems reasonable.

preetmishra:
@zee-bit Hey, is this ready for another review iteration?

zee-bit:
@preetmishra Yes, I have already updated the PR according to your review, so it should be ready for the next iteration. :)

zee-bit:

preetmishra (commented):
@zee-bit Thanks for making the amendments! I have tested this locally (except the last two commits, see my last point in the previous review) and it seems to well. :+1:

@neiljp The first four commits look good to me! We could have them in if everything looks good to you as well.

zulipterminal/core.py
neiljp: @zee-bit Could you point to the particular server code you're referring to?

tests/core/test_core.py
```diff
@@ -266,18 +266,25 @@ def test_narrow_to_all_mentions(self, mocker, controller, index_all_mentions):
...
         # Set DISPLAY environ to be able to run test in CI
         os.environ["DISPLAY"] = ":0"
-        mock_get = mocker.patch("webbrowser.get", mocker.Mock())
-        mock_open = mocker.patch("webbrowser.open", mocker.Mock())
-        message_id = 123456
-        controller.model.server_url = "https://foo.zulipchat.com/"
+        controller.model.server_url = server_url
+        message_id = message_fixture["id"]
+        mocked_get = mocker.patch(CORE + ".webbrowser.get")
+        mocked_open = mocker.patch(CORE + ".webbrowser.open")
```
neiljp, line 282: These lines look better than the original, but there's no reason to change the names and move them around - the commit diff is bigger and less clear because of it.

zulipterminal/core.py
neiljp: Re-reading this PR now I see that we've got `server_url.py` providing this functionality, which as @zee-bit says is from the server code :+1:

zulipterminal/ui_tools/views.py
```diff
@@ -1410,6 +1415,8 @@ def keypress(self, size: urwid_Size, key: str) -> str:
                 message_links=self.message_links,
                 time_mentions=self.time_mentions,
             )
+        elif is_command_key("VIEW_IN_BROWSER", key):
+            self.controller.view_in_browser(self.msg["id"])
```
neiljp: I'd suggest adding the functionality first, then hooking it up in a separate commit, which here could combine the small first commit with the UI which uses it.
(new popups tend to have a more complex structure since a new view often requires the key first)

neiljp (commented):
@zee-bit Thanks for pushing this feature forward again and @preetmishra for reviewing recently. The core features are looking good.

However, reading this there's a lot of intermediate refactoring of the *new code* added in this PR, which indicates the commits might be more "as explored" rather than as cleanly laid out with what we know what we want and can achieve - the tests are moved around quite a bit too. I'm not sure if more of the older commits were kept as they were for consistency, but you can always add `Co-authored-by: <author>` to the commit text to share authorship if things work better when split up, which you may want to do if you significantly add to (author) a commit in any case as the 'commit name' might be replaced by whoever merges the commits eventually. Of course if you completely rewrite, that's all you :)

I'm not completely opposed to the current structure and happy to discuss in the stream, but I'd suggest something like the following might be clearer (possibly with intermediate commits which we can squash together later after discussion):
* Add output suppression functionality (same commit, by Puneeth)
* Add open url in browser feature in near-final form, using output suppression & UI error handling (part of 2nd, adapted as per 5th?)
* Add key+UI for opening message (current first commit + part of second + easy update using our library from the 4th)
* General url opening implementation (to be discussed)

For sanity we could add an assert and/or tests to the url-opening method to ensure it's from the server, before we add general link opening, but it depends how we expect to proceed with general links - was the idea to maybe have a prompt like with media?

zulipterminal/core.py
```diff
@@ -304,6 +305,24 @@ def show_edit_history(
             "area:msg",
         )
 
+    def view_in_browser(self, message_id: int) -> None:
+        url = "{}#narrow/near/{}".format(self.model.server_url, message_id)
+
+        # Don't try to open web browser if running without a GUI
```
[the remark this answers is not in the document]
zee-bit: I would like to explore more on how text-browsers work for different sites and our webapp, so I am leaving this as a TODO for now. :)

zulipterminal/core.py
```diff
@@ -304,6 +305,24 @@ def show_edit_history(
...
+            self.view.set_footer_text(
+                "No DISPLAY environment variable specified. This could "
+                "likely mean the ZT host is running without a GUI.",
+                3,
+            )
+            return
+        # Set a footer text if no runnable browser is located
+        try:
+            webbrowser.get()
+            webbrowser.open(url)
```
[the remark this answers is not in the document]
zee-bit: Yes, `open_new_tab` sounds better. Although I do wonder what's the difference between `.open(new=0)`(default) and `new=2`(same as new_tab), they both seem to behave exactly the same to me. If both do come out to be the same then we don't need to change the above, right?

In any case, added a success message here as asked. :)

zulipterminal/ui_tools/views.py
```diff
@@ -1300,6 +1300,7 @@ def __init__(
         date_and_time = controller.model.formatted_local_time(
             msg["timestamp"], show_seconds=True, show_year=True
         )
+        view_in_browser_keys = ", ".join(map(repr, keys_for_command("VIEW_IN_BROWSER")))
```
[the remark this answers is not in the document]
zee-bit: `repr` gives the string representation here. So, if the key is `'v'`, the same will be printed (with quotes).
I think that's how we do in other such places throughout our codebase?
Or, am I misunderstanding your question?

zulipterminal/core.py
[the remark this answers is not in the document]
zee-bit: Well, I tried `lynx` on a remote server, and though some simple website were rendered OK, I couldn't get it to work for our webapp properly. As in, it loads the CZO website, and I can even log-in but it cant fetch any messages apparently. I am just left with the top-left panel items aligned haphazardly and a few other options such as log-out and server name...

neiljp (commented):
@zee-bit This is super-close, just left a few small points :+1:

Maybe we'll squash the 3rd and 4th commit, and if we're planning to drop the last commit into another PR we might want to do that in this next (hopefully final!) iteration.

I'm pretty excited at the upcoming rework of the last commit too - with the work on topic links too we'll be able to open linkified PRs/issues in the browser much more easily :tada:

zulipterminal/core.py
```diff
@@ -304,6 +305,34 @@ def show_edit_history(
...
+            return
+        try:
+            # Checks for a runnable browser in the system and returns
+            # its browser controller, if found, else reports an error
+            browser_controller = webbrowser.get()
+            # Suppress stdout and stderr when opening browser
+            with suppress_output():
+                browser_controller.open(url)
+                self.view.set_footer_text(
+                    f"The link was successfully opened in {browser_controller.name}", 3
```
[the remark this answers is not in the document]
zee-bit: Hmm...That's odd.
Locally I get `...google-chrome`, and on the remote server running `lynx`, I get `...generic-browser`. I am aware that webbrowser module uses `xdg-open` under the hood, but it shouldn't be returning that as the name?

Any chance you're on Python v<=3.6, I found a few StackOverflow answers which suggest that its a Python issue, which was fixed on 3.6+?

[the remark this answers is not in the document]
zee-bit: In any case, if we can't grab the actual name it's better to output a general message like: `The link was successfully opened in the default browser.`?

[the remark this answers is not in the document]
neiljp: I appear to have 3.7.3 here. As I said above, perhaps let's simply switch to `using` in place of `in` in the text, which reads slightly better for both `using chrome` and `using xdg-open`?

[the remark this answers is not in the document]
neiljp: (just to keep things simple for now)

tests/core/test_core.py
```diff
@@ -265,6 +267,39 @@ def test_narrow_to_all_mentions(self, mocker, controller, index_all_mentions):
...
+        assert called_url == url
+        controller.view.set_footer_text.assert_called_once_with(
+            f"The link was successfully opened in {mock_get.return_value.name}", 3
+        )
+
+        # When webbrowser.get() fails to find a browser controller
+        error_context = "no runnable browser found!"
+        mock_get.side_effect = webbrowser.Error(error_context)
+
+        controller.open_in_browser(url)
```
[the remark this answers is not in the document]
zee-bit: No, we don't need to parametrize over URL. Thanks. :+1:

zulipterminal/core.py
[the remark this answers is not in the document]
zee-bit: Sure, I'll use using.
I just refreshed this page so didn't saw your comment earlier. :)

zee-bit:
Removed the last commit from here, will be introducing it in a new PR as requested. :)

neiljp:
@zee-bit Thanks for the adjustments; I did that squash, adjusted some commit text and simplified the tests slightly and am merging now! :tada:

Thanks for @punchagan and @preetmishra for starting this and keeping this going too!

PR #994, opened 2021-04-13
Add warning on startup if locale suggests UTF-8 not enabled

### LINKED ISSUE #256: Add warning on startup if locale/language suggests missing "icons"
At least one user reported that the screen had question marks or similar, and it turned out their locale did not include UTF-8.

We do note that eg. `en_US.UTF-8` is suggested in the FAQs, though I personally use `en_GB.UTF-8` and the drawing lines and so on all work fine. However, if I run `LANG=en_GB zulip-term` then the top lines are instead `????`, and the bullets for the top-left buttons and users are also `?`.

Given that we can detect the `LANG` setting by accessing the environment variables, we should be able to give a warning on startup that the icons/lines may be incorrect/unexpected.

Beyond this, in future we could aim to be adaptable:
- detect non-UTF-8 and use different characters for the top lines or bullets
- better font (UTF-16? some other way of detecting?) and use more complex icons

Ezio-Sarthak:
**What does this PR do!?**
 - Shows a warning message on startup if the language/locale settings don't include `utf-8` encoding.

**Commit flow**
 - Add helper to return whether required `utf-8` encoding is enabled in locale.
 - Show warning message on the loading screen if required `utf-8` encoding is disabled.

Fixes #256.

**Screenshots/GIFs**
warning_message

### DIFF
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -124,6 +124,48 @@ def test_valid_zuliprc_but_no_connection(capsys, mocker, minimal_zuliprc,
     assert captured.err == ""
 
 
+def test_warning_utf_encoding_not_enabled(capsys, mocker, monkeypatch,
+                                          minimal_zuliprc,
+                                          server_connection_error="sce"):
+    mocker.patch('zulipterminal.core.Controller.__init__',
+                 side_effect=ServerConnectionFailure(server_connection_error))
+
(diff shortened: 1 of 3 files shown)


neiljp:
@Ezio-Sarthak Thanks for looking into this but as we discussed and agreed in the stream, we likely don't need this now as per my comment in and closing of #256.

PR #996, opened 2021-04-13
bugfix: Restrict content-length for compose box.

zee-bit:
This PR restricts input beyond certain max-limit for compose
box, as specified by the corresponding parameters in initial_data's
realm object. ZT didn't had this restriction earlier which resulted
in certain long topic names/messages getting cropped from the end
without the user noticing, until the message was sent.
These parameters were recently added in ZFL 53 in [`4a3ad0d`](https://github.com/zulip/zulip/commit/4a3ad0da06b3e8db511e9690b31507b565f7f2fc), so we allow for
backporting by defining hard-coded values for these parameters.

The content restriction of compose boxes was made possible only
after urwid-readline merged a patch in v0.13, which added support
for specifying a `max_char` argument to their `ReadlineEdit` widget.
Related patch: https://github.com/rr-/urwid_readline/pull/18.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -7,7 +7,13 @@ import pytest
 from zulip import ZulipError
 
 from zulipterminal.helper import initial_index, powerset
-from zulipterminal.model import Model, ServerConnectionFailure
+from zulipterminal.model import (
+    MAX_MESSAGE_LENGTH,
+    MAX_STREAM_NAME_LENGTH,
+    MAX_TOPIC_NAME_LENGTH,
+    Model,
+    ServerConnectionFailure,
+)
 
 
 class TestModel:
@@ -627,6 +633,40 @@ class TestModel:
 
         assert (model.modernize_message_response(messages)
                 == expected_messages_response)
+
+    @pytest.mark.parametrize('feature_level, to_vary_in_initial_data', [
+        (None, {}),
+        (27, {}),
+        (52, {}),
+        (53, {
+            'max_stream_name_length': 30,
+            'max_topic_length': 20,
+            'max_message_length': 5000,
+        })], ids=[
+            'Zulip_2.1.x_ZFL_None_no_restrictions',
+            'Zulip_3.1.x_ZFL_27_no_restrictions',
+            'Zulip_4.0.x_ZFL_52_no_restrictions',
+            'Zulip_4.0.x_ZFL_53_with_restrictions',
+    ])
+    def test__store_content_length_restrictions(
+            self, model, initial_data, feature_level, to_vary_in_initial_data):
+        initial_data.update(to_vary_in_initial_data)
+        model.initial_data = initial_data
+        model.server_feature_level = feature_level
+
+        model._store_content_length_restrictions()
+
+        if to_vary_in_initial_data:
+            assert (model.max_stream_name_length
+                    == to_vary_in_initial_data['max_stream_name_length'])
+            assert (model.max_topic_length
+                    == to_vary_in_initial_data['max_topic_length'])
+            assert (model.max_message_length
+                    == to_vary_in_initial_data['max_message_length'])
+        else:
+            assert model.max_stream_name_length == MAX_STREAM_NAME_LENGTH
+            assert model.max_topic_length == MAX_TOPIC_NAME_LENGTH
+            assert model.max_message_length == MAX_MESSAGE_LENGTH
 
     def test_get_message_false_first_anchor(
             self, mocker, messages_successful_response, index_all_messages,
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -26,6 +26,9 @@ class TestWriteBox:
         write_box = WriteBox(self.view)
         write_box.view.users = users_fixture
         write_box.model.user_dict = user_dict
+        write_box.model.max_stream_name_length = 60
+        write_box.model.max_topic_length = 60
+        write_box.model.max_message_length = 10000
         write_box.model.user_group_names = [
             groups['name'] for groups in user_groups_fixture]
 
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -52,6 +52,14 @@ from zulipterminal.ui_tools.utils import create_msg_box_list
 
 OFFLINE_THRESHOLD_SECS = 140
 
+# Adapted from zerver/models.py
+# These fields have migrated to the API inside the Realm object
+# in ZFL 53. To allow backporting to earlier server versions, we
+# define these hard-coded parameters.
+MAX_STREAM_NAME_LENGTH = 60
+MAX_TOPIC_NAME_LENGTH = 60
+MAX_MESSAGE_LENGTH = 10000
+
 
 class ServerConnectionFailure(Exception):
     pass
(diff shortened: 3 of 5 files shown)


zulipterminal/model.py
```diff
@@ -171,6 +179,14 @@ def __init__(self, controller: Any) -> None:
         self.active_emoji_data = OrderedDict(sorted(all_emoji_data,
                                                     key=lambda e: e[0]))
 
+        # New in server version 4.0, ZFL 53.
+        self.max_stream_name_length = self.initial_data.get(
+            'max_stream_name_length', MAX_STREAM_NAME_LENGTH)
+        self.max_topic_length = self.initial_data.get(
+            'max_topic_length', MAX_TOPIC_NAME_LENGTH)
+        self.max_message_length = self.initial_data.get(
+            'max_message_length', MAX_MESSAGE_LENGTH)
```
neiljp, line 188: This would benefit from testing over various ZFL.

neiljp (commented):
@zee-bit This looks to work quite well, though splitting out the model changes and adding a test over ZFL may be useful.

The other point which looks more important is that Zulip text (topics, content) is I believe typically stripped so that ` a ` becomes `a`. However, the readline limit is fixed - leading spaces in pasted text wouldn't disappear, but the end of it would be "blocked" by the new feature?

This issue aside, the content-length restriction seems like a good v1, but it doesn't give much feedback?

We can discuss, but having an explicit area colored as the edit area, or perhaps marking (not limiting?) the text beyond the limit may be some options to pursue?

neiljp:
It's worth noting that if we don't go for the full UI approach initially due to concerns there, or perhaps only for different input boxes, we could always have a popup or similar which asks the user what they wish to do with the extra text before sending, or similar.

tests/model/test_model.py
```diff
@@ -628,6 +628,33 @@ def test_modernize_message_response(self, model, messages,
...
+            'Zulip_4.0.x_ZFL_53_with_restrictions',
+    ])
+    def test__store_content_length_restrictions(
+            self, model, initial_data, feature_level, to_vary_in_initial_data):
+        initial_data.update(to_vary_in_initial_data)
+        model.initial_data = initial_data
+        model.server_feature_level = feature_level
+        EXPECTED_MAX_STREAM_NAME_LENGTH = 60
+        EXPECTED_MAX_TOPIC_LENGTH = 60
+        EXPECTED_MAX_MESSAGE_LENGTH = 10000
```
neiljp, line 650: We can import these directly from model.py.

tests/model/test_model.py
```diff
@@ -628,6 +628,33 @@ def test_modernize_message_response(self, model, messages,
         assert (model.modernize_message_response(messages)
                 == expected_messages_response)
 
+    @pytest.mark.parametrize('feature_level, to_vary_in_initial_data', [
+        (None, {}),
+        (27, {}),
+        (53, {
+            'max_stream_name_length': 60,
+            'max_topic_length': 60,
+            'max_message_length': 10000,
```
neiljp, line 637: This looks good, but we want to ensure that these are not the same as the defaults, to ensure that we process the values correctly - so maybe values relative to the old defaults?

tests/model/test_model.py
```diff
@@ -628,6 +628,33 @@ def test_modernize_message_response(self, model, messages,
         assert (model.modernize_message_response(messages)
                 == expected_messages_response)
 
+    @pytest.mark.parametrize('feature_level, to_vary_in_initial_data', [
+        (None, {}),
+        (27, {}),
```
neiljp, line 633: A value between is good - I see that this is for 3.1, though I wouldn't know from reading the code.

Of course here we're defining the input *plus* the feature level. Like you said elsewhere, we could even skip the feature level since we don't rely on it right now - though it's useful to include for documentation? If we do include an extra feature level I'd also/instead suggest the level just before the change.

zulipterminal/model.py
```diff
@@ -171,6 +179,7 @@ def __init__(self, controller: Any) -> None:
         self.active_emoji_data = OrderedDict(sorted(all_emoji_data,
                                                     key=lambda e: e[0]))
 
+        self._store_content_length_restrictions()
```
neiljp, line 182: This is cleaner, though I'm often torn between seeing the higher level details and explicit setting of attributes in `__init__`.

tests/ui_tools/test_boxes.py
```diff
@@ -26,6 +26,9 @@ def write_box(self, mocker, users_fixture, user_groups_fixture,
         write_box = WriteBox(self.view)
         write_box.view.users = users_fixture
         write_box.model.user_dict = user_dict
+        write_box.model.max_stream_name_length = 60
+        write_box.model.max_topic_length = 60
+        write_box.max_message_length = 10000
```
neiljp, line 31: This typo (?) concerns me - do we even need these values?

neiljp (commented):
@zee-bit This is looking good as the first version - we did discuss other user feedback to explore in the future.

I do think we could later explore initial-whitespace trimming like in the search boxes if we're going to use this urwid feature, though we did have a debate then over the lack of feedback when typing whitespace and seemingly getting no response.

tests/ui_tools/test_boxes.py
zee-bit: These are necessary for the tests to pass since the `ReadlineEdit` gets these values from model, and if we don't define this ourselves, `Mock` will be used in their place.

tests/model/test_model.py
zee-bit: I initially tried to keep the fixtures matching with the server versions that I have tested locally. :)

Re skipping feature level, I think we should keep that since it indicates that these fields are actually dependent on the server version, and adding a feature level just before the change would also indicate precisely when these fields were added in the server.

tests/ui_tools/test_boxes.py
neiljp: I meant specifically the last line - which seems like a typo in that you have no `model` in there. Do mock values just work anyhow?

zee-bit: Ahh...Right, my bad I missed that.

I just tested with mock values for all, and it fails as expected, for the above two but not for the `max_message_length` field. On looking at the test summary, it appears to fail because the above fields pass an `edit_text` parameter to the `ReadlineEdit`, but the `msg_write_box` does not. And, passing this parameter leads to the below condition in `urwid_readline/readline_edit.py` being positive, and `max_char` being of type `Mock`, can't be used as an index in slicing:
```python3
if max_char and "edit_text" in kwargs:
    kwargs["edit_text"] = kwargs["edit_text"][:max_char]
```
So, `Mock` values should work here if we don't specify `edit_text` parameter to ReadlineEdit initially. I have updated the PR fixing the above typo.

neiljp:
@zee-bit Thanks for noticing and fixing this - just merging now :tada:

Could you file a follow-up issue for ideas we have so far around improving this? We may want to discuss further before proceeding.

The numbers we have from this have relevance to aspects like #905 (compose header lengths), which could also relate to the new issue, as well as later perhaps things like maximum left panel widths, or the maximum search length?

We don't have information for names (& emails for the current UI), and I'm not sure if the server intends to extend the coverage to include that level of parametrization, but that could open up scope for applying the same thoughts to the right panel width and compose area lengths.

tests/model/test_model.py
```diff
@@ -628,6 +634,40 @@ def test_modernize_message_response(self, model, messages,
...
+    ])
+    def test__store_content_length_restrictions(
+            self, model, initial_data, feature_level, to_vary_in_initial_data):
+        initial_data.update(to_vary_in_initial_data)
+        model.initial_data = initial_data
+        model.server_feature_level = feature_level
+
+        model._store_content_length_restrictions()
+
+        if to_vary_in_initial_data:
```
neiljp, line 659: This kind of branching suggests we could try hoisting values into the parametrizations, but this is fine.

PR #997, opened 2021-04-14
Handle read events to the same server from different clients

mkp6781:
Taking forward @kaustubh-nair 's work on #535.
I have added a commit to update `set_count` function to use index[`unread_msgs`] along with necessary changes to make sure that the application does not crash.


mkp6781:

PR #998, opened 2021-04-18
Gruvbox theme: Add true color version.

### LINKED ISSUE #282: Add 24-bit theme support
This is a tracking issue for https://github.com/urwid/urwid/pull/284. Once merged, I will extend gruvbox to 24-bit, and also this will make adding other true color themes easier.

rht:
This fixes #282.

gruvbox_dark:
2021-04-18-055109_985x654_scrot
gruvbox_dark24 when color depth is 2**24:
2021-04-18-055137_991x662_scrot
gruvbox_dark24 when color depth is 256 (i.e. wrong setting):
2021-04-18-055316_985x655_scrot

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ footlinks=disabled
 maximum-footlinks=3
 # Notify defaults to 'disabled', but can be set to 'enabled' to display notifications (see next section).
 notify=enabled
-# Color depth defaults to 256, but can be set to 1 (for monochrome) or 16.
+# Color depth defaults to 256, but can be set to 1 (for monochrome), or 16 or 24bit.
 color-depth=256
 ```
 
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -92,7 +92,7 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
                         action="store_true",
                         help='list all the color themes.')
     parser.add_argument('--color-depth',
-                        choices=['1', '16', '256'],
+                        choices=['1', '16', '256', '24bit'],
                         help="Force the color depth "
                              f"(default {DEFAULT_SETTINGS['color-depth']}).")
     notify_group = parser.add_mutually_exclusive_group()
@@ -388,7 +388,11 @@ def main(options: Optional[List[str]]=None) -> None:
         if args.color_depth:
             zterm['color-depth'] = (args.color_depth, 'on command line')
 
-        color_depth = int(zterm['color-depth'][0])
+        color_depth_str = zterm['color-depth'][0]
+        if color_depth_str == '24bit':
+            color_depth = 2 ** 24
+        else:
+            color_depth = int(color_depth_str)
 
         if args.notify:
             zterm['notify'] = (args.notify, 'on command line')
@@ -422,7 +426,7 @@ def main(options: Optional[List[str]]=None) -> None:
         valid_settings = {
             'autohide': ['autohide', 'no_autohide'],
             'notify': ['enabled', 'disabled'],
-            'color-depth': ['1', '16', '256']
+            'color-depth': ['1', '16', '256', '24bit']
         }
         boolean_settings: Dict[str, bool] = dict()
         for setting, valid_values in valid_settings.items():
--- a/tests/config/test_themes.py
+++ b/tests/config/test_themes.py
@@ -10,7 +10,7 @@ from zulipterminal.config.themes import (
 
 
 expected_complete_themes = {
-    'zt_dark', 'gruvbox_dark', 'zt_light', 'zt_blue',
+    'zt_dark', 'gruvbox_dark', 'zt_light', 'zt_blue', 'gruvbox_dark24'
 }
 
 
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -92,6 +92,23 @@ GRAY = 'h244'  # gray_244
 LIGHTMAGENTA = 'h132'  # neutral_purple
 LIGHTMAGENTABOLD = f"{LIGHTMAGENTA}, bold"
 
+# Colors used in gruvbox_dark24
+BLACK24 = '#1d2021'  # dark0_hard
+WHITE24 = '#a89984'  # light4
+WHITEBOLD24 = f"{WHITE24}, bold"
+DARKBLUE24 = '#076678'  # faded_blue
+DARKRED24 = '#9d0006'  # faded_red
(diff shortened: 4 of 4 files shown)


rht:
I'd say, the true color version with the right setting has a warmer color, and is closer to the color of the screenshot in https://github.com/morhetz/gruvbox (dark mode).

We probably need to add an assertion that if a truecolor theme is used, you must also use the right color depth.

rht:
How do I fix https://github.com/zulip/zulip-terminal/pull/998/checks?check_run_id=2373629708#step:9:65 ?

zee-bit:
> How do I fix https://github.com/zulip/zulip-terminal/pull/998/checks?check_run_id=2373629708#step:9:65 ?

You need to add the new theme(i.e. `gruvbox_dark24`) to the set of `expected_complete_themes` in `test_themes.py`. :)

prah23:
While you're at it, would you please also add a theme alias under `THEME_ALIASES` in `zulipterminal\config\themes.py`?

prah23:
While you're at it, would you please also add a theme alias under `THEME_ALIASES` in `zulipterminal/config/themes.py`?

zulipterminal/cli/run.py
```diff
@@ -92,7 +92,7 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
                         action="store_true",
                         help='list all the color themes.')
     parser.add_argument('--color-depth',
-                        choices=['1', '16', '256'],
+                        choices=['1', '16', '256', '2**24'],
```
zee-bit, line 95: Note that the new `color-depth` option needs to be wrapped in quotes when passing to the CLI parameter. This is not the case with other options, which makes the new one inconsistent and a bit long compared to other options.

Alternatively, we could use something like `2^24` which is shorter and don't need to be wrapped in quotes?

zee-bit (commented):
@rht Thanks for bringing this up. This looks very nice. :+1: 
I have tested it across all `color-depths` and it works just as expected.

One thing we may need to consider is, this theme works best with a `color-depth` of `2**24`, so we may need to specify this somewhere(in the docs, maybe). A better approach might be to implicitly use `2**24` for this theme, unless the user explicitly supplies a different `color-depth` option?

zulipterminal/cli/run.py
```diff
@@ -388,7 +388,11 @@ def main(options: Optional[List[str]]=None) -> None:
         if args.color_depth:
             zterm['color-depth'] = (args.color_depth, 'on command line')
 
-        color_depth = int(zterm['color-depth'][0])
+        color_depth_str = zterm['color-depth'][0]
+        if color_depth_str == '2**24':
+            color_depth = 2 ** 24
```
neiljp, line 393: This would be better as `24bit`; we can migrate others across to support similar naming too, perhaps.

neiljp (commented):
@rht Thanks for keeping your eye on this :) 

It's been a while since I've looked at these palette rules, so I was expecting we might simply have an extra entry in each style/palette for the 24-bit values. Unfortunately that seems not the case, and I do recall challenges with this in the past.

While gruvbox is the only theme that has a 24-bit version at the moment via this PR, I'd really prefer that we don't require the user to specify a compatible bit-depth to go with the style or just error-out, but instead just make it work transparently.

The two approaches I'd look at would be:
* simplest, if it works: If you run your gruvbox24 theme at 256, does it look close enough - or better - than the existing gruvbox at 256? If so, can we replace the old theme with the values in the new one?
* OK, if urwid doesn't work well enough with the first approach: extend our palette entries to include a non-urwid 24-bit value, which we reduce at runtime to an urwid-compatible one depending on the bit-depth we're running at
(we could extend the latter to simplify our theme datastructure in any case, given how we handle 1-bit/mono)

There may be a different urwid approach to this, but I've not dug into the code, and this is new and not best documented?

Once this is settled we could add an issue for tracking other potential themes to offer which may render better with this mode - light gruvbox? solarized dark/light?

zulipterminal/cli/run.py
rht: Resolved as `24bit`

rht:
> While you're at it, would you please also add a theme alias under `THEME_ALIASES` in `zulipterminal/config/themes.py`?

I think `gruvbox_dark24` is pretty explicit about what it is. Did you mean to alias it to `gruvbox24`?

rht:
> simplest, if it works: If you run your gruvbox24 theme at 256, does it look close enough - or better - than the existing gruvbox at 256? If so, can we replace the old theme with the values in the new one?

`gruvbox24` with 256 color depth looks worse. The white color is off, if you see my first message of this PR.

rht:
If you want to reduce boilerplate code, I do think it is possible to write a generic abstract theme function, which can generate a specific theme that translates e.g. `white` into `GRUVBOXWHITE24`.

neiljp:
@rht I saw the screenshots, but wasn't sure if they had a particular meaning, which was why I asked, so thanks for clarifying. Do you think we should take something like the second approach I suggested, then?

If we work with a 24-bit version of gruvbox, then I don't think we need an alias. In any case, the aliases were added for backwards compatibility, so we would want to discuss whether we want to use them as shortcuts rather than for historical benefit only.

I'm not sure how much you want to commit to theme refactoring - we could certainly simply work with a 256/24bit reduction only. That would provide a nice first step from `palette is style` to a theme spec in any case.

neiljp:
For the boilerplate, there is various scope for reduction, but I think you'd still need eg `"selected": [("white", "black"), (WHITE, BLACK), (WHITE24, BLACK24)]` for a given style in a theme. Pairing foreground/background would be optional but potentially cleaner.

rht:
I will do the refactor in a separate PR. The new theme in this PR works and can be used as data/test for the refactor.

neiljp:
This looks almost good to go :+1: We could do with a quick update of this in the README, from a quick git-grep of places we mention color depth - and the commit messages don't match the repo style.

README.md
```diff
@@ -113,7 +113,7 @@ footlinks=disabled
 maximum-footlinks=3
 # Notify defaults to 'disabled', but can be set to 'enabled' to display notifications (see next section).
 notify=enabled
-# Color depth defaults to 256, but can be set to 1 (for monochrome) or 16.
+# Color depth defaults to 256, but can be set to 1 (for monochrome), or 16 or 24bit.
```
rht, line 116: Added this to README.md in my most recent force push.

rht:
Updated the commit messages.

PR #999, opened 2021-04-19
refactor: boxes: Refactor soup2mark for better readability.

Rohitth007:
This PR on the whole is to make soup2mark _cleaner_, more _readable_ and _easier to understand_.

**Commit 1: Shorten soup2mark conditionals.**
* This commit extracts commonly used code in conditional statements
into variables so as to reduce repeatition and improve readability.
HTML semantics like `tag`, `html_attrs`, `css_classes` are used to
name these variables.
* These variables are only extracted when the element is a `Tag` as
`NavigableString`'s don't have these attributes.

**Commit 2: Move css_classes into a single block.**
* This commit moves all code related to css_classes into a single
block which makes for better organization of code.
   * Six conditionals have been moved to a single block.
   * Now these six conditionals can be eliminated in a single shot and
also makes the corresponding conditionals shorter.

**Commit 3: Improve long condition statements.**
* This commit makes the conditions for `MATH TEXT` and `MENTIONS`
shorter and more readable. It follows a pattern similar to
`unrendered_div_classes`.
* `_` underscore is used as a placeholder for css classes and 2 new
class lists were defined which are `math_classes` and
`mention_classes` respectively for the two conditionals.

**Commit 4: Replace `cls` with `_` to remove confusion.**
* This commit replaces the local variable `cls` in the comprehension
with the python variable placeholder `_` so that it is not confused
with the main `cls` variable of the `classmethod` representing a
python class instead of a CSS class.

**Commit 5: Clean whitespace in `code` and `codehilite`.**
* This commit just removes whitespace in `markup.append` argument
of `code` and `codehilite` tag and class respectively

**Commit 6: Simplify UNRENDERED DIV CLASSES code.**
* This comment simplifies the code for UNRENDERED DIV CLASSES by
changing the variable `matching_class` with `match`. This allows for
removal of unnecessary brackets.

**Commit 7: Simplify code for `time` tag.**
* This commit simplifies code for the time tag:
   * Brackets for `source_text` have been removed as the space allows for
it.
   * `markup.append` has been moved to the end to keep it similar and
consistent with the rest of the code.

**Collapsed Screenshot:**

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -955,8 +955,8 @@ class MessageBox(urwid.Pile):
                     continue
                 markup.append(element)
             elif (tag == 'div'
-                  and any(cls in tag_classes
-                          for cls in unrendered_div_classes)):
+                  and (set(tag_classes)
+                       & set(unrendered_div_classes))):
                 # UNRENDERED DIV CLASSES
                 matching_class = (set(unrendered_div_classes)
                                   & set(tag_classes))
@@ -981,13 +981,13 @@ class MessageBox(urwid.Pile):
                 # EMOJI
                 markup.append(('msg_emoji', tag_text))
             elif (tag == 'span'
-                  and ('katex-display' in tag_classes
-                       or 'katex' in tag_classes)):
+                  and ({'katex-display', 'katex'}
+                       & set(tag_classes))):
                 # MATH TEXT
                 markup.append(tag_text)
             elif (tag == 'span'
-                  and ('user-group-mention' in tag_classes
-                       or 'user-mention' in tag_classes)):
+                  and ({'user-group-mention', 'user-mention'}
+                       & set(tag_classes))):
                 # USER MENTIONS & USER-GROUP MENTIONS
                 markup.append(('msg_mention', tag_text))
             elif tag == 'a':
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -954,7 +954,7 @@ class MessageBox(urwid.Pile):
                     metadata['bq_len'] -= 1
                     continue
                 markup.append(element)
-            elif (tag == 'div' and tag_attrs
+            elif (tag == 'div'
                   and any(cls in tag_classes
                           for cls in unrendered_div_classes)):
                 # UNRENDERED DIV CLASSES
@@ -976,16 +976,16 @@ class MessageBox(urwid.Pile):
             elif tag in ('p', 'del'):
                 # PARAGRAPH, STRIKE-THROUGH
                 markup.extend(cls.soup2markup(element, metadata)[0])
-            elif (tag == 'span' and tag_attrs
+            elif (tag == 'span'
                   and 'emoji' in tag_classes):
                 # EMOJI
                 markup.append(('msg_emoji', tag_text))
-            elif (tag == 'span' and tag_attrs
+            elif (tag == 'span'
                   and ('katex-display' in tag_classes
                        or 'katex' in tag_classes)):
                 # MATH TEXT
                 markup.append(tag_text)
-            elif (tag == 'span' and tag_attrs
+            elif (tag == 'span'
                   and ('user-group-mention' in tag_classes
                        or 'user-mention' in tag_classes)):
                 # USER MENTIONS & USER-GROUP MENTIONS
@@ -1069,7 +1069,7 @@ class MessageBox(urwid.Pile):
                 markup.append((
                     'msg_code', tag_text
                 ))
-            elif (tag == 'div' and tag_attrs
+            elif (tag == 'div'
                     and 'codehilite' in tag_classes):
                 # CODE (BLOCK?)
                 markup.append((
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -954,17 +954,16 @@ class MessageBox(urwid.Pile):
                     metadata['bq_len'] -= 1
                     continue
                 markup.append(element)
-            elif (tag == 'div'
-                  and (set(tag_classes)
-                       & set(unrendered_div_classes))):
+            elif tag == 'div' and (
+                 set(tag_classes) & set(unrendered_div_classes)):
                 # UNRENDERED DIV CLASSES
-                matching_class = (set(unrendered_div_classes)
-                                  & set(tag_classes))
-                text = unrendered_div_classes[matching_class.pop()]
+                # NOTE: Though `matches` is generalized for multiple
+                # matches it is very unlikely that there would be any.
+                matches = set(unrendered_div_classes) & set(tag_classes)
+                text = unrendered_div_classes[matches.pop()]
                 if text:
                     markup.append(unrendered_template.format(text))
-            elif (tag == 'img'
-                  and tag_classes == ['emoji']):
+            elif tag == 'img' and tag_classes == ['emoji']:
                 # CUSTOM EMOJIS AND ZULIP_EXTRA_EMOJI
                 emoji_name = tag_attrs.get('title', [])
                 markup.append(('msg_emoji', f":{emoji_name}:"))
@@ -976,18 +975,15 @@ class MessageBox(urwid.Pile):
             elif tag in ('p', 'del'):
                 # PARAGRAPH, STRIKE-THROUGH
                 markup.extend(cls.soup2markup(element, metadata)[0])
-            elif (tag == 'span'
-                  and 'emoji' in tag_classes):
+            elif tag == 'span' and 'emoji' in tag_classes:
                 # EMOJI
                 markup.append(('msg_emoji', tag_text))
-            elif (tag == 'span'
-                  and ({'katex-display', 'katex'}
-                       & set(tag_classes))):
+            elif tag == 'span' and (
+                    {'katex-display', 'katex'} & set(tag_classes)):
                 # MATH TEXT
                 markup.append(tag_text)
-            elif (tag == 'span'
-                  and ({'user-group-mention', 'user-mention'}
-                       & set(tag_classes))):
+            elif tag == 'span' and (
+                    {'user-group-mention', 'user-mention'} & set(tag_classes)):
                 # USER MENTIONS & USER-GROUP MENTIONS
                 markup.append(('msg_mention', tag_text))
             elif tag == 'a':
@@ -1066,15 +1062,10 @@ class MessageBox(urwid.Pile):
                 ))
             elif tag == 'code':
                 # CODE (INLINE?)
-                markup.append((
-                    'msg_code', tag_text
-                ))
-            elif (tag == 'div'
-                    and 'codehilite' in tag_classes):
+                markup.append(('msg_code', tag_text))
+            elif tag == 'div' and 'codehilite' in tag_classes:
                 # CODE (BLOCK?)
-                markup.append((
-                    'msg_code', tag_text
-                ))
+                markup.append(('msg_code', tag_text))
             elif tag in ('strong', 'em'):
                 # BOLD & ITALIC
                 markup.append(('msg_bold', tag_text))
@@ -1143,9 +1134,7 @@ class MessageBox(urwid.Pile):
                     'msg_time', f" {TIME_MENTION_MARKER} {time_string} "
                 ))
 
-                source_text = (
-                    f"Original text was {tag_text.strip()}"
-                )
+                source_text = f"Original text was {tag_text.strip()}"
                 metadata['time_mentions'].append((time_string, source_text))
             else:
                 markup.extend(cls.soup2markup(element, metadata)[0])
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -925,6 +925,9 @@ class MessageBox(urwid.Pile):
         markup: List[Union[str, Tuple[Optional[str], Any]]] = ['']
         if soup is None:  # This is not iterable, so return promptly
             return markup, metadata['message_links'], metadata['time_mentions']
+
+        math_classes = ['katex', 'katex-display']
+        mention_classes = ['user-mention', 'user-group-mention']
         unrendered_tags = {  # In pairs of 'tag_name': 'text'
             # TODO: Some of these could be implemented
             'br': '',  # No indicator of absence
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -11,7 +11,7 @@ from urllib.parse import urljoin, urlparse
 import dateutil.parser
 import urwid
 from bs4 import BeautifulSoup
-from bs4.element import NavigableString
+from bs4.element import NavigableString, Tag
 from tzlocal import get_localzone
 from urwid_readline import ReadlineEdit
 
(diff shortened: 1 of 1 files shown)


neiljp:
@Rohitth007 I'm sure this code could be tidier, so a refactoring is potentially useful to make this more readable :+1:

That said, given how critical this function is to how messages are rendered - even though we have some tests - I would definitely prefer to see such refactoring split into smaller steps (ie. commits) so that it is clearer when code is simply moved around, when code is moved into a block, minor adjustments are made and so on.

Regarding the specific changes you listed:
* Extracting the classes seems promising, but looking at your collapsed code screenshot, I'd consider doing it for the name, and perhaps even for all elements to make the structure simpler.
* The `cls` definitely seems like a good separate change for clarity or possibly a bugfix, though I'm not familiar with a standard python idiom of using `_` in a comprehension. I'm not sure if it's worth producing a testcase for illustrating that cls is causing a bug, but you may be able to find one to go with that change.

Rohitth007:
@neiljp Thanks for the review! I'll split the code into smaller steps.
Re `cls` I don't think the cls in the comprehension interferes with the other cls. It's just confusing to have 2 `cls` within the same section of code. Unless I'm missing something that's causing a bug?

Rohitth007:
@zulipbot remove "PR awaiting review"

Rohitth007:

zulipterminal/ui_tools/boxes.py
```diff
@@ -1099,14 +1099,12 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
                 # TODO: Address 12-hour format support with application-wide
                 # support for different formats.
                 time_string = local_time.strftime('%a, %b %-d %Y, %-H:%M (%Z)')
+                source_text = f"Original text was {raw_text.strip()}"
+                metadata['time_mentions'].append((time_string, source_text))
```
neiljp, line 1103: I'd suggest squashing this with the "shorten lines" commit that I mentioned floating to just after the first commit in the other comment - the rename is in the first commit, and this is a simplification enabled by that first commit.

You have moved code, which is a minor extra change we could include too.

zulipterminal/ui_tools/boxes.py
```diff
@@ -940,54 +940,59 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
         }
         unrendered_template = '[{} NOT RENDERED]'
         for element in soup:
+            if isinstance(element, Tag):
```
neiljp, line 947: I really like the simplification this enables, though my only thought would be how this isn't part of the if/elif chain. Use of the `tag` prefix names elsewhere imply that this is how it works, but we might consider a comment to the effect of we're caching/naming entries used in all but the first element of the chain to improve legibility/lookup, and a blank line between this conditional and the main if/elif chain.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1129,9 +1129,8 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
                 elif tag == 'div' and any(
                         _ in css_classes for _ in unrendered_div_classes):
                     # UNRENDERED DIV CLASSES
-                    matching_class = (set(unrendered_div_classes)
-                                      & set(css_classes))
-                    text = unrendered_div_classes[matching_class.pop()]
+                    match = set(unrendered_div_classes) & set(css_classes)
+                    text = unrendered_div_classes[match.pop()]
```
neiljp, line 1131: This seems like something we could include in that new simplifying commit, though I'd suggest `matches` since that's more general. We should likely have a note here about the (unlikely) potential for multiple matches.

neiljp (commented):
@Rohitth007 Having the commits separate - even if small - makes it much easier to read, review, pick them individually or combine them :+1: From the 7 commits there's some simpler clearly high value improvements we can extract first before looking at some smaller changes in there, that are not as obviously warranted.

See the inline comments, but basically if we can slightly tidy the naming in the first commit, and combine the line-tidying elements from some later ones into a (new) second commit then it'd be great to get those changes merged (including the current 6th, with my comment addressed). If you rebase the other changes as commits after that then I can pick off and potentially merge the first two and we can discuss the other changes which should then sit cleanly on the new main branch.

Some notes on the remaining changes:
* we can perhaps simplify class checks in any case, since `css_classes` (`tag_classes`) defaults to `[]`. So I'm not sure if we need a check on that variable, as it's clearly a reasonable sequence/container
* I like the simplification from the math/mention changes; however, the benefit isn't clear since there are only two strings and the two elements are now split at different places in the function
* using `_` seems workable, but doesn't seem a much discussed/used idiom in code I've seen for a *wanted* variable, as opposed to one that one is ignoring

zulipterminal/ui_tools/boxes.py
```diff
@@ -940,54 +940,59 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
         }
         unrendered_template = '[{} NOT RENDERED]'
         for element in soup:
+            if isinstance(element, Tag):
+                tag = element.name
+                html_attrs = element.attrs
+                css_classes = html_attrs.get('class', [])
```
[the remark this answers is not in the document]
Rohitth007: Yes having a common `tag_` prefix seems to make sense and also looks clean. Also `raw_text` seems like the odd-one-out here so `tag_text` is clearly a better name.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1037,9 +1037,7 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
                 ))
             elif tag == 'code':
                 # CODE (INLINE?)
-                markup.append((
-                    'msg_code', raw_text
-                ))
+                markup.append(('msg_code', raw_text))
```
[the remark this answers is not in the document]
Rohitth007: The blank line was supposed to be removed in the first commit itself, do you suggest we keep the blank line?

zulipterminal/ui_tools/boxes.py
Rohitth007: :+1:

zulipterminal/ui_tools/boxes.py
Rohitth007: Yes I was considering adding a comment. This would certainly help first-timers :+1:

zulipterminal/ui_tools/boxes.py
Rohitth007: :+1:

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
Rohitth007: From a later comment a blank line seems to be a good choice :+1:

zulipterminal/ui_tools/boxes.py
```diff
@@ -940,54 +940,59 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
...
                 if text:
                     markup.append(unrendered_template.format(text))
-            elif element.name in ('p', 'del'):
+            elif tag in ('p', 'del'):
                 # PARAGRAPH, STRIKE-THROUGH
                 markup.extend(cls.soup2markup(element, metadata)[0])
-            elif (element.name == 'span' and element.attrs
-                  and 'emoji' in element.attrs.get('class', [])):
+            elif (tag == 'span' and html_attrs
+                  and 'emoji' in css_classes):
```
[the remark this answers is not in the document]
Rohitth007: :+1:

zulipterminal/ui_tools/boxes.py
```diff
@@ -1144,6 +1105,43 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
                     f"Original text was {raw_text.strip()}"
                 )
                 metadata['time_mentions'].append((time_string, source_text))
+            elif css_classes:
```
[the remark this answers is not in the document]
Rohitth007: Yes, I just realized that the check for `css_classes` is unnecessary now. Thought that having all classes in the same place would be better. Even the speed gains won't be that helpful as very few messages out of many being loaded will recursively use soup2mark.
Re saving line-length from not indenting would still lead to re-splitting. Still, linear if-else seems like the right way to go considering ordering constraints.

zulipterminal/ui_tools/boxes.py
```diff
@@ -945,48 +945,18 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
...
-                # MATH TEXT
-                markup.append(raw_text)
-            elif (tag == 'span' and html_attrs
-                  and ('user-group-mention' in css_classes
-                       or 'user-mention' in css_classes)):
-                # USER MENTIONS & USER-GROUP MENTIONS
-                markup.append(('msg_mention', raw_text))
+            elif tag in ('strong', 'em'):
+                # BOLD & ITALIC
+                markup.append(('msg_bold', raw_text))
```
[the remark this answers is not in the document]
Rohitth007: unrendered tags has to come after the `img` tag is checked for and since img is inside css_classes, it was moved down.

zulipterminal/ui_tools/boxes.py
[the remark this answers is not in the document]
Rohitth007: I think the ordering should be done according to most used tags/classes so that the loop ends faster, which it is not doing right now.
| Currently | Proposed |
| :----------:  | :-----------: |
| NavString | NavString |
| Unrendered Div classes | Paragraph |
| Custom emoji | Bold-Italics |
| Unrendered Tags | mentions |
| Paragraph | block quotes |
| emoji | Links |
| math | unrendered-tags(has images) |
| mentions | emoji |
| links | custom emoji |
| blockquote | code |
| code | codeblock|
| block-code | lists list-items |
| bold-italics | table |
| lists and list-items | unrendered div(inline img, inline ref, twitter) |
| table | math |
| time | time |

Rohitth007:
@neiljp I have divided the commits into, one, for renaming, a second for line shortening as suggested.
Here I have used `sets` instead of `_` as a different approach. Though to make it even smaller the only way is to extract those 2 list.

I have even added a reordering commit according to this [message](https://github.com/zulip/zulip-terminal/pull/999#discussion_r617483368)

Rohitth007:

Rohitth007:
@neiljp I think we can remove `katex-display` too because `katex` is contained inside `katex-display`

zulipterminal/ui_tools/boxes.py
```diff
@@ -975,18 +975,16 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
...
+            elif tag == 'span' and (
+                 set(('katex-display', 'katex')) & set(tag_classes)):
                 # MATH TEXT
                 markup.append(tag_text)
-            elif (tag == 'span' and tag_attrs
-                  and ('user-group-mention' in tag_classes
-                       or 'user-mention' in tag_classes)):
+            elif tag == 'span' and (
+                 set(('user-group-mention', 'user-mention'))
+                 & set(tag_classes)):
```
neiljp, line 969: I do like how the set approach in the three cases keeps the strings together without requiring the explicit loops from before and like with the unrendered-div case.

Thoughts:
* simpler with `{...}` instead of `set` for the mentions?
* use a standard indent?
* minor: I'm used to us having operators on the following line, but this is OK

zulipterminal/ui_tools/boxes.py
```diff
@@ -954,39 +954,29 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
...
                 # PARAGRAPH, STRIKE-THROUGH
                 markup.extend(cls.soup2markup(element, metadata)[0])
-            elif tag == 'span' and 'emoji' in tag_classes:
-                # EMOJI
-                markup.append(('msg_emoji', tag_text))
-            elif tag == 'span' and (
-                 set(('katex-display', 'katex')) & set(tag_classes)):
-                # MATH TEXT
-                markup.append(tag_text)
+            elif tag in ('strong', 'em'):
```
neiljp, line 964: Could you explain your prioritization of each tag?

neiljp (commented):
@Rohitth007 The diffs and the intention behind them are clearer to read now, without being very small (though of course that's not always bad!) :+1:

I hope you can see the first commit diff on github is highlighted nicely and shows the changes on each line are only specific parts, which is one reason I'd suggest moving that one line change into the second commit.

The second commit does have a lot of slightly different restructuring that might fall into different categories and so different commits; you're welcome to explore that if you can see a pattern now the code is different from the previous iteration.

I'm not convinced about the reordering being necessary, though maybe things like emoji could go together; I suppose this is a text. With the `set` approach then the logic is simpler and the strings are closer together, so I'm not sure that hoisting the strings up into variables (that may not get used) helps.

Regarding the katex lines, that will need more investigation - we currently only do the minimal work to try and include just text, which is not ideal but we need to ensure we don't miss cases with the changes. We could work on including a link to the image that's generated in future.

This is `soup2markup` - check your commit messages :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -940,54 +940,61 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
...
                 # UNRENDERED DIV CLASSES
                 matching_class = (set(unrendered_div_classes)
-                                  & set(element.attrs.get('class')))
+                                  & set(tag_classes))
                 text = unrendered_div_classes[matching_class.pop()]
                 if text:
                     markup.append(unrendered_template.format(text))
-            elif (element.name == 'img'
-                  and element.attrs.get('class', []) == ['emoji']):
+            elif tag == 'img' and tag_classes == ['emoji']:
```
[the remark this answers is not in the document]
Rohitth007: I might have missed this one, I'll move it to the next commit :+1:

zulipterminal/ui_tools/boxes.py
```diff
@@ -1138,14 +1131,12 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
                 # TODO: Address 12-hour format support with application-wide
                 # support for different formats.
                 time_string = local_time.strftime('%a, %b %-d %Y, %-H:%M (%Z)')
+                source_text = f"Original text was {tag_text.strip()}"
+                metadata['time_mentions'].append((time_string, source_text))
```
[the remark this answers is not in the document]
Rohitth007: Yes this reasoning is clearer. I'll only keep the tidying whitespace part then. :+1:

zulipterminal/ui_tools/boxes.py
Rohitth007: wow, I thought we always had to use `set()`. I think now everything might fit in a single line :crossed_fingers:

zulipterminal/ui_tools/boxes.py
Rohitth007: I think the ordering should be done according to most used tags/classes so that the loop ends faster, which it is not doing right now. Many of the message tags are going to be `p` tags as it is text. Since bold and italics are also text related, I have grouped them together. Next most used are `mentions` and `blockquotes` I  feel after which come emoji's. I have also grouped both the emojis together. I wanted to put Links and Images(unrendered tags) before emoji but since custom-emoji's have an `img` tag it had to be processed before. Towards the end I have put least used (time, math, twitter preview) before which I have put code, lists and tables which are moderately used. Position of `unrendered_div` is still debatable  

| Currently | Proposed |
| :----------:  | :-----------: |
| NavString | NavString |
| Unrendered Div classes | Paragraph |
| Custom emoji | Bold-Italics |
| Unrendered Tags | mentions |
| Paragraph | block quotes |
| emoji | emoji |
| math | custom emoji |
| mentions | Links |
| links | unrendered-tags(has images) |
| blockquote | code |
| code | codeblock|
| block-code | lists list-items |
| bold-italics | table |
| lists and list-items | unrendered div(inline img, inline ref, twitter) |
| table | math |
| time | time |

zulipterminal/ui_tools/boxes.py
```diff
@@ -975,18 +975,16 @@ def soup2markup(cls, soup: Any, metadata: Dict[str, Any],
             elif tag in ('p', 'del'):
                 # PARAGRAPH, STRIKE-THROUGH
                 markup.extend(cls.soup2markup(element, metadata)[0])
-            elif (tag == 'span' and tag_attrs
-                  and 'emoji' in tag_classes):
+            elif tag == 'span' and 'emoji' in tag_classes:
```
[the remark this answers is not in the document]
Rohitth007: Yes, I actually meant to say `tag_class` default to [] so `tag_attrs` was unnecessary. I have fixed it

zulipterminal/ui_tools/boxes.py
Rohitth007: No, it doesn't fit in a single line :disappointed: but it looks much better now.

Rohitth007:
@neiljp I have cleaned up the commits as suggested, the 2 commits that you initially wanted to merge are now the first 4 commits to make the commits more readable. Hopefully they are ready to be merged.

neiljp:
@Rohitth007 I just merged the first four commits, which have the greatest impact, ending with d4c17887e117782c6b1b03598b4ec4754368d426 :tada:

For the ordering commit, while I understand the desire to optimize, I've not seen (or heard of) any significant issues with this being too slow - and certainly not enough to move things around as much as in that commit. I'm also not convinced that many of the elements have extra use compared to others. If reordering, I'd prioritize only based on readability at this point (grouping similar entries) - which is one reason why the commits merged already are so helpful :+1:
(the one exception may be `<p>` since that's possibly in every response, though that will depend on the parser I believe)

The last commit wasn't updated to be useful in your last rebase, but with the text being adjacent in each `set` in each conditional, that feels sufficient and so this commit doesn't feel warranted in any case.

Given the above, I'm closing this now - thanks for tidying this up !:+1:

Rohitth007:
:tada: Thanks for merging this, this should make soup2markup much cleaner now!

PR #1000, opened 2021-04-20
helper: Sync unread pm message count in all_pms and user button count.

### LINKED ISSUE #438: Sync unread counts of PMs between All_PM_Buttona and the respective user button.
This is very likely a regression due to recent changes in `user list`. When a new message is sent in a one-to-one PM, the unread count of `All_PM` button is updated immediately. But the unread_count of the user_button for that respective user is delayed for quite some time. Only after waiting, is the user button unread_count also seen to increase.

The possible reason for this behaviour is due to the user_list updating which happend every 60 seconds conflicting with the normal UI updation.

mkp6781:
User buttons list is updated using `update_user_list` before count of user
buttons is modified in `_set_count_in_view`. Now, we do not have to wait
for the presence update before displaying the correct unread count.

Fixes #438

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -167,6 +167,10 @@ def _set_count_in_view(controller: Any, new_count: int,
     if is_open_topic_view:
         topic_buttons_log = controller.view.topic_w.log
         toggled_stream_id = controller.view.topic_w.stream_button.stream_id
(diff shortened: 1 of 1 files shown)


mkp6781:
@zulipbot addd "PR needs review"

mkp6781:

PR #1004, opened 2021-04-21
Add support for stream + topic autocomplete

Ezio-Sarthak:
The PR does the following tasks:
 - Fixes a bug re space-separated typeaheads (commit 1),
 - Extracts all regexes to a new file `regexes.py` (commit 2-5)
 - Adds support for stream+topic autocompletes. (commit 6-9).

Ticks a check box in #448.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -86,9 +86,13 @@ class TestWriteBox:
             ("#**Stream 1>", -2, True, "#**Stream 1>This is a topic**"),
             # Fenced prefix
             ("#**Stream 1**>T", 0, True, "#**Stream 1>Topic 1**"),
+            # Unfenced prefix
+            ("#Stream 1>T", 0, True, "#**Stream 1>Topic 1**"),
+            ("#Stream 1>T", 1, True, "#**Stream 1>This is a topic**"),
             # Invalid stream
             ("#**invalid stream>", 0, False, None),
             ("#**invalid stream**>", 0, False, None),
+            ("#invalid stream>", 0, False, None),
             # Invalid prefix format
             ("#**Stream 1*>", 0, True, None),
             ("#*Stream 1>", 0, True, None),
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -84,9 +84,13 @@ class TestWriteBox:
             ("#**Stream 1>", 1, True, "#**Stream 1>This is a topic**"),
             ("#**Stream 1>", -1, True, "#**Stream 1>Hello there!**"),
             ("#**Stream 1>", -2, True, "#**Stream 1>This is a topic**"),
+            # Fenced prefix
+            ("#**Stream 1**>T", 0, True, "#**Stream 1>Topic 1**"),
             # Invalid stream
             ("#**invalid stream>", 0, False, None),
+            ("#**invalid stream**>", 0, False, None),
             # Invalid prefix format
+            ("#**Stream 1*>", 0, True, None),
             ("#*Stream 1>", 0, True, None),
             # Complex autocomplete prefixes.
             ("(#**Stream 1>", 0, True, "(#**Stream 1>Topic 1**"),
@@ -94,6 +98,7 @@ class TestWriteBox:
             ("@#**Stream 1>", 0, True, "@#**Stream 1>Topic 1**"),
             ("@_#**Stream 1>", 0, True, "@_#**Stream 1>Topic 1**"),
             (":#**Stream 1>", 0, True, ":#**Stream 1>Topic 1**"),
+            ("(#**Stream 1**>", 0, True, "(#**Stream 1>Topic 1**"),
         ],
     )
     def test_generic_autocomplete_stream_and_topic(
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -69,6 +69,51 @@ class TestWriteBox:
         assert not write_box.model.send_typing_status_by_user_ids.called
 
     @pytest.mark.parametrize(
+        "text, state, is_valid_stream, required_typeahead",
+        [
+            ("#**Stream 1>T", 0, True, "#**Stream 1>Topic 1**"),
+            ("#**Stream 1>T", 1, True, "#**Stream 1>This is a topic**"),
+            ("#**Stream 1>T", 2, True, None),
+            ("#**Stream 1>T", -1, True, "#**Stream 1>This is a topic**"),
+            ("#**Stream 1>T", -2, True, "#**Stream 1>Topic 1**"),
+            ("#**Stream 1>T", -3, True, None),
+            ("#**Stream 1>To", 0, True, "#**Stream 1>Topic 1**"),
+            ("#**Stream 1>H", 0, True, "#**Stream 1>Hello there!**"),
+            ("#**Stream 1>Hello ", 0, True, "#**Stream 1>Hello there!**"),
+            ("#**Stream 1>", 0, True, "#**Stream 1>Topic 1**"),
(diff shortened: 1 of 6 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -480,6 +497,27 @@ def autocomplete_streams(self, text: str, prefix_string: str
...
+
+        stream = m.group(1) if m else ''
+
+        if self.model.is_valid_stream(stream):
+            stream_id = self.model.stream_id_from_name(stream)
+            topic_names = self.model.topics_in_stream(stream_id)
+        else:
+            topic_names = []
+
+        stream_prefix = f"#**{stream}>"
```
mkp6781, line 512: Is there a reason you aren't using `prefix_string` directly here. On doing a quick check, I don't think we need this new variable.

mkp6781:
@Ezio-Sarthak Thanks for taking this up. I tested this locally and it seems to work fine 👍🏼.

zulipterminal/ui_tools/boxes.py
Ezio-Sarthak: Thanks. I updated the PR to get the required prefix using `prefix_string`

Ezio-Sarthak:

zulipterminal/ui_tools/boxes.py
```diff
@@ -459,6 +467,25 @@ def generic_autocomplete(self, text: str, state: Optional[int]) -> Optional[str]
         #        anything; this implementation simply chooses the right-most
         #        match of the longest length
         prefix_indices = {prefix: text.rfind(prefix) for prefix in autocomplete_map}
+
+        # Matches autocomplete for stream + topic at the end of a text.
+        m = re.search(REGEX_AUTOCOMPLETE_STREAM_TOPIC, text)
+        m_fenced = re.search(REGEX_AUTOCOMPLETE_STREAM_TOPIC_FENCED, text)
```
prah23, line 473: minor: You could use more descriptive aliases, maybe match and fenced_match? (just to improve readability)

tests/ui_tools/test_boxes.py
```diff
@@ -401,13 +451,13 @@ def test_generic_autocomplete_set_footer(
...
             ("@Human", 1, "@**Human 1**"),
-            ("@Human 1", 0, "@**Human 1**"),
+            ("@Human 1", 0, "@**Human 1**"),  # Space-containing prefix
             ("@_H", 1, "@_**Human 1**"),
             ("@_Hu", 1, "@_**Human 1**"),
             ("@_Hum", 1, "@_**Human 1**"),
             ("@_Huma", 1, "@_**Human 1**"),
             ("@_Human", 1, "@_**Human 1**"),
-            ("@_Human 1", 0, "@_**Human 1**"),
+            ("@_Human 1", 0, "@_**Human 1**"),  # Space-containing prefix
```
prah23, line 460: Can we call these "space-containing text" instead, consistent with their parameter's name?

prah23 (changes requested):
Thanks for working on this @Ezio-Sarthak!
This work and the extension to the autocomplete looks good to me and functions well too :+1:! I've left a couple of minor suggestions in-line.

zulipterminal/config/regexes.py
```diff
@@ -0,0 +1,6 @@
+# All regexes with (*) are inspired from webapp's
+# `static/js/composebox_typeahead.js`
+
+
+# (*) Example text: #**stream name>Topic name
+REGEX_AUTOCOMPLETE_STREAM_TOPIC = r"#\*\*([^*>]+)>([^*]*)$"
```
neiljp, line 6: I would add this to the next commit, but use this commit as a refactor of at least some other regexes, particularly those where the content of the regex is not as critical to understand in the subsequent code.

This constant is not different from in @prah23's feedback?

tests/ui_tools/test_boxes.py
```diff
@@ -67,6 +67,52 @@ def test_not_calling_typing_method_without_recipients(self, mocker, write_box):
...
+            ("#**Stream 1>Hello ", 0, "#**Stream 1>Hello there!**", True),
+            ("#**Stream 1>", 0, "#**Stream 1>Topic 1**", True),
+            ("#**Stream 1>", 1, "#**Stream 1>This is a topic**", True),
+            ("#**Stream 1>", -1, "#**Stream 1>Hello there!**", True),
+            ("#**Stream 1>", -2, "#**Stream 1>This is a topic**", True),
+            # Invalid stream
+            ("#**invalid stream>", 0, None, False),
+            # Invalid prefix format
+            ("#*Stream 1>", 0, None, True),
+            ("#Stream 1>", 0, None, True),
```
neiljp, line 93: This seems like it should complete?

tests/ui_tools/test_boxes.py
```diff
@@ -83,9 +83,12 @@ def test_not_calling_typing_method_without_recipients(self, mocker, write_box):
             ("#**Stream 1>", 1, "#**Stream 1>This is a topic**", True),
             ("#**Stream 1>", -1, "#**Stream 1>Hello there!**", True),
             ("#**Stream 1>", -2, "#**Stream 1>This is a topic**", True),
+            # Fenced prefix
+            ("#**Stream 1**>T", 0, "#**Stream 1>Topic 1**", True),
```
neiljp, line 87: For reference, the web app also supports `#**Stream 1** >`, presumably for the stream complete generating a space after it?

I think this could be an extra feature on top, depending how straightforward it is, since we don't add a space.

neiljp (commented):
@Ezio-Sarthak This looks really useful: it works and that first commit looks like it does fix that issue without an obvious drawback?

tests/ui_tools/test_boxes.py
Ezio-Sarthak: Well, for a user, autocompleting a stream name would be handier than fully writing it, I believe. Also, this doesn't cope up with the web app's way, I think.

tests/ui_tools/test_boxes.py
Ezio-Sarthak: Yeah. I just quickly tested some autocompletes in the web app. All of them added a space in the end. For us, that might be a future normal, I think :)

zulipterminal/config/regexes.py
Ezio-Sarthak: Ah yes. Adding small understandable regexes and integrating them would be more apt. :+1:

Re your latter point, I think Hari was referring to the local variable renaming, particularly, `m_fenced` to `match_fenced` that are used in the next commit.

neiljp (commented):
@Ezio-Sarthak The refactor makes this cleaner and close to merging :+1: 
As per inline, we don't need to add the regexes until we use them though.

We discussed the `#streamname>` autocomplete being OK on the stream; could you file an issue or start a discussion regarding the space-after-completion point? Maybe there's discussion re the web app that's relevant?

zulipterminal/config/regexes.py
```diff
@@ -0,0 +1,24 @@
...
+
+# (*) Stream and topic regexes
+REGEX_STREAM_NAME = r"([^*>]+)"
+REGEX_TOPIC_NAME = r"([^*]*)"
+
+
+# (*) Example: #**zulip-terminal
+REGEX_STREAM_FENCED_HALF = r"#\*\*{stream}".format(stream=REGEX_STREAM_NAME)
+# (*) Example: #**zulip-terminal**
+REGEX_STREAM_FENCED = r"#\*\*{stream}\*\*".format(stream=REGEX_STREAM_NAME)
```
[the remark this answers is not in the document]
neiljp: Some of these still seem to be present?

neiljp:
@Ezio-Sarthak Thanks for getting this done and making the updates to this :+1: I've shifted the stream/topic regexes to the first commit they're used to hopefully make it clearer, made a few commit text adjustments and am merging now :tada:

Issue #965, opened 2021-03-28
TRACKING: Update to use modern API elements (retaining support to 2.1)


neiljp:
There are various situations where we currently use older elements of the API which have been deprecated. We should migrate to the newer forms, while possibly retaining (or extending) compatibility down to our stated lowest supported version if the server is older (from feature level `None` to `1+`, ie. the current dev release).

Some API elements to consider: (more likely can be added)
* [ ] message destinations (recipient id rather than name)
* [ ] various stream properties
* [ ] various message properties
* [ ] `subject` -> `topic` for stream messages
* [ ] handling `subject_links` (we may be missing support for 2.1-2.2)

Other than the number of different API elements to review and update, the other aspect of this work is considering where the external and internal data names are updated/handled. Currently these are the same in many cases, but this may not be straightforward to support across multiple server versions and would be stable if there was a mapping layer between them in any case (rather than just using the raw server data).

Related tasks: (possibly to expand into separate issues)
* Add a TypedDict for useful objects
* Adjust TypedDicts to be total (in the API)
* Adjusting fixtures to match

zulipbot:
Hello @zulip/server-refactoring members, this issue was labeled with the "area: refactoring" label, so you may want to check it out!

Ezio-Sarthak:
Makes sense. I'd also like the intent of refactoring different API formats to restrict performing amendments in a single dedicated file (for instance, `api_types.py`).

Re the handling of `subject_links` (last point), I've added some initial refactor commits providing support for `modernizing message API response` (as discussed), in an unrelated PR #867. This could be extended in general for different API responses and could lend a helping hand in retaining support to 2.1+ while migrating to newer forms :)

neiljp:
@Ezio-Sarthak Re your first point, I'm not sure quite what you're saying, but I expect we'll end up with feature-level-dependent types in `api_types.py`. Were you meaning to have more than that?

neiljp:
The first point in the list looks like #579.

Ezio-Sarthak:
> @Ezio-Sarthak Re your first point, I'm not sure quite what you're saying, but I expect we'll end up with feature-level-dependent types in `api_types.py`. Were you meaning to have more than that?

@neiljp To clarify, my intent was that we could have mapping layers between different feature-level dependencies, such that the same feature level data (TypedDicts) would be in a single file, so as to head for the `total=True` aspect for TypedDicts.

[...]

PR #1006, opened 2021-04-21
Use `user_id`s as message destinations

prah23:
There are 4 commits to this PR:
- The first commit refactors `update_recipient_emails()` to `update_recipients()`, also updating the recipients' `user_id`s now, which is a preparatory step for the third commit. A test has also been added for this method.
- The second commit adds a test fixture for `user_id_email_dict` as a prep commit for the third commit.
- The third commit migrates from using delivery emails as message destinations to using `user_id`s instead. Tests updated.
- The fourth commit deprecates the `emails` parameter from `private_box_view()`. Tests updated.
 
The abstraction between the narrow model and the Composition API type is maintained by modifying the corresponding connections to the narrow model such that the delivery emails are still in use in the narrow model currently, making this a message-destination-only change. 

This change also serves as a prelude to the larger scale narrow model refactoring, which needs `Model` and narrow structure abstraction.

This would accommodate the first point under #965.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -339,6 +339,34 @@ class TestWriteBox:
         )
 
     @pytest.mark.parametrize(
+        "header, expected_recipient_emails, expected_recipient_user_ids",
+        [
+            case(
+                "Human 1 <person1@example.com>",
+                ["person1@example.com"],
+                [11],
+                id="single_recipient",
+            ),
+            case(
+                "Human 1 <person1@example.com>, Human 2 <person2@example.com>",
+                ["person1@example.com", "person2@example.com"],
+                [11, 12],
+                id="multiple_recipients",
+            ),
+        ],
+    )
+    def test_update_recipients(
+        self, write_box, header, expected_recipient_emails, expected_recipient_user_ids
+    ):
+        write_box.private_box_view()
+        write_box.to_write_box.edit_text = header
+
+        write_box.update_recipients(write_box.to_write_box)
+
+        assert write_box.recipient_emails == expected_recipient_emails
+        assert write_box.recipient_user_ids == expected_recipient_user_ids
+
+    @pytest.mark.parametrize(
         "text, state",
         [
             ("Plain Text", 0),
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -69,7 +69,7 @@ class TestWriteBox:
 
     def test_not_calling_typing_method_without_recipients(self, mocker, write_box):
         write_box.model.send_typing_status_by_user_ids = mocker.Mock()
-        write_box.private_box_view(emails=[], recipient_user_ids=[])
+        write_box.private_box_view(recipient_user_ids=[])
         # Set idle_status_tracking to True to avoid setting off the
         # idleness tracker function.
         write_box.idle_status_tracking = True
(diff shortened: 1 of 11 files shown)


preetmishra (commented):
@prah23 Thanks for taking this up. This seems to work well and the changes look good to me as well! :+1: 

With `user_id`s in place and given that we could build back `recipient_email`s from the `user_id`s, should we drop the `recipient_email` attribute in boxes.py? 

We could potentially squash the changes pre-merge. I have left a few minor in-line comments.

prah23:
Thanks for the review, and the educative point about the AAA pattern, @preetmishra!
I've incorporated your suggestions and have added a 4th commit deprecating the `emails` parameter from `private_box_view`.

preetmishra (commented):
@prah23 Thanks for the update! The last commit is a good cleanup. We could label it as a `refactor:`. Overall, the PR looks good to me! :+1:

There are two concerns that I would like to point out:
* When opening a saved draft with an invalid user email, the `To:` field renders empty. 
* When sending a PM to an invalid recipient, the footer message gets updated to "Cannot send message without specifying recipients." than "Invalid recipient(s) - ...".

This is likely because you're skipping out invalid emails while creating the user_ids from the regex grepped emails in `update_recipients()`.

prah23:
This PR is now up-to-date @preetmishra.

Thanks to #937, both of your concerns should no longer persist since we restrict the user from sending a message and saving a draft too if the recipients are invalid. Only if the recipients are valid, they will be updated in `self.recipient_user_ids` and `self.recipient_emails` and then the relevant action will proceed.

tests/ui_tools/test_boxes.py
```diff
@@ -196,6 +196,46 @@ def test_footer_notification_on_invalid_recipients(
...
+                ["person1@example.com", "person2@example.com"],
+                [11, 12],
+                id="multiple_recipients",
+            ),
+            case(
+                "Human 1person1@example.com",
+                ["1person1@example.com"],
+                [],
+                id="invalid_recipient",
+            ),
```
prah23, line 219: I'm removing these test cases, since `self.update_recipients()` will only be called if all the recipients are valid, eliminating the possibility of invalid emails being passed into it.

zulipterminal/ui_tools/boxes.py
```diff
@@ -725,14 +727,10 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
                         return key
-                    # We extract emails into self.recipient_emails only once we know
+                    # We extract recipients' user_ids and emails only once we know
                     # that all the recipients are valid, to avoid including any
                     # invalid ones.
-                    self.update_recipient_emails(self.to_write_box)
-                    users = self.model.user_dict
-                    self.recipient_user_ids = [
-                        users[email]["user_id"] for email in self.recipient_emails
-                    ]
```
prah23, line 735: This is part of the `update_recipients` method now, so removing these lines as well.

prah23 (commented):
I've made a couple of changes since the #937 merge, I'm leaving my thoughts on these below, please let me know if you think they need changes.

tests/ui_tools/test_boxes.py
prah23: I seemed to have marked the wrong lines here, I've let the "multiple_recipients" case stay, I've removed another case below "invalid_recipient", which was valid and invalid mixed together.

preetmishra (commented):
@prah23 Thanks for carrying this forward and resolving the issues that were related elsewhere!  :+1: This works well locally.

@neiljp Other than the minor in-line comments, this looks good to me.

prah23:
Thanks for the reviews, @preetmishra!
Updated as per review.

neiljp:
@prah23 I made a minor test change and commit text change and am merging this now - thanks! :tada:

Other than the ongoing migration towards ids and improving the narrow structure, I expect we'll want to wrap the access to the user_id_email_dict with a method, but this refactoring is useful to have right now and we can build on this :+1:

PR #1017, opened 2021-04-29
docs: Add autocomplete info section in FAQ.

Ezio-Sarthak:
This PR adds a section `How do I use autocompletes in ZT?`, which lists and explains the working of currently supported autocompletes in ZT, grouped by different compose boxes
 - Message body autocompletes
 - Other (Stream/Topic/PM recipients) autocomplete options

**NOTE:** Currently I'm not sure if the first bullet title is user-friendly, and so would appreciate feedback/suggestions for the same :)

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -68,6 +68,56 @@ Since we expect the above to be straightforward for most users and it allows the
 However, we are certainly likely to move towards a system to make access of the different servers simpler, which should be made easier through work such as [zulip-terminal#678](https://github.com/zulip/zulip-terminal/issues/678).
 In the longer term we may move to multiple servers per session, which is tracked in [zulip-terminal#961](https://github.com/zulip/zulip-terminal/issues/961).
 
+## How do I use autocompletes in ZT?
+
+Composing messages could be a tiring task when you have to type all things by yourself. Autocompletes come to the rescue here! ZT currently supports some basic autocompletes, for different compose boxes.
+
+### Hotkeys triggering autocomplete
+
+We use `ctrl` + **`f`** and `ctrl` + **`r`** for cycling through autocompletes (**forward** & **reverse** respectively).
+
+**NOTE:** We don't use `tab`/`shift tab` (although it is widely used elsewhere) for cycling through matches. However, recall that we do use `tab` to cycle through recipient and content boxes. (See [hotkeys for composing a message](https://github.com/zulip/zulip-terminal/blob/main/docs/hotkeys.md#composing-a-message))
+
+### Example: Using autocomplete to add a recognized emoji in your message content
+
+1. Type a prefix designated for an autocomplete (e.g., `:` for autocompleting emojis).
+2. Along with the prefix, type the initial letters of the text (e.g., `air` for `airplane`).
+3. Now hit the hotkey. You'd see suggestions being listed in the footer (a maximum of 10) if there are any.
+4. Cycle between different suggestions as described in above hotkey section. (Notice that a selected suggestion is highlighted)
+5. Cycling past the end of suggestions goes back to the prefix you entered (`:air` for this case).
+
+![selected_footer_autocomplete](https://user-images.githubusercontent.com/55916430/116669526-53cfb880-a9bc-11eb-8073-11b220e6f15a.gif)
+
+### Autocomplete in the message content
+
+As in the example above, a specific prefix is required to indicate which action to perform (what text to insert) via the autocomplete:
+
+|Action|Prefix text(s)|Autocompleted text format|
+| :--- | :---: | :---: |
+|Mention user or user group|`@`|`@*User group*` or `@**User name**`|
+|Mention user|`@**`|`@**User name**`|
+|Mention user group|`@*`|`@*User group*`|
+|Mention user (silently)|`@_`, `@_**`|`@_**User name**`|
+|Link to stream|`#`, `#**`|`#**stream name**`|
+|Insert emoji|`:`|`:emoji_name:`|
+
+**NOTE:** Currently, the above autocompletes ignore a text containing whitespaces if it is specified as a prefix for autocompletion. (e.g., `#test he` wouldn't lead to `#**test here**`) See also [#925](https://github.com/zulip/zulip-terminal/issues/925).
+
+### Autocomplete of message recipients
+
+Since each of the stream (1), topic (2) and private message recipients (3) areas are very specific, no prefix must be manually entered and values provided through autocomplete depend upon the context automatically.
+
+![Stream header](https://user-images.githubusercontent.com/55916430/118403323-8e5b7580-b68b-11eb-9c8a-734c2fe6b774.png)
+
+**NOTE:** The stream box prefix is automatic:
+* `#` (with stream color) if the stream is valid and public,
+* `P` (with stream color) if the stream is valid and private,
+* `✗` if the stream is invalid.
+
+![PM recipients header](https://user-images.githubusercontent.com/55916430/118403345-9d422800-b68b-11eb-9005-6d2af74adab9.png)
+
+**NOTE:** If a PM recipient's name contains comma(s) (`,`), they are currently treated as comma-separated recipients.
+
 ## Unable to render symbols
 
 If some symbols don't render properly on your terminal, it could likely be because of the symbols not being supported on your terminal emulator and/or font.


Ezio-Sarthak:
@zulipbot add "area: documentation"

docs/FAQ.md
```diff
@@ -66,6 +66,58 @@ One session of Zulip Terminal represents a connection of one user to one Zulip s
...
+
+![selected_footer_autocomplete](https://user-images.githubusercontent.com/55916430/116669526-53cfb880-a9bc-11eb-8073-11b220e6f15a.gif)
+
+### List of designated prefixes currently supported in ZT
+
+Now that you have a basic understanding of how autocompletes in ZT work, let's explore what all sorts of autocomplete prefixes ZT support, as of now. Note that the kinds of prefix supported also differ within each compose box. Thus, the prefixes listed below are grouped by type of compose box.
+
+You could refer to the above section to try out each of the below prefixes!
+
+#### Message write box
```
prah23, line 97: This could maybe simply be referred to as "Message compose" or in a verbose manner, "Message compose autocomplete"? Inferring autocomplete that can be used during composing. Since message compose is limited to `msg_write_box`, it would sound simpler to understand with our intent being conveyed.

docs/FAQ.md
```diff
@@ -66,6 +66,58 @@ One session of Zulip Terminal represents a connection of one user to one Zulip s
...
+| :--- | :---: | :---: |
+|Mention a user|`@`, `@**`|`@**User name**`|
+|Mention a user group|`@*`|`@*User group*`|
+|Silently mention a user|`@_`, `@_**`|`@_**User name**`|
+|Get a stream|`#`, `#**`|`#**stream name**`|
+|Get an emoji|`:`|`:emoji_name:`|
+
+**NOTE:** Currently, the above autocompletes ignore a text containing whitespaces if it is specified as a prefix for autocompletion. (e.g., `#test he` wouldn't lead to `#**test here**`) See also [#925](https://github.com/zulip/zulip-terminal/issues/925).
+
+#### Other write boxes
```
prah23, line 109: Maybe "other autocomplete options"?

docs/FAQ.md
```diff
@@ -66,6 +66,58 @@ One session of Zulip Terminal represents a connection of one user to one Zulip s
...
+
+All other write boxes in compose box don't require a prefix, i.e., the prefix for them is an empty string (`''`). Autocompletes here are contextual and specific to a particular write box.
```
prah23, line 111: I'd suggest explaining how message recipients work here (stream, topic, pm recipients), as in a mention that we have "headers" or "fields" where we enter these (a screenshot might be helpful here like the GIF you have above), so that it serves as context to the table below, keeping the table compact as it currently stands.

"Fields" is a fairly easy term to understand for something that accepts input.

prah23 (changes requested):
@Ezio-Sarthak, this looks fairly informative 👍

Leaving a couple of comments inline as suggestions to make this more user oriented, let me know how they sound to you.

docs/FAQ.md
Ezio-Sarthak: Right, I overlapped with that thought too recently. Though since selecting streams and topics may also be considered in "composing a message", I wondered if that's an apt option. Nevertheless, we could explore it with feedback. I'd share this suggestion and one other I had in mind for this in CZO, and would appreciate suggestions from others as well.

docs/FAQ.md
Ezio-Sarthak: Seeing your below point, I am a bit more inclined to `Other autocomplete fields`. What do you think about it?

prah23: That seems a little confusing, if you want to go in that direction maybe `Other autocomplete-enabled fields`?

Ezio-Sarthak: Ah, that makes sense. Let's go with "Other autocomplete options" then :+1:.

docs/FAQ.md
```diff
@@ -66,6 +66,65 @@ One session of Zulip Terminal represents a connection of one user to one Zulip s
...
+
+|S. No.|Field|Autocompleted text format|
+| :---: | :--- | :---: |
+|1|Stream|`Stream name`|
+|2|Topic|`Topic name`|
+|3|PM recipient **\***|`Name <email>`|
+
+**\*** If a PM recipient's name contains comma(s) (`,`), they would be treated as comma-separated recipients.
+
+Stream message compose (for **1** and **2**):
```
prah23, line 122: Since we're being specific, it might be better if we use specific terms as discussed on the stream, like "specifying stream and topic" and "specifying private message recipients". 

The other point I just noticed is, the cursor in your screenshots remains on the message content, could you perhaps put screenshots with the cursor on these fields? It might help with emphasizing the field we're describing here.

neiljp (commented):
@Ezio-Sarthak Apparently I started a review earlier, so here's what I wrote then for reference - I think most points are still relevant.

docs/FAQ.md
```diff
@@ -66,6 +66,58 @@ One session of Zulip Terminal represents a connection of one user to one Zulip s
 
 Since we expect the above to be straightforward for most users and it allows the code to remain dramatically simpler, we are unlikely to support multiple Zulip servers within the same session in at least the short/medium term. However, we are certainly likely to move towards a system to make access of the different servers simpler, which should be made easier through work such as [zulip-terminal#678](https://github.com/zulip/zulip-terminal/issues/678).
 
+## How do I use autocompletes in ZT?
+
+Composing messages could be a tiring task when you have to type all things by yourself. Autocompletes come to rescue here! ZT currently supports some basic autocompletes, for different compose boxes.
+
+### Hotkeys triggering autocomplete
+
+In most places, hitting `tab` allows toggling autocompletes. In ZT, however, `ctrl f` and `ctrl r` are used for toggling purpose:
```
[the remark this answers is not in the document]
Ezio-Sarthak: Re shitf+tab, could you elaborate our use-case for it? I tried some things but wasn't able to figure out its working.

docs/FAQ.md
```diff
@@ -66,6 +66,58 @@ One session of Zulip Terminal represents a connection of one user to one Zulip s
...
+* `ctrl f` allows switching between autocomplete suggestions in a forward manner (see the GIF below).
+* `ctrl r` does the same job but in reverse manner.
+
+### Understanding the working
+
+Below are some basic steps to work with autocompletes in ZT (assuming that the user is composing a message):
+1. Type a prefix designated for an autocomplete (e.g., `:` for autocompleting emojis).
+2. Along with the prefix, type the initial letters of the text (e.g., `air` for `airplane`).
+3. Now hit the hotkey. You'd see suggestions being listed in the footer (if there are any).
+4. Cycle between different suggestions as described in above hotkey section. (Notice that a selected suggestion is colored in **black**)
```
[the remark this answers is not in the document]
Ezio-Sarthak: No, I would say, as we use the default `"code"` palette for this, uniform throughout all themes.

docs/FAQ.md
```diff
@@ -66,6 +66,58 @@ One session of Zulip Terminal represents a connection of one user to one Zulip s
...
+
+Now that you have a basic understanding of how autocompletes in ZT work, let's explore what all sorts of autocomplete prefixes ZT support, as of now. Note that the kinds of prefix supported also differ within each compose box. Thus, the prefixes listed below are grouped by type of compose box.
+
+You could refer to the above section to try out each of the below prefixes!
+
+#### Message write box
+
+|Prefix Type|Prefix text(s)|Autocompleted text format|
+| :--- | :---: | :---: |
+|Mention a user|`@`, `@**`|`@**User name**`|
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yeah. `@**` is assigned to `autocomplete_users` and `@` is assigned to `autocomplete_mentions`. While some are specified explicitly in the autocomplete_map, I see it's better to include the `@` to user groups too :+1:. Here's what we have:
```python
        autocomplete_map = OrderedDict([
                ('@_', self.autocomplete_users),
                ('@_**', self.autocomplete_users),
                ('@', self.autocomplete_mentions),
                ('@*', self.autocomplete_groups),
                ('@**', self.autocomplete_users),
                ('#', self.autocomplete_streams),
                ('#**', self.autocomplete_streams),
                (':', self.autocomplete_emojis),
            ])
```

docs/FAQ.md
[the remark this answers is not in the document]
Ezio-Sarthak: Do you mean grouping the recipient headers (as done currently)?

docs/FAQ.md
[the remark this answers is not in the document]
neiljp: We don't currently use `shift tab`, but in my past experience it's used for cycling backwards, ie. opposite of `tab`.

Originally I thought we could mention it since if there was a reason we didn't use `shift tab` then that may help explain our current key setup in this kind of document. Unfortunately not :(

docs/FAQ.md
```diff
@@ -78,6 +78,18 @@ In ZT, we use `ctrl f` and `ctrl r` for cycling through autocompletes:
 
 **NOTE:** We don't use `tab` for cycling through matches (although it is widely used for the same elsewhere). However, recall that we do use `tab` to cycle through recipient and content boxes. (See [hotkeys for composing a message](https://github.com/zulip/zulip-terminal/blob/main/docs/hotkeys.md#composing-a-message))
 
+### Understanding the working
+
+Below are some basic steps to work with autocompletes in ZT (assuming that the user is composing a message):
```
neiljp, line 83: The text before the numbered list says little except composing text in the "content box", so we can probably combine the section title and this text together.

Maybe: (too long?)
"Example: Using autocomplete to add a recognized emoji in your message body"
(the extra text re the gif could then be removed too)

docs/FAQ.md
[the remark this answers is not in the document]
neiljp: It might be simpler to indicate the selected value is highlighted?

We could also add a 5th entry to indicate that cycling "past the end" goes back to the prefix you enter.

zulipterminal/ui.py
```diff
@@ -83,6 +83,7 @@ def set_footer_text(self, text_list: Optional[List[Any]]=None,
             text = self.get_random_help()
         else:
             text = text_list
+            print(text)
```
neiljp, line 86: Oops :)

docs/FAQ.md
```diff
@@ -90,6 +90,24 @@ See the below GIF to better sync with above steps.
 
 ![selected_footer_autocomplete](https://user-images.githubusercontent.com/55916430/116669526-53cfb880-a9bc-11eb-8073-11b220e6f15a.gif)
 
+### List of designated prefixes currently supported in ZT
+
+Now that you have a basic understanding of how autocompletes in ZT work, let's explore what all sorts of autocomplete prefixes ZT support, as of now. Note that the kinds of prefix supported also differ within each compose box. Thus, the prefixes listed below are grouped by type of compose box.
+
+You could refer to the above section to try out each of the below prefixes!
```
neiljp, line 97: This style of text is probably more appropriate in a tutorial.

Probably all we need is text somewhere which indicates that:
* in the message body (I don't think that word came up, but it seems OK?) we use prefixes
* for the recipient boxes (1, 2, 3) the prefixes are assumed

Note that with recipient boxes there is no "prefix" (as above bullets) so perhaps this title could be "Areas with autocomplete support" ?

docs/FAQ.md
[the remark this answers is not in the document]
neiljp: I wonder if it's worth treating `@` with it's own row, indicating it matches users and groups (eg. `@term` on czo).

docs/FAQ.md
```diff
@@ -108,6 +108,29 @@ You could refer to the above section to try out each of the below prefixes!
 
 **NOTE:** Currently, the above autocompletes ignore a text containing whitespaces if it is specified as a prefix for autocompletion. (e.g., `#test he` wouldn't lead to `#**test here**`) See also [#925](https://github.com/zulip/zulip-terminal/issues/925).
 
+#### Other autocomplete options
+
+All other message composing areas that support autocompletion don't require a prefix, i.e., the prefix for them is an empty string (`''`). Autocompletes here are mostly contextual and specific. See the below snapshots to better distinguish between various compose fields.
```
neiljp, line 113: Rather than prefixes being the empty string, we can just not mention prefixes in this case, or that they're taken from the context (you have "contextual" :+1:).

docs/FAQ.md
```diff
@@ -108,6 +108,29 @@ You could refer to the above section to try out each of the below prefixes!
...
+
+|S. No.|Field|Autocompleted text format|
+| :---: | :--- | :---: |
+|1|Stream|`Stream name`|
+|2|Topic|`Topic name`|
+|3|PM recipient|`Name <email>`|
```
neiljp, line 119: Given you have two images for Streams and PMs, I'd suggest skipping the text above and this table, and having two sections referring to 1, 2 and 3 in the text in each section. The table doesn't really give much extra information here.

The notes can also fit into the sections.

neiljp (commented):
@Ezio-Sarthak This is looking close, just rather large and verbose for an FAQ answer, so if we can trim it and keep it accurate and useful, that would be great :+1: (my ideas for that are in the comments)

docs/FAQ.md
Ezio-Sarthak: Makes sense :+1:

zulipterminal/ui.py
Ezio-Sarthak: Apologies :sweat_smile:

docs/FAQ.md
Ezio-Sarthak: Sounds good. I was just framing the two bullets in a similar way. I wanted to query if recipient boxes sound internal? Does "recipient header" make sense?

docs/FAQ.md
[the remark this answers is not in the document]
Ezio-Sarthak: Agreed.

neiljp:
@Ezio-Sarthak Thanks for this - I'm sure users will benefit from this :+1: Merging this now after your agreement on the minor adjustments I made via the branch I referred to in the stream :tada:

Issue #1005, opened 2021-04-21
TRACKING: Issues for smaller terminal sizes down to 80x24 (columns x rows)


neiljp:
Known issues, in approximate priority order:
* [ ] 'All messages' is not accessible from a panel context
* [ ] Depending on user name/email and organization name/url, this can disappear from the top of the UI in small widths
* [ ] Message display is too narrow without autohide mode (manually) enabled
* [ ] Popups do not scale down appropriately if window sizes are adjusted #1008 
* [ ] Autohide: clicking on a message does not hide side panels [minor? mouse] 
* [x] Autohide: side panels not hidden when narrowing with shortcut keys (f/#/P) [fixed by #1100]
* [x] Help menu does not render properly (descriptions in left column are long and do not wrap) [fixed by #1018]

If you experience other issues, please add comments below.

PR #1018, opened 2021-04-30
popup: core: Use linear scaling for small terminal sizes.

Rohitth007:
This commit is a step towards solving #1005 especially the first
2 points. This changes the fixed scaling fraction of 3/4 to one that
is linear in the range 80 - 100 column width:

* The scaling fraction increases from 3/4 to 1 (full-width) as the
available size increases from 80 to 100. This increases the range of
sizes for which the popup is rendered properly.
* For width's smaller than 80 it defaults to full width.
* For width's greater than 100 it defaults to 3/4.

The values 80 and 100 are given by:
* MIN_SUPPORTED_POPUP_WIDTH = 80
* MAX_LINEAR_SCALING_WIDTH = 100

The popup height has a fixed scaling factor of 2/3.

Tests added

80 < Width < 100

Width = 80

Width > 100 (Full screen)

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -343,3 +343,19 @@ class TestController:
             num_after=0, num_before=30, anchor=10000000000)
         create_msg.assert_called_once_with(controller.model, msg_ids)
         assert controller.model.index == {'search': msg_ids}
+
+    @pytest.mark.parametrize('screen_size, expected_popup_size', [
+        ((150, 90), (3 * 150 // 4, 3 * 90 // 4)),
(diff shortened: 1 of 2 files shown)


neiljp (commented):
@Rohitth007 See my feedback in the stream, but I'm essentially interested in whether we should scale the height based on the width, or treat it differently based on another rule.

neiljp:
@Rohitth007 Adjusted comment and expressions for clarity, and commit text slightly, and just merged :tada:
(the change should be visible here)

PR #1019, opened 2021-05-01
WIP: Action for dev env setup test

neiljp:
I just threw this together based on an extension of the pytest matrix, to see if we can test installing the build environment using the methods we currently suggest, ie. pip, pipenv, make.

This was inspired by the discussion at #**zulip-terminal>Dev Setup Installation**.

This raises a number of points:
* current pip has a clever back-tracking dependency resolver - but so much that it checks ~30 versions of coverage!
* pipenv fails to resolve the conflict with the latest versions of autopep8 and flake8 matching our pinned versions, though one is possible via pip
* pipenv fails differently on macos?
* unsurprisingly make is both simple and works - it uses the pip approach underneath
* pypy cannot be used as a standard dev environment, as it lacks support for mypy via typed-ast

We can add something like this as an on-demand action (workflow_dispatch) to avoid running on every push and failing CI, or perhaps also on push to main branches once this workflow is passing?

Issues to consider arising from this:
* How best to handle these dev dependencies? (or others generally?)
* Add a note that pypy cannot be used for development as per above reason
* Is all the pip backtracking something we can simplify? (maybe just with better versions/dependency options?)
* Do we really want to keep pipenv?

I'd prefer if the matrix used in this could be reused, but this needs to be in a separate file if running via workflow_dispatch and not on push?

### DIFF
--- a/.github/workflows/setup-dev-env.yml
+++ b/.github/workflows/setup-dev-env.yml
@@ -1,0 +1,50 @@
+name: Test setting up dev environment
+
+on: [push, workflow_dispatch]
+
+jobs:
+  setup:
+    strategy:
+      # Not failing fast allows all matrix jobs to try & finish even if one fails early
+      fail-fast: false
+      matrix:
+        env:
+          - {PYTHON: 3.6, OS: ubuntu-latest, NAME: "CPython 3.6 (ubuntu)"}
+          - {PYTHON: 3.7, OS: ubuntu-latest, NAME: "CPython 3.7 (ubuntu)"}
+          - {PYTHON: 3.8, OS: ubuntu-latest, NAME: "CPython 3.8 (ubuntu)"}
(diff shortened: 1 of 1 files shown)


neiljp:
For results see: https://github.com/neiljp/zulip-terminal/actions/runs/801239198

neiljp:
Following the merge of commits to bump the versions of pytest+plugins (including a rather outdated pytest-cov) and flake8, this PR is closer to passing more cleanly to ensure a dev setup works on different platforms:
* no pip coverage backtracking (pytest-cov upgrade related?)
* no pip flake8/autopep8 backtracking (new flake8 version)
* pipenv now passes with Linux python 3.7+, but still fails on 3.6 and macos

Latest action results are available here: https://github.com/neiljp/zulip-terminal/actions/runs/807568894

PR #1020, opened 2021-05-01
Tidy up reaction tests (prep for reaction event improvements)

neiljp:
Feedback welcome!

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1472,13 +1472,18 @@ class TestModel:
         assert model.index == previous_index
         assert not model._update_rendered_view.called
 
-    def test__handle_reaction_event_add_reaction(
+    @pytest.mark.parametrize("op, expected_number_after", [
+        ("add", 2),
+        ("remove", 1),  # Removed emoji doesn't match, so length remains 1
+    ])
+    def test__handle_reaction_event_for_msg_in_index(
         self, mocker, model,
         reaction_event_factory, reaction_event_index_factory,
+        op, expected_number_after,
         event_message_id=1,
     ):
         reaction_event = reaction_event_factory(
-            op="add", message_id=event_message_id,
+            op=op, message_id=event_message_id,
         )
         model.index = reaction_event_index_factory(
             [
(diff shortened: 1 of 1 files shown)


tests/model/test_model.py
```diff
@@ -1416,11 +1416,15 @@ def reaction_event_index_factory(self):
...
         """
-        def _factory(msgs: List[Tuple[int, List[Tuple[int, str, str, str]]]]):
+        MsgsType = List[Tuple[int, Optional[List[Tuple[int, str, str, str]]]]]
+
+        def _factory(msgs: MsgsType):
             return {
                 'messages': {
-                    message_id: {
+                    message_id:
+                    {} if reactions is None else {
```
zee-bit: This is a pretty neat logic for simplifying generating both indexed/non-indexed reactions. :+1:

tests/model/test_model.py
```diff
@@ -1513,8 +1495,8 @@ def test__handle_reaction_event_remove_reaction(
 
         model._handle_reaction_event(reaction_event)
 
-        # Removed emoji doesn't match, so length remains 1
-        assert len(model.index['messages'][event_message_id]['reactions']) == 1
+        end_reactions = model.index['messages'][event_message_id]['reactions']
+        assert len(end_reactions) == expected_number_after
```
zee-bit, line 1499: "reactions_after_event" might be a better indication than "end_reactions" here?

zee-bit (commented):
This is a very clean refactoring of the tests. I like the idea of combining the tests for both `op` :+1:
Below is a few minor feedback, but I approve the changes made in this PR in general.

tests/model/test_model.py
```diff
@@ -1391,20 +1391,31 @@ def test__update_rendered_view_change_narrow(self, mocker, model, subject,
...
+              'type': 'reaction',
+              'op': op,
+            }
+        return _factory
+
+    @pytest.fixture
+    def reaction_event_index(self):
+        """
+        Minimal index to test reaction events
+        """
```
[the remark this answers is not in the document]
neiljp: Perhaps eventually. Keeping fixtures locally can make it simpler, but also:
* for the initial refactoring it is hopefully clearer in transitioning from a parametrize to a fixture in terms of comparing lines
* they are only used in this module, so one doesn't need to seek them out across files
* they won't be imported in other tests (other files)
* if they are needed elsewhere, we can always move them across - meanwhile we know they're *only* used here
* we could even consider moving other fixtures *out* of conftest.py (notably API fixtures?)

tests/model/test_model.py
neiljp: I agree we could use that. I normally do err towards more verbose, particularly in code, but when we're pushed for space in tests and the variable is used on the next line this is not so critical.

There is always the option we could move towards a longer line length like ~88, but we might want to wait for any potential transition to using something like `black`.

neiljp:
@zee-bit Thanks for the feedback :+1: Merging this now so we can get closer to developing them further to test the improvements/fixes themselves.

PR #1023, opened 2021-05-06
buttons: Change text_color to text_style.

klfoulk16:
Rename TopButton arg text_color to text_style to better align with what it actually represents. This may also require renaming the `color` variable that is used throughout buttons.py and views.py (along with various tests) to `style`.


neiljp:
@klfoulk16 Thanks for looking at this last year :+1: I recently merged #1088, which is a larger tidy-up along these lines, so I'm closing this now.

PR #1024, opened 2021-05-08
Improve katex/math text rendering in ZT

Ezio-Sarthak:


Ezio-Sarthak:

neiljp (commented):
@Ezio-Sarthak This should definitely improve rendering, though without good math rendering directly (if possible!) I'm not sure how strong ZT is for maths. We should probably add an issue to indicate this limitation.

As per my inline comment I'd suggest replacing the existing test case with the new one since it more accurately reflects the behavior we want, and let's combine the first and second commits. I know I mentioned TDD, and it's good to note in the PR text to note you developed that way, but for now we're sticking with each commit passing CI without xfails - I think we discussed that, but we can follow-up on czo to discuss further if necessary.

This also needs a black rebase.

neiljp:
@Ezio-Sarthak I just made a few text adjustments and re-pushed here, and will merge shortly :tada:

Other than the reduced noise from the math in triplicate, the styling separates the text well too :+1:

PR #1025, opened 2021-05-09
Add message formatting tips help menu

### LINKED ISSUE #623: Show message formatting tips
The web app has a section under keyboard shortcuts that shows a list of message formatting help tips. This, in particular, is helpful for users that are relatively new to markdown syntax.

It would be good to make it context-dependent.

Conversation link: https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Message.20formatting.20tips

**Screenshot**:

image

Ezio-Sarthak:
**What does this PR do?!**
 - Adds a message formatting help menu.
 - Creates a new config file `markdown.py` for markdown-related material.
 - Fixes #623.

**Commit flow**
1. Add hotkey `meta m ` to show markdown popup.
2. Create a new file `markdown_examples.py` containing markdown help data.
3. refactor: Amend the type annotation for PopUpView contents.
4. Structure UI in MarkdownHeloView class.
5. Add UI elements to markdown help view on toggling.
6. Add support for accessing the popup from compose area.

**Doubts/Concerns**
 - The Popup name, i.e., `MarkdownHelpView` could be replaced with one having `message formatting` instead?.
 - The markdown help data might be migrated to `ui_mappings`

**Screenshot(s)/GIF(s)**
improved_header

### DIFF
--- a/zulipterminal/config/markdown_examples.py
+++ b/zulipterminal/config/markdown_examples.py
@@ -1,0 +1,118 @@
+from typing import List
+
+from typing_extensions import TypedDict
+
+
+class MarkdownElements(TypedDict):
+    name: str
+    raw_text: str
+    html_element: str
+
+
+MARKDOWN_ELEMENTS: List[MarkdownElements] = [
+    {
+        # BOLD TEXT
+        "name": "Bold text",
+        "raw_text": "**bold**",
+        "html_element": "<strong>bold</strong>",
+    },
+    {
+        # EMOJI
+        "name": "Emoji",
+        "raw_text": ":heart:",
+        "html_element": '<span class="emoji">:heart:</span>',
+    },
+    {
+        # MESSAGE LINKS
+        "name": "Message links",
+        "raw_text": "[Zulip website]\n(https://zulip.org)",
+        "html_element": '<a href="https://zulip.org">Zulip website</a>',
+    },
+    {
+        # BULLET LISTS
+        "name": "Bullet lists",
+        "raw_text": "* Milk\n* Tea\n  * Green tea\n  * Black tea\n"
+        "  * Oolong tea\n* Coffee",
+        "html_element": "<ul><li>Milk</li><li>Tea<ul><li>Green tea</li>"
+        "<li>Black tea</li><li>Oolong tea</li></ul>"
+        "</li><li>Coffee</li>",
+    },
+    {
+        # NUMBERED LISTS
+        "name": "Numbered lists",
+        "raw_text": "1. Milk\n2. Tea\n3. Coffee",
+        "html_element": "<ol><li>Milk</li><li>Tea</li><li>Coffee</li></ol>",
+    },
+    {
+        # USER MENTIONS
+        "name": "User mentions",
+        "raw_text": "@**King Hamlet**",
+        "html_element": '<span class="user-mention">@King Hamlet</span>',
+    },
+    {
+        # USER SILENT MENTIONS
+        "name": "User silent mentions",
+        "raw_text": "@_**King Hamlet**",
+        "html_element": '<span class="user-mention silent">King Hamlet</span>',
+    },
+    {
+        # NOTIFY ALL RECIPIENTS
+        "name": "Notify all recipients",
+        "raw_text": "@**all**",
+        "html_element": '<span class="user-mention">@all</span>',
+    },
+    {
+        # LINK TO A STREAM
+        "name": "Link to a stream",
+        "raw_text": "#**announce**",
+        "html_element": '<a class="stream" data-stream-id="6" '
+        'href="/#narrow/stream/6-announce">#announce</a>',
+    },
+    {
+        # STATUS MESSAGE
+        "name": "Status message",
+        "raw_text": "/me is busy writing code.",
+        "html_element": "<strong>{user}</strong> is busy writing code.",
+    },
+    {
+        # INLINE CODE
+        "name": "Inline code",
+        "raw_text": "Some inline `code`",
+        "html_element": "Some inline <code>code</code>",
+    },
+    {
+        # CODE BLOCK
+        "name": "Code block",
+        "raw_text": "```\ndef zulip():\n    print 'Zulip'\n```",
+        "html_element": '<div class="codehilite"><pre><span></span><code>\n'
+        "def zulip():\n    print 'Zulip'</code></pre></div>",
+    },
+    {
+        # QUOTED TEXT
+        "name": "Quoted text",
+        "raw_text": ">Quoted",
+        "html_element": "<blockquote>░ Quoted</blockquote>",
+    },
+    {
+        # QUOTED BLOCK
+        "name": "Quoted block",
+        "raw_text": "```quote\nQuoted block\n```",
+        "html_element": "<blockquote>\n░ Quoted block</blockquote>",
+    },
+    {
+        # TABLE RENDERING
+        "name": "Table rendering",
+        "raw_text": "|Name|Id|\n|--|--:|\n|Robert|1|\n|Mary|100|",
+        "html_element": (
+            "<table>"
+            "<thead>"
+            '<tr><th align="left">Name</th><th align="right">Id</th></tr>'
+            "</thead>"
+            "<tbody>"
+            '<tr><td align="left">Robert</td><td align="right">1</td></tr>'
+            '<tr><td align="left">Mary</td><td align="right">100</td></tr>'
+            "</tbody>"
+            "</table>"
+        ),
+    },
+]
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -275,6 +275,9 @@ class View(urwid.WidgetWrap):
             # Show help menu
             self.controller.show_help()
             return key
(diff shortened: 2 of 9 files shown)


zulipterminal/ui.py
```diff
@@ -259,6 +259,10 @@ def keypress(self, size: Tuple[int, int], key: str) -> Optional[str]:
             # Show help menu
             self.controller.show_help()
             return key
+        elif is_command_key('MARKDOWN_HELP', key):
+            # Show markdown help menu
+            self.controller.show_markdown_help(self.model.server_url)
```
mkp6781, line 264: We can get `server_url` in a similar way in `show_markdown_help`. In that case, do we need to pass it as an argument here?

zulipterminal/config/markdown.py
```diff
@@ -0,0 +1,86 @@
...
+                        "\ndef zulip():\n    print 'Zulip'</code></pre></div>",
+    }, {
+        'help_text': 'Quoted text',
+        'raw_text': '>Quoted',
+        'html_element': '<blockquote>░ Quoted',
+    }, {
+        'help_text': 'Quoted block',
+        'raw_text': '```quote\nQuoted block\n```',
+        'html_element': '<blockquote>\n░ Quoted block',
+    }, {
```
mkp6781, line 77: Closing tag for blockquote is missing.

zulipterminal/ui_tools/views.py
```diff
@@ -1060,6 +1061,30 @@ def __init__(self, controller: Any, title: str) -> None:
...
+                element['html_element'],
+                server_url
+            )
+
+            raw_menu_content.append((element['raw_text'], element['raw_text']))
+            rendered_menu_content.append((element['raw_text'], markup_desc))
+
+        popup_width, column_widths = self.calculate_table_widths(
+            [('', raw_menu_content)], len(title))
+
```
mkp6781, line 1080: Unwanted diff.

mkp6781: Yea just the blank line.

Ezio-Sarthak:
@mkp6781 Thanks for the review! Updated the PR addressing the points :)

zulipterminal/config/markdown.py
```diff
@@ -0,0 +1,86 @@
...
+        'html_element': 'Some inline <code>code</code>',
+    }, {
+        'help_text': 'Code block',
+        'raw_text': "```\ndef zulip():\n    print 'Zulip'\n```",
+        'html_element': '<div class="codehilite"><pre><span></span><code>\ndef'
+                        ' zulip():\n    print \'Zulip\'</code></pre></div>',
+    }, {
+        'help_text': 'Quoted text',
+        'raw_text': '>Quoted',
+        'html_element': '<blockquote>░ Quoted',
```
mkp6781, line 72: Here too!

zulipterminal/config/markdown.py
```diff
@@ -0,0 +1,86 @@
+from typing import List
+
+from typing_extensions import TypedDict
+
+
+class MarkdownElements(TypedDict, total=True):
```
neiljp, line 6: There's no need for `total=True` - that's the default preferred version.

zulipterminal/ui_tools/views.py
```diff
@@ -886,7 +886,7 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
 
 
 PopUpViewTableContent = Sequence[Tuple[str, Sequence[Union[str,
-                                                           Tuple[str, str]]]]]
+                                                           Tuple[Any, Any]]]]]
```
neiljp, line 890: I don't like this change. We should clarify what type this is and define it. I expect it's an urwid type, and possibly one that could be used more widely in ZT.

zulipterminal/config/markdown.py
```diff
@@ -0,0 +1,86 @@
...
+    help_text: str
+    raw_text: str
+    html_element: str
+
+
+MARKDOWN_ELEMENTS: List[MarkdownElements] = [{
+        'help_text': 'Title text',
+        'raw_text': 'You type',
+        'html_element': 'You get',
```
neiljp, line 15: I see how this makes it easier to add the column titles, but it's confusing.

I'd suggest manually adding these column headings; really it would be better if these column headings stayed in place as things scrolled, though that would likely need adjustment in the main popup code.

neiljp (commented):
@Ezio-Sarthak This is a pretty useful feature as it stands; I've not tested it but it looks great in the screenshot :+1:

My concern is that the table titles scroll too, which is a little confusing, particularly if you have a short screen (height). That could be for the next version of this feature though.

I'm not sure of the best place for the markdown data; my gut feeling is that it is more internal to the software than symbols/keys/styles (which are definitely configuration, even if default), but perhaps big enough to have its own file.

This is awaiting a black rebase from what I can tell.

neiljp:
Like with the webapp, we could potentially simplify the UI at a later stage by using tabs or similar to make a more universal help menu - which is a good example of how having a general tabbing or similar widget would be useful, not just for large/raw message rendering :+1:

zulipterminal/ui_tools/views.py
Ezio-Sarthak: Pinged my concerns :+1:

neiljp:
@Ezio-Sarthak I've not looked at the code but I confirmed the headers are nicely fixed now, though we might benefit from styling them more befitting a title - maybe the same as popup categories?

Are the set in markdown.py just a selection you wanted to try, or did you pick them specifically? This can be extended/adjusted later of course :)

This has 'help' for each element; did you plan to extend the table later to have a name (that's what it is right now?) as well as maybe 'notes' (eg. emoji are actually shown in the webapp, italic is shown as bold, ...)?

Ezio-Sarthak:
@neiljp thanks for giving this a go! I agree the header may be made to stand out more. I'll use the `popup_category` palette for this :+1:.

Re the set in `markdown.py`, I just saw what all elements are shown in web app's message formatting tips menu, and listed them in order, ensuring we support that element in ZT as well :)

Re the `help`, partially yes. I initially kept it to let devs understand what a particular dict is about in `markdown.py`(I can document this by comments as well :))

zulipterminal/ui_tools/views.py
```diff
@@ -1211,6 +1213,46 @@ def __init__(self, controller: Any, title: str) -> None:
...
+
+        for element in MARKDOWN_ELEMENTS:
+            content = element["content"]
+
+            # Currently we only re-render user names
+            if element["re_render_required"]:
+                content = controller.model.user_full_name
+
+            wrapped_html_element = format_string([content], element["html_format"])[0]
+            wrapped_raw_content = format_string([content], element["raw_format"])[0]
```
neiljp, line 1229: See my comment on the stream, but I think this (and the dict format) can be simplified.

neiljp (commented):
@Ezio-Sarthak This is getting close :+1:

It was good to investigate the user-customization aspects, but I think this can be much simpler including in the data structures, which I think you could almost revert to the previous version - it's fine to include the extra 'name' fields, and I think we should file an issue for showing the 'name' or some 'notes' like in the web app.

Note that `/me dances` makes sense to use the current user, but we likely want a different user for other mentions, as that's the typical use-case.

If we can use a Union between the existing content type and that provided by the soup2markup method then this seems more explicit than just using `Any`?

Could the window could be wider? Were you restricting the width for a specific reason?

How feasible is this to access from compose?

Minor: maybe `markdown_examples.py` would be more specific?

neiljp (commented):
@Ezio-Sarthak Thanks for addressing most of the points! It's useful we'll be able to access this from compose now :)

This looks close, with as you say the type and width calculation being the sticking points. If we can make the minimal change I suggested in type and add a FIXME referencing a new issue for improving soup2markup typing then I don't see any issue with this as a v1 :+1:

The rebase doesn't look too tricky.

zulipterminal/ui_tools/views.py
```diff
@@ -1211,6 +1212,41 @@ def __init__(self, controller: Any, title: str) -> None:
...
+            rendered_content, *_ = MessageBox.transform_content(
+                html_element, server_url
+            )
+
+            raw_menu_content.append((raw_content, raw_content))
+            rendered_menu_content.append((raw_content, rendered_content))
+
+        popup_width, column_widths = self.calculate_table_widths(
+            [("", raw_menu_content)], len(title)
+        )
```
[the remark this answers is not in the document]
Ezio-Sarthak: Yes, precisely.

zulipterminal/config/markdown_examples.py
```diff
@@ -0,0 +1,112 @@
+from typing import List
+
+from typing_extensions import TypedDict
+
+
+class MarkdownElements(TypedDict):
+    name: str
+    raw_text: str
+    html_element: str
```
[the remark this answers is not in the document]
neiljp: Pending?

neiljp:
@Ezio-Sarthak I made a few minor textual changes but otherwise am merging as you had it - thanks! :tada:
(I do like that the examples are easily customizable in the config file :+1:)

PR #1026, opened 2021-05-09
Realm emoji test update

mkp6781:
This PR is a follow-up to #827. Based on @neiljp suggestions, I have 2 commits in this PR.

1st commit: Test for `handle_update_emoji_event` improved. Specifies the type of expected updated emoji in `model.active_emoji_data`.

2nd commit: Refactor commit to extract type of emoji into `api_types` as EmojiType.


neiljp:
@mkp6781 Thanks for the followup! Made a few minor adjustments to commit text before merging but looked fine otherwise :tada:

PR #1032, opened 2021-05-14
cryptography requires Rust toolchain on Alpine Linux.

fredrikekre:
`cryptography` now requires Rust toolchain on Alpine Linux, xref https://cryptography.io/en/latest/installation/#alpine


neiljp:
@fredrikekre Thanks for noting this and the PR :+1:

We could consider adding a build via docker to CI to avoid these issues in future, particularly since these are only tested manually right now I believe.

PR #1035, opened 2021-05-15
bugfix: boxes: Allow message body editing when limit doesn't exist.

### LINKED ISSUE #1034: Can't edit messages: "Time limit for editing the message body..."
I am running `zulip-term` from docker (https://github.com/zulip/zulip-terminal/blob/main/docker/Dockerfile.buster) and I am not able to edit messages I have sent, even if it is directly after. I see the error
```
Time limit for editing the message body exceeded
```
but I can go to the browser client and edit the message just fine.

Worth noting also that https://julialang.zulipchat.com, where I noticed this, doesn't enforce any time limit on edits.

cc @MasonProtter

prah23:
Currently, a configuration of 0 for `realm_message_content_edit_limit_seconds`
is treated as a complete restriction against message body editing, while
the configuration actually signifies no limit in message body editing.

This PR adds a commit that fixes the bug that doesn't let the users
edit the message body when the configuration sent by the server is 0
for a realm. 

Fixes #1034.

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1901,20 +1901,24 @@ class TestMessageBox:
     @pytest.mark.parametrize(['to_vary_in_each_message',
                               'realm_editing_allowed',
                               'msg_body_edit_enabled',
+                              'msg_body_edit_limit',
                               'expect_editing_to_succeed'], [
-        ({'sender_id': 2, 'timestamp': 45}, True, True, False),
-        ({'sender_id': 1, 'timestamp': 1}, True, False, True),
-        ({'sender_id': 1, 'timestamp': 45}, False, True, False),
-        ({'sender_id': 1, 'timestamp': 45}, True, True, True),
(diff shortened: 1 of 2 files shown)


neiljp:
@prah23 Thanks for the quick fix! :+1: I reordered the test parameters to keep the expected value at the end and have just merged :tada:

I checked the server tests and the zero value meaning is indicated there; as per my message in the stream, if there is not API documentation of this special value then I'd suggest we add an issue or PR for that, or discuss on czo.

prah23:
Thanks for merging it, @neiljp! I think an issue/PR would be prompt as well, as the documentation seems missing for this. I've also responded on CZO to discuss further.
I'll link to that here for anyone reading: [#zulip-terminal > "Any time" message edits #T1035 (fix #T1034)](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/.22Any.20time.22.20message.20edits.20.23T1035.20.28fix.20.23T1034.29).

PR #1038, opened 2021-05-25
bugfix: Fix wildcard_mentions not showing in mentions narrow.

### LINKED ISSUE #1037: All/everyone/stream mentions don't show up in mentions narrow
`@all`, `@everyone` and `@stream` mentions don't seem to appear in the mentions narrow.

I would recommend against testing the anomaly on any stream with a high subscriber count and would recommend using the zulip dev server.

zee-bit:
The server defines several wildcard_mentions like @all/@everyone/@stream, which are not marked with 'mentions' in the message flags, rather 'wildcard_mentions'. This bugfix PR now checks for this flag in the message response from the server and appropriately categorizes these messages as mentioned and also updates the count in the respective narrows.

Fixes #1037

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1311,7 +1311,8 @@ class TestModel:
         assert notify.called == is_notify_called
 
     @pytest.mark.parametrize(
-        "event, expected_times_messages_rerendered, expected_index, topic_view_enabled",
+        "event, expected_times_messages_rerendered, expected_index, topic_view_enabled,"
+        "initial_flags, mentioned_msg_ids",
         [
             case(
                 {  # Only subject of 1 message is updated.
@@ -1328,18 +1329,23 @@ class TestModel:
                             "stream_id": 10,
                             "content": "old content",
                             "subject": "new subject",
+                            "flags": ["read"],
                         },
                         2: {
                             "id": 2,
                             "stream_id": 10,
                             "content": "old content",
                             "subject": "old subject",
+                            "flags": ["read"],
                         },
                     },
                     "edited_messages": {1},
                     "topics": {10: []},
+                    "mentioned_msg_ids": set(),
                 },
                 False,
+                ["read"],
+                set(),
                 id="Only subject of 1 message is updated",
             ),
             case(
@@ -1357,27 +1363,33 @@ class TestModel:
                             "stream_id": 10,
                             "content": "old content",
                             "subject": "new subject",
+                            "flags": ["read"],
                         },
                         2: {
                             "id": 2,
                             "stream_id": 10,
                             "content": "old content",
                             "subject": "new subject",
+                            "flags": ["read"],
                         },
                     },
                     "edited_messages": {1},
                     "topics": {10: []},
+                    "mentioned_msg_ids": set(),
                 },
                 False,
+                ["read"],
+                set(),
                 id="Subject of 2 messages is updated",
             ),
             case(
-                {  # Message content is updated
+                {  # Message content is updated, wildcard-mention added
                     "message_id": 1,
                     "stream_id": 10,
                     "rendered_content": "<p>new content</p>",
+                    "flags": ["read", "wildcard_mentioned"],
                 },
-                1,
+                2,
                 {
                     "messages": {
                         1: {
@@ -1385,29 +1397,35 @@ class TestModel:
                             "stream_id": 10,
                             "content": "<p>new content</p>",
                             "subject": "old subject",
+                            "flags": ["read", "wildcard_mentioned"],
                         },
                         2: {
                             "id": 2,
                             "stream_id": 10,
                             "content": "old content",
                             "subject": "old subject",
+                            "flags": ["read"],
                         },
                     },
                     "edited_messages": {1},
                     "topics": {10: ["old subject"]},
+                    "mentioned_msg_ids": {1},
                 },
                 False,
-                id="Message content is updated",
+                ["read"],
+                set(),
+                id="Message content is updated, wildcard-mention added",
             ),
             case(
-                {  # Both message content and subject is updated.
+                {  # Both message content and subject is updated, mention added.
                     "message_id": 1,
                     "rendered_content": "<p>new content</p>",
                     "subject": "new subject",
                     "stream_id": 10,
                     "message_ids": [1],
+                    "flags": ["read", "mentioned"],
                 },
-                2,
+                3,
                 {  # 2=update of subject & content
                     "messages": {
                         1: {
@@ -1415,19 +1433,24 @@ class TestModel:
                             "stream_id": 10,
                             "content": "<p>new content</p>",
                             "subject": "new subject",
+                            "flags": ["read", "mentioned"],
                         },
                         2: {
                             "id": 2,
                             "stream_id": 10,
                             "content": "old content",
                             "subject": "old subject",
+                            "flags": ["read"],
                         },
                     },
                     "edited_messages": {1},
                     "topics": {10: []},
+                    "mentioned_msg_ids": {1},
                 },
                 False,
-                id="Both message content and subject is updated",
+                ["read"],
+                set(),
+                id="Both message content and subject is updated, mention added",
             ),
             case(
                 {  # Some new type of update which we don't handle yet.
@@ -1442,18 +1465,23 @@ class TestModel:
                             "stream_id": 10,
                             "content": "old content",
                             "subject": "old subject",
+                            "flags": ["read"],
                         },
                         2: {
                             "id": 2,
                             "stream_id": 10,
                             "content": "old content",
                             "subject": "old subject",
+                            "flags": ["read"],
                         },
                     },
                     "edited_messages": {1},
                     "topics": {10: ["old subject"]},
+                    "mentioned_msg_ids": set(),
                 },
                 False,
+                ["read"],
+                set(),
                 id="Some new type of update which we don't handle yet",
             ),
             case(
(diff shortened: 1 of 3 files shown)


neiljp:
@zee-bit This looks effective after a quick manual test :+1:

As per messages on the stream, I think it'd be useful to ensure through a test that the event handling code handles the behavior correctly of adding/removing these message flags. This sort of updates already via scrolling but not automatically, and the event code looks like it won't behave properly in all cases, so this may be updated behavior belonging in a new commit. We likely want it to behave the same as for starred messages, ie. add but not remove, until leaving and returning to the narrow.

A related point to that above is that we could use a `MessageFlags` type to clarify what values to expect in messages, and a more limited related type `ModifiableMessageFlags` perhaps for those we could see in events?

Another related point - perhaps for the first commit? - is that we may want to explore expanding other tests to incorporate this message flag. That'll be both checking for tests with `mention` and expanding to include extra `wildcard_mention` cases, as well as considering cases where we don't check mention at all but just eg. starred/read, and may want to generalize?

zulipterminal/helper.py
```diff
@@ -173,7 +173,8 @@ def _set_count_in_view(controller: Any, new_count: int,
 
         msg_type = message['type']
         add_to_counts = True
-        if 'mentioned' in message['flags']:
+        if ('mentioned' in message['flags']
+                or 'wildcard_mentioned' in message['flags']):
```
neiljp, line 177: I'm merging a commit to master which will extend the line-length to 88 (ie. black's default), so likely your lines should simplify :)

It's not too bad with only two, but we could also consider a `set` approach here to avoid the duplication, like we've done recently in `soup2markup`.

tests/helper/test_helper.py
```diff
@@ -153,32 +153,43 @@ def test_index_starred(mocker,
     assert index_messages(messages, model, model.index) == expected_index
 
 
-@pytest.mark.parametrize('mentioned_messages', [
-    {537286, 537287, 537288},
-    {537286}, {537287}, {537288},
-    {537286, 537287}, {537286, 537288}, {537287, 537288},
+@pytest.mark.parametrize(['mentioned_messages',
+                          'wildcard_mentioned_messages'], [
```
neiljp, line 157: With extra line-length available and the way `black` reformats these lines (indented on the follow line), we should generally have plenty of space available and not need a list for the params compared to just a single string.

zulipterminal/helper.py
zee-bit: Ahh...The `set` approach seems better. I'll use that. :+1:

tests/helper/test_helper.py
```diff
@@ -152,24 +152,34 @@ def test_index_starred(
...
-        {537286, 537287},
-        {537286, 537288},
-        {537287, 537288},
+        ({537286, 537288}, {537287}),
+        ({537286}, {537286}),
+        ({537287}, {537286, 537288}),
+        ({537288}, {537286, 537287, 537288}),
+        ({537286, 537287}, set()),
+        ({537286, 537288}, {537286, 537288}),
+        ({537287, 537288}, {537286, 537287, 537288}),
```
neiljp, line 163: We've lost some test examples here of just mentions; that may not be critical, but even if we just add these mixed cases then adding inline test ids would be helpful to describe what situation each is covering without looking at the values. You may see when reading those ids that you're missing some combinations that might be interesting, such as only-wildcard.

zulipterminal/model.py
```diff
@@ -289,7 +289,13 @@ def current_narrow_contains_message(self, message: Message) -> bool:
             # all messages contains all messages
             not self.narrow
             # mentions
-            or (self.narrow[0][1] == "mentioned" and "mentioned" in message["flags"])
+            or (
+                self.narrow[0][1] == "mentioned"
+                and (
+                    "mentioned" in message["flags"]
+                    or "wildcard_mentioned" in message["flags"]
```
neiljp, line 296: I'm unsure if we should use the same set intersection here, but it perhaps points towards using that internally for simplicity (and slight optimization) via a refactor.

tests/conftest.py
```diff
@@ -743,25 +743,30 @@ def index_all_starred(empty_index, request):
...
-        {537286, 537287},
-        {537286, 537288},
-        {537287, 537288},
+        ({537286, 537288}, {537287}),
+        ({537286}, {537286}),
+        ({537287}, {537286, 537288}),
+        ({537288}, {537286, 537287, 537288}),
+        ({537286, 537287}, set()),
+        ({537286, 537288}, {537286, 537288}),
+        ({537287, 537288}, {537286, 537287, 537288}),
```
neiljp, line 752: If this is similar to the other mention/wildcard combination fixture, then we could consider extracting this as a fixture to be used in both places.

tests/model/test_model.py
```diff
@@ -1870,11 +1871,15 @@ def test_update_star_status_invalid_operation(
...
             ("add", 1, ["read", "starred"], ["read", "starred"]),
+            ("add", 1, ["mentioned"], ["mentioned", "starred"]),
+            ("add", 1, ["wildcard_mentioned"], ["wildcard_mentioned", "starred"]),
             ("remove", -1, [], []),
             ("remove", -1, ["read"], ["read"]),
             ("remove", -1, ["starred"], []),
             ("remove", -1, ["read", "starred"], ["read"]),
             ("remove", -1, ["starred", "read"], ["read"]),
+            ("remove", -1, ["mentioned", "starred"], ["mentioned"]),
+            ("remove", -1, ["wildcard_mentioned", "starred"], ["wildcard_mentioned"]),
```
neiljp, line 1995: These extra cases in this commit are good examples, but are giving quite random coverage that I'm not sure helps dramatically. Think of what the other cases cover.

tests/model/test_model.py
```diff
@@ -1312,7 +1312,8 @@ def test_notify_users_enabled(
         assert notify.called == is_notify_called
 
     @pytest.mark.parametrize(
-        "event, expected_times_messages_rerendered, expected_index, topic_view_enabled",
+        "event, expected_times_messages_rerendered, expected_index, topic_view_enabled"
+        ", mentioned_msg_ids",
```
neiljp, line 1316: I know I simplified a lot of these into single strings; we can discuss the preferred style of whether to break into a multi-line implicitly-combined string like this or a list of strings, but at minimum I would expect a comma to belong with the preceding text?

neiljp (commented):
@zee-bit I started some feedback on this so wanted to get a little more out to you. I've not looked at the last commit.

Having flags internally as a set may make things simpler for us, though I've not examined how much of a refactor it would be.

tests/model/test_model.py
zee-bit: Personally, I would prefer the multiline string approach since that saves a lot of lines compared to the list approach, where each parameter would be shifted to a new line hence increasing the diff. But, the list approach may be a bit more comprehensible than the string version.
I am going with the multiline string approach here, for now, and maybe we could shift this particular discussion to the stream for more collective decision?

zulipterminal/api_types.py
```diff
@@ -5,6 +5,15 @@
 
 EditPropagateMode = Literal["change_one", "change_all", "change_later"]
 EmojiType = Literal["realm_emoji", "unicode_emoji", "zulip_extra_emoji"]
+# Limited flags which could be present in update_message_flag events.
+ModifiableMessageFlag = Literal["read", "starred", "collapsed"]
+MessageFlags = Literal[
```
neiljp, line 10: I'm not sure if you're done with the current iteration and wanting another review yet, but this is still plural (plus commit text).

tests/conftest.py
zee-bit: I am not sure of the ideal place to refactor this, so I am creating an initial commit that extracts this as a fixture and the subsequent commits will just use this wherever needed. :)

tests/model/test_model.py
zee-bit: I'm not sure if the current iteration of my PR addresses this review, but I tried to cover a few other cases that the previous tests were missing. Let me know if these new cases are still not what you were expecting. :)

tests/helper/test_helper.py
```diff
@@ -151,25 +151,23 @@ def test_index_starred(
...
-    [
-        {537286, 537287, 537288},
-        {537286},
-        {537287},
-        {537288},
-        {537286, 537287},
-        {537286, 537288},
-        {537287, 537288},
-    ],
-)
```
neiljp, line 165: It'd be clearer to sweep all of these in the first pass as a direct refactor wherever these are used, then you can expand it after that. As it is it makes it more difficult to compare this removed text with that removed/swapped in the previous commit. Refactoring first and then switching the fixture second should make this clearer.

zulipterminal/api_types.py
```diff
@@ -101,6 +110,7 @@ class UpdateMessageEvent(TypedDict):
     message_ids: List[int]
     subject: str
     stream_id: int
+    flags: List[MessageFlag]
```
neiljp, line 113: It's a bit of an edge case, but this could go in the next commit.

neiljp (commented):
@zee-bit Just a quick bit of feedback tonight while I have time - independently of the bugfix, I'm looking forward to the refactor of mentions fixture as well as the extra api typing :+1:

neiljp:
@zee-bit The refactoring first is great, and I've cherry-picked that first commit :+1:

I don't like that we introduce a broken state after the second commit - and the subsequent commit is a little unclear too in terms of whether it's fixing that broken state or something else. If you can introduce some wildcards without breaking things then you could, then update the params with more values next - but I suspect that's not that useful.

I'd suggest just adding some mixed data to the new fixture and fixing the code to make tests pass - in one commit. Then if there's a separate bugfix to make, add that - along with a new test case?

neiljp:
(Just cherry-picked the first flag types commit too)

neiljp:
@zee-bit Another cherry-pick after cleaning the commit text to simplify the bug text - it should be accurate, but it's easier if you can do this more yourself. I think you referenced various different bugs? I'd recommend checking it out and we can discuss.

I would have merged the rest but we shouldn't depend on just indexed messages for the count to update - if I edit a message which isn't indexed then the mentioned unreads don't change currently or with the remaining commit(s).

zee-bit:
We may need to get the remaining commits (or, at least the second one) in before I can push a fix for the un-indexed message bug, cause the new field that I'm working on locally needs to be updated when flags are changed through `update_message_event`, which we don't do currently (and is added in the second commit of this PR).

zee-bit:
The rest of the PR is protracted to #1068, along with the un-indexed message bugfix.

PR #959, opened 2021-03-26
refactor: tests: Add path prefixes for patching class methods

Ezio-Sarthak:
Currently, in pytests, several statements use hard-coded module path strings for patching purposes.
The intent of this PR is to refactor such statements and as a result, reduce code duplication.

PR #1039, opened 2021-05-27
Migrate to black

neiljp:
This PR comprises:
* many prep commits motivated by - though not all required for - a better post-`black` experience.
* black configuration in `pyproject.toml`
* removal of a conflicting flake8 rule
* addition of black linting to `tools/lint-all` (so `make check` and `make lint`) and fixing to `make fix`
* two commits for migrating structure only, then adjusting to a standard quote style
* enforcement of black linting via CI


tests/model/test_model.py
```diff
@@ -17,6 +17,7 @@
 
 
 CONTROLLER = 'zulipterminal.core.Controller'
+MODEL = 'zulipterminal.model.Model'
```
Ezio-Sarthak: I just wanted to point that I had a PR #959 that does a general simplification of patches in all tests. Let me know if I should make it review-ready if we wish to merge it first.

neiljp: @Ezio-Sarthak Thanks for reminding me of your PR :+1:

I welcome your intent to add more constants to improve readability; the main challenge with using these is that we don't currently have much of a "standard" way of applying these constants - or it seems that way, but isn't quite. In some pre-existing cases we have files referred to, others to classes, and I think this is where we would benefit from standardizing and clarying. You'll note that in my commit I only extracted the classes, since they were both the longest - so more likely to benefit from compacting the long lines - but also that it made clear that others were module-specific, either imports or not in the main class.

zee-bit (approved):

preetmishra (approved):
@neiljp Thanks for putting all the work into this! This looks great. :+1:

neiljp:
Thanks for all the feedback everyone - this has been rather a lot of effort but I'm looking forward to it being useful in future! :tada:

PR #1041, opened 2021-05-29
bugfix: ui: Focus on content_area when draft is opened from side-panels.

zee-bit:
This commit fixes a bug that caused the focus to not move to the content
area if the draft was opened from the side-panels. This is minor bugfix
that should set the focus to the message_column before setting it to
the footer(i.e. content_area).


Rohitth007:
Looks good to go after a rebase :+1:

zee-bit:
Closing this PR. I messed up the local branch because of the black transitioning. Reopened a new one: #1044. :)

Issue #1042, opened 2021-05-29
Composing message when focus is on side-panel leaves it open in `--autohide` mode


zee-bit:
Composing a message using `x` or `c` when the focus is on either of the side-panels, leaves it open when `--autohide` is specified as a command-line arg. A similar case happens when opening a draft with either of the side-panels open, which is discussed briefly in **#zulip-terminal>Bug: Drafts focus**.

This can lead to some easily reproducible cases where both the side-panels are open, which seems a bit strange for `--autohide`. We would like to have a general approach that can handle these corner-cases(and probably other similar cases) gracefully.

neiljp:
I can't reproduce this on `main`, and it looks like this was fixed in c96931a8c in #1100 (August 2021), so closing this now.

PR #1044, opened 2021-06-01
bugfix: ui: Focus on content_area when draft is opened from side-panels.

zee-bit:
This commit fixes a bug that caused the focus to not move to the content
area if the draft was opened from the side-panels. This is minor bugfix
that should set the focus to the message_column before setting it to
the footer(i.e. content_area).

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -320,6 +320,61 @@ class TestView:
             view.stream_w.stream_search_box
         )
 
+    @pytest.mark.parametrize(
+        "draft",
+        [
+            {
+                "type": "stream",
+                "to": "zulip terminal",
+                "subject": "open draft",
+                "content": "this is a stream message content",
+            },
+            {
+                "type": "private",
+                "to": ["foo@zulip.com", "bar@gmail.com"],
+                "content": "this is a private message content",
+            },
+            None,
+        ],
+        ids=[
+            "stream_draft_composition",
+            "private_draft_composition",
+            "no_draft_composition",
+        ],
+    )
+    @pytest.mark.parametrize("key", keys_for_command("OPEN_DRAFT"))
+    def test_keypress_OPEN_DRAFT(self, view, mocker, draft, key, widget_size):
+        view.middle_column = mocker.Mock()
+        view.set_footer_text = mocker.Mock()
+        view.controller.is_in_editor_mode = lambda: False
+        view.model.stream_id_from_name.return_value = 10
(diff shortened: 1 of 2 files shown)


neiljp:
@zee-bit This is a nice simple fix :+1: Are there tests for this that can be amended to ensure this doesn't break?

I'm happy to merge after we get clarification on a test. However, a good follow-up refactor would be to find some of the similar code associated with opening the compose box and ensure we do the same actions each time by wrapping it in new methods. That may also be useful for resolving things like #1042, though we'd need to check if that's quite the right solution.

zee-bit:
Thanks for looking at my PR, Neil.

I `grep`'ed our codebase for drafts, and couldn't find any tests for opening draft compose, so there was nothing to amend. I'll add the tests for this in another commit in this PR(just to keep the bugfix isolated).
Since this topic is brought up, I'd also like to remind that there's an open PR that adds tests for `SAVE_AS_DRAFT` in #950. Although that is not related to this PR in general we might want to address that PR too?

zee-bit:
@neiljp I had already updated this PR, if you wanted to re-review.

I've added the tests for `OPEN_DRAFT` keypress as the first commit such that it will fail, and the second commit with the one-liner bugfix will fix the test eventually. Let me know if there's anything else that needs to be added/updated. :)

neiljp (commented):
@zee-bit Thanks for adding the test! 👍﻿                                     
                                                                              
  Rather than having the first commit break, I'd suggest adding the bulk of   
  the test for the current behavior in the first commit - which should pass - 
  and then amending that test and adding the new behavior in the second       
  commit.                                                                     
                                                                              
  The test logic is rather complex with the mocks, but if we can refactor the 
  enter-compose functionality this should become much simpler, I expect.﻿

zee-bit:
The tests are a bit complex currently, I admit. But I am not sure if we should do the refactoring in this PR too. That would also require changing it for all enter-compose functionality, not just draft.

The other way, we could get rid of those mocks or at least reduce the apparent redundancy(since other tests define similar mocks too) is to define them all once in the init function?

neiljp:
@zee-bit Thanks for rearranging the test elements - that was my main interest for getting this merged initially, which I'm just about to do after some trimming to your commit text (each commit is independent so doesn't need to refer to the other) :tada:

If you want to take up the refactor please could you note that or open an issue?

zee-bit:
@neiljp Thanks for merging.
Noted. I'll create a PR for the refactor soon...

Issue #993, opened 2021-04-12
TRACKING: Event handling improvements


Ezio-Sarthak:
- [ ] #588 : Alert words handling
- [ ] #988 : Update user display settings
- [ ] Update global notifications (currently stream-based priority? #666)
- [ ] #816 Streams (add, delete, update) (#851)
- [ ] #529 Upload/Attachment events
- [ ] Submessages (e.g., polls) (#287)
- [ ] #988: User status updates.
- [ ] Custom profile fields (extension of user info popup #848)
- [ ] Handle Default stream updates (changed by org admin).
- [ ] Delete message event (first event handling, then feature support; #690)
- [ ] User group events (add/modify/delete)
- [ ] #809 : Realm emoji events (#827)

Ezio-Sarthak:
@zulipbot add "area:event handling"

zulipbot:
**ERROR:** Label "area:event handling" does not exist and was thus not added to this issue.

Ezio-Sarthak:
@zulipbot add "area: event handling"

neiljp:
Tagged with GSoC, but this is across multiple areas so as per the tracking nature of this issue, this isn't something one person will likely work on.

abhijeetbodas2001:
> `realm_filters`: Handle a change of configured linkifiers.

I get that this is not a high priority, but whenever in the future ZT adds support for handling events sent after the linkifiers in the realm have changed, we should handle the new `realm_linkifiers` events, since the `realm_filters` event type has been deprecated in feature level 54. #995 has more context. Can you update the issue description?

Ezio-Sarthak:
Thanks! Updated the issue description :)

neiljp:
Also see zulip/zulip-mobile#3408 for a similar mobile tracking issue.

PR #1048, opened 2021-06-06
Add support for handling delete_message events.

zee-bit:
This adds a new event action in model that looks for delete_message events, potentially handling it by removing the message completely from the index, and then updates the rendered_view to handle the event dynamically without needing to switch narrows.

It also looks out for messages being unread before being deleted and updates the count appropriately. Similarly, for messages that were starred before being deleted, this updates the `starred_count` too.

Fixes one check-box of #993.


zee-bit:
I'll add the tests tomorrow, after exploring for further corner-cases. :)

PR #1053, opened 2021-06-16
Remove unsupported `recipient_id` and `(sender)_short_name` fields.

zee-bit:
This PR removes the deprecated `recipient_id` field from the entire
codebase and adds a comment in the api_types to indicate that the field
is no longer used/needed. The server recommends clients to use the
display_recipients field to get info about the list of recipient(s).

The `short_name`  and `sender_short_name` fields are also completely
removed from the codebase after the server repo removed those from
server version 3.1; ZFL 26+.

Tests amended.


zulipterminal/api_types.py
```diff
@@ -27,7 +27,8 @@ class Message(TypedDict, total=False):
     id: int
     sender_id: int
     content: str
-    recipient_id: int
+    # NOTE: Deprecated; a server implementation detail not useful in a client.
+    # recipient_id: int
```
neiljp, line 31: I think it would be reasonable to move this to the bottom of the class into an unsupported/unused section of comments, with details on the status of each (ie. when they were removed/unused/deprecated/etc).

As you note, `recipient_id` is only deprecated and not removed, at least for now.

However, `sender_short_name` was removed in Zulip 3.1 (check the api changelog) - and we haven't used it or similar short names - so it would be good to do other tidying along these lines too.

tests/ui/test_ui_tools.py
```diff
@@ -1901,14 +1900,12 @@ def test_soup2markup(self, content, expected_markup, mocker):
...
                         {
                             "is_mirror_dummy": False,
-                            "short_name": "aaron",
                             "email": "AARON@zulip.com",
                             "id": 1,
                             "full_name": "aaron",
                         },
                         {
                             "is_mirror_dummy": False,
-                            "short_name": "iago",
```
neiljp, line 1911: This was removed at the same time as other short names, but since we have no typeddict for it then nothing is being removed from the api_types. We could definitely improve the typing of `display_recipient` in this way, maybe first to define what we have, though not necessarily.

neiljp (commented):
@zee-bit Thanks for breaking this out - this will tidy some unused lines and start to document what exists/existed but we don't expect to use :+1:

There are quite a few different things being removed here, so it'd be clearer to describe in separate commits, including some extra lines/fields that I see you weren't sure about, that is documented in the api changelog. As always, it's easier to squash commits together than split them apart.

We should definitely type the display_recipient field at some point; it's not a requirement for this removal, but it would make it clear where things are being removed, where we don't currently have that aspect typed. It should also make it possible to eventually remove the `total=False` aspect of `Message`.

You don't seem to have the files in the commit title?

neiljp:
@zee-bit Thanks for the split & slight adjust :+1: I made some minor commit text adjustments and am just about to merge from here :tada:

(see https://github.com/zulip/zulip-terminal#gitlint for the commit text approach I'm following; if that's not clear or could be improved, please follow-up :+1:)

PR #1054, opened 2021-06-17
bugfix: Construct narrow appropriately before verifying and triggering footer notifications.

prah23:
**PR Structure:**

- **refactor: tests: Change req parameter to use a list of recipient emails.**
This commit refactors the test cases of the `req` parameter in
`test_notify_if_message_sent_outside_narrow` from a string of ', '
separated emails to a list of emails, consistent with the current type
of the `to` attribute of PrivateComposition.
Tests updated.

- **bugfix: helper: Use ', ' as separator to construct narrow to verify.**
This commit fixes a bug that previously caused a "Message sent outside
current narrow." message in the footer in group pms (huddles) even when
on the right narrow. This happened because of the wrong separator being
used, ','. This has been changed to the relevant separator - ', '.
Test added.

**Testing and linting:**

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -365,6 +365,16 @@ def test_display_error_if_present(mocker, response, footer_updated):
         case(
             {
                 "type": "private",
+                "to": ["foo@zulip.com", "bar@zulip.com"],
+                "content": "Hi",
+            },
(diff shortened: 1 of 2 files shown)


neiljp:
@prah23 Thanks for noticing and fixing this :+1: I adjusted the commit text to fit our style and be more descriptive (and the new test case id) and am merging now :tada:

PR #1055, opened 2021-06-24
README: Split out Commit Style notes into an expanded section.

neiljp:
Link to Commit Style (GitLint) sections updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,8 +160,10 @@ Zulip Terminal is being built by the awesome [Zulip](https://zulip.com/team) com
 To be a part of it and to contribute to the code, feel free to work on any [issue](https://github.com/zulip/zulip-terminal/issues) or propose your idea on
 [#zulip-terminal](https://chat.zulip.org/#narrow/stream/206-zulip-terminal).
 
-Please read our [commit message guidelines](http://zulip.readthedocs.io/en/latest/contributing/version-control.html) and
-[git guide](http://zulip.readthedocs.io/en/latest/git/index.html). **NOTE** Due to the difference in project scale, git commit titles in the Zulip Terminal project read slightly differently - please review our recent git log for examples and see the [GitLint](#gitlint) section below for more guidelines.
+For commit structure and style, please review the [Commit Style](#commit-style) section below.
+
+If you are new to `git` (or not!), you may benefit from the [Zulip git guide](http://zulip.readthedocs.io/en/latest/git/index.html).
+When contributing, it's important to note that we use a **rebase-oriented workflow**.
 
 A simple [tutorial](https://github.com/zulip/zulip-terminal/blob/main/docs/developer-feature-tutorial.md) is available for implementing the `typing` indicator.
 Follow it to understand how to implement a new feature for zulip-terminal.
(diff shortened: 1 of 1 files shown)


prah23:
Thanks for doing this, @neiljp, the section looks good!
I had a quick suggestion for the commit message body aspect, there seems to be no mention of that. Since we're emphasizing on commit style in it's own section, I think a short description/pointers on what we expect in a commit message's body would be useful to point new contributors to. I'd suggest points like:
- Why a feature is being added in a commit; a brief explanation on how as well.
- In case of bug-fixes, what went wrong prior to the commit and how the commit fixes it (in detail).
- Nuances to the commit that are missed in the commit title.

prah23:
Thanks for doing this, @neiljp, the section looks good!
I had a quick suggestion for the commit message body aspect, there seems to be no mention of that. Since we're emphasizing on commit style in it's own section, I think a short description/pointers on what we expect in a commit message's body would be useful to point new contributors to. I'd suggest points like:
- Why a feature is being added in a commit; a brief explanation on how as well.
- In case of bug-fixes, what went wrong prior to the commit and how the commit fixes it (in detail).
- Nuances to the commit that are missed in the commit title.

README.md
```diff
@@ -243,22 +245,42 @@ If using make with pip, there are corresponding make targets for running linting
...
 
 If the hook is installed as described above, then after completing the text for a commit, it will be checked by gitlint against the style we have set up, and will offer advice if there are any issues it notices. If gitlint finds any, it will ask if you wish to commit with the message as it is (`y` for 'yes'), stop the commit process (`n` for 'no'), or edit the commit message (`e` for 'edit').
 
 Other gitlint options are available; for example it is possible to apply it to a range of commits with the `--commits` option, eg. `gitlint --commits HEAD~2..HEAD` would apply it to the last few commits.
 
-**NOTE** Not all style suggestions are identified by gitlint at this time, including:
-* If modifying code (not just tests), list modified filenames without extensions between slashes in only one area of the commit title (eg. `run/model/core/README: Some description.`)
-* If modifying tests in addition to code, just note this in the body of the commit after a blank line (eg. `Tests updated.`)
```
preetmishra, line 256: We don't check for this in the GitLint, right? Should we keep it here in the README?

README.md
```diff
@@ -243,22 +245,42 @@ If using make with pip, there are corresponding make targets for running linting
...
+
+We aim to follow a standard commit style to keep the `git log` consistent and easy to read.
+
+Much like working with code, it's great to refer to the git log, for the style we're actively using.
+
+Our overall style for commit structure and messages broadly follows the general [Zulip version control guidelines](http://zulip.readthedocs.io/en/latest/contributing/version-control.html), so we recommend reading that first.
+
+Our commit titles have slight variations from the general Zulip style, with each:
+* starting with one or more areas in lower case, followed by a colon and space
+* area being slash-separated modified files without extensions, or the type of the change
```
preetmishra, line 258: Should we recommend filenames (and areas) in an alphabetically sorted fashion, so that the git log is more consistent?

preetmishra (commented):
@neiljp Thanks for pushing this in. It looks good! 👍 I have left two minor concerns as in-line comments. Re the commit body, I believe we pretty much follow the Zulip version control guidelines, so a link to the docs would be sufficient.

README.md
neiljp: Some of these are mentioned in the new section, so how about adjusting the trailing sentence of that from:
```
To make it easier to meet these style guidelines we recommend installing GitLint - see the next section!
```
to
```
To aid in satisfying some of these rules you can use `GitLint`, as described in the following section.
**However**, please check your commits manually versus these style rules, since GitLint cannot check everything - or language or grammar!

README.md
neiljp: I know that some contributors do this, but there are various rules we could use for ordering, and I think it's best to not impose even more rules on contributors in this way? I'm not sure what we gain in alphabetical ordering that isn't achievable with eg. some kind of `grep`. We could sort by "size of change in file contents" or others - and they all impose various extra work which can be determined otherwise.

Instead we could explicitly mention that the order doesn't matter?

(listing files already impose extra burden, but help summarize what has changed quickly in the log titles - and too many entries sometimes indicates if a commit is too complicated - if it's not a refactor, for example)

neiljp:
@prah23 Should I leave a commit message content follow-up to you?

README.md
preetmishra: That sounds good! :+1:

README.md
preetmishra: Right, that does add a bit of resistance; seems reasonable. :+1: We could mention that the order doesn't matter if needed later.

neiljp: as per PM I've slightly hinted that order isn't important in the first example commit title :)

neiljp:
Updated per feedback. Hope to merge in the next <24h.

prah23:
Sure, I'll write it and open another PR.

neiljp:
OK, let's get this merged - feel free to discuss and update while I'm away so we can help clarify anything else people have noticed :+1:

Issue #1014, opened 2021-04-28
Minimize sending no-op message edits to server


neiljp:
Currently there are cases when editing a message, where we send details to the server of a message edit, even if the topic or content have not changed. These 'no op' edits are not currently identified by the server and can lead to unexpected no-change entries in message history.

After zulip/zulip#16317 and zulip/zulip#17193 these should not be a problem and filtered from the server side, but this will remain an issue for running against older servers until the release with these PRs is our minimum supported version.

Even when those server fixes are in place, the checks added as part of resolving this issue will contribute to allowing
* minor network optimizations
* potentially hooking UI alerts regarding trying to send an edit with no changes

zulipbot:
Hello @neiljp, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

PR #1056, opened 2021-06-25
bugfix: Fix topic editing if topic unchanged

neiljp:
Recent work in resolved/unresolved topics has caused a change in the server (not yet in a release), where it more strictly enforces that if a topic is provided but unchanged in stream message editing/updating, then the propagate mode must be set to change_one - otherwise the edit does not proceed.

This PR works around this and has scope to be expanded towards resolving part or all of #1014.

The need for this was discussed in #**design>closing/solving a topic (#11154)**.

I've not added tests for this change in behavior (yet, at least) - we don't seem to have tests for the existing behavior.

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -55,6 +55,7 @@ if typing.TYPE_CHECKING:
 
 class _MessageEditState(NamedTuple):
     message_id: int
+    old_topic: str
 
 
 class WriteBox(urwid.Pile):
@@ -543,10 +544,17 @@ class WriteBox(urwid.Pile):
                     topic = self.title_write_box.edit_text
 
                 if self.msg_edit_state is not None:
+                    trimmed_topic = topic.strip()
+                    # Trimmed topic must be compared since that is server check
+                    if trimmed_topic == self.msg_edit_state.old_topic:
+                        propagate_mode = "change_one"  # No change in topic
+                    else:
+                        propagate_mode = self.edit_mode_button.mode
+
                     args = dict(
                         message_id=self.msg_edit_state.message_id,
-                        topic=topic,
-                        propagate_mode=self.edit_mode_button.mode,
+                        topic=topic,  # NOTE: Send untrimmed topic always for now
+                        propagate_mode=propagate_mode,
                     )
                     if self.msg_body_edit_enabled:
                         args["content"] = self.msg_write_box.edit_text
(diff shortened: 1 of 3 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -539,10 +539,15 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                     topic = self.title_write_box.edit_text
 
                 if self.msg_edit_state is not None:
+                    if topic == self.msg_edit_state[1]:  # FIXME need strip?
```
preetmishra, line 542: For a topic change from "Test" to "Test ", it says invalid prop mode. I think we need to strip here.

preetmishra (commented):
@neiljp Thanks for pushing a patch! It seems to work well locally. :+1: 

Re the state data structure, fetching individual items from a `tuple` (using indices) doesn't tell much about what is being extracted out of the tuple. What do you think about a `dict` instead?

I have left a minor comment regarding a FIXME comment.

neiljp:
Just returned to this after my time away as it blocks our use on czo with simple editing.

@preetmishra Just pushed an updated version addressing your points:
* using a namedtuple instead of a simple tuple (dicts are OK but are more verbose and less constrained unless one uses typeddict, and that's only enforced via mypy)
* uses the trimmed topic for comparison purposes (with notes to explain why - I checked the server code)

The namedtuple change actually found a minor test change that was missing too, so thanks for the suggestion to avoid indices (even if internal to the WriteBox)

zulipterminal/ui_tools/boxes.py
```diff
@@ -543,9 +544,16 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+                    if trimmed_topic == self.msg_edit_state.old_topic:
+                        propagate_mode = "change_one"  # No change in topic
+                    else:
+                        propagate_mode = self.edit_mode_button.mode
+
                     args = dict(
                         message_id=self.msg_edit_state.message_id,
-                        topic=topic,
+                        topic=topic,  # NOTE: Send untrimmed topic always for now
                         propagate_mode=self.edit_mode_button.mode,
```
preetmishra, line 557: This should be `propagate_mode=propagate_mode`.

preetmishra (changes requested):
@neiljp Thanks for pushing the updates! :+1:

The PR doesn't work currently but that's due to a small mishap (see my in-line comment). Other than that, the changes look good to me.

NamedTuple is even better with strict checking. Thanks for moving ahead with it. I second your decision regarding only using a trimmed topic name for equality and sending the topic as is until we have a guideline from the server.

neiljp:
@preetmishra Thanks for the reviews! Merging this now so I can get back to editing messages! :tada:

PR #1057, opened 2021-06-25
boxes: Use indicator attribute to hold compose box status.

prah23:
**PR structure:**
- **boxes: Use an indicator attribute to specify current compose box status.**
This commit introduces an attribute, `compose_box_status` to the WriteBox
class, to indicate the current compose status of the WriteBox, which can
be one of "open_with_private", "open_with_stream" or "closed". Such an
indicator would be a structured way to check for compose status before
performing status specific tasks.
- **boxes: Create a helper method to set WriteBox attributes to defaults.**
This commit adds a helper method to the WriteBox class to reset all
the WriteBox attributes to their defaults after usage by the
`stream_box_view` or `private_box_view`. This helps to ensure that
the state of the attributes do not persist beyond their function.
- **boxes: Configure `*_box_views` to set and reset WriteBox attributes.**
This commit enables the setting of the `compose_box_status` and
resetting it (by setting it to "closed") along with the other WriteBox
attributes, after a compose.
The `private_box_view` and `stream_box_view` set the `compose_box_view`
status appropriately in their call. The `GO_BACK` keypress now calls
the helper method to reset WriteBox attributes like the `stream_id` and
the `to_write_box` among others, to prevent persistence, which is not
ideal here, because these attributes are specific to the particular
compose action they were set in.
Tests updated.

**Testing and linting:**

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -188,6 +188,38 @@ class TestWriteBox:
 
         assert not write_box.model.send_private_message.called
 
+    @pytest.mark.parametrize("key", keys_for_command("GO_BACK"))
+    def test__compose_attributes_reset_for_private_compose(
+        self, key, mocker, write_box, widget_size
+    ):
+        mocker.patch("urwid.connect_signal")
+        write_box.private_box_view(
+            emails=["person1@example.com"], recipient_user_ids=[11]
+        )
+        write_box.msg_write_box.edit_text = "random text"
+
+        size = widget_size(write_box)
+        write_box.keypress(size, key)
+
+        assert write_box.to_write_box is None
+        assert write_box.msg_write_box.edit_text == ""
+        assert write_box.compose_box_status == "closed"
+
+    @pytest.mark.parametrize("key", keys_for_command("GO_BACK"))
+    def test__compose_attributes_reset_for_stream_compose(
+        self, key, mocker, write_box, widget_size
+    ):
+        write_box._set_stream_write_box_style = mocker.Mock()
+        write_box.stream_box_view(stream_id=1)
+        write_box.msg_write_box.edit_text = "random text"
+
+        size = widget_size(write_box)
+        write_box.keypress(size, key)
+
+        assert write_box.stream_id is None
+        assert write_box.msg_write_box.edit_text == ""
+        assert write_box.compose_box_status == "closed"
+
     @pytest.mark.parametrize(
         ["raw_recipients", "tidied_recipients"],
         [
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -1,3 +1,5 @@
+import datetime
+
 import pytest
 from pytest import param as case
 
@@ -54,7 +56,15 @@ class TestWriteBox:
         assert write_box.model == self.view.model
         assert write_box.view == self.view
         assert write_box.msg_edit_state is None
-        assert not write_box.sent_start_typing_status
+        assert write_box.msg_body_edit_enabled is True
+        assert write_box.stream_id is None
+        assert write_box.recipient_user_ids == []
+        assert write_box.typing_recipient_user_ids == []
+        assert write_box.to_write_box is None
+        assert isinstance(write_box.send_next_typing_update, datetime.datetime)
+        assert isinstance(write_box.last_key_update, datetime.datetime)
+        assert write_box.idle_status_tracking is False
+        assert write_box.sent_start_typing_status is False
 
     def test_not_calling_typing_method_without_recipients(self, mocker, write_box):
         write_box.model.send_typing_status_by_user_ids = mocker.Mock()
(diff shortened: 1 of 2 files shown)


neiljp:
@prah23 Is this ready for review? There are two aspects here - the new attribute to indicate when in a certain mode, plus the refactoring of the clearing/resetting - let's keep them separate. So, some thoughts:

* I can see some of the logic/ideas here, and small commits can be good, but I need to look at all the commits to understand how the new value is initialized, set and reset - so let's use a different commit structure
* It's a bit confusing that there are values reset in the _clear method that aren't removed, or previously present when the function is used instead of the explicit attribute setting
* Are there other cases where _clear could be used? (other times the box is closed? sending? initially? drafts? others?)

Specifically
* For the attribute, to make the structure clearer, maybe set a simple version which sets/resets the new attribute in relevant places as a first commit; follow-up commits can then work on increasingly using that attribute
* The _clear function could be a refactor before or after the other - being clear what is being extracted, or maybe set properly where it should have been previously?
* If the _clear function is a 'set to defaults', then call it that - and maybe it could be called from `__init__`?
* Alternatively, is the _clear function really a "hide compose"?

prah23:
I've used a different commit structure now, along what you suggested.
I'm not entirely sure about the second commit, I've renamed the helper to `_set_compose_attributes_to_defaults` and pulled "compose-specific" (excluding attributes  relevant to `FOCUS` and typeahead logic) into the helper, and I'm calling it on `__init__` and a `GO_BACK`keypress. It's produced some weird diffs, but a look at the changed file should make my intent more clear.

To the point about other cases where this helper can be used:
- sending: Once the message is sent, we're only clearing the `msg_write_box`, and triggering an `esc` keypress, so this method is triggered that way.
- drafts: There's no change of state currently on a draft, that behavior remains consistent.

For the hide compose nature of the helper, there's other calls we make, like `main_view` and `send_stop_typing_status`. I think it's better to keep it as a "set to defaults" helper than "hide compose". What are your thoughts there?

zulipterminal/ui_tools/boxes.py
```diff
@@ -220,11 +234,11 @@ def stream_box_view(
         self, stream_id: int, caption: str = "", title: str = ""
     ) -> None:
         self.set_editor_mode()
+        self.compose_box_status = "open_with_stream"
         self.stream_id = stream_id
         self.recipient_user_ids = self.model.get_other_subscribers_in_stream(
             stream_id=stream_id
         )
-        self.to_write_box = None
```
prah23, line 227: This should no longer be necessary now.

zulipterminal/ui_tools/boxes.py
```diff
@@ -680,10 +702,8 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                     self.msg_edit_state = None
                     self.keypress(size, primary_key_for_command("GO_BACK"))
         elif is_command_key("GO_BACK", key):
-            self.msg_edit_state = None
-            self.msg_body_edit_enabled = True
-            self.compose_box_status = "closed"
             self.send_stop_typing_status()
+            self._set_compose_attributes_to_defaults()
             self.view.controller.exit_editor_mode()
             self.main_view(False)
```
neiljp: Would it make sense to remove `main_view` fully now too? That seems part of the reset?

neiljp:
@prah23 This seems in a good state other than my query regarding inlining the main_view function into its calling points, though I wasn't sure if you were wanting to extend this to include more tests after our discussion?

zulipterminal/ui_tools/boxes.py
prah23: Responded on the stream, [#zulip-terminal > Resetting WriteBox attributes to defaults #T1057](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Resetting.20WriteBox.20attributes.20to.20defaults.20.23T1057).

tests/ui_tools/test_boxes.py
```diff
@@ -54,6 +56,14 @@ def test_init(self, write_box):
         assert write_box.model == self.view.model
         assert write_box.view == self.view
         assert write_box.msg_edit_state is None
+        assert write_box.msg_body_edit_enabled is True
+        assert write_box.stream_id is None
+        assert write_box.recipient_user_ids == []
+        assert write_box.typing_recipient_user_ids == []
+        assert write_box.to_write_box is None
+        assert isinstance(write_box.send_next_typing_update, datetime.datetime)
+        assert isinstance(write_box.last_key_update, datetime.datetime)
```
prah23, line 66: This isn't the best way to test these initializations, but I'm unable to patch the `datetime.now` method (I'm not sure if we can), and this seemed like a reasonable alternative.

neiljp: This is a good improvement for now - I think we'd need to adjust how we handle datetime otherwise.

neiljp:
@prah23 This is a great refactor, and one we can also build upon :+1: I'm merging this now, only having adjusted the style of some asserts in test_init to match the additions :tada:

PR #1058, opened 2021-06-28
Add media support in ZT

### LINKED ISSUE #764: TRACKING: Handle message links

Ezio-Sarthak:
Thanks to @preetmishra and @amanagr for their substantial work in #740 and #359, that helped wrapping this PR up with a sober feature version easily.

Partially fixes #764.

**Commit flow:**
-  helper: Add download_media() and open_media() to handle media links.
-  refactor: Use normalized OS-specific file names.
-  buttons: Use process_media() to handle media via MessageLinkButton.
-  helper: Show downloading & downloaded update in process_media().
-  refactor: buttons/helper: Use callbacks to isolate UI-centric parts.
-  views: Extend PopUpConfirmationView to accept location parameter.
-  core/helper: Show PopUpConfirmationView before opening any media.

**Screenshots/GIFs**

media_links_open

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -35,7 +35,11 @@ from zulipterminal.config.regexes import (
     REGEX_QUOTED_FENCE_LENGTH,
 )
 from zulipterminal.config.ui_mappings import StreamAccessType
-from zulipterminal.platform_code import normalized_file_path, successful_GUI_return_code
+from zulipterminal.platform_code import (
+    PLATFORM,
+    normalized_file_path,
+    successful_GUI_return_code,
+)
 
 
 class StreamData(TypedDict):
@@ -719,6 +723,34 @@ def suppress_output() -> Iterator[None]:
     finally:
         os.dup2(stdout, 1)
         os.dup2(stderr, 2)
+
+
+@asynch
+def process_media(controller: Any, link: str) -> None:
+    """
+    Helper to process media links.
+    """
+    if not link:
+        controller.report_error("The media link is empty")
+        return
+
+    media_path = download_media(controller, link)
+    tool = ""
+
+    # TODO: Add support for other platforms as well.
+    if PLATFORM == "WSL":
+        tool = "explorer.exe"
+        # Modifying path to backward slashes instead of forward slashes
+        media_path = media_path.replace("/", "\\")
+    elif PLATFORM == "Linux":
+        tool = "xdg-open"
+    elif PLATFORM == "MacOS":
+        tool = "open"
+    else:
+        controller.report_error("Media not supported for this platform")
+        return
+
+    open_media(controller, tool, media_path)
 
 
 def download_media(controller: Any, url: str) -> str:
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -35,6 +35,7 @@ from zulipterminal.config.regexes import (
     REGEX_QUOTED_FENCE_LENGTH,
 )
 from zulipterminal.config.ui_mappings import StreamAccessType
+from zulipterminal.platform_code import successful_GUI_return_code
 
 
 class StreamData(TypedDict):
@@ -755,7 +756,7 @@ def open_media(controller: Any, tool: str, media_path: str) -> None:
             command, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
         )
         exit_status = process.returncode
-        if exit_status != 0:
+        if exit_status != successful_GUI_return_code():
             error = [
                 " The tool ",
                 ("footer_contrast", tool),
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -3,7 +3,7 @@ import subprocess
 import time
 from collections import OrderedDict, defaultdict
 from contextlib import contextmanager
-from functools import wraps
+from functools import partial, wraps
 from itertools import chain, combinations
 from re import ASCII, MULTILINE, findall, match
 from tempfile import NamedTemporaryFile
@@ -734,7 +734,10 @@ def process_media(controller: Any, link: str) -> None:
         controller.report_error("The media link is empty")
         return
 
-    media_path = download_media(controller, link)
+    show_download_status = partial(
+        controller.view.set_footer_text, "Downloading your media..."
+    )
+    media_path = download_media(controller, link, show_download_status)
     tool = ""
 
     # TODO: Add support for other platforms as well.
@@ -753,7 +756,9 @@ def process_media(controller: Any, link: str) -> None:
     open_media(controller, tool, media_path)
 
 
-def download_media(controller: Any, url: str) -> str:
+def download_media(
+    controller: Any, url: str, show_download_status: Callable[..., None]
+) -> str:
     """
     Helper to download media from given link. Returns the path to downloaded media.
     """
@@ -771,8 +776,11 @@ def download_media(controller: Any, url: str) -> str:
             for chunk in response.iter_content(chunk_size=8192):
                 if chunk:  # Filter out keep-alive new chunks.
                     file.write(chunk)
+                    show_download_status()
 
+        controller.report_success([" Downloaded ", ("bold", media_name)])
         return normalized_file_path(local_path)
+
     return ""
 
 
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -35,7 +35,7 @@ from zulipterminal.config.regexes import (
     REGEX_QUOTED_FENCE_LENGTH,
 )
 from zulipterminal.config.ui_mappings import StreamAccessType
-from zulipterminal.platform_code import successful_GUI_return_code
+from zulipterminal.platform_code import normalized_file_path, successful_GUI_return_code
 
 
 class StreamData(TypedDict):
@@ -740,7 +740,7 @@ def download_media(controller: Any, url: str) -> str:
                 if chunk:  # Filter out keep-alive new chunks.
                     file.write(chunk)
 
-        return local_path
+        return normalized_file_path(local_path)
     return ""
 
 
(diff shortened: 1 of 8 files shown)


zulipterminal/helper.py
```diff
@@ -732,3 +734,52 @@ def suppress_output() -> Iterator[None]:
...
+        process = subprocess.run(
+            command, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, shell=True
+        )
+        exit_status = process.returncode
+        if exit_status != 0:
+            error = [
+                " The tool ",
+                ("bold", tool),
+                " did not run successfully" ". Exited with ",
+                ("bold", str(exit_status)),
```
neiljp, line 784: `"bold"` as a style doesn't exist - we have footer_contrast now.

zulipterminal/helper.py
```diff
@@ -732,3 +734,52 @@ def suppress_output() -> Iterator[None]:
...
+            for chunk in r.iter_content(chunk_size=8192):
+                if chunk:  # Filter out keep-alive new chunks.
+                    f.write(chunk)
+
+    if WSL:
+        command = f'explorer.exe `wslpath -w "{media_path}"`'  # Needs testing.
+    elif LINUX:
+        command = f"xdg-open {media_path}"
+    elif MACOS:
+        command = f"open {media_path}"
```
neiljp, line 765: These should really be `List[str]`. There is discussion elsewhere, though I can't seem to dig it up on github right now.

zulipterminal/helper.py
```diff
@@ -732,3 +734,52 @@ def suppress_output() -> Iterator[None]:
...
+        with NamedTemporaryFile(
+            mode="wb", delete=False, prefix="zt-", suffix="-{}".format(media_name)
+        ) as f:
+            media_path = f.name
+            for chunk in r.iter_content(chunk_size=8192):
+                if chunk:  # Filter out keep-alive new chunks.
+                    f.write(chunk)
+
+    if WSL:
+        command = f'explorer.exe `wslpath -w "{media_path}"`'  # Needs testing.
```
neiljp, line 761: Let's leave this out until we know we have a version that works, and we're sure is good, particularly with that quoting.

While we support only these three, we should handle the `else` case (or equivalent) in any case, so we'll need `command` to be defined, or have some meaning, and perhaps pop up an 'unsupported' notice or report an error. That means it can apply to the WSL case for now too.

zulipterminal/helper.py
```diff
@@ -732,3 +734,52 @@ def suppress_output() -> Iterator[None]:
...
+    tool = command.split()[0]
+    open_media(controller, command, tool)
+
+
+@asynch
+def open_media(controller: Any, command: str, tool: str) -> None:
+    error = []
+    try:
+        process = subprocess.run(
+            command, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, shell=True
```
neiljp, line 776: It would be good to avoid `shell=True` for security reasons (eg. https://docs.python.org/3/library/subprocess.html#security-considerations)

zulipterminal/helper.py
```diff
@@ -732,3 +734,52 @@ def suppress_output() -> Iterator[None]:
...
+@asynch
+def process_media(controller: Any, media_link: str) -> None:
+    media_name = media_link.split("/")[-1]
+    client = controller.client
+    auth = requests.auth.HTTPBasicAuth(client.email, client.api_key)
+
+    with requests.get(media_link, auth=auth, stream=True) as r:
+        r.raise_for_status()
+        with NamedTemporaryFile(
+            mode="wb", delete=False, prefix="zt-", suffix="-{}".format(media_name)
```
neiljp, line 750: This can now be an f-string.

zulipterminal/helper.py
```diff
@@ -732,3 +734,52 @@ def suppress_output() -> Iterator[None]:
...
+        command = f"xdg-open {media_path}"
+    elif MACOS:
+        command = f"open {media_path}"
+
+    tool = command.split()[0]
+    open_media(controller, command, tool)
+
+
+@asynch
+def open_media(controller: Any, command: str, tool: str) -> None:
```
neiljp, line 772: This is open_media, so it's surprising that the open-command selection is not present here.

It's been a while since I've looked at this code, but it may be clearer not having one function call the other, ie. download_media(url) -> local_path, open_media(local_path) ?

zulipterminal/ui_tools/buttons.py
```diff
@@ -370,6 +370,11 @@ def handle_link(self, *_: Any) -> None:
         server_url = self.model.server_url
         if self.link.startswith(urljoin(server_url, "/#narrow/")):
             self.handle_narrow_link()
+        elif self.link.startswith(urljoin(server_url, "/user_uploads/")):
+            # Exit pop-up promptly, let the media download in the background.
+            if isinstance(self.controller.loop.widget, urwid.Overlay):
```
neiljp: We should probably refactor this to be a controller method - we already have one case of this style.

zulipterminal/helper.py
```diff
@@ -737,7 +737,9 @@ def suppress_output() -> Iterator[None]:
 
 
 @asynch
-def process_media(controller: Any, media_link: str) -> None:
+def process_media(
+    controller: Any, media_link: str, callback: Callable[..., None]
```
neiljp, line 741: callback is a very general name - it's worth clarifying when it's called in the name. This could be after downloading, during downloading...

zulipterminal/helper.py
```diff
@@ -751,9 +753,7 @@ def process_media(controller: Any, media_link: str) -> None:
             for chunk in r.iter_content(chunk_size=8192):
                 if chunk:  # Filter out keep-alive new chunks.
                     f.write(chunk)
-                    controller.view.set_footer_text(
-                        [" Downloading ", ("bold", media_name)]
-                    )
+                    callback([" Downloading ", ("bold", media_name)])
```
neiljp, line 756: This is not coupled to the controller etc, but you need to know that it's an urwid styled-text being passed in. You could specify the parameter in the calling function, and have this just be `callback()`. A different name would help too - see elsewhere.

zulipterminal/ui_tools/views.py
```diff
@@ -1126,19 +1133,31 @@ def __init__(
         no._w = urwid.AttrMap(urwid.SelectableIcon("No", 4), None, "selected")
         display_widget = urwid.GridFlow([yes, no], 3, 5, 1, "center")
         wrapped_widget = urwid.WidgetWrap(display_widget)
-        prompt = urwid.LineBox(
-            urwid.ListBox(
-                urwid.SimpleFocusListWalker([question, urwid.Divider(), wrapped_widget])
-            )
-        )
+        widgets = [question, urwid.Divider(), wrapped_widget]
+        prompt = urwid.LineBox(urwid.ListBox(urwid.SimpleFocusListWalker(widgets)))
```
neiljp, line 1137: A good follow-up to this would be to improve the styling and unify it with the general popup - the border is not that clear compared to those, in retrospect.

neiljp (commented):
@Ezio-Sarthak It's good to see more work on this :+1:

I've not looked at all the tests in detail, but this is some feedback for this version of the PR - have you checked my/others comments from the previous PRs?

There seem to be some changes from the previous iteration in terms of eg. subprocess, so although I've noted some reasons you may want to check how up to date your PR is with that?

prah23 (changes requested):
Thanks for bringing this work up to speed @Ezio-Sarthak!

I've tested this on WSL2 (Ubuntu), and it works well for the file formats I've tried: JPEG, PNG, PDF, and MP4. There's a slight issue as it stands, the footer notification for the tool not running successfully is triggered every time I open something. I presume there's a different exit code being returned on Windows or something of the like. You might have to handle that based on the platform.

I notice Neil has also just reviewed, I wanted to quickly put in a suggestion regarding this while you're still working on it so it's easier to accommodate.

zulipterminal/helper.py
Ezio-Sarthak: Right. I just passed in the params to show the media name as well. But I think showing just `Downloading your media ...` is good, since I'm prompting the user the filename while opening it in any case :+1:

zulipterminal/helper.py
```diff
@@ -711,3 +715,88 @@ def suppress_output() -> Iterator[None]:
...
+        controller.view.set_footer_text, "Downloading your media..."
+    )
+    media_path = download_media(controller, link, show_download_status)
+    tool = ""
+
+    # TODO: Add support for other platforms as well.
+    if PLATFORM == "WSL":
+        tool = "explorer.exe"
+        # Converting absolute path to WSL path
+        media_path = f'`wslpath "{media_path}"`'
```
andersk, line 739: What is interpreting these backticks? What else is it interpreting? Something is definitely wrong here.

Ezio-Sarthak: Yeah. It is a bad way of nesting bash commands. One should instead use `$()` I think. As I said in the stream, we'd potentially not need this now.

andersk: No, my point is there should not _be_ a Bash. Backticks should not be interpreted here. If they _were_ being interpreted here—even if we’re no longer _using_ them—that indicates a deep problem that we need to investigate, because `media_path` could itself contain backticks or other characters that we really don’t want to be interpreted.

The same problem applies to `$()`, `""`, or any other kind of “quote” construct. The first rule of quoting is that you need to ask, _what if the quoting characters are present inside the string you’re quoting?_

Ezio-Sarthak: Makes sense. Thanks for clarifying :+1:

neiljp:
Merged with slight adjustments in #1223 :tada:

Thanks @amanagr, @preetmishra and @Ezio-Sarthak 👍

PR #1059, opened 2021-06-28
Platform handling improvements and WSL bugfix.

### LINKED ISSUE #791: Improve platform handling
Currently we use `platform` to determine which platform we're running on, currently primarily for notifications.

There are currently some limitations in this:
* while not expected, we don't ensure that these values are mutually exclusive (only one is set)
* we explicitly only check for (and test on) these platforms

The former means - admittedly in likely rare situations - that we may load multiple/incorrect notifications. Multiple platforms should be an error that either the application or user disambiguates.

The latter means that other platforms, such as likely BSD or more broadly POSIX-like systems may not run properly, or at minimum will eg just transparently not show notifications as it stands.

We currently have the platform detection and notification code in `helper.py`; I'd suggest we extract this code and have platform-dependent code be isolated in a separate module, with other code not being otherwise aware of any underlying platform, just whether things are working/available or not.

**edit**
Expected functionality to complete this:
* [X] Ensure mutually exclusive detection (#1059)
* [ ] Isolate platform-specific code into file (mostly via #1059 ?)
* [ ] Output detected platform on startup and in About menu
* [ ] Relate platform detection to platforms we're running CI on (ensure we're running where we think; potentially expand CI?)
* [ ] If possible, ensure functionality that is platform-specific is actually tested on those platforms (only)

zee-bit:
This PR improves our platform handling by defining an isolated module `config/platform_detection.py` that contains the platform detection code and uses a single variable `PLATFORM` to hold the name of the platform that the current ZT instance is loaded on. This fixes a bug in WSL where multiple platforms were detected. Related discussion in: [#zulip-terminal>Platform handling (#T791) and WSL issues](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Platform.20handling.20.28.23T791.29.20.20and.20WSL.20issues)

Fixes #791.

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -4,7 +4,6 @@ import pytest
 from pytest import param as case
 from pytest_mock import MockerFixture
 
-import zulipterminal.helper
 from zulipterminal.api_types import Composition
 from zulipterminal.helper import (
     Index,
@@ -18,6 +17,7 @@ from zulipterminal.helper import (
     notify_if_message_sent_outside_narrow,
     powerset,
 )
+from zulipterminal.platform_code import AllPlatforms, SupportedPlatforms
 
 
 MODULE = "zulipterminal.helper"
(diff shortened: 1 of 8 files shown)


neiljp (commented):
@zee-bit Some initial thoughts: (possibly some are in the issue?)
* I think a good long-term goal would be to have all the platform-specific code in one module, and it's likely easier for that to be where the detection happens too; on that basis and given the detection would then be a detail of the module, I'm not entirely convinced it fits in `config/`.
* While it's not always possible, it would be good to see this first commit not just add unused code but migrate to using it - ie. be a refactor of the detection into the new file, with the old platform variables being set according to `PLATFORM` - the second commit would then become a migration from those to use `PLATFORM` (so be slightly smaller). If we go the approach of one module for minimal platform-specific functions, that could then be a good follow-up.
* We don't currently test the platform detection, which I think we could do with marks/skips; that's maybe not explicitly part of #791 but we should test it more explicitly if we can.
* You've reused the Literal definition in multiple places which would normally make me want to define it instead, along with the multi-part definition I suggested on the stream (like we do with Flags), to leave "unsupported" separate.
* Another follow-up might be to add an extra helper to the module to check the platform is supported and/or give details of that

neiljp:
@zee-bit Thanks for this refactor/bugfix :+1: I'm going to merge this now after a few minor adjustments :tada:

While this migrates almost all `platform` use into the new module, we can discuss other functionality which might be worth having there too (open in browser? open media?)

I've left #791 from being closed by this, since while this does resolve the multiple-platform issue, it also mentions other issues, which I've expanded on in a task list there after the discussion in the stream.

PR #1060, opened 2021-06-29
Panel Search Box UI Enhancements

Rohitth007:
This PR deals UI enhancements related to display and reattempting search.
While reattempting search:
* The caption sets to empty string
* The previous edit_text does not get erased so that users can edit typos.

Display:
* Aligns the text inside search box with the rest of the panel
* Displays edit text on new line, which reads clearer and helps render long searches properly

### DIFF
--- a/zulipterminal/ui.py
+++ b/zulipterminal/ui.py
@@ -225,7 +225,6 @@ class View(urwid.WidgetWrap):
             self.users_view.keypress(size, "w")
             self.show_left_panel(visible=False)
             self.show_right_panel(visible=True)
-            self.user_search.set_edit_text("")
             self.controller.enter_editor_mode_with(self.user_search)
             return key
         elif is_command_key("SEARCH_STREAMS", key) or is_command_key(
@@ -240,7 +239,6 @@ class View(urwid.WidgetWrap):
                 search_box = self.topic_w.topic_search_box
             else:
                 search_box = self.stream_w.stream_search_box
-            search_box.set_edit_text("")
             self.controller.enter_editor_mode_with(search_box)
             return key
         elif is_command_key("OPEN_DRAFT", key):
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -1841,7 +1841,7 @@ class PanelSearchBox(urwid.Edit):
             self.panel_view.keypress(size, "esc")
         elif is_command_key("ENTER", key) and not self.panel_view.empty_search:
             self.panel_view.view.controller.exit_editor_mode()
-            self.set_caption([("filter_results", " Search Results "), " "])
+            self.set_caption([("filter_results", " Search Results \n"), " > "])
             self.panel_view.set_focus("body")
             if hasattr(self.panel_view, "log"):
                 self.panel_view.body.set_focus(0)
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -1806,7 +1806,7 @@ class PanelSearchBox(urwid.Edit):
     ) -> None:
         self.panel_view = panel_view
         self.search_command = search_command
-        self.search_text = f"Search [{', '.join(keys_for_command(search_command))}]: "
+        self.search_text = f" Search [{', '.join(keys_for_command(search_command))}]: "
         self.search_error = urwid.AttrMap(
             urwid.Text([" ", INVALID_MARKER, " No Results"]), "search_error"
         )
@@ -1841,7 +1841,7 @@ class PanelSearchBox(urwid.Edit):
             self.panel_view.keypress(size, "esc")
         elif is_command_key("ENTER", key) and not self.panel_view.empty_search:
             self.panel_view.view.controller.exit_editor_mode()
-            self.set_caption([("filter_results", "Search Results"), " "])
+            self.set_caption([("filter_results", " Search Results "), " "])
             self.panel_view.set_focus("body")
             if hasattr(self.panel_view, "log"):
                 self.panel_view.body.set_focus(0)
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -1811,11 +1811,11 @@ class PanelSearchBox(urwid.Edit):
             urwid.Text([" ", INVALID_MARKER, " No Results"]), "search_error"
         )
         urwid.connect_signal(self, "change", update_function)
-        super().__init__(caption="", edit_text=self.search_text)
+        super().__init__(caption=self.search_text, edit_text="")
 
     def reset_search_text(self) -> None:
-        self.set_caption("")
-        self.set_edit_text(self.search_text)
+        self.set_caption(self.search_text)
+        self.set_edit_text("")
 
     def valid_char(self, ch: str) -> bool:
         # This method 'strips' leading space *before* entering it in the box
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -363,7 +363,7 @@ class StreamsView(urwid.Frame):
         if is_command_key("SEARCH_STREAMS", key):
             _, self.focus_index_before_search = self.log.get_focus()
             self.set_focus("header")
-            self.stream_search_box.set_caption("")
+            self.stream_search_box.set_caption(" ")
             return key
         elif is_command_key("GO_BACK", key):
             self.stream_search_box.reset_search_text()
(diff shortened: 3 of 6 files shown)


Rohitth007:
The last commit fixes a **bug** caused by mouse clicking on search, **but I'm not entirely sure about the solution.**

zulipterminal/ui_tools/boxes.py
```diff
@@ -1726,7 +1726,7 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             self.panel_view.keypress(size, "esc")
         elif is_command_key("ENTER", key) and not self.panel_view.empty_search:
             self.panel_view.view.controller.exit_editor_mode()
-            self.set_caption([("filter_results", "Search Results"), " "])
+            self.set_caption([("filter_results", " Search Results"), " "])
```
preetmishra: I think adding " " as another item would look better as this currently applies the styling to `Search results` as well. :)

zulipterminal/ui_tools/views.py
```diff
@@ -367,7 +367,7 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         if is_command_key("SEARCH_STREAMS", key):
             _, self.focus_index_before_search = self.log.get_focus()
             self.set_focus("header")
-            self.stream_search_box.set_caption("")
+            self.stream_search_box.set_caption(" ")
```
preetmishra, line 370: What do you think about adding an extra " " in a method that overrides urwid's `set_caption` method in `PanelSearchBox` instead?

zulipterminal/ui_tools/boxes.py
```diff
@@ -1726,7 +1726,7 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             self.panel_view.keypress(size, "esc")
         elif is_command_key("ENTER", key) and not self.panel_view.empty_search:
             self.panel_view.view.controller.exit_editor_mode()
-            self.set_caption([("filter_results", " Search Results"), " "])
+            self.set_caption([("filter_results", " Search Results\n"), " > "])
```
preetmishra, line 1729: Let's wait for everybody's views on this in the stream.

zulipterminal/ui_tools/views.py
```diff
@@ -368,6 +368,7 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             _, self.focus_index_before_search = self.log.get_focus()
             self.set_focus("header")
             self.stream_search_box.set_caption(" ")
+            self.view.controller.enter_editor_mode_with(self.stream_search_box)
```
preetmishra, line 371: I understand what the code does but I suspect that the commit description isn't accurate.

zulipterminal/ui.py
```diff
@@ -208,8 +208,7 @@ def keypress(self, size: urwid_Box, key: str) -> Optional[str]:
             or is_command_key("PRIVATE_MESSAGE", key)
         ):
             self.body.focus_col = 1
-            self.middle_column.keypress(size, key)
-            return key
+            return self.middle_column.keypress(size, key)
```
preetmishra, line 211: Instead of minor changes, we could point out what the code does exactly in the commit title. :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -1731,3 +1731,13 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             if hasattr(self.panel_view, "log"):
                 self.panel_view.body.set_focus(0)
         return super().keypress(size, key)
+
+    def mouse_event(
```
preetmishra, line 1735: I am not sure if we should support this in the first place given that we mention in the FAQ that mouse support is limited. I think this could go in another PR if we decide to have it later.

preetmishra (commented):
@Rohitth007 Thanks for your work. This seems to work well functionally. :+1:

That said, I would like to mention a couple of points related to the commits:
* Given that we label all other types of changes in the version control, _feature_ changes do not need labels like enhancement, feature, etc (it is implicit). You could take a look at the recently merged commit guidelines in the README.
* We could improve the commit titles and description. Ideally, we should be able to identify what the commit is doing just by reading the commit title. Generic titles (e.g.: "Improve UI") don't give enough context. Commit descriptions should aim to give an overview than explain what the code is doing literally.

Other than that, the PR looks good overall. I have left a few minor in-line comments. :)

zulipterminal/ui_tools/boxes.py
Rohitth007: I think the keypress refactor would avoid this issue and mouse would work as normal without causing havoc because currently it does.

zulipterminal/ui_tools/views.py
Rohitth007: I was instead thinking, during the Panel refactor I could group this and 1 other common line into a function call `enter_search_mode` inside PanelSearchBox

zulipterminal/ui_tools/boxes.py
neiljp: I definitely agree with this. Generally the extra space looks good though :+1:

zulipterminal/ui_tools/views.py
neiljp: This would be a good function to add during the first commit - it avoids spreading general `set_caption` calls through the code, which may also be used for other things, as well as providing a central place for this kind of change - and it's easy to grep for a unique name :)

zulipterminal/ui_tools/boxes.py
neiljp: Agreed - also if this is separate to other changes that are less controversial then having it last make it easier to merge the others first.

neiljp (commented):
@Rohitth007 You have a range of changes here - design changes (one needing discussion), general refactoring (towards panel simplification/combination), and a mouse bugfix. I'm not sure if they're all connected or dependent - though you may be changing the same files of course - but having them in the same PR isn't going to help this get merged very fast as each aspect has a range of different discussions required, and it makes reviewing more complex and time-consuming.

My impression is that the first 2-3 could be merged fairly quickly with a few changes and some simplification to the commit text - we could discuss writing styles in the ZT stream or the ZT GSoC one, or maybe PMs (but fewer people can contribute that way).

The refactoring is good to see too, but unless the changes are tightly coupled with the others it may be easier to have a separate PR for it, and for the mouse bugfix separately.

zulipterminal/ui_tools/boxes.py
Rohitth007: I think it look weird that way, having spaces on both sides is another option which looks better. (See below)

zulipterminal/ui_tools/views.py
Rohitth007: @preetmishra @neiljp Previously, the other common line that I was talking about was `enter_editor_mode` but now that I see that enter_editor_mode would be removed after the keypress refactor and set_caption would be called only once anyway after the Panel refactor, I don't think this needs to be moved into PanelSearchBox using enter_search_mode or similar.

Rohitth007:
@preetmishra @neiljp I have split the PR as discussed and also dropped the mouse-fix commit as I think it would automatically get resolved with the keypress refactor. This PR only contains the Panel Search UI improvement features. (3 commits)

zulipterminal/ui_tools/boxes.py
Rohitth007: I'm closing this as I'm hoping this would automatically get resolved

Rohitth007:
The rest of the commits can be found at #1072.

zulipterminal/ui.py
Rohitth007: I've made some changes and can be found at #1072

zulipterminal/ui_tools/views.py
Rohitth007: I've made some changes and can be found at #1072

neiljp:
@Rohitth007 I just merged the first two of these separately since they are a definite improvement :tada:

However, there's been little feedback or testing of the other change, which could possibly be more convincing accompanied by other changes - such as adding hotkey hints, using a horizontal bar of color to indicate results, and whether we want a search 'prompt' (the `>`).

There is also scope for unifying some parts of the message search UI, at least in terms of how it is presented, after this change.

neiljp:
I'm marking this as a PR completion candidate, but this will likely need some rounds of iteration to generate a consistent design, or perhaps a proposal and discussion beforehand to eg. consider how hotkeys currently work and how to represent that.

The main hotkey hinting issue here is how once we enter a search state then users get no hints (at a minimum of) how to 
- exit the search
- edit the search

So, dynamically showing the appropriate keys would be useful for that.

For the multi-line layout, my previous concern was re the search UI jumping; I'm not sure there's a way around that.

Re a 'prompt' (like the `>`), at this point I'd be mostly concerned whether it looks like part of the search text. Styling that, or having a indent with a space or two, may be sufficient and clean.

Note that some of the original PR here has been merged already, but was discussed at [#zulip-terminal>Panel search box UI](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Panel.20search.20box.20UI/near/1182198).

PR #1062, opened 2021-07-02
Fix bad rendering of new topics when topics list was open.

zee-bit:
This PR fixes the bug that made incoming new topics rendered badly when the topic list was open due to incorrect parameters being passed to `TopicButton`. This PR also migrates `TopicButton` to accept only named parameters so that there's no scope of bug and connascence between parts of the program is reduced.

The discussion regarding this bug can be found in [#zulip-terminal>New-topic topic-list bug](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/New-topic.20topic-list.20bug/)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -86,6 +86,7 @@ def user_button(mocker):
         controller=mocker.patch("zulipterminal.core.Controller"),
         view=mocker.patch("zulipterminal.ui.View"),
         state_marker="*",
+        count=0,
     )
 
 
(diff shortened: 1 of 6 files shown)


neiljp:
@zee-bit Thanks again for finding and fixing the bug :+1: As per the stream, I've merged the first commit (after blackening), and I suggest we can include removal of the default count value in the remaining commits.

neiljp:
@zee-bit Thanks for the update for the counts :+1: I just pushed a minor commit text adjustment to remove width mention (that was removed already elsewhere?) and add a missing `*`.

I'm waiting for tests to pass and then will merge :tada:

PR #1065, opened 2021-07-05
Improve incomplete theme warning text when all are incomplete.

Rohitth007:
This commit changes the warning text shown at startup if the theme
is incomplete. If there aren't any complete themes to suggest
the warning displays "(all themes are incomplete)".

Tests updated.


neiljp:
@Rohitth007 Thanks for the quick adjustment :tada:

Issue #830, opened 2020-11-18
Adding Issue + PR Templates


HarshCasper:
## Description 

- Add an Issue Template to streamline creating new Issues 
- Add a PR Template to document the changes implemented. 

I would like to work on the same. This would make it easy for the reviewers to review the important points of the PR and make necessary request changes for them.

neiljp:
@HarshCasper Thanks for opening this issue!

Re issue templates, I've noticed some other projects/repos now use the github feature of multiple templates, which helps triage feedback and focus the information requested. While we could support a basic issue template first, the breadth expected likely would make this noisy.

Re PR templates, this seems feasible to work on more directly, though needs review of the existing formats and requests that we typically make of contributors.

I'd welcome your work on this issue, though this may end up needing quite some fine-tuning by contributors more experienced with the project.

neiljp:
I listed some example issue types in #**zulip-terminal>Issue types** for open discussion on chat.zulip.org.

neiljp:
Now that ZT is becoming more popular, I'm bumping the priority of this.

neiljp:
The pull request template has gone through various iterations already, so with #1488 or a followup, this issue will be resolved.

neiljp:
Since we now have a structure in place from #1488, I'm going to close this - there will always be iterations we can do in future!

PR #1071, opened 2021-07-08
github: Add pull request template.

neiljp:
**What does this PR do?**

Adds a pull request template, like I'm using here :)

Take a look at the raw and rendered versions to get a full idea of the contributor experience.

- [X] New Feature?
- [X] Fixes Bug? Partially #830
- [ ] Internal Refactor?

**Tested?** 
- [X] Manually 
- [ ] Tests adapted (if necessary)
- [ ] Tests added/changed for any new behavior
- [ ] Passed linting & tests (overall PR)

**Notes/Doubts/Concerns/Questions** 

This is more detailed than others I've seen, eg. https://raw.githubusercontent.com/zulip/zulip/master/.github/pull_request_template.md

Do we need the top title? Or feature/bug boxes? (except for suggesting what it may fix?)

Do the testing boxes help? Do they just take up space?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,36 @@
+<!-- Please see https://github.com/zulip/zulip-terminal#contributor-guidelines ! -->
+
+**What does this PR do?**  <!-- Overall description goes here -->
+
+<!-- If fixing a filed bug or new feature, add 'Fixes #<issue>' or 'Partial fix for #<issue>' -->
+
+<!-- Add a link to a discussion on chat.zulip.org, if relevant -->
+
+**Tested?** <!-- Fine to leave some of these unchecked if this is a draft/work-in-progress -->
+- [ ] Manually
+- [ ] Existing tests (adapted, if necessary)
+- [ ] New tests added (for any new behavior)
+- [ ] Passed linting & tests (each commit)
+<!-- Code must pass CI (GitHub Actions) before merging - look for the green tick! -->
+
+<!-- See https://github.com/zulip/zulip-terminal#commit-style -->
+**Commit flow** <!-- if more than one commit; add/delete/fill-in as appropriate -->
+<!-- For example:
+- first commit doing some thing
+- maybe multiple commits doing similar things
+-->
+
+**Notes & Questions** <!-- if any; add/delete/fill-in as appropriate -->
+<!-- For example:
+- this doesn't include feature X (yet?)
+- unsure about Y
+- should this do Z?
+-->
+
+**Interactions** <!-- if any; add/delete/fill-in as appropriate -->
+<!-- eg.
+- Waiting on #<PR>
+- Blocks #<PR>
+-->
+
+**Visual changes** <!-- if any; add/delete/fill-in with screenshot/diagram as appropriate -->


.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
+<!-- Please see https://github.com/zulip/zulip-terminal#contributor-guidelines ! -->
+
+**What does this PR do?**
+
+Overall description goes here!
+
+- [ ] New Feature?
+- [ ] Fixes Bug? [Fixes <issue>; Partially <issue> | Fixes point <n> of <issue> | Resolves an issue that's not filed]
```
zee-bit, line 8: Perhaps also include the discussion link, if there's no issue filed for the fix?

.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+- [ ] Internal Refactor?
+
+**Tested?** <!-- Fine to leave some of these unchecked! -->
+- [ ] Manually
+- [ ] Tests adapted (if necessary)
+- [ ] Tests added/changed for any new behavior
```
zee-bit, line 14: Does the "adapted" and "changed" carry the same meaning here? Then we could remove the first check-box?

.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+- [ ] Passed linting & tests (overall PR)
+- [ ] Passed linting & tests (each commit)
+
+<!-- See https://github.com/zulip/zulip-terminal#commit-style -->
+**Commit flow** <!-- if more than one commit; add/delete/fill-in as appropriate -->
+* first commit doing some thing
```
zee-bit, line 20: Although I've used this field in some of my own PR's, I wonder if it's really necessary?
The "Commit flow" is essentially just the commit headers, as we have documented in that link, which can be easily accessed from the commits tab at the top?

.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+* maybe multiple commits doing similar things
+* ...
+
+**Notes/Doubts/Concerns/Questions** <!-- if any; add/delete/fill-in as appropriate -->
```
zee-bit, line 24: This field is great. :+1: 
Can we simplify the title to just "Notes/Concerns"?
Doubts/Concerns/Questions sounds pretty similar to me.

zee-bit (commented):
Thanks for getting this done, Neil. It already looks amazing. :+1: 

This is going to be very helpful, particularly, for new contributors. It will also be helpful for reviewers in getting to know what the PR does (abstractly), and how ready it is - is it properly tested, does the linting pass are the test added yet. Reviewers can also provide suggestions for any "Notes/Concerns" that the author has mentioned in the PR template without looking at the code - if possible too.

.github/pull_request_template.md
neiljp: Good point, though I think that's a separate entry to suggest including.

.github/pull_request_template.md
neiljp: The intent is to ensure that existing tests still pass, via slight changes if necessary - which is distinct from covering different new behavior. Perhaps the two should be explicitly "Existing tests pass (adapted if necessary)" and "New tests added (if new behavior)". Maybe they're too long though.

.github/pull_request_template.md
neiljp: I agree 4 is too many. Maybe just "Notes" and push the possible content into the examples.

Of course the titles are editable/removable, but people may not think that, so simple titles may be easier as users are more likely to think to at least fill in their own content I hope :)

.github/pull_request_template.md
neiljp: I agree the commit list can be seen in the PR itself - and does change over time, often, and it's not necessary to keep updating the text here as that happens.

I do feel we could benefit from having something here though - maybe others would like to weigh in.

.github/pull_request_template.md
Rohitth007: Maybe we could follow the same wording as in the commit style guide?
Tests added (new feature)
Tests updated (refactor/feature that needs minor amending in existing tests)

neiljp:
I'm more inclined to remove the first 3 checkboxes as GitHub includes them as a list of tasks - which they are not. I'm not sure there's an easy alternative, but we can simply hint at these elements in the description prompt, or in the commit flow perhaps too?

.github/pull_request_template.md
Ezio-Sarthak: @Rohitth007 I think that'd require the contributor to explicitly write details in parenthesis? I believe we're keener on the tests in the PR in general, and specific additions/amendments are addressed in commits in any case.

Rohitth007: @Ezio-Sarthak No, I just used the parenthesis for explaining added vs updated here

.github/pull_request_template.md
neiljp: @Rohitth007 That's what the examples are intended to indicate, ie. that it's a summary. I meant to include that they're only ~ commit titles, but I don't want to motivate just copy and pasting those either.

.github/pull_request_template.md
```diff
@@ -0,0 +1,34 @@
...
+**Notes & Questions** <!-- if any; add/delete/fill-in as appropriate -->
+<!-- For example:
+- this doesn't include feature X (yet?)
+- unsure about Y
+- should this do Z?
+-->
+
+**Interactions** <!-- if any; add/delete/fill-in as appropriate -->
+- Waiting on other PR(s)? What numbers?
+- Blocks other PR(s)? What numbers?
```
preetmishra, line 32: Should these two points exist as a comment like the above two sections?

preetmishra (commented):
@neiljp Thanks for working on this! :+1: This looks great. I have left a minor in-line comment.

PR #1063, opened 2021-07-03
[WIP] Refactor keypress to follow urwid rules

Rohitth007:
This PR changes the current keypress flow to follow urwid rules.
This is built on PR #974  would need PR #1060 to proceed.

PR #1072, opened 2021-07-08
Panel refactor

Rohitth007:
**What does this PR do?**  
Currently only contains prep commits that help with both Panel refactor and keypress refactor.

**Tested?** 
- [x] Manually

**Commit flow** 

- **Commit 1**: refactor: ui/views: Move enter_editor_mode closer to search box.
- **Commit 2**: refactor: ui: Simplify key redirection to directly return key. 
- **Commit 3**: refactor: view: Group related lines together in __init__.

**Notes & Questions** 

It would be useful to merge the first two.

**Interactions** 

- Blocks #1063

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -284,48 +284,34 @@ class TestView:
         view.body = mocker.Mock()
         view.controller.autohide = autohide
         view.body.contents = ["streams", "messages", mocker.Mock()]
-        view.user_search = mocker.Mock()
         view.left_panel = mocker.Mock()
         view.right_panel = mocker.Mock()
         size = widget_size(view)
-
-        super_view = mocker.patch(MODULE + ".urwid.WidgetWrap.keypress")
         view.controller.is_in_editor_mode = lambda: False
-
         view.body.focus_position = None
 
         view.keypress(size, key)
 
         view.users_view.keypress.assert_called_once_with(size, key)
         assert view.body.focus_position == 2
-        view.controller.enter_editor_mode_with.assert_called_once_with(view.user_search)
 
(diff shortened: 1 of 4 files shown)


tests/ui/test_ui.py
```diff
@@ -292,7 +290,7 @@ def test_keypress_autohide_users(self, view, mocker, autohide, key, widget_size)
 
         view.keypress(size, key)
 
-        view.users_view.keypress.assert_called_once_with(size, key)
+        view.right_panel.keypress.assert_called_once_with(size, key)
```
neiljp, line 293: Replacing users_view with right_panel seems good naming here, though the commit text doesn't indicate the same name?

neiljp (commented):
@Rohitth007 I've just resolved a conflict from the test-typing changes and merged the first commit - it's a very nice tidy, moving the action closer to the data :+1:
(also avoiding that conditional just by knowing which state we're in!)

For the second commit, there's renaming and the moving to a direct return of the key - is the latter intended to help towards the keypress-flow restructuring? There's no mention of that, but the style isn't mentioned with that (or anything else) in mind, so I wasn't clear.

I've not resolved the conflicts in the second commit onwards; it looks like there still may be a need to improve the patching in the new version like in the old one, now that we're checking the types? (@prah23 ?)

tests/ui/test_ui.py
Rohitth007: Oh, I change that right away.

Rohitth007:
Re second commit, it definitely helps with the keypress refactoring by allowing the return of unhandled keys back to urwid in the end but it doesn't directly affect the functionality in that PR as this is the outer most widget.

PR #1073, opened 2021-07-08
tests: run: Use the `tmp_path` fixture instead of `tmpdir`.

prah23:
**Commit structure:**
- **tests: run: Use the `tmp_path` fixture instead of `tmpdir`.**
The `tmpdir` fixture under `pytest` is a `py.path.local` object,
belonging to the `py` library, which is currently under maintenance
mode, mentioned in the note here:
https://py.readthedocs.io/en/latest/path.html#py-path.
This commit migrates from `tmpdir` to using the `tmp_path` fixture
instead, which is a `pathlib.Path` object.
Reference: https://docs.pytest.org/en/6.2.x/tmpdir.html

**Testing and linting:**

### DIFF
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -100,12 +100,12 @@ def test_main_help(capsys, options):
 
 
 @pytest.fixture
-def minimal_zuliprc(tmpdir):
-    zuliprc_path = str(tmpdir) + "/zuliprc"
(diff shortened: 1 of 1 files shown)


neiljp:
@prah23 Thanks for this - I just tagged it with a `refactor:` but otherwise looks good.

Another good refactor sweep to look at would be to be consistent with using using `/` vs `os.path.join`.

PR #1076, opened 2021-07-11
tests: run: Add type annotations.

prah23:
**What does this PR do?**  
Adds type annotations to the `test_run.py` file. Includes this file under the `run-mypy` tool to type-check.

**Tested?** 
- [x] Manually
- [x] Existing tests (adapted, if necessary)
- [x] New tests added (for any new behavior)
- [x] Passed linting & tests (each commit)

**Commit flow** 

- **refactor: tests: run: Add type annotations.**
This commit adds paramter and return type annotations or hints to the
`test_run.py` file, that contains tests for it's counterpart `run.py`
from  the `zulipterminal` module, to make mypy checks consistent and
improve code readability.
- **tools: Include test_run.py under files to be type-checked with mypy.**
This commit adds a new "tests" key to the `repo_python_files` dict to
let mypy know which files to type check. A new list,
`type_consistent_testfiles` is also added, to hold the list of test
files that have been type-annotated incrementally till all the test
files are consistent.

**Notes & Questions** 

I've referred to the `pytest` and `pytest-mock` docs for the type annotations on some of the fixtures, is a reference link in the commit necessary? (one reference would then be to a CHANGELOG)

PR #1077, opened 2021-07-11
tests: core: Add type annotations.

prah23:
**What does this PR do?**  
Adds type annotations to the test_core.py file. I've also added a few refactor/bugfix commits. 

**Tested?** 
- [x] Manually

**Commit flow**
- **bugfix: tests: core: Replace incorrect return type annotation.**
- **bugfix: tests: core: Assign ThemeSpec instance to controller theme.**
- **refactor: tests: core: Use mocker.patch for model.is_muted_topic.**
- **bugfix: tests: core: Assign a set to model's muted_streams.**
- **refactor: tests: core: Use mocker.patch for `report_*` methods.**
- **tests: conftest: Add an index_search_messages fixture.**
- **refactor: tests: core: Use index_search_messages in search tests.**
- **refactor: tests: core: Add type annotations.**

**Interactions** 
- Waiting on #1076 to extend the changes made to the `run-mypy` tool.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -847,6 +847,12 @@ def index_all_mentions(empty_index, mentioned_messages_combination):
     return index
 
 
+@pytest.fixture()
+def index_search_messages(empty_index):
+    """Expected initial index when search contains the message_id 500."""
+    return dict(empty_index, **{"search": {500}})
+
+
 @pytest.fixture
 def user_profile(logged_on_user):
     return {  # FIXME These should all be self-consistent with others?
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -149,42 +149,43 @@ class TestController:
         self,
         mocker,
         controller,
-        msg_box,
         index_multiple_topic_msg,
         initial_narrow,
         initial_stream_id,
         anchor,
         expected_final_focus,
-    ):
+        stream_name="PTEST",
+        topic_name="Test",
+        stream_id=205,
+    ) -> None:
         expected_narrow = [
-            ["stream", msg_box.stream_name],
-            ["topic", msg_box.topic_name],
+            ["stream", stream_name],
+            ["topic", topic_name],
         ]
         controller.model.narrow = initial_narrow
         controller.model.index = index_multiple_topic_msg
         controller.model.stream_id = initial_stream_id
         controller.view.message_view = mocker.patch("urwid.ListBox")
         controller.model.stream_dict = {
-            205: {
+            stream_id: {
                 "color": "#ffffff",
-                "name": "PTEST",
+                "name": stream_name,
             }
         }
         controller.model.muted_streams = set()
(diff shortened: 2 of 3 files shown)


tests/core/test_core.py
```diff
@@ -444,8 +481,12 @@ def set_msg_ids(*args, **kwargs):
         ],
     )
     def test_maximum_popup_dimensions(
-        self, mocker, controller, screen_size, expected_popup_size
-    ):
+        self,
+        mocker: MockerFixture,
+        controller: Controller,
+        screen_size: Tuple[int, int],
+        expected_popup_size: int,
```
neiljp, line 488: This should be being picked up; I'm investigating.

neiljp (commented):
@prah23 I've rearranged these commits slightly to be clearer:
* fixes to the tests
* combined applying mocker.patch (prep for mypy)
* combined index-messages add+use
* type annotation adding (including original first commit, as it was another type annotation correction that we weren't checking)

There is at least one case that I would expect to be caught, so I'm looking into configuration for that to catch it automatically.

tests/conftest.py
```diff
@@ -850,6 +850,12 @@ def index_all_mentions(empty_index, mentioned_messages_combination):
     return index
 
 
+@pytest.fixture()
+def index_search_messages(empty_index):
+    """Expected initial index when search contains the message_id 500."""
+    return dict(empty_index, **{"search": {500}})
```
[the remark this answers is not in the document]
prah23: Yeah, it seemed to raise an error for all the keys that didn't exist, likely because it's being typed as `Index`.

tests/core/test_core.py
```diff
@@ -419,7 +463,9 @@ def set_msg_ids(*args, **kwargs):
             num_after=0, num_before=30, anchor=10000000000
         )
         create_msg.assert_called_once_with(controller.model, msg_ids)
-        assert controller.model.index == {"search": msg_ids}
+        assert controller.model.index == dict(
+            index_search_messages, **{"search": msg_ids}
+        )
```
prah23, line 468: This was originally a mypy error when I was making an assignment like this to `controller.model.index` where mypy mentioned that the types did not match (`Index` and `Dict[str, object]`); I don't think mypy catches type errors when the only interaction ever being made is an `assert`.

prah23:
I've removed the need for `UserButton` and `MessageBox` completely, but the `StreamButton` has to be imported for the `stream_button` fixture used in `test_stream_muting_confirmation_popup`, where the button itself is being passed into the method being tested.

neiljp:
@prah23 Thanks for this second test typing PR :) I pushed a mypy config improvement first, and am now merging your PR with only a few minor name adjustments in the narrowing test improvements :tada:

A good related follow-up, other than the other typing work, would be to decouple that last case of streambutton you noticed in the code and fix the tests accordingly :+1:

PR #900, opened 2021-01-30
model/views: Add notification checkboxes in Stream Info popup.

zee-bit:
- Added Visual desktop notification checkbox in StreamInfo.

Gradually Fixes #887.

PR #1080, opened 2021-07-13
Clearly indicate nuances in Visual notification settings.

zee-bit:
**What does this PR do?**  
This indicates more explicitly subtle details about the recently introduced Visual notification change via #900. Namely, we indicate that the "Visual notification" setting directly affects other clients such as Web app and desktop. Also, we remind the user
to configure/enable notifications if the checkbox is disabled.

**Tested?** 
- [x] Manually

**Commit flow**
- views: Indicate visual notifications setting affects other clients.
- views: Indicate configuring/enabling notifications if checkbox disabled.

**Notes & Questions** 
I have used "msg_mention" style attribute for the suffix text when the checkbox is disabled. Although this looks good visually on all themes, I'm concerned that the attr-name does not align with our context?


zulipterminal/ui_tools/views.py
```diff
@@ -1299,10 +1299,18 @@ def __init__(self, controller: Any, stream_id: int) -> None:
             footlinks_width,
             len(visual_notification.label) + 4,
         )
+        visual_notification = [visual_notification]
```
preetmishra, line 1302: This is a bit unclear since you're reassigning another value in the var and later assigning another value conditionally. Could we call it something else?

preetmishra: Perhaps append `_setting`?

zulipterminal/config/themes.py
```diff
@@ -32,7 +32,6 @@
     'column_title'    : 'bold',
     'time'            : '',
     'bar'             : 'standout',
-    'popup_contrast'  : 'standout',
```
preetmishra, line 35: commit-description: so that it's*

preetmishra (commented):
@zee-bit Thanks for pushing this promptly. While this is a small addition, I believe it would help improve the user experience significantly. :+1: 

I have tested the PR locally and it seems to work well. I have left some in-line comments.

neiljp:
@zee-bit Thanks for the prompt update following #900, I agree with @preetmishra that this will improve the UX :tada:

PR #1082, opened 2021-07-15
refactor: views: Make mouse scroll-speed in side-panels consistent.

zee-bit:
**What does this PR do?**  
 * Adds MIDDLE_COLUMN_MOUSE_SCROLL_LINES and SIDE_PANELS_MOUSE_SCROLL_LINES
   global variables in views to store the scroll values for both columns.
 * Assigns mouse scroll values of 1 and 5 to Middle column and Side panels
   respectively.
 * Uses these values in mouse_events to skip defined number of lines on
   each scroll, making scroll lines for all side-panels consistent with each
   other.

**Tested?** 
- [x] Manually

**Notes & Questions** 

Can we simplify the names of the global variables?

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -18,6 +18,7 @@ from zulipterminal.config.symbols import (
 from zulipterminal.helper import powerset
 from zulipterminal.ui_tools.boxes import MessageBox
 from zulipterminal.ui_tools.views import (
+    SIDE_PANELS_MOUSE_SCROLL_LINES,
     LeftColumnView,
     MessageView,
     MiddleColumnView,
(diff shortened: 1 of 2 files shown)


tests/ui/test_ui_tools.py
```diff
@@ -552,7 +553,8 @@ def test_mouse_event(self, mocker, stream_view, mouse_scroll_event, widget_size)
         row = 1
         focus = "WIDGET"
         stream_view.mouse_event(size, event, button, col, row, focus)
-        stream_view.keypress.assert_called_once_with(size, key)
+        stream_view.keypress.assert_called_with(size, key)
+        assert stream_view.keypress.call_count == SIDE_PANELS_MOUSE_SCROLL_LINES
```
neiljp, line 557: These only assert the last call is called a certain way, and that overall it's a certain number of times. You likely want `has_calls`.

neiljp (commented):
@zee-bit Thanks for quick work on this following the tests :+1:

This is small and I could merge now, but I'll leave you to investigate the more specific test approach here first.

neiljp:
@zee-bit Looks good now :tada:

PR #1083, opened 2021-07-15
tests: helper: Add type annotations. 

prah23:
**What does this PR do?**  
Adds type annotations to `test_helper.py`. I've also added a prior bugfix commit changing `muted_streams`' data structure from List to Set and a final follow up commit to include `test_helper.py` in the list of files being checked by `run-mypy`.

**Tested?** 
- [x] Manually

**Commit flow** 
- bugfix: tests: helper: Use Sets as params for muted_streams.
This commits changes the previous incorrect configuration of `Lists`
to the `muted_streams` parameter in `test_classify_unread_counts`
which is assigned model's muted_streams, to `Set`s instead, which is
the appropriate data structure.

- refactor: tests: helper: Add type annotations.
This commit adds parameter and return type annotations or hints
to the `test_helper.py` file, that contains tests for its counterpart
`helper.py` from  the `zulipterminal` module, to make mypy checks
consistent and improve code readability.

- tools: Include test_helper.py to be checked by mypy.
This commit adds `test_helper.py` to the `type_consistent_testfiles`

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -213,7 +213,7 @@ def test_powerset(iterable, map_func, expected_powerset):
     "muted_streams, muted_topics, vary_in_unreads",
     [
         (
-            [99],
+            {99},
             [["Some general stream", "Some general unread topic"]],
             {
                 "all_msg": 8,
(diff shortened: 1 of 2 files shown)


neiljp:
@prah23 Looks good after the small Composition tidy I mentioned, so merging now :tada:

PR #1084, opened 2021-07-15
Migrate emoji storage to group aliases together for each canonical name.

zee-bit:
**What does this PR do?**  

Migrates our emoji storage structure in `unicode_emojis.py` to a new structure where
for each `canonical_name` we keep its aliases grouped together, instead of considering
them as separate emojis.

**Tested?** 
- [x] Manually

**Notes & Questions** 

I need to test the reactions PR on top of this to determine the authenticity of this migration.

**Interactions** 

- Blocks #913


setup.cfg
```diff
@@ -42,6 +42,8 @@ ignore=
   E221,
   # E252 - Spaces in parameters, a style deliberately avoided
   E252,
+  # E501 - Ignore max-line-length property - this is already defined by black
+  E501,
```
neiljp, line 46: This commit needs to be earlier for the previous commit to pass linting?

zee-bit: Right, I had a bit of difficulty deciding if I should keep the first 2 commits together (they look related - the second one was the result of first?), or move this commit in between. Now, I realize that passing linting for each commit is more important and that's what my PR description indicates too. :+1:

preetmishra (commented):
@zee-bit Thanks for working on this. This looks good overall. :+1: Though, as there has been a discussion on CZO, generating the new file midway or separately in a commit would bring discrepancies in `model.py`.

tools/convert-unicode-emoji-data
```diff
@@ -25,24 +25,26 @@ SCRIPT_NAME = PurePath(__file__).name
...
         "EMOJI_DATA = OrderedDict(\n    [\n"
     )
-    for emoji_code, emoji_name in ordered_emojis.items():
-        # {'smile': {'code': '263a'}}
-        f.write(f'        ("{emoji_code}", {{"code": "{emoji_name}"}}),\n')
-    f.write("    ]\n)\n")
+    for emoji_name, emoji in ordered_emojis.items():
+        # {'smile': {'code': '263a', 'aliases': []}}
+        f.write(f'        ("{emoji_name}", {emoji}),\n')
+    f.write("    ]\n)\n" "# fmt: on\n")
```
neiljp, line 47: Black combination issue?

zulipterminal/model.py
```diff
@@ -180,6 +180,7 @@ def __init__(self, controller: Any) -> None:
         self.active_emoji_data = self.generate_all_emoji_data(
             self.initial_data["realm_emoji"]
         )
+        self.all_emoji_names = self.generate_all_emoji_names()
```
neiljp, line 183: We update emoji when we get a realm emoji change event, so we at least want to call both in each case, or have one method that returns a tuple and updates both at the same time.

neiljp (commented):
@zee-bit You've got the correct principle here. To ease the migration and avoid the mid-commit regression (which should be breaking tests?), a possible refactoring would be to first adjust the emoji data generation function to return an additional set of data ie. the pre-prepared emoji names, adapting the autocomplete to use it. You can then regenerate the emoji file and the emoji manipulation internal to the model at the same time in the next commit.

tools/convert-unicode-emoji-data
zee-bit: Yes, those were in different lines earlier but black moved them in one line and didn't combine them.

neiljp:
@zee-bit This flows much cleaner with no mid-PR regressions - great! Merging this now :tada:

PR #1090, opened 2021-07-19
bugfix: views: Ensure EditModeView works with new frame-based popup.

neiljp:
**What does this PR do?**  

This bug was introduced by the first commit of #874 which was just cherry-picked, and hidden since the button
used in the EditModeView tests did not have any state to trigger the code in question.

Tests extended.

**Tested?** 
- [X] Manually


preetmishra (approved):
@neiljp Thanks for fixing this up promptly! This does fix the bug and the tests are a great addition too. :+1:

PR #1091, opened 2021-07-19
tests: ui: Add type annotations. 

prah23:
**What does this PR do?**  
Adds type annotations to `test_ui.py` and includes it under files to be checked by mypy in the run-mypy tool. Also added a couple of prior commits to use appropriate patches.

**Tested?** 
- [x] Manually

**Commit flow** 

- refactor: ui: Patch *_box_views in OPEN_DRAFT tests.
This commit assigns patches `stream_box_view` and `private_box_view`
using `mocker.patch.object` to maintain type consistency since their
mocked nature's assert methods are being called.

- refactor: ui: Use patch for non-existent attributes instead of Mock.
This commit uses `mocker.patch` with the `create` parameter set as
True for attributes of the `View` fixture instance that don't exist
without their corresponding generation methods being called. `View`'s
`middle_column` is generated by the `middle_column_view()` method
and `users_view` is generated by the `right_column_view()` method.

- refactor: tests: ui: Add type annotations.
This commit adds parameter and return type annotations or hints to the
`test_ui.py` file, that contains tests for its counterpart `ui.py`
from the `zulipterminal` module, to make mypy checks consistent and
improve code readability.

- tools: Include test_ui.py to be checked by mypy.
This commit adds `test_ui.py` to the `type_consistent_testfiles`

### DIFF
--- a/tests/ui/test_ui.py
+++ b/tests/ui/test_ui.py
@@ -361,17 +361,21 @@ class TestView:
             "foo@zulip.com": {"user_id": 1},
             "bar@gmail.com": {"user_id": 2},
         }
+        mocked_stream_box_view = mocker.patch.object(view.write_box, "stream_box_view")
+        mocked_private_box_view = mocker.patch.object(
+            view.write_box, "private_box_view"
+        )
 
         size = widget_size(view)
         view.keypress(size, key)
 
         if draft:
             if draft["type"] == "stream":
-                view.write_box.stream_box_view.assert_called_once_with(
+                mocked_stream_box_view.assert_called_once_with(
                     caption=draft["to"], title=draft["subject"], stream_id=10
                 )
             else:
-                view.write_box.private_box_view.assert_called_once_with(
+                mocked_private_box_view.assert_called_once_with(
                     emails=draft["to"], recipient_user_ids=[1, 2]
                 )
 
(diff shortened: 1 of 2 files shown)


tests/ui/test_ui.py
```diff
@@ -1,5 +1,10 @@
+from typing import Any, Callable, List, Optional, Tuple
+
 import pytest
+from pytest_mock import MockerFixture
+from urwid import Widget
```
prah23, line 5: I'm not very confident in this being the most relevant type for the `widget` parameter in `widget_size`, does this fit?

neiljp: This seems fine for now - it is the base widget class after all?

tests/ui/test_ui.py
```diff
@@ -266,21 +266,22 @@ def test_keypress_ALL_MENTIONS(self, view, mocker, key, widget_size):
 
     @pytest.mark.parametrize("key", keys_for_command("STREAM_MESSAGE"))
     def test_keypress_STREAM_MESSAGE(self, view, mocker, key, widget_size):
-        view.middle_column = mocker.Mock()
+        mocker.patch(VIEW + ".middle_column", create=True)
+        mocked_middle_column_keypress = mocker.patch(VIEW + ".middle_column.keypress")
```
neiljp, line 319: Could we simplify this to use:
* mocker.patch.object on view with "middle_column"? (create=True still probably necessary?)
* assert on that object .keypress?

Same for the users_view case.

tests/ui/test_ui.py
prah23: Yeah, verified by the class's docstring.

neiljp:
@prah23 Just amended one commit text to take into account the extra mock removal, and merging now - thanks! :tada:

PR #1092, opened 2021-07-21
Blacken/lint rest of tools/

neiljp:
**What does this PR do?**  

This is a follow-up to #1087, which did not fully achieve its aim since python scripts without a `.py` ending were not identified.

Other scripts are now updated to be covered, with an additional script providing a list of tools in python, given certain known exclusions. This script is used as a module in `tools/lint-all`, and separately in CI to explicitly include those files.

A broader approach is likely possible with zulint or using lister.py in future, but having these files covered and tested, and not accidentally losing compliance is the intent here.

**Tested?** 
- [X] Manually
- [X] Passed linting & tests (each commit) **within limited local test ability right now**

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install with linting tools
         run: pip install .[linting]
       - name: Run flake8
-        run: flake8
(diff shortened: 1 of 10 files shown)


neiljp:
Just pushed a minimal update here to be compatible with the run-mypy change related to the typing work from @prah23 and apply some commit text polish.

Will merge very soon.

PR #1101, opened 2021-07-27
refactor: Add and use type annotations for urwid Text markups

Ezio-Sarthak:
**What does this PR do?**  
Adds and uses type annotations for `urwid.Text` markups.

**Tested?** 
- [x] Manually

**Commit flow** 
- refactor: urwid_types: Add type annotation for `urwid.Text` markups.
- refactor: core: Use type annotation for urwid text markups.
- refactor: ui: Use type annotation for urwid text markups.
- refactor: boxes: Use type annotation for urwid text markups.
- refactor: views: Use type annotation for urwid text markups.

**Notes & Questions** 

**Interactions** 

**Visual changes**

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -4,7 +4,7 @@ import unicodedata
 from collections import Counter, OrderedDict, defaultdict
 from datetime import date, datetime, timedelta
 from time import sleep, time
-from typing import Any, Callable, Dict, List, NamedTuple, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, NamedTuple, Optional, Tuple
 from urllib.parse import urljoin, urlparse
 
 import dateutil.parser
@@ -46,7 +46,7 @@ from zulipterminal.helper import (
 from zulipterminal.server_url import near_message_url
 from zulipterminal.ui_tools.buttons import EditModeButton
 from zulipterminal.ui_tools.tables import render_table
-from zulipterminal.urwid_types import urwid_Size
+from zulipterminal.urwid_types import urwid_Size, urwidTextMarkup
 
 
 if typing.TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -1079,11 +1080,13 @@ def footlinks_view(
     def soup2markup(
         cls, soup: Any, metadata: Dict[str, Any], **state: Any
     ) -> Tuple[
-        List[Any], "OrderedDict[str, Tuple[str, int, bool]]", List[Tuple[str, str]]
+        List[urwidTextMarkup],
```
zee-bit, line 1083: Do we need to wrap `urwidTextMarkup` inside a `List`?
`urwidTextMarkup` already contains an outer layer of `Sequence` type which should cover `List`?

zulipterminal/ui_tools/views.py
```diff
@@ -944,7 +944,7 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         return super().keypress(size, key)
 
 
-PopUpViewTableContent = Sequence[Tuple[str, Sequence[Union[str, Tuple[str, str]]]]]
+PopUpViewTableContent = Sequence[Tuple[str, urwidTextMarkup]]
```
zee-bit, line 947: Similarly here, shouldn't just `urwidTextMarkup` suffice? Am I misunderstanding something?

zee-bit (commented):
@Ezio-Sarthak This looks great. :+1:

I had a couple of doubts related to the new types since they got really complex in a few places. This may not need updating depending on the legitimacy of my doubts below, but you'll need to rebase to get around with conflicts in any case. :)

zulipterminal/ui_tools/boxes.py
Ezio-Sarthak: I'm skeptical about this too. I think `Sequence` might only be applicable as a type if the contents in the sequence are not changing? (I think I discussed this on stream too?)

PR #1104, opened 2021-07-28
bugfix: views: ZT crashes on muting a stream.

Rohitth007:
Tests updated.

**What does this PR do?**  
ZT crashes because LEFT_WIDTH is no longer stored as a View constant.
This happened because of the recent merge of Autohide Tabs.

**Tested?** 
- [x] Manually

**Notes & Questions** 

LEFT_WIDTH is redefined and used in views.py to fix the issue as
importing from ui.py would cause a cyclic import error.


neiljp:
@Rohitth007 Thanks for finding this and proposing a fix! I've done the refactor to extract the size parameters and merged this manuall as ae1081101 with the refactor :tada:

PR #1105, opened 2021-07-28
refactor: tests: buttons: Add type annotations.

prah23:
**What does this PR do?**  
Adds type annotations to test_buttons.py and adds the file to the list in run-mypy. Also has a prior refactor commit to assign `mocker.patch`s to variables.

**Tested?** 
- [x] Manually

**Commit flow** 

- refactor: tests: buttons: Assign method mocker.patchs to variables.
This commit assigns fixture method attribute `mocker.patch`s to
variables, so that the corresponding asserts can be called on the
mocked nature, as was the original intention.

- refactor: tests: buttons: Add type annotations.
This commit adds parameter and return type annotations or hints to the
`test_buttons.py` file, that contains tests for it's counterpart
`buttons.py` from  the `zulipterminal` module, to make mypy checks
consistent and improve code readability.

- tools: Include test_buttons.py to be checked by mypy.
This commit adds `test_buttons.py` to the `type_consistent_testfiles`

**Waiting on**:
Refactor work to improve type annotations for urwid markup.

### DIFF
--- a/tests/ui_tools/test_buttons.py
+++ b/tests/ui_tools/test_buttons.py
@@ -1,8 +1,9 @@
-from typing import Any, Dict
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import pytest
 from pytest import param as case
-from urwid import AttrMap, Overlay
+from pytest_mock import MockerFixture
+from urwid import AttrMap, Overlay, Widget
 
 from zulipterminal.config.keys import keys_for_command
 from zulipterminal.ui_tools.buttons import (
@@ -10,10 +11,12 @@ from zulipterminal.ui_tools.buttons import (
     MessageLinkButton,
     ParsedNarrowLink,
     StarredButton,
+    StreamButton,
     TopButton,
     TopicButton,
     UserButton,
 )
+from zulipterminal.urwid_types import urwid_Size
 
 
 MODULE = "zulipterminal.ui_tools.buttons"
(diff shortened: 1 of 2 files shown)


tests/ui_tools/test_buttons.py
```diff
@@ -314,8 +359,12 @@ def test_init(self, mocker):
         ],
     )
     def test_update_widget(
-        self, mocker, caption, expected_cursor_position, display_attr=None
-    ):
+        self,
+        mocker: MockerFixture,
+        caption: str,
+        expected_cursor_position: int,
+        display_attr: Optional[str] = None,
```
neiljp, line 366: If this is just `None` then the type is None, though it may warrant a comment to explain why we're only testing that value.

neiljp (commented):
@prah23 This looks broadly good with only a few issues I could see from first review :+1:

tests/ui_tools/test_buttons.py
prah23: This was typed the same way even in the method in the main code, wouldn't it be more consistent to type it the same way? This is something I've followed even in the annotations for `test_boxes.py` with the `state` parameter we pass into the autocomplete. I'm not sure what the best approach is here, but wouldn't it be easier to read as well if it's typed the same way in the tests, irrespective of whether only a subtype is being fulfilled by the cases?

neiljp: The type does give a suggestion that it should be expanded, so I can see this as an alternative :+1:

prah23:
Thanks for taking a look, @neiljp! I've accommodated the changes suggested, only the second point on the review is left, which is pending on other urwid-markup typing refactor work (I've also updated the PR description accordingly).

prah23:
Thanks for taking a look, @neiljp! I've accommodated the changes suggested, only the second point on the review is left, which is pending on other urwid-markup typing refactor work (I've also updated the PR description accordingly).

neiljp:
@prah23 Thanks for this and the updates :+1: Merging with slight commit message tweaks shortly! :tada:

Issue #333, opened 2019-03-30
Right line mismatch with messages containing certain emoji's. 


sumanthvrao:
When messages with certain emoji's appear, the right border line of Middle Column, seems to be mismatched. One example of such an emoji is . (have not done an extensive search of which emoji's cause this issue)
 For example emoji's like this --> 👍🏽 
![right_line](https://user-images.githubusercontent.com/30312566/55270947-93061980-52cb-11e9-991c-b049235bafef.png)

kaustubh-nair:
@sumanthvrao This wouldn't happen from the web app since all emojis there are automatically converted to text. I think this is happening from the mobile version. When I try to send unsupported emojis, from my keypad's emoji picker, ZT just recieves it as unicode which it can't decode.
I guess we can have a generic indicator saying that it is an emoji? Like this? :emoji:

kaustubh-nair:
Or we can try to extract the name of the emoji from unicode using a lookup module. Something like this https://docs.python.org/3/library/unicodedata.html#unicodedata.name

neiljp:
This may be fixed upstream in urwid/urwid#225, but could be resolved via aspects like #687, applied to all names where emoji can take place.

zulipbot:
Hello @zulip/server-emoji members, this issue was labeled with the "area: emoji" label, so you may want to check it out!

PR #1114, opened 2021-08-08
Make certain Emojis not break urwid rendering

Rohitth007:
**What does this PR do?**  
Real Emojis!! Please work!!

**Tested?** 
- [x] Manually

**Commit flow** 

**Notes & Questions** 

**Interactions** 

**Visual changes**


neiljp:
@Rohitth007 Thanks for digging into this - there's an outstanding issue with some progress in urwid (eg. urwid/urwid#225) but as long as we're avoiding emoji by using text then this approach seems reasonable (at least for the emoji selectors).

There's one point where stream replies are broken by this, but my concern more generally is that we likely want to preserve the data and render differently - if that's possible - as otherwise we may end up replying to eg. streams/topics with a different string due to the removal of the extra characters.

neiljp:
This would likely fix #333.

neiljp:
There's also at least one case in #**test here** topics that still is corrupting the display.

I'd recommend we open this up on the broader streams somewhere to help with interoperability?

PR #1115, opened 2021-08-08
Define and use Stream data fixture from conftest

zee-bit:
**What does this PR do?**  
This PR defines the `general_stream` and `secret_stream` dictionaries in conftest to pytest fixtures, allowing us to re-use them in tests that need stream data. The defined fixtures are currently used in tests for StreamInfo popup.

**Tested?** 
- [x] Manually

**Commit flow** 

- refactor: conftest: Define general_stream and secret_stream as fixtures.
- conftest: Add rendered_description field to stream fixtures.
- tests: popups: Use stream fixture instead of defining custom streams.

**Interactions** 

- Blocks #880

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -192,45 +192,50 @@ def logged_on_user():
     }
 
 
-general_stream = {
-    "name": "Some general stream",
-    "invite_only": False,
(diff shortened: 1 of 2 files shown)


neiljp:
@zee-bit Thanks for the reworked & extended prep from #880 - merging now :tada:

PR #1116, opened 2021-08-10
Use urwid's CommandMap to manage navigational keys.

Rohitth007:
**What does this PR do?**  
This PR modifies urwid's `CommandMap` class object `command_map`
to map alternate navigational keys to standard urwid navigational
aliases. This is used in `ui.py` and `PopupView`.

Users can also now add custom nav-keys by adding it to KEY_BINDINGS.

**Tested?** 
- [x] Manually

### DIFF
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -132,10 +132,7 @@ class TestPopUpView:
         self.pop_up_view.keypress(size, "cmd_key")
         assert self.controller.exit_popup.called
 
-    def test_keypress_navigation(
-        self, mocker, widget_size, navigation_key_expected_key_pair
-    ):
-        key, expected_key = navigation_key_expected_key_pair
+    def test_keypress_navigation(self, mocker, navigation_key, widget_size):
         size = widget_size(self.pop_up_view)
         # Patch `is_command_key` to not raise an 'Invalid Command' exception
         # when its parameters are (self.command, key) as there is no
@@ -148,8 +145,10 @@ class TestPopUpView:
                 else is_command_key(command, key)
             ),
         )
-        self.pop_up_view.keypress(size, key)
-        self.super_keypress.assert_called_once_with(size, expected_key)
+
+        self.pop_up_view.keypress(size, navigation_key)
+
+        self.super_keypress.assert_called_once_with(size, navigation_key)
 
 
 class TestAboutView:
@@ -188,15 +187,6 @@ class TestAboutView:
         size = widget_size(self.about_view)
         self.about_view.keypress(size, key)
         assert not self.controller.exit_popup.called
-
-    def test_keypress_navigation(
-        self, mocker, widget_size, navigation_key_expected_key_pair
-    ):
-        key, expected_key = navigation_key_expected_key_pair
-        size = widget_size(self.about_view)
-        super_keypress = mocker.patch(MODULE + ".urwid.Frame.keypress")
-        self.about_view.keypress(size, key)
-        super_keypress.assert_called_once_with(size, expected_key)
 
     def test_feature_level_content(self, mocker, zulip_version):
         self.controller = mocker.Mock()
@@ -338,15 +328,6 @@ class TestUserInfoView:
         self.user_info_view.keypress(size, key)
         assert not self.controller.exit_popup.called
 
-    def test_keypress_navigation(
-        self, mocker, widget_size, navigation_key_expected_key_pair
-    ):
-        key, expected_key = navigation_key_expected_key_pair
-        size = widget_size(self.user_info_view)
-        super_keypress = mocker.patch(MODULE + ".urwid.Frame.keypress")
-        self.user_info_view.keypress(size, key)
-        super_keypress.assert_called_once_with(size, expected_key)
-
 
 class TestFullRenderedMsgView:
     @pytest.fixture(autouse=True)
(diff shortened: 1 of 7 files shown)


zulipterminal/config/keys.py
```diff
@@ -373,6 +374,16 @@ class KeyBinding(TypedDict, total=False):
     ]
 )
 
+STD_NAV_CMDS = [
```
preetmishra, line 377: While it is shorter to type, STD is not a well-known shorthand. How about `BASIC_NAV_COMMANDS`/`PRIMARY_NAV_COMMANDS`?

zulipterminal/config/keys.py
```diff
@@ -415,3 +426,19 @@ def commands_for_random_tips() -> List[KeyBinding]:
         for key_binding in KEY_BINDINGS.values()
         if not key_binding.get("excluded_from_random_tips", False)
     ]
+
+
+# Maps alternate nav keys to standard nav keys.
+# Refer urwid/command_map.py
+for cmd in STD_NAV_CMDS:
+    std_key, *alt_keys = keys_for_command(cmd)
```
preetmishra, line 434: I know that we have the following comment in the file. However, it is not "safe" to assume that the first key would be the `primary/std` key; perhaps we should use `primary_key_for_command` to fetch the `primary/std` key.
```python3
    # Key that is displayed in the UI is determined by the method
    # primary_key_for_command. (Currently the first key in the list)
```

zulipterminal/config/keys.py
```diff
@@ -415,3 +426,19 @@ def commands_for_random_tips() -> List[KeyBinding]:
...
+
+    for key in alt_keys:
+        if std_key == "end":
+            urwid_cmd = "cursor max right"
```
preetmishra, line 438: We could import these constants from urwid instead.
```python3
...
CURSOR_MAX_RIGHT = 'cursor max right'
...
```

preetmishra (commented):
@Rohitth007 Thanks for taking this up! This seems to work well locally and reduces code duplication. :+1:

I've left some comments.

zulipterminal/config/keys.py
Rohitth007: `primary_key_for_command` also picks the first key, it's basically the same thing ryt?

zulipterminal/config/keys.py
Rohitth007: That would only work for 2 cases ryt? It won't be consistent with the `else` conditional?

zulipterminal/config/keys.py
Rohitth007: Maybe `BASIC` is a better option that `PRIMARY` as it might be confused with primary keys. I'll change this.

zulipterminal/config/keys.py
preetmishra: We could have a dict that maps them accordingly. It would be more future proof since we would be importing from urwid directly.

zulipterminal/config/keys.py
preetmishra: Well, there could be discrepancies later down the road. The code is currently relying on the _fact_ that the first key is the primary key for now. However, that might change later warranting a manual change in this code block.

preetmishra (commented):
@Rohitth007 Thanks for the update! :+1: The `dict` approach seems better. The PR looks good to me. Left a minor comment.

zulipterminal/config/keys.py
neiljp: I agree with Preet; if we later changed the feature a lot, it would also be a lot simpler to look for uses of `primary_key_for_command` than cases like this - and if we decided to make the last key important then we just have to change one function, not the function and this extra position.

zulipterminal/config/keys.py
```diff
@@ -415,3 +435,9 @@ def commands_for_random_tips() -> List[KeyBinding]:
         for key_binding in KEY_BINDINGS.values()
         if not key_binding.get("excluded_from_random_tips", False)
     ]
+
+
+# Refer urwid/command_map.py
+for zt_cmd, urwid_cmd in ZT_TO_URWID_CMD_MAPPING.items():
```
neiljp, line 441: This specifically adds to the existing mapping, so it'd be useful to note this in a comment.

tests/ui_tools/test_popups.py
```diff
@@ -131,10 +135,10 @@ def test_keypress_command_key(self, mocker, widget_size):
         self.pop_up_view.keypress(size, "cmd_key")
         assert self.controller.exit_popup.called
 
-    def test_keypress_navigation(
-        self, mocker, widget_size, navigation_key_expected_key_pair
-    ):
-        key, expected_key = navigation_key_expected_key_pair
+    @pytest.mark.parametrize(
+        "key", [key for cmd in ZT_TO_URWID_CMD_MAPPING for key in keys_for_command(cmd)]
```
neiljp, line 139: Potential fixture as `navigation_key` or similar?

neiljp (commented):
@Rohitth007 This is a great cleanup :+1: Was this mixed into another PR previously? (it seems simpler on its own, if so)

This seems to work well. Other than the inline comments, my other concerns would be:
* we currently have distinct general navigation (ui) and up/down navigation (popups), and are now simply generalizing this - though if a widget (eg popup) doesn't support 'left' then I suppose it just doesn't trigger, so that should be fine?
* removing the key-pair (key-binding -> "actual" key) fixtures and tests simplifies things well, but we're not testing the new code - I'm not meaning key dispatch, but just that given an input command map and the mapping and bindings, what comes out?

zulipterminal/config/keys.py
```diff
@@ -415,3 +435,9 @@ def commands_for_random_tips() -> List[KeyBinding]:
...
+    for key in keys_for_command(zt_cmd):
+        command_map.__setitem__(key, urwid_cmd)
```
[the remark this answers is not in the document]
Rohitth007: Oh, did not know that it can be called like a dict directly. I've only seen it for lists.

zulipterminal/config/keys.py
Rohitth007: @neiljp Do you mean in `ui_mappings.py`?

Rohitth007:
@neiljp Re1st point, I think for every widget we have to let urwid take care of navigation and let the widget decide. In cases we want custom behavior we can intercept the keys but let the focus change be taken care of by urwid.
Re 2nd point I've added a test in `test_keys` testing the updated `command_map`

neiljp:
@Rohitth007 Thanks for the refactor - I'm merging this now :tada:

PR #1122, opened 2021-08-17
Improve print-debugging (no `flush=True` required; require `-d`/`--debug`)

### LINKED ISSUE #1066: add option to disable debug.log
Currently, `debug.log` is created wherever the current directory is, and if there are no errors, it is a 0B empty file, even if zulip-term is launched without the `-d` flag.

I would like the creation of `debug.log` to be tied to that `-d` flag, as it gets messy and annoying after a while, deleting that file again and again, i.e. if the debug flag is passed then use debug.log, else don't create it.

If not that, then at least an option to disable creating that file.

neiljp:
**What does this PR do?**  

This mainly improves print debugging:
* just use `print("something")` instead of `print("something", flush=True)` (README updated)
* only open the `debug.log` file that `print` is redirected to, if `-d` or `--debug` is enabled (README updated)

There is also an intermediate refactor commit to require Controller to have named arguments.

Fixes #1066 (via the second part)

**Tested?** 
- [x] Manually

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,13 +321,13 @@ For additional help with pytest options see `pytest -h`, or check out the [full 
 
 #### Output using `print`
 
-The stdout for zulip-terminal is redirected to `./debug.log` by default.
+The stdout (standard output) for zulip-terminal is redirected to `./debug.log` if debugging is enabled at run-time using `-d` or `--debug`.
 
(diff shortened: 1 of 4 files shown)


neiljp:
@zee-bit Thanks for the feedback!
@preetmishra Thanks for the suggestion to output the debug.log info when starting 👍 

I've adjusted for that and merging now - let's get to using this! :)

PR #1124, opened 2021-08-20
Indent to distinguish resolved topics.

Rohitth007:
**What does this PR do?**  
This PR indents all TopicButtons to make space for the resolved
topic symbol and easily distinguish between them.

To make topic name length the same as before, the default suffix is
made from 2 spaces to 1 space and the LEFT_WIDTH is increased by 1.
This is equivalent to no difference in topic name length and 2 char
gain in stream/special narrow buttons.

**Tested?** 
- [x] Manually

**Visual changes** 
image


zulipterminal/ui_tools/buttons.py
```diff
@@ -297,11 +297,18 @@ def __init__(
             stream_name=self.stream_name,
             topic_name=self.topic_name,
         )
+
+        topic_prefix = " "
+        topic_name = self.topic_name
+        if self.topic_name.startswith(CHECK_MARK + " "):
```
neiljp, line 303: This is defined more by the API than our internal symbols, so I'd suggest we put the constant there (eg. `RESOLVED_PREFIX`)

zulipterminal/config/symbols.py
```diff
@@ -6,7 +6,7 @@
 MESSAGE_CONTENT_MARKER = "▒"  # Options are '█', '▓', '▒', '░'
 QUOTED_TEXT_MARKER = "░"
 MESSAGE_HEADER_DIVIDER = "━"
-CHECK_MARK = "✓"
+CHECK_MARK = "✔"
```
neiljp, line 9: This seems like a separate change, particularly given it's part of the API. This might simplify writing the tests, depending if we only use these constants.

zulipterminal/ui_tools/buttons.py
```diff
@@ -81,7 +81,7 @@ def update_widget(
         if count_text[1]:
             suffix = [" ", count_text, " "]
         else:
-            suffix = ["  "]
+            suffix = [" "]
```
neiljp, line 84: The combination of this and the increase of only 1 in the left panel width, means that only fully-read topics return to the original length, so I'd suggest just using 29 for the width.

The other side effect of this change, which we may have discussed in another PR is that longer topics often show an ellipsis in the digits column, which is somewhat unclear when aligned with unread counts.

zulipterminal/ui_tools/buttons.py
```diff
@@ -297,11 +297,18 @@ def __init__(
             stream_name=self.stream_name,
             topic_name=self.topic_name,
         )
+
+        topic_prefix = " "
```
neiljp, line 301: Minor: I'd consider a comment here, simply since this is one space but ends up adding two due to there now being a prefix character.

neiljp (commented):
@Rohitth007 This looks good, though it would be simpler with a separate API-specific addition for the resolved prefix.

My only other concern is that a long list of unresolved topics looks significantly indented - maybe more so with the text flowing closer to the right hand side.

zulipterminal/ui_tools/buttons.py
Rohitth007: When the count is double digit then this is going to happen anyway right? Atleast a letter doesn't come below the count digit.

tests/ui_tools/test_buttons.py
```diff
@@ -336,11 +336,11 @@ def test_keypress_emoji_button(
...
 class TestTopicButton:
     @pytest.mark.parametrize(
-        "count, stream_id, title, stream_name",
+        "count, stream_id, title, stream_name, is_resolved",
         [
-            (2, 86, "topic1", "Django"),
-            (1, 14, "topic2", "GSoC"),
-            (1000, 205, "topic3", "PTEST"),
+            (2, 86, "topic1", "Django", False),
+            (1, 14, "✔ topic2", "GSoC", True),
```
preetmishra, line 342: Since we have a constant defined for the Unicode, let's use it here as well.

PR #1127, opened 2021-08-22
Stream info view

kingjuno:
**What does this PR do?**  
Include more elements to StreamInfo View  like :
- Stream id 
- Type of the stream (Public/Private)
- Posting policy 
- Availability of history
- User's role within the stream

Also fixed convert-unicode-emoji-data from displaying wrong file name when `zulipterminal.unicode_emoji_dict` isn't downloaded.

**Tested?** 
- [X] Manually


zee-bit:
@kingjuno The second commit seems unrelated to the title of the PR. Could you please extract that commit into a new PR?

zulipterminal/ui_tools/views.py
```diff
@@ -1330,6 +1330,19 @@ def __init__(self, controller: Any, stream_id: int) -> None:
         ]
 
         total_members = len(stream["subscribers"])
+        Stream_id = stream["stream_id"]
```
zee-bit, line 1333: We already have the stream_id being passed as a parameter to StreamInfo view, so there's no need to re-define it again here.

zulipterminal/ui_tools/views.py
```diff
@@ -1330,6 +1330,19 @@ def __init__(self, controller: Any, stream_id: int) -> None:
...
+        type_of_stream = "Private" if stream["invite_only"] is True else "Public"
+        stream_posting_policy = {
+            1: "Any user can post",
+            2: "Only administrators can post",
+            3: "Only full members can post",
+            4: "Only moderators can post",
+        }
+        posting_policy = stream_posting_policy[stream["stream_post_policy"]]
```
zee-bit, line 1341: As per API docs, the `stream_post_policy` field was added in Server version 3.0; there's no feature level mentioned there, so you can either query that in the #api-design stream in chat.zulip.org or you could just check if the field is absent so use the older `is_announcement_only` field.

tests/ui_tools/test_popups.py
```diff
@@ -1077,42 +1077,42 @@ def test_keypress_stream_members(self, key, widget_size):
                 {"date_created": None},
                 "74 [Organization default]",
                 None,
-                12,
+                18,
```
zee-bit, line 1080: The increase in popup height does not seem to match the number of fields you added below? i.e. 12 + 5 = 17.

zulipterminal/ui_tools/views.py
```diff
@@ -1330,6 +1330,19 @@ def __init__(self, controller: Any, stream_id: int) -> None:
...
+        availability_of_history = (
+            "Yes" if stream["history_public_to_subscribers"] else "No"
+        )
+        user_role = "Stream administrator" if stream["role"] == 20 else "Subscriber"
```
zee-bit, line 1345: I think we decided to leave the subscriber role for now?

zulipterminal/ui_tools/views.py
```diff
@@ -1365,6 +1378,11 @@ def __init__(self, controller: Any, stream_id: int) -> None:
                         "Stream email",
                         f"Press {email_keys} to copy Stream email address",
                     ),
+                    ("Stream ID", f"{Stream_id}"),
+                    ("Type of Stream", f"{type_of_stream}"),
+                    ("Posting Policy", f"{posting_policy}"),
+                    ("History Available to Subscribers", f"{availability_of_history}"),
+                    ("User Role", f"{user_role}"),
```
zee-bit, line 1385: We may want to re-order some of the fields here. for e.g. Stream Id and type of stream seems more significant than others so we could have those at the top?

zee-bit (changes requested):
@kingjuno Thanks for working on this. This is a nice addition to our StreamInfo view and it works fine. :+1:

I've mostly focused on the code for this review and below are a few points for the first iteration. When working on features that leverage a new field from API responses, it's always important to check if that field was added sometimes later in the server, in which case, you'd need to gracefully handle it for older server versions. Other than that, others are minor points, and considering this is your first PR in this repo, Great work! :tada:

tests/ui_tools/test_popups.py
kingjuno: It was because the key name of last row was long . So I reduced the size.

zulipterminal/ui_tools/views.py
kingjuno: Removed that.

zulipterminal/ui_tools/views.py
kingjuno: Removed.

zulipterminal/ui_tools/views.py
kingjuno: Rearranged stream id and type to the top.

neiljp (commented):
@kingjuno This looks good - thanks for giving this a go 👍 

As @zee-bit mentions, some of these fields are dependent on the server version, so may fail badly depending on which server one uses (we claim to support 2.1+ if you check the README). Due to that you may want to limit this first PR to those which are fully backwards compatible to that point, to avoid needing to get into version/feature detection initially.

For tests, as I note, you've made the existing tests (and parts of tests - "case"s) pass, but there are not yet new ones which cover the range of new values which are being output. It's true that the existing test only examines the height, which doesn't give much detail, but perhaps we can improve upon that.

tests/ui_tools/test_popups.py
```diff
@@ -1077,42 +1077,42 @@ def test_keypress_stream_members(self, key, widget_size):
...
                 id="ZFL=None_no_date_created__no_retention_days",
             ),
             case(
                 {"date_created": None},
                 "74 [Organization default]",
                 16,
-                12,
+                18,
```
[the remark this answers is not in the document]
kingjuno: Oh sorry about that. :)

PR #1130, opened 2021-08-23
Update index's `topic_msg_ids` accurately with topic updates.

prah23:
**What does this PR do?**  
Prior to this PR, if a message's topic was changed, the
corresponding `message_id` was not removed from the old topic's set
and added to the new one under the "topic_msg_ids" key of the model's
index. This PR updates the relevant sets to ensure accuracy.

**Tested?** 
- [x] Manually

**Commit flow** 

- api_types: Include the orig_subject field in UpdateMessageEvent. 
This commit introduces another field to the UpdateMessageEvent class
in `api_types.py` to accommodate this field in the event handler
method for update message events.

- bugfix: model: Remove updated message_ids from their old topic's set.
Prior to this commit, if a message's topic was changed, the
corresponding `message_id` was not removed from the old topic's set
under the "topic_msg_ids" key of the model's index. This commit removes
the updated `message_id` from that set, ensuring the accuracy of the
set.
Tests updated.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1624,6 +1624,7 @@ class TestModel:
             case(
                 {  # Only subject of 1 message is updated.
                     "message_id": 1,
+                    "orig_subject": "old subject",
                     "subject": "new subject",
                     "stream_id": 10,
                     "message_ids": [1],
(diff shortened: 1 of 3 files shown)


neiljp:
@prah23 This change looks reasonable, though I'll read the tests again before merging, but as per `#zulip-terminal>Topic move bug? #T1130` does this fix a particular situation that can also be manually detected/tested?

neiljp:
@prah23 Thanks for the update! 🎉 I tacked a refactor on the end and added a comment to a test, and used a more detailed commit from another PR for the first commit since we can use it later and it was otherwise very tiny :)

Issue #1133, opened 2021-09-02
Support more Themes


neiljp:
Since themes are now isolated, we hope that adding a new theme should be easier for a new contributor!

Please review our existing theme documentation and theme source files. Currently all themes are integrated in the source.

Themes we could consider adding/developing: 
* [ ] dracula
* [ ] nord
* [ ] gruvbox light
* [ ] solarized light
* [ ] solarized dark

(please discuss in the stream first, if you think of others and want to work on them)

NOTE: To support multiple new contributors working on this, it would be useful if you only work on one of these yourself and leave other themes for others to work on.

Things to consider:
* please include a reference to the theme origin, ie. where you found the colors to use
* we aim to work in different color depths, so try testing those as well as the default 256-color depth
* some styles have common named colors; it would be good to reuse these
* supporting syntax highlighting would be useful, but could be worked on later

felipecpassos:
Hello, i'm new to contributing to OS projects, and would like to develop the Nord Theme for Zulip Terminal.

I would like to be inspired in the VS Code Nord Theme: 

[Example](https://raw.githubusercontent.com/arcticicestudio/nord-docs/develop/assets/images/ports/visual-studio-code/ui-overview-jsx.png)

neiljp:
Hi @felipecpassos! Welcome to the project!

For the colors themselves, you may wish to refer to https://www.nordtheme.com/
For application of the colors, that site may also have insights as to which kinds of UI elements might fit best with different colors?

Other than the colors and assigning them to the UI styles in a new theme file, you'll need to add the connections for the new theme to enable it to be selected at run-time, though that should be a case of looking for the names of existing themes and adding in your new theme file name.

mfonism:
Hi @felipecpassos,

I'd like to work on adding the solarized light theme.

I'm currently looking through https://ethanschoonover.com/solarized/ to have a good feel for the theme.

CypherpunkSamurai:
Hello sir, I'm new to open-source contributions and I would like to create new theme `OneDark` for zulip. Can you please guide me through this to get started.

Thank You Sir

Dishti-Oberai:
Hi all,
Can I please work on gruvbox light theme?

neiljp:
@Dishti-Oberai Hopefully you saw that @amanagr and myself responded in #zulip-terminal.

[...]

PR #1139, opened 2021-10-18
Nord theme

felipecpassos:
**What does this PR do?**  

Added support for Nord theme.

Partial fix for #1133

**Tested?** 
- [X] Manually

**Commit flow** 

**Notes & Questions** 

- https://www.nordtheme.com/

**Interactions** 

**Visual changes** 

image

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -3,7 +3,7 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 from pygments.token import STANDARD_TYPES
 
 from zulipterminal.config.color import term16
-from zulipterminal.themes import gruvbox, zt_blue, zt_dark, zt_light
+from zulipterminal.themes import gruvbox, nord, zt_blue, zt_dark, zt_light
 
 
 StyleSpec = Union[
@@ -85,6 +85,7 @@ REQUIRED_META = {
 # fmt: on
 
 THEMES = {
+    "nord": nord,
     "gruvbox_dark": gruvbox,
     "zt_dark": zt_dark,
     "zt_light": zt_light,
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -85,8 +85,8 @@ REQUIRED_META = {
 # fmt: on
 
 THEMES = {
-    "nord": nord,
     "gruvbox_dark": gruvbox,
+    "nord": nord,
     "zt_dark": zt_dark,
     "zt_light": zt_light,
     "zt_blue": zt_blue,
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -86,7 +86,7 @@ REQUIRED_META = {
 
 THEMES = {
     "gruvbox_dark": gruvbox,
-    "nord": nord,
+    "nord_dark": nord,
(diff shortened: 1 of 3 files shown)


neiljp (commented):
@felipecpassos Thanks for looking into this - I've been away for a bit so just coming back to it now! :+1:

You don't state this, but this is effectively a "dark ambiance" theme, to use the terminology on the Nord website?
If so, while the main Nord website doesn't seem to have much evidence of "white ambiance" designs in screenshots, it might be useful to use the `_dark` suffix for our theme like we have `zt_dark` and `gruvbox_dark` (which might eventually have the matching `gruvbox_light`, for example).

It's good to see the tests appear to run OK and manually the style looks to be rendering visually OK after a quick check, though I've not examined whether the nord colors match the recommended usage in detail.

For the color specification, for easier comparison with the source material, and also referencing within the styles, I feel it would be cleaner to order the NordColor entries in the same order as in their documentation, as well as use the descriptive prefixes eg. maybe `FROST_7` and `AURORA_ORANGE_12`.

There are a few extra colors which aren't in the Nord palette - It may be challenging but I'd hope that we'd be able to avoid using them?

felipecpassos:
> For the color specification, for easier comparison with the source material, and also referencing within the styles, I feel it would be cleaner to order the NordColor entries in the same order as in their documentation, as well as use the descriptive prefixes eg. maybe `FROST_7` and `AURORA_ORANGE_12`.

About the color names, i gave them the same name from Nord pallete:

image

I called that color NORD_7. Would you prefer me to call it FROST_7 or FROST_1 ?

zulipterminal/config/themes.py
```diff
@@ -86,7 +86,7 @@
 
 THEMES = {
     "gruvbox_dark": gruvbox,
-    "nord": nord,
+    "nord_dark": nord,
```
neiljp, line 89: This is fine until we have a light version :)

zulipterminal/config/themes.py
```diff
@@ -95,6 +95,7 @@
 THEME_ALIASES = {
     "default": "zt_dark",
     "gruvbox": "gruvbox_dark",
+    "nord": "nord_dark",
```
neiljp, line 98: This section was intended for historical aliases; I'd prefer we avoid supporting "nord" as a theme name for now, which is what this line would do.

neiljp (commented):
@felipecpassos I added a few notes based on the changes you made already.

Re the numeric suffix, I used the `<n>` from `nord<n>` as a suffix since that's how the color is referenced in that document. I can see it being beneficial both ways. In my AURORA example I used ORANGE, so for the others we could include a primary index instead of a subtle color-name, so my examples might become `FROST_1_n7` and `AURORA_ORANGE_n12`. Those last digits are the nord index, which we could also potentially list in an aligned comment at the end of each line, as they won't line up in the names.

For merging note that we'll likely want to squash the commits together manually before merging due to the rebase-oriented workflow we use.

neiljp:
@felipecpassos Do you need more support with this?

neiljp:
@felipecpassos This will need a little more work after a few changes to styles, and I'd recommend for you or anyone else taking this up where there may be multiple versions (eg dark/light) to split the common color palette into a separate file, as we've now done for gruvbox.

PR #1140, opened 2021-10-30
Translation of developer feature tutorial to pt_BR

felipecpassos:
**What does this PR do?**

Added translation for developer feature tutorial documentation in pt_BR (Brazil).

**Tested?** 
- [X] Manually

**Commit flow** 

**Notes & Questions**

### DIFF
--- a/docs/developer-feature-tutorial-ptBR.md
+++ b/docs/developer-feature-tutorial-ptBR.md
@@ -1,0 +1,173 @@
+## Tutorial - Adicionando o indicador de 'digitando'
+
+Este tutorial mostra como o indicador de 'digitando' foi implementado no cliente. O processo de adicionar uma nova feature ao zulip terminal varia bastante dependendo da feature. Esse tutorial tem o objetivo de te tornar familiar ao processo geral.
+
+Como os dados do indicador de 'digitando' para o outro usuário em uma mensagem privada não pode ser gerado localmente, ele deve ser recebido do cliente.
+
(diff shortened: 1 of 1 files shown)


neiljp:
@felipecpassos Thanks for this - did you translate this manually?

We don't have documentation translations in the terminal documentation, so I'm not sure of the best practice for handling translations like this. If it would have been useful to you then that provides a good starting point of course :) My main concern is keeping things synchronized if we update documents, and how far things go before we definitely need to update (or even remove) the translations due to inaccuracy.

felipecpassos:
Yes i did it manually :sweat_smile: 

The synchronization issue is really a problem, and i can't think of a way to solve it.
I had my doubts about the usefulness of this PR, because there was no other translated file.

I made this for a assignment in a class in Software Engineering, and you can discard if it has no use (not a lot of Brazilians in zulip terminal community i believe). 

Sorry for the time taken to review this PR.

neiljp:
@felipecpassos No problem, thanks for working on it! Maybe the wider Zulip community has more information on translations of this kind of document.

zormit:
Considering the discussion above and the age, I think this PR can be closed.

neiljp:
@zormit Agreed, this seems sensible at this point :+1:

@felipecpassos Thanks for your work on this in any case! I'm unsure if you're using or otherwise involved in Zulip, but you might consider the translations in the apps with translations if you've not already - or for ourselves when we add that! Thanks again!

PR #1147, opened 2022-01-06
Display of starred and unstarred messages enhanced

### LINKED ISSUE #171: Fix issues with message merging (hiding content headers) based on star status

Dishti-Oberai:
**What does this PR do?**  
This PR enhances the display of consecutive starred and/or unstarred messages which have identical dates, times, and authors.
Earlier all such messages which had similar star status (whether starred or unstarred) appeared together and messages which had different star status were separated by a line. However, this led to difficulty in identifying the star status of messages (which were starred and thus appeared together). The following screenshot depicts the same-
image

Now, all starred messages and messages which have different star statuses have been separated by a blank line whereas those which are all unstarred appear together. This leads to a better visual display and interpretation of the star status of messages.
The following screenshot demonstrates the present scenario-
image

(The first two messages are starred and the third one is unstarred in both the above screenshots)

Fixes #171 

A detailed explanation of the issue and discussion related to it can be read at - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Message.20merging.20-.20star.20status.20.28issue.20.23T171.29

**Tested?** 
- [x] Manually

**Notes & Questions** 
Should any other changes be made?

**Interactions** 
The changes have been made (developed on) in addition to the following PR - https://github.com/zulip/zulip-terminal/pull/335

**Visual changes** 
Earlier-
image
image
image

Presently-
image
image
image
image

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -2300,21 +2300,38 @@ class TestMessageBox:
     @pytest.mark.parametrize(
         "to_vary_in_each_message",
         [
+            {"flags": ["starred"]},
+        ],
+        ids=[
+            "no_stars",
+        ],
+    )
+    def test_main_view_compact_output_no_star(
+        self, mocker, message_fixture, to_vary_in_each_message
+    ):
+        message_fixture.update({"id": 0})
+        varied_message = dict(message_fixture, **to_vary_in_each_message)
+        msg_box = MessageBox(varied_message, self.model, varied_message)
+        view_components = msg_box.main_view()
+        assert len(view_components) == 2
+        assert isinstance(view_components[0], Columns)
+
+    @pytest.mark.parametrize(
+        "to_vary_in_each_message",
+        [
             {"sender_full_name": "bob"},
             {"timestamp": 1532103779},
             {"timestamp": 0},
             {},
-            {"flags": ["starred"]},
         ],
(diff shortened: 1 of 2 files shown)


neiljp:
@Dishti-Oberai I've marked this as awaiting update, since I assume you're still working on tests.

Dishti-Oberai:
Thanks for reminding me regarding this @neiljp !
I am working on it presently but it might take some time as I have exams upcoming at my institute too :)
Will try my level best to complete this at the earliest.

Dishti-Oberai:

neiljp (commented):
@Dishti-Oberai I left comments in the stream.

The test changes so far make the tests pass, but they don't yet cover the behavior, which is why I can change the new logic and still get the tests to pass.

PR #1149, opened 2022-01-29
Restrict posting messages on streams having moderators/admins only

### LINKED ISSUE #682: Restrict sending messages to announcement streams
`is_announcement_only` flag is present in `model.stream_data`. The user should be notified ( in the footer ) if they don't have permissions to post to the stream.

mounilKshah:
**What does this PR do?** 
- Added `is_unauthorised_to_post()` function to verify if a user is allowed to message in the given stream or not
- If the user is not allowed, restrict compose box from opening
- Print a warning in the footer
- Test functions (can post and can-not post) for `is_unauthorised_to_post()` function
Partly fixes #682 

**Discussion on developer channel:**
[#zulip-terminal> [Restrict sending messages to streams #T1149 #T682](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Restrict.20sending.20messages.20to.20streams.20.23T1149.20.23T682)
[#api-documentation>stream_post_policy (API docs)](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/stream_post_policy.20.28API.20docs.29)

**Tested?**
- [x] Manually
- [x] Existing tests
- [x] New tests added

**Commit flow** 
Only one commit which contains all the changes made
- Creating `is_unauthorised_to_post()` in `model.py`
- Using that function in `boxes.py`, the compose box is opened only when the user is authorised to send a message to the stream
- Two new test functions added to `test_model.py`:
  -`test_is_unauthorised_to_post_in_stream_can_post()` and 
  -`test_is_unauthorised_to_post_in_stream_cannot_post()`

**Notes & Questions**
The current changes have covered the following cases:
- Reply to a message in stream (r/enter)
- Reply quoting the current message text (>)
- Reply (in the same stream) mentioning the sender of the current message (@)

Use cases is still pending:
- New message to a stream
- Draft whose stream post policies have been changed
- _Full_ member

### Before implementing the changes
before_changes

Here, upon pressing enter/r for replying to the message in the announce stream, the compose box opens even though the user does not have permission to post on that stream

### After implementing the changes
after_changes

Upon trying to reply to the message in the stream, no compose box pops up and a warning is displayed in the footer informing the user about the restrictions due to the stream post policy.

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -37,7 +37,11 @@ from zulipterminal.api_types import (
 )
 from zulipterminal.config.keys import primary_key_for_command
 from zulipterminal.config.symbols import STREAM_TOPIC_SEPARATOR
-from zulipterminal.config.ui_mappings import ROLE_BY_ID, StreamAccessType
+from zulipterminal.config.ui_mappings import (
+    ROLE_BY_ID,
+    STREAM_POST_POLICY,
+    StreamAccessType,
+)
 from zulipterminal.helper import (
     Message,
     NamedEmojiData,
@@ -549,6 +553,60 @@ class Model:
         if message_was_sent:
             notify_if_message_sent_outside_narrow(composition, self.controller)
         return message_was_sent
+
+    def is_unauthorised_to_post_in_stream(self, stream_id: int) -> Optional[str]:
+        """
+        Identify whether a user is authorised to send message to the given
+        stream or not using stream_post_policy and/or is_announcement_policy
+        and accordingly update the footer
+        """
+        if stream_id in self.stream_dict:
+            stream = self.stream_dict[stream_id]
+        else:
+            return "Stream ID is invalid"
+        user_info = self._all_users_by_id.get(self.user_id, None)
+        if user_info is None:
+            # Check whether the given user id is valid or not
+            return "User does not exist"
+
+        if "role" in user_info:
+            role = user_info["role"]
+        else:
+            if user_info.get("is_owner"):
+                role = 100
+            elif user_info.get("is_admin"):
+                role = 200
+            elif user_info.get("is_guest"):
+                role = 600
+            else:
+                role = 400
+
+        if stream.get("is_announcement_only") or stream.get("stream_post_policy") == 2:
+            # For admins and owners only (not using user role)
+            if role <= 200:
+                return None
+            else:
+                return STREAM_POST_POLICY[2]
+        elif stream.get("stream_post_policy") is not None:
+            # is_announcement_only is deprecated in Zulip 3.0+ -> stream_post_policy (ZFL 1)
+            if stream.get("stream_post_policy") == 4:
+                # Check for moderators only stream
+                if role <= 300:
+                    return None
+                else:
+                    return STREAM_POST_POLICY[4]
+            elif stream.get("stream_post_policy") == 3:
+                # Check for 'members' in the stream
+                if role <= 400:
+                    return None
+                else:
+                    return STREAM_POST_POLICY[3]
+            else:
+                # Anyone can post
+                return None
+        else:
+            # Zulip version < 3.0 and is_announcement_only == False, ie, anyone can post
+            return None
 
     def update_private_message(self, msg_id: int, content: str) -> bool:
         request = {
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -474,7 +474,11 @@ class WriteBox(urwid.Pile):
         self, text: str, state: Optional[int]
     ) -> Optional[str]:
         streams_list = self.view.pinned_streams + self.view.unpinned_streams
-        streams = [stream["name"] for stream in streams_list]
+        streams = [
+            stream["name"]
+            for stream in streams_list
+            if self.model.is_unauthorised_to_post_in_stream(stream.get("id")) is None
+        ]
 
         # match_streams takes stream names and typeaheads,
         # but we don't have typeaheads here.
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -1763,11 +1763,18 @@ class MessageBox(urwid.Pile):
                     recipient_user_ids=self.recipient_ids,
                 )
             elif self.message["type"] == "stream":
-                self.model.controller.view.write_box.stream_box_view(
-                    caption=self.message["display_recipient"],
-                    title=self.message["subject"],
-                    stream_id=self.stream_id,
+                is_unauthorised_warning = self.model.is_unauthorised_to_post_in_stream(
+                    self.message["stream_id"]
                 )
+                if is_unauthorised_warning is not None:
+                    self.model.controller.view.set_footer_text(is_unauthorised_warning, "task:warning", 8)
+                    return key
+                else:
+                    self.model.controller.view.write_box.stream_box_view(
+                        caption=self.message["display_recipient"],
+                        title=self.message["subject"],
+                        stream_id=self.stream_id,
+                    )
         elif is_command_key("STREAM_MESSAGE", key):
             if len(self.model.narrow) != 0 and self.model.narrow[0][0] == "stream":
                 self.model.controller.view.write_box.stream_box_view(
@@ -1833,7 +1840,22 @@ class MessageBox(urwid.Pile):
                 recipient_user_ids=[self.message["sender_id"]],
             )
         elif is_command_key("MENTION_REPLY", key):
-            self.keypress(size, primary_key_for_command("REPLY_MESSAGE"))
+            is_unauthorised_warning = self.model.is_unauthorised_to_post_in_stream(
+                self.message["stream_id"]
+            )
+            if is_unauthorised_warning is not None:
+                self.model.controller.view.set_footer_text(is_unauthorised_warning, "task:warning", 8)
+                self.model.controller.view.write_box.stream_box_view(
+                    caption="",
+                    title="",
+                    stream_id="",
+                )
+            else:
+                self.model.controller.view.write_box.stream_box_view(
+                        caption=self.message["display_recipient"],
+                        title=self.message["subject"],
+                        stream_id=self.stream_id,
+                    )
             mention = f"@**{self.message['sender_full_name']}** "
             self.model.controller.view.write_box.msg_write_box.set_edit_text(mention)
             self.model.controller.view.write_box.msg_write_box.set_edit_pos(
@@ -1841,7 +1863,22 @@ class MessageBox(urwid.Pile):
             )
             self.model.controller.view.middle_column.set_focus("footer")
         elif is_command_key("QUOTE_REPLY", key):
-            self.keypress(size, primary_key_for_command("REPLY_MESSAGE"))
+            is_unauthorised_warning = self.model.is_unauthorised_to_post_in_stream(
+                self.message["stream_id"]
+            )
+            if is_unauthorised_warning is not None:
+                self.model.controller.view.set_footer_text(is_unauthorised_warning, "task:warning", 8)
+                self.model.controller.view.write_box.stream_box_view(
+                    caption="",
+                    title="",
+                    stream_id="",
+                )
+            else:
+                self.model.controller.view.write_box.stream_box_view(
+                        caption=self.message["display_recipient"],
+                        title=self.message["subject"],
+                        stream_id=self.stream_id,
+                    )
 
             # To correctly quote a message that contains quote/code-blocks,
             # we need to fence quoted message containing ``` with ````,
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -808,6 +808,263 @@ class TestModel:
             )
 
     @pytest.mark.parametrize(
+        "user_role, stream_properties",
+        [
+            case(
+                {"is_admin": True},
+                {"is_announcement_only": True},
+                id="is_admin:announcement_only:Zulip2.1",
+            ),
+            case(
+                {"is_admin": True},
+                {"is_announcement_only": False},
+                id="is_admin:not_announcement_only:Zulip2.1",
+            ),
+            case(
+                {"is_admin": False},
+                {"is_announcement_only": False},
+                id="is_not_admin:not_announcement_only:Zulip2.1",
+            ),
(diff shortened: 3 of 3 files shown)


neiljp (commented):
@mounilKshah Thanks for working on this! :+1:

I've left some feedback inline, let me know if you have questions.

Once the code is moved and the logic is isolated, I'd highly recommend adding some tests for the new model function, as that will ensure we cover a lot (or all) of the combinations of stream settings, server levels (if necessary) and user roles.

zulipterminal/helper.py
```diff
@@ -485,6 +485,43 @@ def classify_unread_counts(model: Any) -> UnreadCounts:
...
+    """
+    if model.stream_dict[stream_identifier].get(
+        "stream_post_policy"
+    ) == 2 or model.stream_dict[stream_identifier].get("is_announcement_only"):
+        if model.get_user_info(model.user_id).get("role") <= 200:
+            return False
+        else:
+            print("Admins only")
+            model.controller.report_warning("Only Admins and owners are allowd to type")
+            return True
```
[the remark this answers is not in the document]
mounilKshah: Thank you for the feedback! I will modify the function as per your suggestions and move it to `model.py`.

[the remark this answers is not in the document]
neiljp: Extra detail in each case of why they cannot post would be useful to a user. Note that you don't need to have the error message in the check function, but you may be able to output the warning as a string with one or more elements filled in from the 'can I post' function return value(s); either approach can work.

zulipterminal/model.py
```diff
@@ -516,6 +516,45 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
...
+        update the footer
+        """
+        if self.stream_dict[stream_identifier].get(
+            "stream_post_policy"
+        ) == 2 or self.stream_dict[stream_identifier].get("is_announcement_only"):
+            # checks whether the user is an admin/owner
+            if (
+                self.get_user_info(self.user_id).get("role") <= 200
+                or self.initial_data.get("is_admin", True)
+                or self.initial_data.get("is_owner", True)
```
neiljp, line 533: I'm guessing `get_user_info` could work here, but the type is wrong so mypy is complaining.

Do we need more than the `get_user_info`?

In principle we could track the role of the user who is running the app separately, but we already have the logic in `get_user_info`. That method does currently do a lot of work which we won't use here, however.

zulipterminal/model.py
```diff
@@ -516,6 +516,45 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
...
+                self.get_user_info(self.user_id).get("role") <= 300
+                or self.initial_data.get("is_admin", True)
+                or self.initial_data.get("is_owner", True)
+            ):
+                return None
+            else:
+                return "Only Admins, moderators and owners are allowd to type"
+        elif self.stream_dict[stream_identifier].get("stream_post_policy") == 3:
+            # checks whether the user is an admin/owner/moderator/full member
+            if self.get_user_info(self.user_id).get("role") <= 400:
```
neiljp, line 550: This includes all users; a "full user" is subtly different and can be time-limited (eg. so new members cannot write to some streams). Check the zulip help documentation.

zulipterminal/model.py
```diff
@@ -516,6 +516,45 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
             notify_if_message_sent_outside_narrow(composition, self.controller)
         return message_was_sent
 
+    def prevent_compose_box_in_stream(self, stream_identifier: Any) -> bool:
```
neiljp, line 519: If you're unsure of a type, you can check similar methods that use it. This is an `int`, and we normally use `stream_id`.

neiljp (commented):
@mounilKshah This is looking cleaner already, including after that `black` issue!

Once you've extracted some variables this will read much simpler :)

Given the function behaves a little like a boolean, we could consider naming it with an `is_` prefix or similar, and of course this doesn't need to refer to the compose box specifically now.

As per my earlier review, I'd recommend investigating tests of the model method now, or perhaps once you've pushed a version with these changes. The user roles should be already independent of server version via the function you're already calling, but a parametrised test will help show confidence in behavior.

neiljp:
@mounilKshah Just a note that generally it's fine (at least in Zulip) to force push a rebased version to a PR branch (where we also avoid merge commits). It's fine to show branch progress as you're coding as extra commits if it's something being explored, but otherwise it's better to show the current status of commits that you'd like to see merged. We can discuss on czo if you'd like to know more.

zulipterminal/model.py
mounilKshah: Yes, that is the definition which had been [suggested](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/stream_post_policy.20.28API.20docs.29/near/1319730) on the #api-documentation channel.
However, as suggested by Tim in the [#api-documentation>stream_post_policy](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/stream_post_policy.20.28API.20docs.29/near/1319836), I checked with the [help center documentation](https://zulip.com/help/roles-and-permissions) and here they have mentioned that a _full member_ is

> Some organization settings allow an organization to restrict the permissions of new members; Members who do not have those restrictions are called full members.

So it is not limited to the new members who are members for the `waiting_period_threshold`. Hence, I kept a separate conditional statement for this as well.

neiljp: Ah, OK, so you know about it but you've left the `waiting_period_threshold` part unimplemented at this point?

zulipterminal/model.py
mounilKshah: I too thought of eliminating the use of the function by extracting value of the `role` variable alone. However, going through the changelogs, I noticed that `role` was introduced in Zulip 4.0 (ZFL 59) and the moderator role was added in Zulip 4.0 (ZFL 60), so using `role` alone can cause problems with respect to backwards compatibility. We would also have to use the `is_admin` and `is_owner` attributes to make sure it runs smoothly for earlier versions as well.

One way to do it would be following your suggestion about assigning the user information to a variable. After assigning, we can verify whether the given user exists or not (as the mypy error is due to the `Optional` return type of `get_user_info`). I tried this on my local machine and mypy didn't throw any errors for it.

zulipterminal/model.py
mounilKshah: I hadn't actually thought about its implementation yet. However, I'll look into it.
On a side note, I couldn't find any references to `waiting_period_threshold` in the terminal repository nor in the (API and Developer) documentation. Can you please guide me on that?

mounilKshah:
> @mounilKshah Just a note that generally it's fine (at least in Zulip) to force push a rebased version to a PR branch (where we also avoid merge commits). It's fine to show branch progress as you're coding as extra commits if it's something being explored, but otherwise it's better to show the current status of commits that you'd like to see merged. We can discuss on czo if you'd like to know more.

I am sorry for these extra commits. I will make sure to avoid any unnecessary forced pushes.
I'll rebase them before making the next commit.

zulipterminal/model.py
```diff
@@ -516,6 +516,45 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
...
+        """
+        Identify whether a user is authorised to send message to the given
+        stream or not using stream_post_policy and/or is_announcement_policy
+        and allow/restrict the compose box from opening and accordingly
+        update the footer
+        """
+        if self.stream_dict[stream_identifier].get(
+            "stream_post_policy"
+        ) == 2 or self.stream_dict[stream_identifier].get("is_announcement_only"):
```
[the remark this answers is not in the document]
neiljp: You've done this, but note that you've replaced a `.get` with an index; this may or may not be a problem, but this will be clearer once you have test cases for different server versions.

zulipterminal/model.py
neiljp: I'm a little wary of duplicating the logic in this new method. That said, let's get it working and tested against some known server responses, and we can refactor next.

zulipterminal/model.py
neiljp: I see you posted about that; you could check the server code in the meantime.

neiljp (commented):
@mounilKshah As per another comment, this looks a lot cleaner 👍 

Now seems a good time to write some tests for the function added to `model.py` for different user/stream combinations, for different server versions. That will ensure that the behavior is as we expect without manually testing every combination.

zulipterminal/model.py
mounilKshah: Considering your suggestion regarding `get_user_info` doing a lot more than what we need here, I considered this logic as an alternative to using the `get_user_info` function within the new method.

zulipterminal/model.py
```diff
@@ -516,6 +516,59 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
...
+        stream_dict = self.stream_dict[stream_id]
+        user_info = self._all_users_by_id.get(self.user_id, None)
+        if not user_info:
+            # Check whether the given user id is valid or not
+            return "User does not exist"
+        if (
+            stream_dict.get("is_announcement_only")
+            or stream_dict.get("stream_post_policy") == 2
+        ):
+            # For admins and owners only (not using user role)
```
neiljp, line 534: In future servers there will likely be only `stream_post_policy`, so maybe we should check for the role here?

Could we simplify the conditionals overall, since `== 2` and `is_announcement_only` are the same?

zulipterminal/model.py
```diff
@@ -516,6 +516,59 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
...
+                return "Only Admins and owners can type"
+        elif stream_dict.get("stream_post_policy") is not None:
+            # is_announcement_only is deprecated in Zulip 3.0+ -> stream_post_policy (ZFL 1)
+            if stream_dict.get("stream_post_policy") == 4:
+                # Check for moderators only stream
+                if (
+                    user_info.get("role", 400) <= 300
+                    or user_info.get("is_admin") is True
+                    or user_info.get("is_owner") is True
+                ) and (user_info.get("is_guest") is False):
```
neiljp, line 551: This may be clearer if we hoist the `.get` results into variables above the conditionals.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1734,6 +1734,14 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                     recipient_user_ids=self.recipient_ids,
                 )
             elif self.message["type"] == "stream":
+                is_unauthorised_warning = self.model.is_unauthorised_to_post_in_stream(
+                    self.message["stream_id"]
+                )
+                if is_unauthorised_warning is not None:
+                    print(is_unauthorised_warning)
```
neiljp, line 1741: Maybe we should lint for print statements :)

neiljp (commented):
@mounilKshah This is looking a lot clearer :+1:

zulipterminal/model.py
```diff
@@ -516,6 +516,59 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
             notify_if_message_sent_outside_narrow(composition, self.controller)
         return message_was_sent
 
+    def is_unauthorised_to_post_in_stream(self, stream_id: int) -> Optional[str]:
+        """
+        Identify whether a user is authorised to send message to the given
+        stream or not using stream_post_policy and/or is_announcement_policy
+        and accordingly update the footer
+        """
+        stream_dict = self.stream_dict[stream_id]
```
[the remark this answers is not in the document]
mounilKshah: Okay, I'll rename it to `stream`.

I will add a case within the `is_unauthorised_to_post` function to check whether the `stream_id` is valid or not and return a string in case it is invalid.

zulipterminal/model.py
mounilKshah: Actually, `role` falls under ZFL 59, whereas `is_admin`, `is_guest` and `is_owner` are under ZFL 8/10. Hence, I felt that we could do the checking without using `role`. However, I'll add `role`.

`is_announcement_only` got deprecated after [ZFL 1](https://zulip.com/api/update-stream#parameter-stream_post_policy) right? Hence, I decided to keep both so that it can function the same even after we stop support for v2.1.

zulipterminal/model.py
mounilKshah: Alright, I'll make the required changes.

zulipterminal/ui_tools/boxes.py
mounilKshah: Oh sorry. I completely forgot to remove it before pushing the commit.

zulipterminal/model.py
```diff
@@ -549,6 +549,60 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
...
+            else:
+                role = 400
+
+        if stream.get("is_announcement_only") or stream.get("stream_post_policy") == 2:
+            # For admins and owners only (not using user role)
+            if role <= 200:
+                return None
+            else:
+                return STREAM_POST_POLICY[2]
+        elif stream.get("stream_post_policy") is not None:
```
mounilKshah, line 585: @neiljp 
In my opinion the condition can be changed to – `"stream_post_policy" in stream`
Apart from this, the rest of the function seems fine to me in the current condition.

mounilKshah:

neiljp (commented):
@mounilKshah Thanks for getting back to this :+1:

I'd suggest splitting the core function+test into an initial commit and then connecting the UI in one or more later commits - for now the second commit is completely removing some code you added in the first commit, which is not how we'd approach step-by-step commits in a PR branch. It also is easier to see an isolated function and the tests for it separately.

Regarding the new behavior, I tried some of this out and we can clarify in the stream, but broadly I'd suggest:
- we disallow simple replies and user mention replies (as you had before)
- we open the compose box for quoted replies (as you have now, to allow content quoting to another stream)

In any case I think we need to extend the time the warnings are shown?

I'm not sure if the filter in the comprehension you added in the second commit is for the reply processing, or a step towards further work? It limits autocomplete, but doesn't handle the case of `c` while in the stream, and also entering text manually doesn't stop compose either.

srdeotarse (changes requested):
The PR looks good with proper commits 👍. I will have a look at failing test cases more closely, but here are some suggestions.

srdeotarse (changes requested):

PR #1161, opened 2022-03-07
Rebase and simplification of #870

neiljp:
**What does this PR do?**  

This is a rebase and simplification of #870, which is a partial fix for #774.

I'd appreciate feedback from anyone who contributed to that issue.

**Tested?** 
- [x] Manually

**Commit flow** 

After rebasing and fixing formatting, this splits the original PR into a refactor and feature commit. Around these there are various simplifying commits, most of which I expect to squash with those two before merging - though I may leave them in this PR for future reference by others in terms of tidying a previous PR.

**Notes & Questions** 

This doesn't fully fix the issue, since it doesn't handle private messages, so anyone is welcome to take that up separately.

This also exposes how we could further simplify and refactor the compose code, though considering older PRs may be good to do to avoid them bit-rotting further too :)

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -1412,7 +1412,7 @@ class TestWriteBox:
                 True,
                 True,
                 "CONTAINER_HEADER",
-                "HEADER_BOX_STREAM",
+                "HEADER_BOX_TOPIC",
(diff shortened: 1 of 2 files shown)


neiljp:
This is now merged as 80bb9c0d75f56db98b06da96af411425c75a7304 and the two prior commits, so I'm going to manually close this.

PR #1163, opened 2022-03-12
Respect PM typing notification settings

neiljp:
**What does this PR do?**  

**Tested?** 
- [x] Manually

**Commit flow** 

- Introduce & test new centralized user settings in model for new property (from register)
- Update the new property from events
- Use the updating property to determine whether to send typing events

**Notes & Questions** 

This introduces user settings, which is intended as a new central location for this. This new setting is set and updated via `user_settings`, whereas older settings may be in multiple locations dependent on parameter passed to the server (`user_settings_object`)

I have initially assumed that this setting is only available in the new structure, but that's not completely clear to me.

**Visual changes** 

For servers which support this setting, ZT currently shows warning notices in the footer, returned from the server. If requested via the setting, this PR avoids sending notifications, so the server doesn't send a notice, which then isn't shown.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -14,6 +14,7 @@ from zulipterminal.model import (
     MAX_TOPIC_NAME_LENGTH,
     Model,
     ServerConnectionFailure,
+    UserSettings,
 )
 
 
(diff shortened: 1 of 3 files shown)


neiljp:
Merging this now after discussion in #**api design > typing notification endpoint**.

Issue #1119, opened 2021-08-14
Include more elements in StreamInfo view.


zee-bit:
#880 and #900 added several important elements in the `StreamInfo` view almost matching the "Stream settings" of the web app. However, there are possibilities for adding more fields to the popup, which includes:
- Stream id (maybe the web app is planning  to add this too(?))
- Type of the stream (Public/Private)
- [Posting policy](https://chat.zulip.org/help/stream-sending-policy) (e.g. announcement only streams)
- Availability of history i.e. whether the history of the stream is public to its subscribers
- User's role within the stream? - different from the user's organization-level role

The API documentation for the [`/subscription`](https://zulip.com/api/get-subscriptions#response) endpoint should prove helpful in tackling this PR.

kingjuno:
@zulipbot claim

zulipbot:
Welcome to Zulip, @kingjuno! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

avats-dev:
@kingjuno are you working on this?

zee-bit:
@avats-dev Some of those fields have already been added by @kingjuno via #1127. I have edited the issue summary to use checkboxes instead of bullets so that it's easier to indicate completed tasks.

The remaining tasks either need to be made backward compatible, which requires some knowledge of the zulip-api and manual research, or are not much of a priority right now. See [#zulip-terminal>Include more elements in StreamInfo view](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Include.20more.20elements.20in.20StreamInfo.20view) for discussions related to this issue.

PR #1165, opened 2022-03-14
Added stream_post_policy to the streaminfo view

mounilKshah:
**What does this PR do?**
This PR contains commit which add the `stream_post_policy` to the streaminfo view. Test case for the same have been added under `tests/ui_tools/test_popups.py`
Partial fix for #1119.

Link to discussion on chat.zulip.org:
[#zulip-terminal>Adding elements to stream info](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Adding.20elements.20to.20stream.20info.20.28issue.20.23T1119.29)

**Tested?**
- [x] Manually

**Commit flow**
- Contains code for adding `stream_post_policy` and the updated `test_popup_height` function

**Visual changes**

## Before the commt:
image

## After the commit:
image


zulipterminal/ui_tools/views.py
```diff
@@ -1307,6 +1307,20 @@ def __init__(self, controller: Any, stream_id: int) -> None:
...
+            3: "Only members can post",
+            4: "Only owners, admins and moderators can post",
+        }
+        if stream.get("stream_post_policy") is None:
+            if stream.get("is_announcement_only"):
+                stream_policy = "Only admins can post"
+            else:
+                stream_policy = "Any user can post"
+        else:
+            stream_policy = stream_posting_policy[stream.get("stream_post_policy")]
```
neiljp, line 1322: This is what I meant about the start of the other PR code, where there is something simpler.

neiljp (commented):
@mounilKshah Good start, with lots of overlap with the other PR.

One thing we can do in these situations is, where appropriate, extract things like functions we use (or will use) in separate earlier commits in one of these PRs. If we're only adjusting existing code with no behavioral change, this is a `refactor:` commit. Myself or someone else with merge rights can push a few commits using `git cherry-pick`, making multiple PRs which depend on it easier to handle as they can both then be rebased against the new `main` branch. Or, of course, you could just make another PR for that and rebase once that is merged.

zulipterminal/ui_tools/views.py
```diff
@@ -1307,6 +1307,20 @@ def __init__(self, controller: Any, stream_id: int) -> None:
         self.controller = controller
         stream = controller.model.stream_dict[stream_id]
 
+        stream_posting_policy = {
+            1: "Any user can post",
+            2: "Only owners and admins can post",
+            3: "Only members can post",
+            4: "Only owners, admins and moderators can post",
+        }
```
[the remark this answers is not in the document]
mounilKshah: Okay, I will add this under a separate `refactor` commit for `ui_mappings.py`.

zulipterminal/ui_tools/views.py
mounilKshah: I will go through it once again and make the required changes.

zulipterminal/config/ui_mappings.py
```diff
@@ -40,3 +40,10 @@
     400: {"bool": "", "name": "Member"},
     600: {"bool": "is_guest", "name": "Guest"},
 }
+
+STREAM_POST_POLICY = {
```
neiljp, line 44: This is not the same as in the title of the commit?

zulipterminal/config/ui_mappings.py
```diff
@@ -40,3 +40,10 @@
...
+    1: "Any user can post.",
+    2: "Only organization administrators can send to this stream.",
+    3: "Guests cannot send to this stream.",
+    4: "Only organization administrators and moderators can send to this stream.",
```
neiljp, line 48: Let's skip the trailing `.` - no other fields in the popup have this, and we can always add it uniformly later.

zulipterminal/ui_tools/views.py
```diff
@@ -1376,6 +1385,7 @@ def __init__(self, controller: Any, stream_id: int) -> None:
                         f"Press {email_keys} to copy Stream email address",
                     ),
                     ("History of Stream", f"{availability_of_history}"),
+                    ("Posting Policy", f"{stream_policy}"),
```
neiljp, line 1388: We're adding new functionality here, so it is not a refactor commit. In any case we should mention the file being updated, as in the zulip-terminal code style section.

zulipterminal/ui_tools/views.py
```diff
@@ -1329,6 +1330,14 @@ def __init__(self, controller: Any, stream_id: int) -> None:
             )
         ]
 
+        if stream.get("stream_post_policy") is None:
```
neiljp, line 1333: Where we're using default values, it is clearer if you use `"text" in some_dict`, as that indicates we're testing for membership. The cost is similar, but "fetch and return an error value, which we test against" is not as obvious as explicitly using `in`.

neiljp (commented):
@mounilKshah This looks close :+1:

At some point I think it'd be good to migrate to modern versions when we get the data itself like we do with some other data, but this version works well for now.

I'm unclear if the final part of #1119 is functional on the server at this point (or will be replaced by a new implementation instead), but I suspect this can be extracted into a separate issue once we know what's happening there.

As a follow-up to this it might be useful to discuss if we want to group/combine the related stream visibility properties to make it clearer.
(we could also add a way to copy a link to the stream for web usage)

zulipterminal/config/ui_mappings.py
mounilKshah: This was added under the first commit which made changes only to `ui_mappings.py` so that it could be merged using `git cherry-pick` if required.

mounilKshah:
> I'm unclear if the final part of #1119 is functional on the server at this point (or will be replaced by a new implementation instead), but I suspect this can be extracted into a separate issue once we know what's happening there.

Putting it under a separate issue seems to be a good idea to me. As you had mentioned [here](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Restrict.20sending.20messages.20to.20streams.20.23T1149.20.23T682/near/1331823), we can have a separate section for the user details in the stream. Apart from the user role, we can also add details like average number of messages sent by user in the stream, number of topics the user has posted on (out of total topics in the stream), etc. 

> As a follow-up to this it might be useful to discuss if we want to group/combine the related stream visibility properties to make it clearer. (we could also add a way to copy a link to the stream for web usage)

This can be a good addition to the stream info!

neiljp:
@mounilKshah While I think in principal we could improve these tests, I'll be happy to merge my local slightly adjusted version if we can agree on the text for the `3` policy value, as per my message in the stream.

neiljp:
@mounilKshah Thanks for this! Merging now after some minor commit text adjustments 🎉

PR #1167, opened 2022-03-15
Ensure color codes are valid at runtime if running in 16-color mode.

### LINKED ISSUE #1158: Ensure color codes are valid at runtime if running in 16-color mode
Zulip terminal runs in 16 color depth mode only if the color codes used in themes for the same are acceptable by urwid. Codes of those 16 acceptable colors are mentioned in the [documentation](https://github.com/zulip/zulip-terminal/blob/main/zulipterminal/themes/THEME_CONTRIBUTING.md). 
In case the contributor uses some other color code instead of the aforementioned codes, the terminal shall not run. The prompted error in such a case, (as of now) doesn't indicate the exact reason for the error. 
Construct a test in the codebase to check if the 16 color codes used are acceptable ones and prompt an error otherwise. 
PS: This issue is a further development of PR #1154.
Related discussions on CZO: https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/16-color.20names.20in.20themes

srdeotarse:
**What does this PR do?**  
In ``/zulip-terminal/zulipterminal/config/themes.py`` code is added in ``generate_theme`` function to check if valid color codes are used in color files for specific themes. For eg. (``colors_gruvbox.py`` for ``gruvbox_dark.py`` theme)

Fixes #1158 

**Tested?** 
- [x] Manually

**Commit flow** 

**Notes & Questions** 

**Interactions** 

**Visual changes**

### DIFF
--- a/tests/config/test_themes.py
+++ b/tests/config/test_themes.py
@@ -211,6 +211,8 @@ def test_add_pygments_style(
 def test_validate_colors(theme_name: str, color_depth: int) -> None:
     theme = THEMES[theme_name]
 
+    header_text = f"Invalid 16-color codes in theme '{theme_name}':\n"
+
     # No invalid colors
     class Color(Enum):
         # color          =  16code          256code   24code
@@ -233,10 +235,7 @@ def test_validate_colors(theme_name: str, color_depth: int) -> None:
     theme.Color = Color1
     with pytest.raises(InvalidThemeColorCode) as e:
         validate_colors(theme_name, 16)
-    assert (
-        str(e.value)
-        == f"Check 16-color-code for DARK0_HARD = (blac) in theme - {theme_name}\n"
-    )
+    assert str(e.value) == header_text + "- DARK0_HARD = blac"
 
     # Two invalid colors
     class Color2(Enum):
@@ -250,9 +249,7 @@ def test_validate_colors(theme_name: str, color_depth: int) -> None:
     with pytest.raises(InvalidThemeColorCode) as e:
         validate_colors(theme_name, 16)
     assert (
-        str(e.value)
-        == f"Check 16-color-code for DARK0_HARD = (blac) in theme - {theme_name}\n"
-        + f"Check 16-color-code for GRAY_244 = (dark_gra) in theme - {theme_name}\n"
+        str(e.value) == header_text + "- DARK0_HARD = blac\n" + "- GRAY_244 = dark_gra"
     )
 
     # Multiple invalid colors
@@ -268,8 +265,9 @@ def test_validate_colors(theme_name: str, color_depth: int) -> None:
         validate_colors(theme_name, 16)
     assert (
         str(e.value)
-        == f"Check 16-color-code for DEFAULT = (defaul) in theme - {theme_name}\n"
-        + f"Check 16-color-code for DARK0_HARD = (blac) in theme - {theme_name}\n"
-        + f"Check 16-color-code for GRAY_244 = (dark_gra) in theme - {theme_name}\n"
-        + f"Check 16-color-code for LIGHT2 = (whit) in theme - {theme_name}\n"
+        == header_text
+        + "- DEFAULT = defaul\n"
+        + "- DARK0_HARD = blac\n"
+        + "- GRAY_244 = dark_gra\n"
+        + "- LIGHT2 = whit"
     )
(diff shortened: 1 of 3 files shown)


tests/config/test_themes.py
```diff
@@ -213,3 +215,25 @@ def test_add_pygments_style(
...
+
+# Validate 16-color-codes
+@pytest.mark.parametrize(
+    "color_depth, theme_name",
+    [
+        (16, "zt_dark"),
+        (16, "gruvbox_dark"),
+        (16, "gruvbox_light"),
+        (16, "zt_light"),
+        (16, "zt_blue"),
```
neiljp, line 228: It is useful to use an existing theme as a basis for a test, but I'm not sure we benefit from testing against all the existing themes.

Instead, I'd suggest exploring different `Color` sets, eg. ensuring that a range of names work, as well as what output one gets if multiple colors in `Color` are invalid.

tests/config/test_themes.py
```diff
@@ -213,3 +215,25 @@ def test_add_pygments_style(
...
+)
+def test_validate_colors(theme_name: str, color_depth: int) -> None:
+    class Color(Enum):
+        WHITE = "whit  #256  #24"
+
+    theme = THEMES[theme_name]
+    theme.Color = Color
+
+    with pytest.raises(Check16ColorCode):
+        validate_colors(theme_name, 16)
```
neiljp, line 239: This is a good basic test, but it would be useful to also confirm what text is in the exception (particularly for different cases).

neiljp (commented):
@srdeotarse Thanks for looking into this :+1:

I've left some feedback inline. We'll likely squash the test commit with the other one when merging.

Please do check the commit style guidelines in the README :)
You may also find installing gitlint useful, if you haven't already.

zulipterminal/config/themes.py
```diff
@@ -138,6 +163,26 @@ def generate_theme(theme_name: str, color_depth: int) -> ThemeSpec:
...
+        invalid_16_color_code = ""
+        for c in theme_colors:
+            color_code = ""
+            color_code = c.value.split()[0]
+            if color_code not in aliases_16_color:
+                invalid_16_color_code = str(c.name)
+                failure_text = "Check 16-color-code for {} in theme - {}".format(
+                    invalid_16_color_code, theme_name
+                )
+                raise Check16ColorCode("".join(failure_text))
```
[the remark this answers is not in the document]
neiljp: This is an improvement, though note that we now *always* raise, even with valid colors! It seems like you've written your tests to match the code.

zulipterminal/config/themes.py
```diff
@@ -103,6 +103,30 @@
     "blue": "zt_blue",
 }
 
+aliases_16_color = [
```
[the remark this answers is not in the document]
neiljp: Please do migrate this over in a first commit, and ensure everything works (eg. using `make check`) at that point. This will make the second commit smaller and easier to read - and won't have the duplication you have currently.

zulipterminal/config/themes.py
```diff
@@ -138,6 +164,27 @@ def generate_theme(theme_name: str, color_depth: int) -> ThemeSpec:
...
 
+def validate_colors(theme_name: str, color_depth: int) -> None:
+    """
+    This function validates 16-color-codes in color files for theme selected.
+    If any color is not in accordance with urwid default 16-color-code then the function raises Check16ColorCode error
+    which prints colors for which 16-color-code is invalid.
+    """
+    theme_colors = THEMES[theme_name].Color
+    failure_text = []
+    new_line = "\n"
```
neiljp, line 175: You can join the errors with this instead.

zulipterminal/config/themes.py
```diff
@@ -138,6 +164,27 @@ def generate_theme(theme_name: str, color_depth: int) -> ThemeSpec:
...
+    if color_depth == 16:
+        for c in theme_colors:
```
neiljp, line 177: Please use meaningful names.

zulipterminal/config/themes.py
```diff
@@ -138,6 +164,27 @@ def generate_theme(theme_name: str, color_depth: int) -> ThemeSpec:
...
+            color_code = c.value.split()[0]
+            if color_code not in valid_16_color_codes:
+                invalid_16_color_code = str(c.name)
+                invalid_16_color_code_name = str(c.value).split()[0]
```
neiljp, line 181: This is just `color_code`?

neiljp:
@srdeotarse Thanks for the update; as per discussion in the stream I slightly adjusted some text and added a commit to tidy slightly. Merging now :tada:

PR #1175, opened 2022-03-22
[WIP] themes: Add solarized theme

siddharthc30:
**What does this PR do?**  

Partial FIx for [#1133](https://github.com/zulip/zulip-terminal/issues/1133)
Adds Solarized theme support for zulip terminal app.
As of Now, has only colors_solarized.py file. Need to add the rest of the files

**Tested?** 
Not tested as of now
- [ ] Manually

**Commit flow** 

**Notes & Questions** 

**Interactions** 

**Visual changes**

### DIFF
--- a/zulipterminal/themes/colors_solarized.py
+++ b/zulipterminal/themes/colors_solarized.py
@@ -1,0 +1,40 @@
+"""
+COLORS FOR SOLARIZED THEMES
+-------------------------
+Contains color definitions or functions common accross solarized themes.
+For further details on themefiles look at the theme contribution guide.
+
+This file uses the official solarized colors.
+For color reference see:
+    https://github.com/altercation/solarized/blob/master/vim-colors-solarized/colors/solarized.vim
+"""
+
+from enum import Enum
+
+from zulipterminal.config.color import DefaultBoldColor, color_properties
+
+class SolarizedColor(Enum):
+    # color          =  16code          256code   24code
+    base03           = 'dark_gray       h234      #002b36'
+    base02           = 'black           h235      #073642'
+    base01           = 'light_green     h239      #586e75'
+    base00           = 'dark_gray       h240      #657b83'
+    base0            = 'light_blue      h244      #839496'
+    base1            = 'light_cyan      h245      #93a1a1'
+    base2            = 'light_gray      h187      #eee8d5'
+    base3            = 'white           h230      #fdf6e3'
+    yellow           = 'yellow          h136      #b58900'
+    orange           = 'light_red       h166      #cb4b16'
+    red              = 'dark_red        h124      #dc322f'
+    magenta          = 'dark_magenta    h125      #d33682'
+    violet           = 'light_magenta   h61       #6c71c4'
+    blue             = 'dark_blue       h33       #268bd2'
+    cyan             = 'dark_cyan       h37       #2aa198'
+    green            = 'dark_green      h64       #859900'
+
+
+
+
+DefaultBoldColor = color_properties(SolarizedColor, "BOLD")
+
+    


zulipterminal/themes/solarized_dark.py
```diff
@@ -0,0 +1,90 @@
...
+
+"""
+
+from pygments.styles.solarized import SolarizedDarkStyle
+
+from zulipterminal.themes.colors_solarized import DefaultBoldColor as Color
+
+STYLES = {
+    # style_name       :  foreground                   background
+    None               : (Color.base0,                 Color.base03),
```
neiljp, line 14: While the dark and light themes use some similar colors from the solarized palette, `base03` is not very descriptive when we're writing the styles. My suggestion would be to write two separate enums for dark (and later light) versions which use the values from the main palette. We could still store the main colors in an enum if that's simpler.

This would allow eg. `Color.bg`, `Color.fg` (or longer, like background/foreground) to be different values based on which enum we import, which is clearer than `base0`.

If we keep the highlight colors the same, this could potentially also enable complete reuse of `STYLES` for light after importing a different enum for `Color`.

zulipterminal/themes/solarized_dark.py
```diff
@@ -0,0 +1,90 @@
...
+    'name'             : (Color.red__BOLD,             Color.base03),
+    'unread'           : (Color.violet,                Color.base03),
+    'user_active'      : (Color.green,                 Color.base03),
+    'user_idle'        : (Color.yellow,                Color.base03),
+    'user_offline'     : (Color.base01,                Color.base03),
+    'user_inactive'    : (Color.base01,                Color.base03),
+    'title'            : (Color.base0__BOLD,           Color.base03),
+    'column_title'     : (Color.base0__BOLD,           Color.base03),
+    'time'             : (Color.base01,                Color.base03),
+    'bar'              : (Color.base0,                 Color.base00),
```
neiljp, line 29: According to the solarized spec, we shouldn't have `base00` in the dark theme? This would be more obvious if we stick to meaningful names like `background`, `foreground` and so on with the separate enum approach I suggest above.

zulipterminal/themes/solarized_dark.py
```diff
@@ -0,0 +1,90 @@
...
+    'msg_math'         : (Color.base02,                Color.base00),
+    'msg_mention'      : (Color.orange__BOLD,          Color.base03),
+    'msg_link'         : (Color.blue,                  Color.base03),
+    'msg_link_index'   : (Color.blue__BOLD,            Color.base03),
+    'msg_quote'        : (Color.base1,                 Color.base03),
+    'msg_code'         : (Color.base03,                Color.base0),
+    'msg_bold'         : (Color.base0__BOLD,           Color.base03),
+    'msg_time'         : (Color.base03,                Color.base1),
+    'footer'           : (Color.base03,                Color.base2),
+    'footer_contrast'  : (Color.base2,                 Color.base03),
```
neiljp, line 43: Similarly `base2` does not belong in dark?

neiljp (commented):
@siddharthc30 Thanks for looking at this - this seems to run OK now we have the styles :+1:

I've left some comments inline. For the tests failing, some things may be fixed by running `make fix`, or individually running isort or black on the files to reformat them automatically.

I've not examined the color allocation at this point, as I think that'll be clearer with more meaningful names for the colors, and also avoid using colors from the overall palette that don't belong in dark (and then later also light).

zulipterminal/config/themes.py
```diff
@@ -99,6 +100,7 @@
 THEME_ALIASES = {
     "default": "zt_dark",
     "gruvbox": "gruvbox_dark",
+    "solarized": "solarized_dark",
```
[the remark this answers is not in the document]
siddharthc30: Ok, updated

zulipterminal/themes/colors_solarized.py
```diff
@@ -0,0 +1,40 @@
...
+For further details on themefiles look at the theme contribution guide.
+
+This file uses the official solarized colors.
+For color reference see:
+    https://github.com/altercation/solarized/blob/master/vim-colors-solarized/colors/solarized.vim
+"""
+
+from enum import Enum
+
+from zulipterminal.config.color import DefaultBoldColor, color_properties
```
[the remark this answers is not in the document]
siddharthc30: Updated

zulipterminal/themes/colors_solarized.py
```diff
@@ -0,0 +1,40 @@
...
+
+class SolarizedColor(Enum):
```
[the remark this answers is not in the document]
siddharthc30: Ok, so Is it fine as of now?

zulipterminal/themes/solarized_dark.py
siddharthc30: yeah, I will look into it

zulipterminal/themes/colors_solarized.py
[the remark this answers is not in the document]
neiljp: Once we have the semantic color labels then it will be clearer to read all of this. I'll take a look at that stage.

zulipterminal/themes/solarized_dark.py
neiljp: A simple approach could be to have the core palette as you have it now - which is easier to compare to the solarized specification - and then define other enums for dark and light which reuse those values, eg.

```python3
class SolarizedDarkColor(Enum):
    background = SolarizedColor.base03.value
    foreground = SolarizedColor.base0.value
    ...
```
Then in the theme file we can import that enum instead, and `Color.foreground` is easier to remember and validate than some number. Similarly the enum as above makes it easier to validate that base0 is the foreground and should be according to the specification.

Please do read the enum documentation yourself, but I believe inheritance of enums with values doesn't work like 'sets' in this way, ie. I don't think you can use a 'mixin enum' to set common colors.

siddharthc30: ok got it, will try this out and update soon

PR #1176, opened 2022-03-23
Use only colors from suggested gruvbox palettes

neiljp:
**What does this PR do?**  

When the light variant was added, we noticed that the gruvbox palettes have the faded/neutral/bright variants of colors, and it was suggested to group these. 

It is noticeable in the palettes linked below that:
- dark uses bright+neutral colors
- light uses faded+neutral colors

The existing dark variant uses some faded colors, and the light variant seems to use some bright ones. This PR is intended to address these issues, and group colors to make this usage clearer.

https://github.com/morhetz/gruvbox#palette

**Tested?** 
- [x] Manually

**Commit flow** 

- Update dark mode, and label/group used color palette elements
- Update light mode, and label/group remaining color palette elements

**Notes & Questions** 

The term 'mode' is used in the color palette comments, based on the gruvbox terminology.

**Visual changes** 

Please do fetch this PR for comparison, as colors can look differently in local systems.

Main elements that changed: (as evident in the diff) <-- DARK MODE
- selected elements: light-on-faded-blue -> dark-on-neutral-blue
- general narrow titles (PMs, mentions, ...): light-on-faded-blue -> dark-on-bright-blue
- quotes: faded-yellow-on-dark -> neutral-yellow-on-dark
- area:error/task:error: light-on-faded-red -> dark-on-bright-red
- area:msg: dark-on-bright-red (ok!) -> dark-on-neutral-purple (avoid area:error conflict)
- task:warning: dark-on-bright-red (ok!) -> dark-on-neutral-purple (avoid task:error conflict)
- area:user (user info): light-on-faded-blue -> dark-on-bright-yellow (bright blue already used for streams)

### DIFF
--- a/zulipterminal/themes/gruvbox_dark.py
+++ b/zulipterminal/themes/gruvbox_dark.py
@@ -19,10 +19,10 @@ from zulipterminal.themes.colors_gruvbox import DefaultBoldColor as Color
 STYLES = {
     # style_name       :  foreground                   background
     None               : (Color.LIGHT2,                Color.DARK0_HARD),
-    'selected'         : (Color.LIGHT2,                Color.FADED_BLUE),
-    'msg_selected'     : (Color.LIGHT2,                Color.FADED_BLUE),
-    'header'           : (Color.NEUTRAL_BLUE,          Color.FADED_BLUE),
-    'general_narrow'   : (Color.LIGHT2,                Color.FADED_BLUE),
+    'selected'         : (Color.DARK0_HARD,            Color.NEUTRAL_BLUE),
+    'msg_selected'     : (Color.DARK0_HARD,            Color.NEUTRAL_BLUE),
+    'header'           : (Color.NEUTRAL_BLUE,          Color.BRIGHT_BLUE),
+    'general_narrow'   : (Color.DARK0_HARD,            Color.BRIGHT_BLUE),
     'general_bar'      : (Color.LIGHT2,                Color.DARK0_HARD),
     'name'             : (Color.NEUTRAL_YELLOW__BOLD,  Color.DARK0_HARD),
     'unread'           : (Color.NEUTRAL_PURPLE,        Color.DARK0_HARD),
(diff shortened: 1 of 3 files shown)


zulipterminal/themes/gruvbox_dark.py
```diff
@@ -19,10 +19,10 @@
 STYLES = {
     # style_name       :  foreground                   background
     None               : (Color.LIGHT2,                Color.DARK0_HARD),
-    'selected'         : (Color.LIGHT2,                Color.FADED_BLUE),
-    'msg_selected'     : (Color.LIGHT2,                Color.FADED_BLUE),
-    'header'           : (Color.NEUTRAL_BLUE,          Color.FADED_BLUE),
-    'general_narrow'   : (Color.LIGHT2,                Color.FADED_BLUE),
+    'selected'         : (Color.DARK0_HARD,            Color.NEUTRAL_BLUE),
```
rht, line 22: From reading the blame, there seems to be a reason why the foreground is light. See https://github.com/zulip/zulip-terminal/commit/0800dfa12a50430a3265f2e98f2106b52b269efa.

neiljp: I believe we should stick to intended palette colors over matching zt_dark, and syntax highlighting mitigates the other reason that change was made?

rht: It might have looked better with white for ZT dark as well?

Dishti-Oberai (approved):

zulipterminal/themes/gruvbox_dark.py
rht: This is what is in zt_dark.py: `    'selected'        : (Color.WHITE,               Color.DARK_BLUE),`. Making it `DARK0_HARD` actually is not making it consistent.

neiljp: @rht I agree it is not consistent.

I experimented with `LIGHT2` with the different blues which should be in gruvbox-dark (as backgrounds), and the contrast wasn't great, which is why I selected `DARK0_HARD`.

For me, the blue used in zt_dark is darker, so white text works better there.

Your terminals and opinions may differ, so please do follow-up :)

rht: OK, I'm good with the `DARK0_HARD` then. It will look the same in my terminal, because 24 bit is true color.

neiljp: @rht To confirm, you're OK with the set of dark theme changes?

neiljp:
I've just made a preliminary equivalent set of changes for the light version.

I'm incrementally rearranging the color ordering/grouping as this happens.

There are still some overlapping colors which I'd like to avoid in both dark and light from these changes.

zulipterminal/themes/gruvbox_dark.py
rht: Yes.

neiljp:
This should be the final version to review now.

Changes since the 2nd iteration are:
- improve comments in the colors file
- use purples for some styles that otherwise didn't need to change, but avoid confusion/conflict with other styles in the same area (`task:*`, `area:*`)

Dishti-Oberai:
I find the changes good enough.

Issue #938, opened 2021-02-28
Support symbol "themes" to allow varying levels of icon/symbol rendering 


neiljp:
Spinning this feature idea out from #178, which this could end up resolving and is a relevant read.

The idea is to add a feature to support terminals with various levels of unicode support, so that the UI can look reasonable on basic terminals, "typical" ones and potentially up to ones with recent emoji support - though rendering of emoji themselves in place of our current text rendering is likely a different (if related) issue.

Currently we have aimed for a middle path of symbols that appear to render fairly consistently across common platforms but make the UI interesting enough. However, there are two extremes around this:
* some platforms/systems/fonts might support as little as basic letters (ascii and/or localized text + limited symbols)
* some platforms/systems/fonts might support way more unicode, and would benefit from using more advanced symbols

Some aspects to consider:
* likely a command-line option, later a config option
* Can we somehow detect if symbols render properly, or the font supports them, and downgrade/upgrade? (override by the above)

neiljp:
This was touched on in #**zulip-terminal > Web public marker #T1135/#T712/#T1177**, where an emoji was proposed for this role. However, that emoji was double-width and did not align with the existing stream markers. In an emoji-capable terminal we could have sets of characters which align well for streams, including eg. 'lock' markers for private streams, while keeping simple characters/letters which also align well together in other terminal 'symbol themes'.

neiljp:
For reference, on Debian(-like) systems one can install `fonts-noto-color-emoji` and get emoji displayed quite well, with other fonts falling back to use this set even if the main font is something else - at least on `gnome-terminal`.

However, this does not suffice for allowing eg. running on a basic linux terminal, at least based on prior experience with lots of other characters not displaying as expected.

Issue #712, opened 2020-07-08
Introduce new prefix character for web public streams


preetmishra:
Currently, both the web public streams and the public streams use '#' as `prefix_character` which makes it hard to portray the distinction. To resolve this, we could either use 'W' (considering how 'P' is used for private streams) or some other character for the web public streams.

#### Screenshot

| Web app | Zulip Terminal |
| ------------ | ------------------- |
|![image](https://user-images.githubusercontent.com/43616959/86891138-fceeaf00-c11b-11ea-8890-e8d3f430bfd6.png) | ![image](https://user-images.githubusercontent.com/43616959/86891492-8f8f4e00-c11c-11ea-8663-2b6c9d88146d.png) |

preetmishra:
This not critically important for now but would be useful to have post https://github.com/zulip/zulip/pull/14638.

neiljp:
This is covered in more detail by the broader but duplicate #1135.

neiljp:
This was fixed by #1177.

PR #1177, opened 2022-03-23
Indicate web-public streams differently

neiljp:
**What does this PR do?**  

This is intended to fix most of #712 (#1135)

See #**zulip-terminal > Web public marker #T1135/#T712**

**Tested?** 
- [X] Manually

**Commit flow** 

- Add model accessor for stream type, given stream id
- Use new accessor outside of model (in UI)
- Add fixtures for a web public stream & adjust existing tests
- Add symbol and extend existing code minimally to indicate symbol in most cases
- Add symbol rendering in stream list
- Refactor duplicated if/elif blocks to use new common dict in ui_mappings (descriptions)
- Refactor if/elif blocks to use same dict for icons too

**Notes & Questions** 

This needs some tests fixed and I'm planning an additional refactor, but pushing now to get feedback.

**update** Tests are now fixed and extended.

**Visual changes** 

Currently, web-public streams are shown as `A`. This is not ideal, but I've not found a reliable symbol/character yet (`W` in bold was very full/blocky). If you fetch the PR you can play with variations by changing the character in `symbols.py`.

**update** After debate on the stream, we currently have `⊚` as an approximation to a narrow (1-character-wide) circular symbol. We can explore this further when we explore symbol themes #938 (maybe #1114)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -54,7 +54,7 @@ def stream_button(mocker: MockerFixture) -> StreamButton:
             "name": "PTEST",
             "id": 205,
             "color": "#bfd56f",
-            "invite_only": False,
+            "stream_access_type": "public",
             "description": "Test stream description",
         },
         controller=mocker.patch("zulipterminal.core.Controller"),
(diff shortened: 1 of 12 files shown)


neiljp:
Merging this is temporarily blocked on reports that the selected symbol doesn't render in WSL, though does in Windows Terminal.

Issue #1172, opened 2022-03-18
Improve reporting upon moving/splitting topics


neiljp:
Currently a message is given to the user when a topic is moved, but we could give more information about this:
- where from?
- where to?
- how many? (all? some?)

For the first version of this, updating the text we output would be sufficient.

As a follow-up, we could consider if there are other ways to indicate eg. split topics.

For comparison, the web app appears to intentionally not sort after splitting topics, to leave both parts in view, whereas ZT does so.

srdeotarse:
@zulipbot claim

zulipbot:
Hello @srdeotarse, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

PR #1178, opened 2022-03-24
model: Improve reporting upon moving/splitting topic.

srdeotarse:
**What does this PR do?**  
This PR improves reporting upon moving a message by editing it's topic.

Partial fix for #1172 

**Tested?** 
- [x] Manually

**Commit flow** 

**Notes & Questions** 

**Interactions** 

**Visual changes** 

https://user-images.githubusercontent.com/53873549/159926734-3f04ffb8-c35f-4836-8d76-ae7401e6fca6.mp4

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -877,6 +877,16 @@ class TestModel:
             case(
                 {
                     "message_id": 1,
+                    "propagate_mode": "change_later",
+                    "content": ":smile:",
+                    "topic": "new_terminal",
+                },
+                "old_terminal",
+                "You changed some messages' topic from #stream > old_terminal to #stream > new_terminal.",
+            ),
+            case(
+                {
+                    "message_id": 1,
                     "propagate_mode": "change_one",
                     "content": "Hey!",
                     "topic": "grett",
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -834,9 +834,9 @@ class TestModel:
         ],
     )
     @pytest.mark.parametrize(
-        "req, old_topic, footer_updated",
+        "req, old_topic, expected_report_success",
         [
-            (
+            case(
                 {
                     "message_id": 1,
                     "propagate_mode": "change_one",
@@ -844,23 +844,27 @@ class TestModel:
                     "topic": "Some topic",
                 },
                 "Some topic",
-                False,
+                None,  # None as footer is not updated.
             ),
-            (
+            case(
                 {
                     "message_id": 1,
                     "propagate_mode": "change_one",
                     "topic": "Topic change",
                 },
                 "Old topic",
-                True,
+                "You changed one message's topic from #stream > Old topic to #stream > Topic change.",
             ),
-            (
-                {"message_id": 1, "propagate_mode": "change_all", "topic": "Old topic"},
+            case(
+                {
+                    "message_id": 1,
+                    "propagate_mode": "change_all",
+                    "topic": "Old topic",
+                },
                 "Old topic",
-                False,
+                None,  # None as footer is not updated.
             ),
-            (
+            case(
                 {
                     "message_id": 1,
                     "propagate_mode": "change_later",
@@ -868,9 +872,9 @@ class TestModel:
                     "topic": "terminal",
                 },
                 "terminal",
-                False,
+                None,  # None as footer is not updated.
             ),
-            (
+            case(
                 {
                     "message_id": 1,
                     "propagate_mode": "change_one",
@@ -878,9 +882,9 @@ class TestModel:
                     "topic": "grett",
                 },
                 "greet",
-                True,
+                "You changed one message's topic from #stream > greet to #stream > grett.",
             ),
-            (
+            case(
                 {
                     "message_id": 1,
                     "propagate_mode": "change_all",
@@ -888,24 +892,33 @@ class TestModel:
                     "topic": "party",
                 },
                 "lets_party",
-                True,
+                "You changed all messages' topic from #stream > lets_party to #stream > party.",
             ),
         ],
     )
     def test_update_stream_message(
-        self, mocker, model, response, return_value, req, old_topic, footer_updated
+        self,
+        mocker,
+        model,
+        response,
+        return_value,
+        req,
+        old_topic,
+        expected_report_success,
+        old_stream_name="stream",
     ):
         self.client.update_message = mocker.Mock(return_value=response)
         model.index["messages"][req["message_id"]]["subject"] = old_topic
-
+        model.index["messages"][req["message_id"]][
+            "display_recipient"
+        ] = old_stream_name
         result = model.update_stream_message(**req)
-
         self.client.update_message.assert_called_once_with(req)
         assert result == return_value
         self.display_error_if_present.assert_called_once_with(response, self.controller)
         report_success = model.controller.report_success
-        if result and footer_updated:
-            report_success.assert_called_once_with("You changed a message's topic.")
+        if result and expected_report_success is not None:
+            report_success.assert_called_once_with(expected_report_success)
         else:
             report_success.assert_not_called()
 
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -437,23 +437,23 @@ class Controller:
         self.is_typing_notification_in_progress = False
         self.view.set_footer_text()
 
-    def report_error(self, text: str) -> None:
+    def report_error(self, text: str, duration: int = 3) -> None:
         """
         Helper to show an error message in footer
         """
-        self.view.set_footer_text(text, "task:error", 3)
+        self.view.set_footer_text(text, "task:error", duration)
 
-    def report_success(self, text: str) -> None:
+    def report_success(self, text: str, duration: int = 3) -> None:
         """
         Helper to show a success message in footer
         """
-        self.view.set_footer_text(text, "task:success", 3)
+        self.view.set_footer_text(text, "task:success", duration)
 
-    def report_warning(self, text: str) -> None:
+    def report_warning(self, text: str, duration: int = 3) -> None:
         """
         Helper to show a warning message in footer
         """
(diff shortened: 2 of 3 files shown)


zulipterminal/model.py
```diff
@@ -567,8 +567,14 @@ def update_stream_message(
...
             old_topic = self.index["messages"][message_id].get("subject", None)
             new_topic = request["topic"]
+            stream_name = self.index["messages"][message_id].get(
+                "display_recipient", None
+            )
+            no_of_messages = len(self.index["edited_messages"])
             if old_topic != new_topic:
-                self.controller.report_success("You changed a message's topic.")
+                self.controller.report_success(
+                    f"You changed {no_of_messages} message's topic from #{stream_name} > {old_topic} to #{stream_name} > {new_topic}."
```
neiljp, line 576: I'm not sure if we have a standard way to do this using `report_success`, but this may be clearer if we can provide a highlight style to group the stream+topic in each case.

This may require changes to existing code, so that could be a follow-up.

zulipterminal/model.py
```diff
@@ -567,8 +567,14 @@ def update_stream_message(
         if response["result"] == "success":
             old_topic = self.index["messages"][message_id].get("subject", None)
             new_topic = request["topic"]
+            stream_name = self.index["messages"][message_id].get(
+                "display_recipient", None
+            )
+            no_of_messages = len(self.index["edited_messages"])
```
neiljp, line 573: This is all the edited messages we know about, so this may work the first time, with no known edited messages, but I expect likely will grow each time and so report the wrong result.

Try testing this manually, and then you can add a test case for this and check it fails, before fixing.

zulipterminal/model.py
```diff
@@ -567,8 +567,14 @@ def update_stream_message(
         if response["result"] == "success":
             old_topic = self.index["messages"][message_id].get("subject", None)
             new_topic = request["topic"]
+            stream_name = self.index["messages"][message_id].get(
+                "display_recipient", None
+            )
```
neiljp, line 572: Let's simplify using a well-named variable for the common code here.

neiljp (commented):
@srdeotarse Thanks for looking at this, including giving the tests a go :+1:

This doesn't appear to pass tests (CI), which is not what you have in the PR description.

Since the re-indexing occurs outside the current method where the message occurs, it's possible we may need to do that elsewhere.

zulipterminal/model.py
srdeotarse: After fixing the format for reporting message upon moving topics, we can look into highlighting as well.

zulipterminal/model.py
srdeotarse: I got your point. Therefore, I have used different method -
1. Getting all stream messages
2. Filtering messages based on topic where message is edited by user.
3. Compare message_id of edited message and topic messages to get recent_edited_messages.

zulipterminal/model.py
srdeotarse: get_message = self.index["messages'] for common code.

srdeotarse:
Rebased with current main branch and updated mypy. Still showing error below @neiljp -
```
Successfully installed autoflake-1.4 autopep8-1.6.0 flake8-4.0.1 isort-5.10.1 jedi-0.18.1 lxml-stubs-0.4.0 mypy-0.942 parso-0.8.3 pudb-2022.1.1 pycodestyle-2.8.0 pyflakes-2.4.0 pytz-2022.1 snakeviz-2.1.1 types-docutils-0.18.0 types-pygments-2.9.17 types-setuptools-57.4.11 zulip-term-0.6.0+git
(zt_venv) srdeotarse@LAPTOP-VC2MC73M:~/zulip-terminal$ make check
================================================================================
Import order (isort)
--------------------------------------------------------------------------------
All files have imports sorted according to rules.
================================================================================
Type consistency (mypy)
--------------------------------------------------------------------------------
Running mypy for `zulipterminal`.
zulipterminal/model.py:584: error: Invalid index type "Optional[int]" for "Dict[int, Set[int]]"; expected type "int"
Found 1 error in 1 file (checked 37 source files)
Running mypy for `tests`.
Success: no issues found in 15 source files
================================================================================
LINTING FAILED
make: *** [makefile:22: lint] Error 1
```

srdeotarse:
Rebased with current main branch and updated mypy. Still showing error below @neiljp -

```
Successfully installed autoflake-1.4 autopep8-1.6.0 flake8-4.0.1 isort-5.10.1 jedi-0.18.1 lxml-stubs-0.4.0 mypy-0.942 parso-0.8.3 pudb-2022.1.1 pycodestyle-2.8.0 pyflakes-2.4.0 pytz-2022.1 snakeviz-2.1.1 types-docutils-0.18.0 types-pygments-2.9.17 types-setuptools-57.4.11 zulip-term-0.6.0+git
(zt_venv) srdeotarse@LAPTOP-VC2MC73M:~/zulip-terminal$ make check
================================================================================
Import order (isort)
--------------------------------------------------------------------------------
All files have imports sorted according to rules.
================================================================================
Type consistency (mypy)
--------------------------------------------------------------------------------
Running mypy for `zulipterminal`.
zulipterminal/model.py:584: error: Invalid index type "Optional[int]" for "Dict[int, Set[int]]"; expected type "int"
Found 1 error in 1 file (checked 37 source files)
Running mypy for `tests`.
Success: no issues found in 15 source files
================================================================================
LINTING FAILED
make: *** [makefile:22: lint] Error 1
```

neiljp:
@srdeotarse Was there a reason you closed/reopened this? Generally there's no need to do so.

zulipterminal/model.py
```diff
@@ -576,10 +576,28 @@ def update_stream_message(
...
+            if propagate_mode == "change_later":
+                recent_moved_msgs = "some"
+            if propagate_mode == "change_one":
+                recent_moved_msgs = "one"
+
             if old_topic != new_topic:
-                self.controller.report_success("You changed a message's topic.")
+                if recent_moved_msgs == "one":
+                    self.controller.report_success(
+                        f"You changed {recent_moved_msgs} message's topic from #{stream_name} > {old_topic} to #{stream_name} > {new_topic}."
```
neiljp, line 595: This is identical except for the apostrophe placement? If so, it'll be simpler to integrate the difference into the variable, and avoid the need for a conditional here.

tests/model/test_model.py
```diff
@@ -888,24 +907,54 @@ def test_update_private_message(
...
+        model.index["messages"][req["message_id"]]["display_recipient"] = stream_name
+        req["propagate_mode"] = propagate_mode
+
+        recent_moved_msgs = ""
+        if propagate_mode == "change_all":
+            recent_moved_msgs = "all"
+        if propagate_mode == "change_later":
+            recent_moved_msgs = "some"
+        if propagate_mode == "change_one":
+            recent_moved_msgs = "one"
```
neiljp, line 942: It's better to put the explicit text into the test cases above, than have conditional code here - particularly if it is exactly the same as the code in the application.

tests/model/test_model.py
```diff
@@ -888,24 +907,54 @@ def test_update_private_message(
...
 
         result = model.update_stream_message(**req)
-
         self.client.update_message.assert_called_once_with(req)
         assert result == return_value
         self.display_error_if_present.assert_called_once_with(response, self.controller)
         report_success = model.controller.report_success
         if result and footer_updated:
-            report_success.assert_called_once_with("You changed a message's topic.")
+            if recent_moved_msgs == "one":
```
neiljp, line 950: By just using a string we can assert on an 'expected string' provided in each parametrize 'case', avoiding the need for extra conditionals here.

tests/model/test_model.py
```diff
@@ -888,24 +907,54 @@ def test_update_private_message(
                     "topic": "party",
                 },
                 "lets_party",
+                "party",
+                "stream",
+                "change_all",
                 True,
             ),
         ],
```
neiljp, line 915: Given the longer test here now, it would be useful to also add inline test ids.

srdeotarse:
@neiljp I was trying to rebase with main to solve mypy error for Optional[int] and accidentally deleted my local branch.

neiljp (commented):

Other than the comments inline, I think we could simplify the code for readability, though it is functional as it is :)

One aspect this brings up is the short time given until the message disappears in the footer. For a simple solution, we could tune this manually with an optional parameter to pass through, which might be warranted as a preparatory commit before the one you have.

I've thought previously we could try to handle this automatically based on word length and reading speed, with some minimum, but that may need more tuning than a simple prep commit.

tests/model/test_model.py
neiljp: See my followup comment in the next review.

zulipterminal/model.py
neiljp: This was what I referred to by simplifying the code in the next review.

neiljp (commented):
@srdeotarse I've mainly focused on the tests here, since once they're well-defined then refactoring the code to give the same behavior should be enforced - and good test-writing technique is a useful skill.

The first commit is good to have separate, but as per my inline comment I'd suggest you keep it very simple to simply allow passing a custom duration in the main commit, with a default of the current 3 as now. The version with a formula is worthy of it's own PR since the implementation is worthy of a separate discussion and this can be merged in the meantime.

zulipterminal/core.py
```diff
@@ -447,7 +447,9 @@ def report_success(self, text: str) -> None:
         """
         Helper to show a success message in footer
         """
-        self.view.set_footer_text(text, "task:success", 3)
+        self.view.set_footer_text(
+            text, "task:success", float(len(text.split()) * 6 / 13)
+        )  # Average reading speed = 130 wpm
```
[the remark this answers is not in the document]
neiljp: We're discussing this in the stream now :+1:

neiljp:
@srdeotarse Thanks for the contribution 👍 I made a few minor textual adjustments which you might be able to identify manually, or find using `git diff` or `git range-diff`, I but merged otherwise as-is :tada:

Issue #1155, opened 2022-02-28
Improvement of presence markers UI


Ezio-Sarthak:
After the successful merge of #987, we support a simple UI for presence markers. Here are some thoughts for some follow-ups that'd need discussion and implement promptly:

- Show bot markers (this was also suggested in role-markers PR)
- Handle "unavailable" visibility and explicit changes (are they events? See #988)
- Only update the status rather than the entire box (maybe not that tricky?)
- Show user presence in PM compose?
- Show user presence in user mentions?

Feel free to hop-in for any discussion on [CZO](https://chat.zulip.org/#narrow/stream/206-zulip-terminal)

Tagging with GSoC, as it might be a good project during summer.

neiljp:
Note that the PM compose and user mention cases also could include autocomplete, which is what the web app currently does - it doesn't include user status in rendered mentions.

PR #1184, opened 2022-04-03
[WIP] Add last active to the PM window, and add more features to PMs.

plugyawn:
**What does this PR do?**  
Aims to add a few more features to the PM window, such as adding a last active message. 

**Tested?** 
- [x] Manually

**Commit flow** 
- added the last active message.

**Notes & Questions** 
What more could we add to the PM Window?

**Interactions** 

**Visual changes**

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -216,8 +216,10 @@ class WriteBox(urwid.Pile):
         self.send_next_typing_update = datetime.now()
         try:
             recipient_ID = int(
-                    recipient_info[recipient_info.find("user") + 4 : recipient_info.find("@")]
-                )
+                recipient_info[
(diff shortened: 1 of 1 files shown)


neiljp:
@plugyawn I'd recommend discussing this on the stream before going too far into this.

The equivalent detail we've discussed previously would be a summary of this via an extension of the message list version which shows the user presence markers (which is not in the web app). The web app has that in the autocomplete/typeahead equivalent, though not in the compose box itself during the rest of composing of PMs. That would basically be part of #1155 for the summary data for what you're looking at here.

plugyawn:
Okay! 

The last seen thing was something that was commonplace on messenger applications, and it's a pretty handy indicator to have, so I guessed it's useful to have it here, too. 

Also attaching an image for general reference.

Screenshot 2022-04-04 at 8 50 17 AM

PR #1186, opened 2022-04-04
ui/boxes: Improve stream sorting for stream typeahead in compose box.

srdeotarse:
**What does this PR do?**  
This PR improves stream sorting for stream typeahead in compose box stream results in footer in the following order -

1) current stream (if matching)
2) pinned and unmuted
3) unpinned and unmuted
4) pinned and muted
5) unpinned and muted

CZO - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Improve.20stream.20sorting.20for.20stream.20typeahead.20in.20compose.20box.2E

**Tested?** 
- [x] Manually

Passing all lint tests.

``test_generic_autocomplete_streams`` test to be updated in ``test_boxes.py``

**Commit flow** 

**Notes & Questions** 

**Interactions** 

**Visual changes** 
Stream sorting for stream typeahead before -
image

Stream sorting for stream typeahead after -
image

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -136,9 +136,11 @@ class TestWriteBox:
         is_valid_stream: bool,
         required_typeahead: Optional[str],
         topics: List[str],
+        stream_dict: Dict[int, Dict[str, Any]],
     ) -> None:
         write_box.model.topics_in_stream.return_value = topics
         write_box.model.is_valid_stream.return_value = is_valid_stream
+        write_box.model.stream_dict = stream_dict
         write_box.model.muted_streams = set()
         typeahead_string = write_box.generic_autocomplete(text, state)
 
@@ -555,10 +557,12 @@ class TestWriteBox:
         state: Optional[int],
         footer_text: List[Any],
         text: str,
+        stream_dict: Dict[int, Dict[str, Any]],
     ) -> None:
         write_box.view.set_typeahead_footer = mocker.patch(
             "zulipterminal.ui.View.set_typeahead_footer"
         )
+        write_box.model.stream_dict = stream_dict
         write_box.model.muted_streams = set()
         write_box.generic_autocomplete(text, state)
 
@@ -875,6 +879,34 @@ class TestWriteBox:
                 },
                 {"pinned": ["Secret stream", "Stream 1"], "muted": {99, 2}},
             ),
+            # With 'Stream 1' as current stream.
+            (
+                "#S",
+                {
+                    0: "#**Stream 1**",
+                    1: "#**Secret stream**",
+                    2: "#**Some general stream**",
+                    3: "#**Stream 2**",
+                    4: "#**Web public stream**",
+                },
+                {"current_stream": 1},
+            ),
+            # With 'Stream 1' and 'Secret stream' pinned, 'Secret stream' and 'Stream 2' muted, 'Stream 2' as current stream.
+            (
+                "#Stream",
+                {
+                    0: "#**Stream 2**",
+                    1: "#**Stream 1**",
+                    2: "#**Secret stream**",
+                    3: "#**Some general stream**",
+                    4: "#**Web public stream**",
+                },
+                {
+                    "pinned": ["Secret stream", "Stream 1"],
+                    "muted": {99, 2},
+                    "current_stream": 2,
+                },
+            ),
         ],
     )
     def test_generic_autocomplete_streams(
@@ -893,6 +925,7 @@ class TestWriteBox:
         for stream in streams_to_pin:
             write_box.view.unpinned_streams.remove(stream)
         write_box.view.pinned_streams = streams_to_pin
+        write_box.stream_id = stream_categories.get("current_stream", None)
         write_box.model.stream_dict = stream_dict
         write_box.model.muted_streams = stream_categories.get("muted", set())
         for state, required_typeahead in state_and_required_typeahead.items():
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -829,7 +829,7 @@ class TestWriteBox:
                     3: "#**Web public stream**",
                     4: "#**Secret stream**",
                 },
-                {"muted": {99}},
+                {"muted": ["Secret stream"]},
             ),
             # With 'Stream 1' and 'Secret stream' muted.
             (
@@ -841,7 +841,7 @@ class TestWriteBox:
                     3: "#**Stream 1**",
                     4: "#**Secret stream**",
                 },
-                {"muted": {99, 1}},
+                {"muted": ["Secret stream", "Stream 1"]},
             ),
             # With 'Stream 1' and 'Secret stream' pinned, 'Secret stream' muted.
             (
@@ -853,7 +853,7 @@ class TestWriteBox:
                     3: "#**Some general stream**",
                     4: "#**Web public stream**",
                 },
-                {"pinned": ["Secret stream", "Stream 1"], "muted": {99}},
+                {"pinned": ["Secret stream", "Stream 1"], "muted": ["Secret stream"]},
             ),
             # With 'Stream 1' and 'Secret stream' pinned, 'Some general stream' and 'Stream 2' muted.
             (
@@ -865,7 +865,10 @@ class TestWriteBox:
                     3: "#**Stream 2**",
                     4: "#**Some general stream**",
                 },
-                {"pinned": ["Secret stream", "Stream 1"], "muted": {1000, 2}},
+                {
+                    "pinned": ["Secret stream", "Stream 1"],
+                    "muted": ["Some general stream", "Stream 2"],
+                },
             ),
             # With 'Stream 1' and 'Secret stream' pinned, 'Secret stream' and 'Stream 2' muted.
             (
@@ -877,7 +880,10 @@ class TestWriteBox:
                     3: "#**Web public stream**",
                     4: "#**Stream 2**",
                 },
-                {"pinned": ["Secret stream", "Stream 1"], "muted": {99, 2}},
+                {
+                    "pinned": ["Secret stream", "Stream 1"],
+                    "muted": ["Secret stream", "Stream 2"],
+                },
             ),
             # With 'Stream 1' as current stream.
             (
@@ -903,7 +909,7 @@ class TestWriteBox:
                 },
                 {
                     "pinned": ["Secret stream", "Stream 1"],
-                    "muted": {99, 2},
+                    "muted": ["Secret stream", "Stream 2"],
                     "current_stream": 2,
                 },
             ),
@@ -927,7 +933,11 @@ class TestWriteBox:
         write_box.view.pinned_streams = streams_to_pin
         write_box.stream_id = stream_categories.get("current_stream", None)
         write_box.model.stream_dict = stream_dict
-        write_box.model.muted_streams = stream_categories.get("muted", set())
+        write_box.model.muted_streams = set(
+            stream["stream_id"]
+            for stream in stream_dict.values()
+            if stream["name"] in stream_categories.get("muted", set())
+        )
         for state, required_typeahead in state_and_required_typeahead.items():
             typeahead_string = write_box.generic_autocomplete(text, state)
             assert typeahead_string == required_typeahead
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -139,7 +139,7 @@ class TestWriteBox:
     ) -> None:
         write_box.model.topics_in_stream.return_value = topics
         write_box.model.is_valid_stream.return_value = is_valid_stream
-
+        write_box.model.muted_streams = set()
         typeahead_string = write_box.generic_autocomplete(text, state)
 
         assert typeahead_string == required_typeahead
@@ -559,6 +559,7 @@ class TestWriteBox:
         write_box.view.set_typeahead_footer = mocker.patch(
             "zulipterminal.ui.View.set_typeahead_footer"
         )
+        write_box.model.muted_streams = set()
         write_box.generic_autocomplete(text, state)
 
         write_box.view.set_typeahead_footer.assert_called_once_with(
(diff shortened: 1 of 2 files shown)


neiljp:
Just to note here that as per discussion in the stream, the intention is to leave the search order unaltered during filtering in the left panel (as in the web app zulip/zulip#20618) and adjust the order for other cases such as the stream-box and stream/topic mentions in the main compose box.

srdeotarse:
@neiljp as per your note, I have changed the PR topic and updated the code to improve stream sorting in compose box stream typeahead. Recent commit only handles stream typeahead sorting in compose box message area.

srdeotarse:
Reopening the PR

srdeotarse:
@zulipbot add "PR needs review'.

srdeotarse:

tests/ui_tools/test_boxes.py
```diff
@@ -819,12 +837,23 @@ def test_generic_autocomplete_streams(
...
         to_pin: List[str],
+        to_mute: Set[int],
     ) -> None:
         streams_to_pin = [{"name": stream_name} for stream_name in to_pin]
         for stream in streams_to_pin:
             write_box.view.unpinned_streams.remove(stream)
         write_box.view.pinned_streams = streams_to_pin
+        write_box.model.muted_streams = to_mute
+        write_box.stream_id = 1
+        write_box.model.stream_dict = {
```
neiljp, line 850: There is a stream_dict fixture that should be suitable for this.

tests/ui_tools/test_boxes.py
```diff
@@ -507,41 +507,23 @@ def test_generic_autocomplete_no_prefix(
...
+            #     "#**Stream",
+            #     0,
+            #     ["Stream 1", "Stream 2", "Secret stream", "Some general stream"],
+            # ),  # Optional 2-stars
+            # (
+            #     "#Stream",
+            #     None,
+            #     ["Stream 1", "Stream 2", "Secret stream", "Some general stream"],
+            # ),
+            # ("#NoMatch", None, []),
```
neiljp, line 526: You didn't mention test_generic_autocomplete_set_footer in the stream either, but this is (partially) commented?

Did you miss adding the web public stream?

tests/ui_tools/test_boxes.py
```diff
@@ -819,12 +837,23 @@ def test_generic_autocomplete_streams(
         text: str,
         state: Optional[int],
         required_typeahead: Optional[str],
+        stream_id: int,
```
neiljp, line 840: I'm maybe missing something, but you're not using this parameter?

neiljp (commented):

The code looks broadly correct, but could be clearer.

I'm a little confused about some of the tests, particularly since the commented out ones appear to be not the ones you named in the stream?

tests/ui_tools/test_boxes.py
srdeotarse: I have added web public stream as well in latest commit.

tests/ui_tools/test_boxes.py
```diff
@@ -743,88 +725,360 @@ def test_generic_autocomplete_user_mentions(
...
+            ("@_#Stream", 0, "@_#**Stream 1**", [], {}),
+            (":#Stream", 0, ":#**Stream 1**", [], {}),
+            ("##Stream", 0, "##**Stream 1**", [], {}),
+            ("##*Stream", 0, None, [], {}),  # NOTE: Optional single star fails
+            ("##**Stream", 0, "##**Stream 1**", [], {}),  # Optional 2-stars
             # With 'Secret stream' pinned.
             (
                 "#Stream",
-                0,
+                1,
```
neiljp, line 794: Why are these existing values changing, with no mutes?

tests/ui_tools/test_boxes.py
neiljp: I was referring to this commented out code, not the other test cases, which I think you added and were referring to just now?

tests/ui_tools/test_boxes.py
```diff
@@ -743,88 +725,360 @@ def test_generic_autocomplete_user_mentions(
         )
 
     @pytest.mark.parametrize(
-        "text, state, required_typeahead, to_pin",
+        "text, state, required_typeahead, to_pin, to_mute",
```
neiljp, line 728: To avoid this adding lots of empty data and having lines spill-over, as well as improving the self-documenting of the test cases, I'm wondering if this would be clearer as a dict which might be:
```
{}  # no pins or mutes
{"pinned": [...]}  # equivalent to to_pin=[...]
{"muted": {...}}  # equivalent to to_mute={...}
```

If so, this logic could be a quick refactor commit before this with data and logic adjusted for just to_pin there, and then extended in this commit.

neiljp (commented):
@srdeotarse Here's another pass.

Assuming no streams are muted, the existing test cases should pass, and shouldn't need altering - unless there is something I'm missing? You might find it useful to split the muting test cases into a separate later commit for now, so you can check that the existing ones pass in one commit and work on others in the next (using rebase or checkout to jump between commits). We may combine the commits later, but that's easier than splitting them apart and wondering why lots of commits fail :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -636,10 +636,31 @@ def autocomplete_streams(
...
+            stream_name
+            for stream_name in matched_streams
+            if stream_name in muted_stream_names
+        ]
+        for stream_name in muted_stream_names:
+            matched_streams.remove(stream_name)
+            if stream_name in [stream["name"] for stream in self.view.pinned_streams]:
+                muted_stream_names.remove(stream_name)
+                muted_stream_names.insert(0, stream_name)
+        matched_streams.extend(muted_stream_names)
```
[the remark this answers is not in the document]
srdeotarse: - When I was writing more test cases, I found that ``muted_stream_ids`` are received in ascending order and not the order of ``left_streams_bar`` (pinned_muted_id, unpinned_muted_id). 
- Therefore had to arrange ``muted_stream_names`` with same order as ``matched_streams`` and then remove ``pinned_muted_streams``, add to start of ``muted_stream_names``.
-  At last, extend it to ``matched_streams`` after removing common streams.

zulipterminal/ui_tools/boxes.py
```diff
@@ -636,10 +636,34 @@ def autocomplete_streams(
...
+        ]
+
+        for muted_stream in muted_stream_names:
+            matched_streams.remove(muted_stream)
+            if muted_stream in [stream["name"] for stream in self.view.pinned_streams]:
+                muted_stream_names.remove(muted_stream)
+                muted_stream_names.insert(0, muted_stream)
+        matched_streams.extend(muted_stream_names)
+
+        current_stream_name = self.model.stream_dict[self.stream_id]["name"]
```
neiljp, line 660: This fails for private messages, and actually causes a crash. There is an outstanding question of what should be the stream to use here - narrow? compose? recent? neither?

neiljp (commented):
@srdeotarse Thanks for giving the test refactor a try - I left some inline code suggestions that may explain what I had in mind better.

We really need to address the no-stream case, which explains some of the test changes I asked about earlier I believe? Let's discuss the best approach in the topic.

tests/ui_tools/test_boxes.py
```diff
@@ -743,74 +743,105 @@ def test_generic_autocomplete_user_mentions(
...
-            ("#Stream", 0, "#**Stream 1**", ["Secret stream", "Stream 1"]),
-            ("#Stream", 1, "#**Secret stream**", ["Secret stream", "Stream 1"]),
-            ("#Stream", 2, "#**Stream 2**", ["Secret stream", "Stream 1"]),
-            ("#Stream", 3, "#**Some general stream**", ["Secret stream", "Stream 1"]),
+            (
+                "#Stream",
+                0,
+                "#**Stream 1**",
+                {"pinned": ["Secret stream", "Stream 1"]},
+            ),  # 1st-word startswith match (pinned).
```
neiljp, line 825: You seem to have added these comments? Without them the test cases fit on one line still?

tests/ui_tools/test_boxes.py
```diff
@@ -743,74 +743,105 @@ def test_generic_autocomplete_user_mentions(
...
-                ["Secret stream"],
+                {"pinned": ["Secret stream"]},
             ),  # 1st-word startswith match (unpinned).
             (
                 "#Stream",
                 2,
                 "#**Stream 2**",
-                ["Secret stream"],
+                {"pinned": ["Secret stream"]},
             ),  # 1st-word startswith match (unpinned).
```
neiljp, line 806: I think it may be clearer if we remove these comments in a separate commit (maybe when adding the web public cases?).

However, see my general review point too.

neiljp (commented):
@srdeotarse I've mainly looked at the first part of this for now.

Your added tests may well be good, but they're difficult to review since they're not in order like others (eg. 0,1,2,3,4 vs 0,1,4,2,3 or similar).

While this is a relatively small thing to put a lot of effort into, one thought to simplify the tests further would be to describe sequences of matches for a range of indices, which would reduce the need to repeat the search text and stream properties. I'm wary of putting extra logic into the test body and having multiple test cases in one, but this could be more expressive - one alternative may be using separate list comprehensions for each set of test cases and adding them together.

neiljp (commented):
@srdeotarse The test refactors clean things up very well - and maybe other similar ones would benefit?

I'd like to merge the early commits, so if you split out the web-public change and address any comments related to those commits, we can get those in first. At least one of the later commits seem to be failing, which may be related to my thought that the output is reversed?

One benefit of the muted/pinned groups are that they're sort of self-documenting, so some of the comments are redundant. That said, the muted streams are by id right now - maybe if we express the tests by name instead then this would be even clearer?

Generally if a single test case has a dedicated comment, then it's worth using an id string instead, since test passes (and failures!) are then much more descriptive.

zulipterminal/ui_tools/boxes.py
```diff
@@ -636,10 +636,29 @@ def autocomplete_streams(
...
+
+        for muted_stream in muted_stream_names:
+            matched_streams.remove(muted_stream)
+            if muted_stream in [stream["name"] for stream in self.view.pinned_streams]:
+                muted_stream_names.remove(muted_stream)
+                muted_stream_names.insert(0, muted_stream)
+        matched_streams.extend(muted_stream_names)
+
+        matched_stream_typeaheads = format_string(matched_streams, "#**{}**")
+        return matched_stream_typeaheads, matched_streams
```
Rohitth007, line 669: I haven't followed this PR a lot but I feel that all of this code should belong inside `match_stream`. Is there any reason for not doing so? We can send in the model variables as arguments into the function and process everything there.

Rohitth007: @neiljp On a design level, I wanted to ask why, in the model, we treat `muted_streams` different from `(un)pinned_streams`, i.e., different data types.

srdeotarse: And about data type of ``muted_streams``, I agree with using ``stream_names`` instead of ``stream_ids`` like we have implemented for ``(un)pinned_streams``

tests/ui_tools/test_boxes.py
```diff
@@ -814,6 +814,66 @@ def test_generic_autocomplete_user_mentions(
                 },
                 {"pinned": ["Secret stream", "Stream 1"]},
             ),
+            # With 'Secret stream' muted.
+            (
+                "#Stream",
```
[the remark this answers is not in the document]
srdeotarse: 3 tests started failing in ``test_helper.py`` after I rebased my current branch with ``upstream/main``. Test cases are as follows -
* ``test_download_media``
* ``test_open_media[0-error0]``
* ``test_open_media[1-error1]``

zulipterminal/ui_tools/boxes.py
Rohitth007: Oh I see, but a boolean like `prioritize_unmuted` might help? Since `match_stream` also does some sorting apart from matching, it's better to place all sorting code together.
* If False, it'll return pinned matches then unpinned matches.
* If True, it'll return pinned unmuted matches, unpinned unmuted matches, pinned muted matches then unpinned muted matches.

tests/ui_tools/test_boxes.py
```diff
@@ -743,57 +743,72 @@ def test_generic_autocomplete_user_mentions(
         )
```
neiljp, line 748: In the commit introducing stream_categories, note that while details can be useful, in the commit text it's more useful to write what you can't easily include in the commit (code) itself, such as why the change is useful/important.

Some of that why might be noting how it might be extended, but until the extra parameters are added, it's not important to refer to them by name - otherwise you couple this commit text to how it *will* be implemented, and either this commit text needs updating later (though it might be merged first!), or else the later code 'must' be named appropriately later (which one cannot always guarantee). Basically, it can be useful to describe what else it might be applied to, or perhaps an example, but otherwise focus the commit text on *this* commit.

I may have given an impression that something wasn't in the commit text which led to this; I believe I was referring to how the commit text didn't describe what this commit was doing (re switching to stream_categories).

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
neiljp: I haven't seen these failures locally after rebasing, only the new tests you added here. Do you still get the errors you saw? (Was that locally or CI, ie. GitHub actions?)

To clarify, it seems that it's one test from each group, perhaps, or at least not an entire case that's failing. While the assert-in-a-loop does detect the error, as I mentioned previously it may be more useful to show the failing pairs all together, which might be straightforward if you eg. build a dict and compare it to that expected?

These new tests apply to the new code, so this should go with the previous commit that changed/added the functionality.

tests/ui_tools/test_boxes.py
```diff
@@ -815,6 +815,66 @@ def test_generic_autocomplete_user_mentions(
...
+                "#Stream",
+                {
+                    0: "#**Stream 1**",
+                    1: "#**Some general stream**",
+                    2: "#**Web public stream**",
+                    3: "#**Secret stream**",
+                    4: "#**Stream 2**",
+                },
+                {"pinned": ["Secret stream", "Stream 1"], "muted": {99, 2}},
+            ),
```
neiljp, line 915: In this commit where you add the test cases, you're not connecting the muted elements to the muted_streams data.

neiljp (commented):
@srdeotarse I focused on reviewing the first 5 commits, since the first ones are close to ready and hopefully the muting one will be when the test situation is clarified :)

zulipterminal/ui_tools/boxes.py
neiljp: @Rohitth007 I agree there are multiple ways of writing the code, and we may even update the stream order later - or add options - but for now I'm happy for this to be focused on the functionality and ensuring this behaves as expected by using the new tests (and test structure). Once we have tests, we can refactor more safely. I think @srdeotarse already rewrote the code, and we still may need to adjust the algorithm.

neiljp: Re muted_streams vs (un)pinned_streams, these are data-structures used for different purposes right now. The former is simply to know if a stream is muted, while the latter is designed for the UI (historically).

Regarding the tests, I would favor having names in the test cases to make them more readable, independently of the data in the model. That is eg. `{"pinned": {"Stream 1"}, "muted": {"Secret Stream"}}` is more readable than the latter being an int.

Rohitth007: The reason for having it inside `match_stream` is it would simplify the logic by a lot. Here we are throwing away the matched typeaheads, resorting an already sorted list given by `match_streams`  and then recreating new matched typeaheads. If this is done inside `match_stream` then it would barely take 5-6 lines and it wouldn't affect any of the tests.

Another reason is we would barely have to do any work to extend this feature to the stream panel.

neiljp:
@srdeotarse I've just pushed the first 3 commits here, with some changes to the commit text reasoning as those ending in https://github.com/zulip/zulip-terminal/commit/898bb2e3722567f6a4f4bcc0730f2ad77cfa9d9f. Rebasing should clear any conflict.

I've not looked at more at this point, as the corresponding tests for the first code change fail, so there is some mismatch between test expectations and code behavior.

tests/ui_tools/test_boxes.py
neiljp: Yes, it's there now, but it had disappeared in the previous iteration it seemed.

neiljp (commented):
@srdeotarse We discussed the changes in the stream, so I'm merging shortly. I had some outstanding comments so I'm posting this as a review.

neiljp:
@srdeotarse As per our discussion in the stream, I squashed your feature code+test commits together, and added some adjustments as per the extra commits, resulting in the branch with separated commits at https://github.com/neiljp/zulip-terminal/tree/review-original-1186-6-with-adjust

I did a final tidy of the commit text, to emphasize the action associated with the feature commits, ie. what changed/why, which you can see by comparison with your original commits in the link above or that you may have locally. This kind of detail can be helpful not just at this stage for merging and reading later, but also for reviewers wanting to know "What's in this commit?" and whether it matches the description.

Thanks for pushing this to completion - as I say, merging now! 🎉 
(with those fixup commits squashed into yours)

PR #1187, opened 2022-04-05
Showing Bot Markers.

plugyawn:
**What does this PR do?**  
Adds bot markers to bots.

Partial fix for #1155. 

**Tested?** 
- [x] Manually

**Commit flow** 
- first commit adds BOT_MARKER to symbols.py. Currently set to robot face, U+1F916 (🤖). 
- second commit adds necessary changes in ui_mappings.py to associate BOT_MARKER with the "bot" status, as compared to the "inactive" status they were assigned earlier.
- third commit checks if it is indeed a bot and associates the marker to it. 

**Notes & Questions** 
- unsure about how to fix the black background against the robot emoji. I'm guessing it's because it's not a traditional character? I couldn't really choose a better character.

**Visual changes** 

Screenshot 2022-04-06 at 12 00 38 AM


neiljp:
@plugyawn I mentioned about the symbol in the stream. Otherwise please use the standard commit text style for zulip-terminal as indicated in the README (and see gitlint), and adjust the commit flow ready for merging.

This looks straightforward; I'm not sure if we explicitly need tests for this.

In a bot-related way, you might consider how the user popup can be improved for the case of bots, as a follow-up.

plugyawn:
On it! I'll update the commit text and let you know asap.

neiljp:
@plugyawn I'm mainly waiting for a commit reflow here, which may be simple, but just checked functionality again as it stands as this seems a quick PR. However, it seems that the symbol you're using is styled to be black on a white background somehow, and that is bleeding into the user search box names too. That doesn't seem to be intentional, so I wonder if it's part of the symbol itself somehow?

plugyawn:
@neiljp updated the commit flow.

zulipterminal/config/symbols.py
```diff
@@ -19,5 +19,6 @@
 STATUS_IDLE = "◒"
 STATUS_OFFLINE = "○"
 STATUS_INACTIVE = "•"
+BOT_MARKER = "🤖"
```
neiljp: This is not a pawn unicode character, in this first commit.

zulipterminal/config/symbols.py
```diff
@@ -19,6 +19,6 @@
 STATUS_IDLE = "◒"
 STATUS_OFFLINE = "○"
 STATUS_INACTIVE = "•"
-BOT_MARKER = "🤖"
+BOT_MARKER = "♟"
```
neiljp, line 22: Was this supposed to be in the first commit?

zulipterminal/config/ui_mappings.py
```diff
@@ -4,14 +4,14 @@
...
 from zulipterminal.config.symbols import (
+    BOT_MARKER,
     STATUS_ACTIVE,
     STATUS_IDLE,
     STATUS_INACTIVE,
     STATUS_OFFLINE,
     STREAM_MARKER_PRIVATE,
     STREAM_MARKER_PUBLIC,
     STREAM_MARKER_WEB_PUBLIC,
-    BOT_MARKER
```
neiljp: This change looks like a reformat that belongs in another commit?

zulipterminal/config/ui_mappings.py
```diff
@@ -28,7 +28,7 @@
     "idle": STATUS_IDLE,
     "offline": STATUS_OFFLINE,
     "inactive": STATUS_INACTIVE,
-    "bot": BOT_MARKER
+    "bot": BOT_MARKER,
```
neiljp, line 31: Same here with the reformat. This should go where the original code was.

neiljp (commented):
@plugyawn This is tidier, but as per my inline notes, please keep the tidying (and intentional changes) that go with each set of changes in the associated commit. Reading your changes through in the github interface, as if reviewing your own code, will likely show obvious issues like the pawn vs robot unicode codepoint.

We don't always end up needing the commits to be so small, but keeping logical changes separate makes it much easier to read.

neiljp:
Also note that the formatting being in the wrong commit would be picked up if you ran `tools/check-branch`.

neiljp (commented):
@plugyawn This looks much tidier :)

I think the first two commits may be squashable together later, as they're very small.

This appears to work fine on one machine, but I tested on my new machine (python3.9, debian) and it doesn't show the symbols, but that may be a local issue.

My remaining concerns are if we have other locations where the status is otherwise used which haven't been updated. This is evident since like I mentioned in the last group chat, many locations directly access this data and so it's rather fragile.

One thing we might do to address this is to define the types of the elements of user_dict, and mypy may then be able to work with that, particularly if we use Literal types for the status.

neiljp:
See other feedback in the stream. I resolved my issue with it not working at all.

srdeotarse:
I have reviewed the PR and have posted some questions in the stream.

mounilKshah (commented):
Upon running the PR, it worked without any glitches/crashes. It didn't work for _Notification Bot_ though but I think this has been discussed in the stream already. The changes look good to me.

neiljp:
@plugyawn This PR would be good to close out, now you've got feedback too?

zulipterminal/model.py
```diff
@@ -949,6 +949,15 @@ def get_all_users(self) -> List[Dict[str, Any]]:
...
                 "status": status,
             }
+
+            if user["is_bot"]:
+                self.user_dict[user["email"]] = {
+                    "full_name": user["full_name"],
+                    "email": email,
+                    "user_id": user["user_id"],
+                    "status": "bot",
+                }
```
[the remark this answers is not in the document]
neiljp: Did you understand my suggestion? You've not replied or updated for it.

neiljp (commented):
@plugyawn The system bots appear to work now (eg. Notification bot), though I assume you'll check yourself again since you left a comment :)

As you said in the stream, the styles are still strange, which is likely related to the way styles are picked for each user status - like the green/yellow/white. These are connected somewhere, so we likely want to extend the styles to cover the bot case you added.

zulipterminal/model.py
```diff
@@ -1012,12 +1021,18 @@ def get_all_users(self) -> List[Dict[str, Any]]:
             for properties in self.user_dict.values()
             if properties["status"] == "inactive"
         ]
+        bot = [
+            properties
+            for properties in self.user_dict.values()
+            if properties["status"] == "bot"
+        ]
```
neiljp, line 1028: This is different behavior than what you had before, and what we have now, so it at least warrants mentioning what changed in the commit text. I'd almost suggest this could be a separate commit, since it's a different change - the ordering, not the icon.

neiljp (commented):
@plugyawn This looks cleaner now :+1:

I've left more feedback inline, but also please try and keep the linting clean, and adapt tests appropriately where necessary, ideally on a per-commit basis - keeping those up to date means that the PR is closer to a mergeable state at any given time. Have you been using black or `make fix` on commits to tidy these? That should keep the manual tidying load down.

There was a previous comment you didn't address, but that could be a refactor later - it's just us avoiding calculating presence for bots from memory.

zulipterminal/model.py
plugyawn: Rebased anda added a new commit.

neiljp:
@plugyawn Thanks for working on this :+1: I reworked this slightly after a preliminary refactor and a few adjustments, and this work is effectively merged as #1377 :tada:

PR #1189, opened 2022-04-07
feature: Add presence markers to PM compose.

plugyawn:
**What does this PR do?**  

Adds presence markers to PM compose.

Previously the only way to know if the person on the other side was online was to check through the info window or through the sidebar, which are big hassles. In the future, for non-group PMs, we could add a good last-seen implementation that doesn't take up unnecessary space. 

Partial fix for #1155.

**Tested?** 
- [x] Manually

**Commit flow** 

- Add presence markers to PM compose

**Notes & Questions** 

- markers don't have colours yet. 

**Visual changes** 
Screenshot 2022-04-07 at 7 24 27 PM

### DIFF
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -201,7 +201,7 @@ class WriteBox(urwid.Pile):
             ]
             recipient_info = ", ".join(
                 [
-                    f"""{self.model.user_dict[email]['full_name']} <{email}>"""
+                    f"""{STATE_ICON[self.model.user_dict.get(email, None).get("status","inactive") if self.model.user_dict.get(email,None) else "inactive"]}""" + " " + f"""{self.model.user_dict[email]['full_name']} <{email}>"""
                     for email in self.recipient_emails
                 ]
             )
@@ -219,7 +219,7 @@ class WriteBox(urwid.Pile):
 
         self.send_next_typing_update = datetime.now()
         self.to_write_box = ReadlineEdit(
-            "To: " + recipient_markers + " ", edit_text=recipient_info
+            "To: ", edit_text=recipient_info
         )
         self.to_write_box.enable_autocomplete(
             func=self._to_box_autocomplete,
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -200,10 +200,12 @@ class WriteBox(urwid.Pile):
                 for user_id in self.recipient_user_ids
             ]
             # In the format "{state marker} {recipient name} <{recipient email}> for all recipients.
-            # Avoids issue where all state markers appear back-to-back. 
+            # Avoids issue where all state markers appear back-to-back.
             recipient_info = ", ".join(
                 [
-                    f"""{STATE_ICON[self.model.user_dict.get(email, None).get("status","inactive") if self.model.user_dict.get(email,None) else "inactive"]}""" + " " + f"""{self.model.user_dict[email]['full_name']} <{email}>"""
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -201,16 +201,14 @@ def private_box_view(
             ]
             recipient_info = ", ".join(
                 [
-                    f"{self.model.user_dict[email]['full_name']} <{email}>"
+                    f"""{STATE_ICON[self.model.user_dict.get(email, None).get("status","inactive") if self.model.user_dict.get(email,None) else "inactive"]} {self.model.user_dict[email]['full_name']} <{email}>"""
```
prah23, line 204: The text being added here is validated before any action is performed, such as sending a message, saving a message as a draft, or even cycling out of the recipient header. Only adding an icon here will not bypass the validation check and hence, prevent the user from performing any of the aforementioned actions.

zulipterminal/ui_tools/boxes.py
```diff
@@ -201,16 +201,14 @@ def private_box_view(
...
                     for email in self.recipient_emails
                 ]
             )
         else:
             self._set_regular_and_typing_recipient_user_ids(None)
             self.recipient_emails = []
             recipient_info = ""
-
-        self.send_next_typing_update = datetime.now()
```
prah23, line 213: Unrelated?

prah23 (changes requested):
Thanks for the PR, @plugyawn.

Left a couple of comments inline, but more broadly, I'd suggest testing the proposed feature as a user thoroughly, so that any unintended behavior is detected and dealt with. The change proposed here currently prevents fundamental user abilities such as sending a private message.

zulipterminal/ui_tools/boxes.py
plugyawn: @prah23 thanks for pointing that out! Sorry, I thought I fixed this already, but I just realised I forgot to push it. I'll do so as soon as I get access to my computer again. My bad.

zulipterminal/ui_tools/boxes.py
plugyawn: Unrelated. Forgot to add a commit to fix the errors that sprung up.

plugyawn:
@prah23 Sorry, could you check again now?

zulipterminal/ui_tools/boxes.py
```diff
@@ -201,17 +201,26 @@ def private_box_view(
             ]
             recipient_info = ", ".join(
                 [
-                    f"{self.model.user_dict[email]['full_name']} <{email}>"
+                    f"""{self.model.user_dict[email]['full_name']} <{email}>"""
+                    for email in self.recipient_emails
+                ]
```
prah23, line 206: There's no change here; it can be removed from the commit.

prah23 (changes requested):
Thanks for the update.

The current iteration would make it hard to understand which status marker belongs to which user when there's a group PM of > 4-5 people, since all the markers are shown at the beginning and the names + emails come afterwards.  There is no update in the number of status markers either when a new recipient is added; if I add a third recipient to a group PM with 2 current recipients, there are still only 2 markers.

I would suggest discussing the design before updating the PR, so that you have more feedback to work with beforehand.

plugyawn:
Screenshot 2022-04-08 at 6 27 07 PM

Fixed the issue with Group PMs!

zulipterminal/ui_tools/boxes.py
plugyawn: Oops -- will do.

plugyawn:
Needs some changes in the tests.

PR #1194, opened 2022-04-10
Support 'Narrow to current compose box recipient'.

### LINKED ISSUE #1182: Support 'Narrow to current compose box recipient'
This is useful if starting a new topic - currently we give a notice that the sent message is outside of the current narrow, but there is no easy way of jumping to it.

The web app supports <kbd>ctrl</kbd>+<kbd>.</kbd> for achieving this, so this is a logical feature for us to also add.

Adding the key should show it in the help menu, but we could also show it in the message that is shown when messages are sent outside of the current narrow.

srdeotarse:
**What does this PR do?**  
This PR adds support to 'Narrow to current compose box recipient' after sending a message outside current narrow.
Shortcut - ``ctrl``+``.``

Fixes #1182 

CZO - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Support.20'Narrow.20to.20current.20compose.20box.20recipient'

**Tested?** 
- [x] Manually

**Commit flow** 

**Notes & Questions** 

Currently I have used shortcut - `.` instead of ``ctrl``+``.`` as ``ctrl``+``.`` doesn't seem to work in my local development environment (WSL2 Windows 10) 

**Interactions** 

**Visual changes**

### DIFF
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -80,6 +80,7 @@
 |Save current message as a draft|<kbd>meta</kbd> + <kbd>s</kbd>|
 |Autocomplete @mentions, #stream_names, :emoji: and topics|<kbd>ctrl</kbd> + <kbd>f</kbd>|
 |Cycle through autocomplete suggestions in reverse|<kbd>ctrl</kbd> + <kbd>r</kbd>|
+|Narrow to compose box message recipient|<kbd>meta</kbd> + <kbd>.</kbd>|
 |Jump to the beginning of line|<kbd>ctrl</kbd> + <kbd>a</kbd>|
 |Jump to the end of line|<kbd>ctrl</kbd> + <kbd>e</kbd>|
 |Jump backward one word|<kbd>meta</kbd> + <kbd>b</kbd>|
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -5,6 +5,7 @@ from pytest import param as case
 from pytest_mock import MockerFixture
 
 from zulipterminal.api_types import Composition
+from zulipterminal.config.keys import primary_key_for_command
 from zulipterminal.helper import (
     Index,
     canonicalize_color,
@@ -428,8 +429,10 @@ def test_notify_if_message_sent_outside_narrow(
     notify_if_message_sent_outside_narrow(req, controller)
 
     if footer_updated:
+        key = primary_key_for_command("NARROW_MESSAGE_RECIPIENT")
         report_success.assert_called_once_with(
-            "Message is sent outside of current narrow."
+            f"Message is sent outside of current narrow. Press [{key}] to narrow to conversation.",
+            duration=6,
         )
     else:
         report_success.assert_not_called()
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -164,6 +164,11 @@ KEY_BINDINGS: 'OrderedDict[str, KeyBinding]' = OrderedDict([
         'help_text': 'Narrow to the topic of the current message',
         'key_category': 'msg_actions',
     }),
+    ('NARROW_MESSAGE_RECIPIENT', {
+        'keys': ['meta .'],
+        'help_text': 'Narrow to compose box message recipient',
+        'key_category': 'msg_compose',
+    }),
     ('TOGGLE_NARROW', {
         'keys': ['z'],
         'help_text':
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -26,6 +26,7 @@ from urllib.parse import unquote
 from typing_extensions import TypedDict
 
 from zulipterminal.api_types import Composition, EmojiType, Message
+from zulipterminal.config.keys import primary_key_for_command
 from zulipterminal.config.regexes import (
     REGEX_COLOR_3_DIGIT,
     REGEX_COLOR_6_DIGIT,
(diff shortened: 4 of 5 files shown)


neiljp (commented):
@srdeotarse This is a good start, thanks! :+1:

This does appear to suffer from `ctrl .` not working for me too, which may be an issue with terminals in general. I'd suggest `meta .` (ie. `alt .`) in the meantime, unless/until we can find out why it is failing.

There is an issue where exiting compose seems to trigger going to all messages which we need to resolve before this can be merged. This appears to only happen if the new action is triggered and then the compose is cancelled/closed (or sent and then cancelled/closed).

Tests would also be good if possible.

docs/hotkeys.md
```diff
@@ -54,6 +54,7 @@
 |Show/hide Emoji picker popup for current message|<kbd>:</kbd>|
 |Narrow to the stream of the current message|<kbd>s</kbd>|
 |Narrow to the topic of the current message|<kbd>S</kbd>|
+|Narrow to compose box message recipient.|<kbd>.</kbd>|
```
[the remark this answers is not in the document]
srdeotarse: Definitely, you just have to change the key in `keys.py`` and run the script to update ``hotkeys.md`` instead of changing manually. Really helpful.

zulipterminal/helper.py
```diff
@@ -652,7 +653,11 @@ def check_narrow_and_notify(
         and current_narrow != outer_narrow
         and current_narrow != inner_narrow
     ):
-        controller.report_success("Message is sent outside of current narrow.")
+        controller.report_success(
+            "Message is sent outside of current narrow. Press 'Ctrl .' to narrow to message recipient. {}".format(
+                NARROW_TO_MESSAGE_RECIPIENT,
+            )
```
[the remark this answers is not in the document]
srdeotarse: You had mentioned an UI element being proposed in #design > button to go to conversation. Should that arrow like symbol appear in compose box when messaging outside current narrow?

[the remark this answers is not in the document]
srdeotarse: So, if we change the shortcut for narrowing to current message recipient, the key in the message should also change. Right?

zulipterminal/config/keys.py
```diff
@@ -164,6 +164,11 @@ class KeyBinding(TypedDict, total=False):
         'help_text': 'Narrow to the topic of the current message',
         'key_category': 'msg_actions',
     }),
+    ('NARROW_MESSAGE_RECIPIENT', {
+        'keys': ['meta .'],
+        'help_text': 'Narrow to compose box message recipient.',
+        'key_category': 'msg_actions',
```
neiljp, line 170: This is really during compose.

tests/helper/test_helper.py
```diff
@@ -429,7 +431,10 @@ def test_notify_if_message_sent_outside_narrow(
 
     if footer_updated:
         report_success.assert_called_once_with(
-            "Message is sent outside of current narrow."
+            "Message is sent outside of current narrow. Press {} to narrow to message recipient. {}".format(
+                KEY_BINDINGS["NARROW_MESSAGE_RECIPIENT"]["keys"][0],
+                NARROW_TO_MESSAGE_RECIPIENT,
```
neiljp, line 436: As per discussion in the stream, let's skip the icon for now; I think it'd be a representation of the key somehow if we did, a little like we have `[P]`.

neiljp (commented):
@srdeotarse Thanks for the improved iteration, though it's still not quite performing right for me.

We've discussed this a little in the stream which should help with the next iteration.

You mentioned the hotkeys script; we don't lint for this and I'm not sure we have an issue for that, if you want to take that up. It would be useful in CI (and locally) to ensure that any changes are synchronized, and could be part of `make fix` too.

srdeotarse:

srdeotarse:
@zulipbot remove "PR needs review".

zulipterminal/helper.py
```diff
@@ -646,13 +647,16 @@ def check_narrow_and_notify(
...
     if (
         current_narrow != []
         and current_narrow != outer_narrow
         and current_narrow != inner_narrow
     ):
-        controller.report_success("Message is sent outside of current narrow.")
+        key = primary_key_for_command("NARROW_MESSAGE_RECIPIENT")
+
+        controller.report_success(
+            f"Message is sent outside of current narrow. Press [{key}] to narrow to conversation."
```
neiljp, line 658: If we keep this message this long, I'd welcome this being displayed a little longer (eg. the 6 we used in the other PR).

However, my remaining query here is whether we want to customize this logic/message a little further:
- in all-messages narrow, we never see this, but may still want to go to the conversation
- in stream (or all-PM) narrows which hold the message, we may also still want to go to the conversation
- if a message is outside the narrow, we always want to show this

To address these combinations we may want to split this messaging change out into a followup commit and/or PR, since the functionality itself will be present and consistent with the help menu and documentation already in the rest of this commit. This is an extra hint, and it'd be good to handle it well.

zulipterminal/ui_tools/boxes.py
```diff
@@ -773,6 +773,27 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                 if self.msg_edit_state is not None:
                     self.keypress(size, primary_key_for_command("GO_BACK"))
                     assert self.msg_edit_state is None
+        elif is_command_key("NARROW_MESSAGE_RECIPIENT", key):
+            if not self.to_write_box:
```
neiljp, line 777: These branches would be more readable using `compose_box_status`.

neiljp (commented):
@srdeotarse If we assume #1201 is reasonable, this PR seems functional :tada:

I've noted a few outstanding points, but otherwise this is close to merging.

We could add a test for the UI function, but the code is rather simple and UI code can be challenging to test and fragile.

srdeotarse:
@zulipbot "PR needs review".

srdeotarse:

neiljp:
@srdeotarse Thanks for updating this :+1: I adjusted the other conditional to use compose_box_status and will be merging shortly. This is going to be very helpful day to day 🎉

zulipterminal/helper.py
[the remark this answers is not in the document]
neiljp: The arrow issue is now at #1203.

tests/helper/test_helper.py
neiljp: Again, this is now #1203

PR #1201, opened 2022-04-15
views: Avoid sending GO_BACK to duplicate child widgets unnecessarily.

neiljp:
**What does this PR do?**  

This conditional appears unnecessary, and interception here can cause multiple triggers of the keypress in corner cases.

Test removed.

Discussed in #**zulip-terminal > Avoid double Esc keypress #T1201**

**Tested?** 
- [x] Manually

**Interactions** 

This appears to also resolve the outstanding issue with #1194.


neiljp:
@Rohitth007 Thanks for the comments via czo :+1:

PR #1204, opened 2022-04-17
Display 'Message sent outside current narrow' when narrowing from all_messages and all_PMs.

srdeotarse:
Follow up PR #1194

**What does this PR do?**  
This PR displays ``Message sent outside current narrow`` even if narrowing from ``all_messages`` or ``all_PMs``

**Tested?** 
- [x] Manually

**Commit flow** 

**Notes & Questions** 

**Interactions** 

CZO - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/.E2.9C.94.20Support.20'Narrow.20to.20current.20compose.20box.20recipient'.20.23T1194

Cases for showing success message - https://github.com/zulip/zulip-terminal/pull/1194#discussion_r851485357

**Visual changes**

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -339,7 +339,7 @@ def test_display_error_if_present(
         case(
             {"type": "private", "to": [1], "content": "bar"},
             [["is", "private"]],
-            False,
+            True,
             id="all_private__pm__not_notified",
         ),
         case(
@@ -386,7 +386,7 @@ def test_display_error_if_present(
                 "content": "Hi `|new_stream|`",
             },
             [],
-            False,
+            True,
             id="all_messages__stream__not_notified",
         ),
         case(
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -648,11 +648,7 @@ def check_narrow_and_notify(
 ) -> None:
     current_narrow = controller.model.narrow
 
-    if (
-        current_narrow != []
-        and current_narrow != outer_narrow
-        and current_narrow != inner_narrow
-    ):
+    if current_narrow != inner_narrow:
         key = primary_key_for_command("NARROW_MESSAGE_RECIPIENT")
 
         controller.report_success(
@@ -666,7 +662,6 @@ def check_narrow_and_notify(
 def notify_if_message_sent_outside_narrow(
     message: Composition, controller: Any
 ) -> None:
-    current_narrow = controller.model.narrow
 
     if message["type"] == "stream":
         stream_narrow = [["stream", message["to"]]]
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -108,6 +108,9 @@ class WriteBox(urwid.Pile):
         self.idle_status_tracking: bool
         self.sent_start_typing_status: bool
 
+        # To display narrow to conversation message.
+        self.check_stream_topic_edit_txt_flag: bool = True
+
         self._set_compose_attributes_to_defaults()
 
         # Constant indices into self.contents
@@ -257,6 +260,8 @@ class WriteBox(urwid.Pile):
         start_period_delta = timedelta(seconds=TYPING_STARTED_WAIT_PERIOD)
         stop_period_delta = timedelta(seconds=TYPING_STOPPED_WAIT_PERIOD)
 
+        self.check_stream_topic_edit_txt(self.check_stream_topic_edit_txt_flag)
+
         def on_type_send_status(edit: object, new_edit_text: str) -> None:
             if new_edit_text and self.typing_recipient_user_ids:
                 self.last_key_update = datetime.now()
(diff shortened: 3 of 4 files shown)


srdeotarse:

neiljp:
From #1194:
> However, my remaining query here is whether we want to customize this logic/message a little further:
>     - in all-messages narrow, we never see this, but may still want to go to the conversation
>     - in stream (or all-PM) narrows which hold the message, we may also still want to go to the conversation
>     - if a message is outside the narrow, we always want to show this

I may not have been clear in this comment, but what I meant is that the two messages are not necessarily correlated since the conditions vary:
- we still want the 'outside of narrow' behavior to stay the same
- we want to also show the hint if not fully narrowed, ie. what the shortcut would do (though this could become too much, we'll see)

zulipterminal/helper.py
```diff
@@ -664,7 +660,6 @@ def check_narrow_and_notify(
 def notify_if_message_sent_outside_narrow(
     message: Composition, controller: Any
 ) -> None:
-    current_narrow = controller.model.narrow
```
neiljp, line 667: This appears a valid one-off tidy-up, but if it's not related to the bulk of the code then it's good to note it in the commit text body. More tidies => maybe it belongs in it's own commit.

srdeotarse:

neiljp:
@srdeotarse Left messages in stream for discussion. Also please address comments in a PR via changes or comments, or discuss in stream, before requesting review again :)

srdeotarse:
add "PR needs review".

zulipterminal/ui_tools/boxes.py
```diff
@@ -403,12 +408,51 @@ def stream_box_view(
...
-
         # Use and set a callback to set the stream marker
         self._set_stream_write_box_style(None, caption)
         urwid.connect_signal(
             self.stream_write_box, "change", self._set_stream_write_box_style
         )
+        self.check_stream_topic_edit_txt(self.check_stream_topic_edit_txt_flag)
+
+    @asynch
+    def check_stream_topic_edit_txt(
```
neiljp, line 419: Minor perhaps, but this handles private cases too.

zulipterminal/ui_tools/boxes.py
```diff
@@ -403,12 +408,51 @@ def stream_box_view(
...
+        self, check_stream_topic_edit_txt_flag: bool
+    ) -> None:
+        narrow_key = primary_key_for_command("NARROW_MESSAGE_RECIPIENT")
+        narrow_markup = [
+            ("footer", "Help(?): "),
+            ("footer_contrast", f" {narrow_key} "),
+            ("footer", " Narrow to conversation."),
```
neiljp, line 426: This styling (and below) is extracted from the other location. It would be better to use a common functionality.

zulipterminal/ui_tools/boxes.py
```diff
@@ -403,12 +408,51 @@ def stream_box_view(
...
+        ]
+        send_key = primary_key_for_command("SEND_MESSAGE")
+        send_markup = [
+            ("footer", "Help(?): "),
+            ("footer_contrast", f" {send_key} "),
+            ("footer", " Send a message."),
+        ]
+        while check_stream_topic_edit_txt_flag:
```
neiljp, line 434: We don't want to use a busy loop for this, if we do take this approach. This sends the zulip-term CPU usage to 100%+.

An urwid on-change signal may be better, though note that while potentially helpful, `alt .` will not (or shouldn't) do anything if the recipients aren't valid, so we can avoid showing the notice in that case.

zulipterminal/helper.py
```diff
@@ -648,11 +648,7 @@ def check_narrow_and_notify(
 ) -> None:
     current_narrow = controller.model.narrow
 
-    if (
-        current_narrow != []
-        and current_narrow != outer_narrow
-        and current_narrow != inner_narrow
-    ):
+    if current_narrow != inner_narrow:
```
neiljp, line 651: This is the correct condition for the narrowing hint, but please see https://github.com/zulip/zulip-terminal/pull/1204#issuecomment-1100809732

There are two separate conditions; one handles the 'outside of narrow' part of the message (doesn't need updating), another separate one handles the showing of the second (hint) part.

tests/helper/test_helper.py
```diff
@@ -339,7 +339,7 @@ def test_display_error_if_present(
         case(
             {"type": "private", "to": [1], "content": "bar"},
             [["is", "private"]],
-            False,
+            True,
             id="all_private__pm__not_notified",
```
neiljp, line 343: We may need to update either the test or the ids, to represent what we're expecting here.

At the very least, this change is now inconsistent: True => notified :)

However, there are different aspects being notified here, so a boolean doesn't fully encapsulate the results any more.

neiljp (commented):
@srdeotarse The extra commit here looks interesting, but I'd be inclined to split it out into another PR since it's really extending the "random" help to be more contextual.

The first commit doesn't quite address one of my original points; I referenced that inline, and it likely requires the test structure to be slightly updated to indicate the output. Hopefully you'll find that easier after working on that other PR.

Issue #1174, opened 2022-03-21
Check formatting of markdown in notifications


neiljp:
Following #1173 it would be useful to look through other markdown in Zulip and see how it is output in notifications, and attempt to improve it if possible.

Possible issues:
- realm emoji?
- tables (also an issue in regular zulip notifications?)
- rules (also in zulip?)
- ...

zulipbot:
Hello @zulip/server-notifications members, this issue was labeled with the "area: notifications" label, so you may want to check it out!

plugyawn:
@claim

plugyawn:
@zulipbot claim.

@neiljp There's also issues with the \me mention, time, tasks, in addition to tables, rules, etc, as you mentioned. Would appreciate advice on how to resolve the time and tasks problems

zulipbot:
Welcome to Zulip, @plugyawn! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

plugyawn:
Fixed tables (with spaces and line-breaks), bullets, and emojis. Fixed linting issues.

zulipbot:
Hello @plugyawn, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

plugyawn:
@zulipbot claim

zulipbot:
[...]

PR #1207, opened 2022-04-18
model:  added placeholder for tables in notifications.

plugyawn:
**What does this PR do?**  
Added a soup_to_formatted_soup function that takes the soup and formats it so the notification doesn't contain any unwanted text.

Currently, it only handles tables (sets it to [MxN table attached]) but in the future it will handle emojis, \me, etc too.

Partially fixes issue #1174. 

https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Check.20formatting.20of.20markdown.20in.20notifications.20.23T1174.20.23T1183
**Tested?** 
- [x] Manually

**Commit flow** 

- first commit adds soup_to_formatted_soup() function to boxes.py and calls it from the notify_user function.

**Notes & Questions** 

**Interactions** 

**Visual changes** 
Present:
image
Past:
image

For this message:
Screenshot 2022-04-18 at 4 30 20 PM

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -26,6 +26,8 @@ from typing import (
 from urllib.parse import unquote
 
 import requests
+from bs4 import BeautifulSoup
+from bs4.element import Tag
 from typing_extensions import TypedDict
 
 from zulipterminal.api_types import Composition, EmojiType, Message
@@ -41,6 +43,7 @@ from zulipterminal.platform_code import (
     normalized_file_path,
     successful_GUI_return_code,
 )
+from zulipterminal.ui_tools.tables import parse_html_table
 
 
 class StreamData(TypedDict):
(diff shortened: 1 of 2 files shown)


neiljp (commented):
@plugyawn Thanks for looking into this in a smaller iteration than before.

Could you discuss why you think this belongs in `MessageBox` as a classmethod?

I think it may be simpler to use a helper non-member function (ie. not method) which processes the content of a message for the purposes of notification, so a first step might be to refactor the existing spoiler process into such a function first (or a function to be called from it), then extend to take into account the extra behavior.

Each piece of behavior could itself be a function call from there, though I can see you have things set up for a design mirroring the soup2markup function right now?

There are specific variables in the current code that don't appear at first glance to be used yet, which makes this commit more confusing than necessary to read. Those should be added when they are required.

The added bonus of a specific function should be that adding a test for that function should be a lot simpler, and could be applied to spoilers too in future.

plugyawn:
> Could you discuss why you think this belongs in `MessageBox` as a classmethod?

My bad, I was working with soup2markup and it missed my mind that it doesn't fit into `MessageBox`. I'll put it in helpers.py.

> a first step might be to refactor the existing spoiler process into such a function first (or a function to be called from it), then extend to take into account the extra behavior.

This I was planning on doing, but was confused whether to add that in a separate PR and focus this PR on only fixing the table versus making one PR to both refactor and then fix the tables.

>There are specific variables in the current code that don't appear at first glance to be used yet, which makes this commit more confusing than necessary to read. Those should be added when they are required.

Understood, I'll change those. 

>The added bonus of a specific function should be that adding a test for that function should be a lot simpler, and could be applied to spoilers too in future.

Got it! I've been looking forward to writing tests for a while now. : )
So this works out nicely.

> Each piece of behavior could itself be a function call from there, though I can see you have things set up for a design mirroring the soup2markup function right now?

I have, yes, but right now my plan is to use individual functions to fix each of the issues with tables, emojis, etc, and then wrap all of them in another function, so it's neater and easier to read. 
I am concerned about some amount of pointless verbosity, though, so I'd like your opinion.

plugyawn:
Just updated. Please ignore the lines where the out-of-narrow code are adjusted, I'll fix that.

plugyawn:
@neiljp I changed the commit flow, could you check if it's fine?

zulipterminal/helper.py
```diff
@@ -721,3 +724,56 @@ def suppress_output() -> Iterator[None]:
...
+            conversion_parameter = True
+    return soup
+
+
+def notification_formatter(soup: Any) -> Any:
+    """
+    Formatter for message notifications. Takes soup, returns soup.
+    """
+    soup = notification_formatter_table(soup)
+    soup = notification_formatter_spoiler(soup)
```
neiljp, line 778: These functions actually modify the arguments, right?

If so, let's name them better, eg. `reformat_tables_in_soup`, and not include the redundant return value?

zulipterminal/model.py
```diff
@@ -1318,20 +1319,7 @@ def notify_user(self, message: Message) -> str:
                 text = content
             else:
                 soup = BeautifulSoup(content, "lxml")
```
neiljp, line 1321: This could nicely go into the notification_formatter function instead; leaving it here exposes the details of how we're doing it, rather than just passing the message content to the function.

neiljp (commented):
@plugyawn You asked about the commit flow here, but you only have one commit, so I'm a bit confused. I've done a quick review, but it would be cleaner if we had a refactor commit to move the existing formatting into eg. the general formatter and spoiler-specific function, then another commit to add the table functionality.

Otherwise having the separate functions for each process, and one to call them from seems fine, and you could add tests for the overall function you extract in the first commit in that commit, then build on those tests when you add more functionality.

zulipterminal/helper.py
plugyawn: Yep, alright!

PR #1208, opened 2022-04-19
[WIP]  Add docstrings to each source file and add linting for developer-file-overview.

### LINKED ISSUE #892: Automate developer overview of files by extracting module docstrings
Currently overviews of each file are manually described in `docs/developer-file-overview.md`.

For developers it would be useful to have each of these descriptions at the top of a given file as a docstring; with that, this document could then be generated automatically and checked to be up to date in CI.

* [x] Add docstrings to each source file that match the overview file
* [ ] Add a CI Action which ensures the document matches the expectations from the docstrings
* [x] Generate the document from docstrings rather than manually

mounilKshah:
**What does this PR do?**
- This PR adds docstrings to the source files in the codebase
- Add a CI Action which ensures the document matches the expectations from the docstrings
- Generate the document from docstrings

Fixes #892 

**Tested?** 
- [x] Manually

**Commit flow**
- first commit: Adds docstrings to the source files
- second commit: Adds linting for developer-file-overview.md

### DIFF
--- a/docs/developer-file-overview.md
+++ b/docs/developer-file-overview.md
@@ -1,3 +1,6 @@
+<!--- Generated automatically by tools/lint-docstring.py -->
+<!--- Do not modify -->
+
 ## Overview
 
 Zulip Terminal uses [Zulip's API](https://zulip.com/api/) to store and retrieve all the information it displays. It has an [MVC structure](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) overall. Here is a description of some of its files:
--- a/tools/lint-all
+++ b/tools/lint-all
@@ -23,6 +23,7 @@ tools = {
         "./tools/generate_hotkeys.py",
         "--check-only",
     ],
+    "Docstring linting & docs sync check (lint-docstring)": "./tools/lint-docstring",
 }
 
 HEADER = 80 * "="
--- a/tools/lint-docstring
+++ b/tools/lint-docstring
@@ -35,9 +35,78 @@ DESC_FOR_NO_FILE_FOLDERS = {
 FOLDERS_TO_EXCLUDE = ["__pycache__"]
 
 
-def main() -> None:
-    create_file_overview_doc()
-    print(f"Generated {DEVELOPER_DOC_NAME} successfully.")
+def main(fix_file: bool) -> None:
+    if fix_file:
+        create_file_overview_doc()
+        print(f"Generated {DEVELOPER_DOC_NAME} successfully.")
+    else:
+        error_messages = lint_file_overview()
+        if error_messages:
+            for file_name, error in error_messages.items():
+                print(f"{file_name}: {error} ")
+            print(
+                f"\nRun './tools/lint-docstring --fix' to add the docstring to {DEVELOPER_DOC_NAME}"
+            )
+            sys.exit(1)
+        else:
+            print(f"{DEVELOPER_DOC_NAME} has been linted successfully.")
+
+
+def lint_file_overview() -> Dict[str, str]:
+    """
+    Used for linting of the developer-file-overview.md document
+    """
+    folder_file_docstring = generate_folder_file_docstrings_dict()
+    existing_doc_dict = extract_docstrings_from_file_overview()
+    # folders and files which return error while linting are stored under
+    # error_message_dict with their respective error messages
+    error_message_dict = {}
+    for folder in folder_file_docstring:
+        # check if the folder for the docstring is present or not
+        if folder in existing_doc_dict:
+            for file in folder_file_docstring[folder]:
+                # check if the file under the folder is present or not
+                if file in existing_doc_dict[folder]:
+                    # check if docstrings match or not
+                    if (
+                        folder_file_docstring[folder][file]
+                        != existing_doc_dict[folder][file]
+                    ):
+                        error_message_dict[
+                            f"{folder}/{file}"
+                        ] = f"Docstrings do not match those listed in {DEVELOPER_DOC_NAME}"
+                    del existing_doc_dict[folder][file]
+                else:
+                    error_message_dict[
+                        f"{folder}/{file}"
+                    ] = f"File does not exist in {DEVELOPER_DOC_NAME}"
+            # if the folder dictionary is empty, delete it from the existing_doc_dict
+            if not existing_doc_dict[folder]:
+                del existing_doc_dict[folder]
+        else:
+            error_message_dict[folder] = f"Folder not present in {DEVELOPER_DOC_NAME}"
+
+    # for docstrings in DESC_FOR_NO_FILE_FOLDERS
+    for folder_name in DESC_FOR_NO_FILE_FOLDERS:
+        if (
+            DESC_FOR_NO_FILE_FOLDERS[folder_name]
+            != existing_doc_dict[folder_name]["no_file_present"]
+        ):
+            error_message_dict[
+                folder_name
+            ] = f"Docstrings do not match those listed in {DEVELOPER_DOC_NAME}"
+        del existing_doc_dict[folder_name]
+
+    # check if there are any docstrings in the overview file
+    # for which folder/file has been removed
+    if existing_doc_dict:
+        for folder_name in existing_doc_dict:
+            if existing_doc_dict[folder_name]:
+                for file_name in existing_doc_dict[folder_name]:
+                    error_message_dict[
+                        f"{folder_name}/{file_name}"
+                    ] = "Folder/File for this docstring has been removed"
+    return error_message_dict
 
 
 def create_file_overview_doc() -> None:
@@ -120,9 +189,41 @@ def extract_folder_file_structure() -> Dict[str, List[str]]:
     return folders_and_files
 
 
+def extract_docstrings_from_file_overview() -> Dict[str, Dict[str, str]]:
+    """
+    Reads the developer-file-overview.md document and creates
+    a dictionary containing folder name which in turn is a dictionary
+    containing files and their respective descriptons
+    """
+    with open(DEVELOPER_DOC_PATH, "r") as file:
+        doc_data = file.readlines()
+
+    existing_doc_dict: Dict[str, Dict[str, str]] = {}
+
+    # table is present from line 6 to (length - 1)
+    for doc_row in doc_data[TABLE_OFFSET - 1 :]:
+        _, folder, filename, docstring, _ = doc_row.split("|")
+        folder, filename = folder.strip(), filename.strip()
+        if folder:
+            # For files under a folder but no folder value in overview file (eg. core.py)
+            folder_name = folder
+            existing_doc_dict[str(folder_name)] = {}
+        if docstring.strip():
+            if filename:
+                existing_doc_dict[folder_name][filename] = docstring.strip()
+            else:
+                existing_doc_dict[folder_name]["no_file_present"] = docstring.strip()
+    return existing_doc_dict
+
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
-        description=f"Regenerate the {DEVELOPER_DOC_NAME} file according to the docstrings"
+        description=f"Lint {DEVELOPER_DOC_NAME} as per docstrings across all files in zulipterminal"
+    )
+    parser.add_argument(
+        "--fix",
+        action="store_true",
+        help=f"Regenerate {DEVELOPER_DOC_NAME} according to the docstrings",
     )
     args = parser.parse_args()
-    main()
+    main(args.fix)
--- a/tools/lint-docstring
+++ b/tools/lint-docstring
@@ -8,7 +8,7 @@ from pathlib import Path
 from typing import Dict, List
 
 
-TABLE_OFFSET = 7
+TABLE_OFFSET = 10
 COLUMN_WIDTHS = (23, 20, 103)
 SPACEBAR = " "
 
@@ -200,7 +200,7 @@ def extract_docstrings_from_file_overview() -> Dict[str, Dict[str, str]]:
 
     existing_doc_dict: Dict[str, Dict[str, str]] = {}
 
-    # table is present from line 6 to (length - 1)
+    # table is present from line TABLE_OFFSET to (length - 1)
     for doc_row in doc_data[TABLE_OFFSET - 1 :]:
         _, folder, filename, docstring, _ = doc_row.split("|")
         folder, filename = folder.strip(), filename.strip()


mounilKshah:

tools/lint-docstring
```diff
@@ -0,0 +1,220 @@
...
+    Returns a dictionary containing folder name which in turn
+    is a dictionary containing files and their respective descriptons
+    """
+    folder_paths = {
+        "zulipterminal": join(ROOT_DIRECTORY, "zulipterminal"),
+        "zulipterminal/cli": join(ROOT_DIRECTORY, "zulipterminal", "cli"),
+        "zulipterminal/config": join(ROOT_DIRECTORY, "zulipterminal", "config"),
+        "zulipterminal/ui_tools": join(ROOT_DIRECTORY, "zulipterminal", "ui_tools"),
+    }
+    total_files = {}
```
neiljp, line 76: total_files is really the input to this function; you might make it a constant, perhaps, if we don't calculate it dynamically at the moment.

To put it a different way this 'get' or 'extract' of the docstrings is separate from the calculation of which paths to extract them from.

neiljp (commented):
@mounilKshah Thanks for giving this a go - I think this will help new devs with understanding individual files while also keeping the summary file synchronized :+1:

That said, I couldn't get this working, but I've left a preliminary review - it's quite large and in one commit, so I've focused on the structure.

tools/lint-docstring
```diff
@@ -0,0 +1,220 @@
...
+from typing import Dict, cast
+
+
+INITIAL_INDEX = 7
+
+# absolute path to zulip-terminal
+ROOT_DIRECTORY = pathlib.Path(__file__).resolve().parent.parent
+
+# new doc file has been created for the time being to compare it with the original
+DEVELOPER_DOC = join(ROOT_DIRECTORY, "docs", "developer-file-overview_new.md")
```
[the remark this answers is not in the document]
neiljp: This also applies to other pathlib options to other functions.

zulipterminal/api_types.py
```diff
@@ -1,3 +1,8 @@
+"""
+Preliminary Zulip API types defined in python,
+to allow type checking
```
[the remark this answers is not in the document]
mounilKshah: It was to avoid horizontal scrolling across those files. I'll remove the line breaks.

tools/lint-docstring
```diff
@@ -0,0 +1,220 @@
...
+    containing files and their respective descriptons
+    """
+    with open(DEVELOPER_DOC, "r") as file:
+        doc_data = file.readlines()
+
+    computed_doc_dict: Dict[str, Dict[str, str]] = {}
+
+    folder_name = ""
+    # table is present from line 6 to (length - 1)
+    for i in range(INITIAL_INDEX - 1, len(doc_data) - 4):
```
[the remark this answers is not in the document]
mounilKshah: The last 4 lines of file-overview are [hard coded](https://github.com/zulip/zulip-terminal/pull/1208/files#diff-2ed47285f7c803803527b645e26c80f37db9583f2c8146b9eeeace9d969bc925R137) and not extracted from any file as there aren't any file(s) corresponding to the `zulipterminal/scripts` and `zulipterminal/themes` folders. Their description is added at the end of the file (the last 4 lines to be precise). Hence, while extracting the docstrings for linting, I avoided taking them into consideration.

tools/lint-docstring
```diff
@@ -0,0 +1,220 @@
...
+    if fix_file:
+        create_file()
+        print("Generated developer-file-overview_new.md successfully.\n")
+    else:
+        lint_files()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Lint and fix docstrings" "in all ZulipTerminal files"
```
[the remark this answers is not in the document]
mounilKshah: I'll rectify this.

tools/lint-docstring
```diff
@@ -0,0 +1,220 @@
...
+    )
+    parser.add_argument(
+        "-fix",
+        action="store_true",
+        help="""Update developer-file-overview with
```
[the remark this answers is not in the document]
mounilKshah: It does. I'll rectify this.

tools/lint-docstring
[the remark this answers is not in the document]
mounilKshah: I guess the file path strings vary depending upon the OS right? Windows might have `\` while Unix and Linux have `/`? Considering these differences, I thought of using something that was independent of the user's system.

mounilKshah:

tools/lint-docstring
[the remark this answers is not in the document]
neiljp: To clarify, I was referring to the behavior of `/` for Path and PurePath.
(https://docs.python.org/3/library/pathlib.html#operators)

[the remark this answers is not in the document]
mounilKshah: Oh. I tried this out. It works pretty well. I'll replace all instances of `join()` with `/`.

docs/developer-file-overview.md
```diff
@@ -17,7 +17,8 @@ Zulip Terminal uses [Zulip's API](https://zulip.com/api/) to store and retrieve
 |                        |                     |                                                                                                        |
 | zulipterminal/cli      | run.py              | Marks the entry point into the application                                                             |
 |                        |                     |                                                                                                        |
-| zulipterminal/config   | color.py            | Color definitions or functions common across all themes                                                |
+| zulipterminal/config   | color.py            | Contains color definitions or functions common across all themes. For further details on themefiles    |
+|                        |                     | look at the theme contribution guide.                                                                  |
```
neiljp, line 21: The overview file doesn't need the extra line. For now I'd suggest simplifying your code and moving it to a comment in the file.

docs/developer-file-overview.md
```diff
@@ -26,7 +27,7 @@ Zulip Terminal uses [Zulip's API](https://zulip.com/api/) to store and retrieve
 |                        | ui_mappings.py      | Relationships between state/API data and presentation in the UI                                        |
 |                        | ui_sizes.py         | Fixed sizes of UI elements                                                                             |
 |                        |                     |                                                                                                        |
-| zulipterminal/ui_tools | boxes.py            | UI boxes for displaying messages and entering text, such as `MessageBox`, `SearchBox`, `WriteBox`, etc.|
+| zulipterminal/ui_tools | boxes.py            | UI boxes for displaying messages and entering text,such as `MessageBox`, `SearchBox`, `WriteBox`, etc. |
```
neiljp, line 30: Note that github is indicating a deleted space here.

tools/lint-docstring
```diff
@@ -0,0 +1,273 @@
...
+from typing import Dict, Tuple, cast
+
+
+INITIAL_INDEX = 7
+
+# absolute path to zulip-terminal
+ROOT_DIRECTORY = pathlib.Path(__file__).resolve().parent.parent
+
+# absolute path to zulip-terminal/zulipterminal to be passed as parameter
+ZULIPTERMINAL = join(ROOT_DIRECTORY, "zulipterminal")
```
neiljp, line 18: This can written as `ROOT_DIRECTORY / "zulipterminal"` per my other point.

tools/lint-docstring
```diff
@@ -0,0 +1,273 @@
...
+
+# absolute path to the folders under zulipterminal/ to be skipped while parsing
+THEMES_PATH = join(ZULIPTERMINAL, "themes")
+SCRIPTS_PATH = join(ZULIPTERMINAL, "scripts")
```
neiljp, line 22: This does work, but is fragile - if we change this behavior or add other folders like this we need to do extra work, and we're already doing it twice. It is cleaner to make a data-structure that represents special cases and then process it. That will allow it to include the 4 other constants you defined lower down.

tools/lint-docstring
```diff
@@ -0,0 +1,273 @@
...
+    folder_name = folder_text + folder_remaining_spaces
+    file_name = file_text + file_remaining_spaces
+    file_desc = desc_text + desc_remaining_spaces
+    return (folder_name, file_name, file_desc)
+
+
+def split_string(long_string):
+    """
+    In case the docstring is longer than the length of description string (>104),
+    this function breaks it into two lines.
```
neiljp, line 62: Generally this should represent a 'bug' - we don't want descriptions to be "too long", otherwise the file doesn't represent an overview and the descriptions could get long.

I think you added this to handle the case of the text being multiple lines in one case? If so, as per a comment there, simply make the non-matching text into a comment for now.

zulipterminal/api_types.py
```diff
@@ -1,6 +1,5 @@
 """
-Preliminary Zulip API types defined in python,
-to allow type checking
+Preliminary Zulip API types defined in python, to allow type checking
```
neiljp, line 2: I'm not sure why the change in this 3rd (?) commit isn't in the original commit?

neiljp (commented):
@mounilKshah This runs for me now :+1: That said, I ran with -fix (this should probably be `--fix`) and it leaves a change to the state of the PR.

This isn't a full review, but I left some feedback inline. The commit flow is not what I would expect. It currently represents the way in which you've done the development rather than adding in stages to illustrate the changes. This can take some getting used to.

I'm excited about this tool, and the globbing results look good so far!

tools/lint-docstring
mounilKshah: Yes. It was to handle the case of `color.py` which had a description too long to fit within the docstring column.

Re bug: I'll make changes to throw an error in case the docstring is longer than the column width.

zulipterminal/api_types.py
mounilKshah: The original commit had line breaks for longer docstrings. After your comments, I removed those line breaks, which were added in the third commit. As said above, I'll make the commit flow cleaner.

zulipterminal/config/color.py
```diff
@@ -1,9 +1,7 @@
 """
-COLOR
------
-Contains color definitions or functions common across all themes.
-For further details on themefiles look at the theme contribution guide.
```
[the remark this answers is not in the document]
Rohitth007: Maybe lets add the "For further details..." line in the `developer-file-overview.md`. Linking to the theme contribution guide from the overview doc make sense.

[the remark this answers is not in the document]
mounilKshah: Are you suggesting to do something similar to [this](https://github.com/zulip/zulip-terminal/pull/1208/files#diff-0bb745d3e470b3a0c93da4298bceebbdf4d2f1f0cf8a7624c0f2f4e5991e5ac9R20)?

Rohitth007:
As a suggestion, it would be more readable if you reverse the order of all your function.
Basically, first `main()` then `lint_files()` then `create_file()` and so on.
If you first tell the reader where the function is used and then tell them what the function does, then it's a much more readable.

zulipterminal/config/color.py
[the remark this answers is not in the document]
Rohitth007: Seems like Neil has a different opinion.
In general I feel the theme contributing guide is not accessible from anywhere unless you open the `themes` directory or `color.py`. (Is it linked from anywhere?) Maybe a link that directs interested users/developers from the README/FAQ or any other relevant documentation could help.

Adding a link might be a problem while parsing though.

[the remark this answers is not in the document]
neiljp: We can always document these folders or improve this process later, and of course other documentation linkage is great to add otherwise.

This issue/PR is about decentralizing the data in this file and making it easy to keep it in sync.

tools/lint-docstring
```diff
@@ -0,0 +1,246 @@
...
+    for folder, files in sorted(total_files.items()):
+        folder_file_docstring[str(folder)] = {}
+        for file in files:
+            imported_file = importlib.import_module(
+                f'{folder.replace("/",".")}.{file[:-3]}'
+            )
+            extracted_docstring = cast(str, imported_file.__doc__)
+            docstring = extracted_docstring.strip().replace("\n", " ")
+            if len(docstring) >= 104:
+                raise ValueError(f"{file} has docstring longer than 104")
```
neiljp, line 110: It's good to avoid hard-coded numbers; it's fairly obvious which variable this comes.

tools/lint-docstring
```diff
@@ -0,0 +1,183 @@
...
+    """
+    total_files = extract_folder_file_structure()
+    folder_file_docstring: Dict[str, Dict[str, str]] = {}
+    for folder, files in sorted(total_files.items()):
+        folder_file_docstring[str(folder)] = {}
+        for file in files:
+            imported_file = importlib.import_module(
+                f'{folder.replace("/",".")}.{file[:-3]}'
+            )
+            extracted_docstring = cast(str, imported_file.__doc__)
```
neiljp, line 151: Before you cast, check what type it is, and whether you really need to cast. Then if you do, I'd recommend always adding a comment unless it's really obvious.

What is the type here? Why did you add a cast, given we're not running mypy on this by default?

tools/lint-all
```diff
@@ -19,6 +19,7 @@ tools = {
     "Type consistency (mypy)": "./tools/run-mypy",
     "PEP8 & more (flake8)": ["flake8"] + python_sources,
     "Formatting (black)": ["black", "--check"] + python_sources,
+    "Docstring generation (developer-file-overview is generated using docstrings)": "./tools/lint-docstring",
```
neiljp: I wouldn't expect something which modifies things to be a linter, so save this for the next commit only.

neiljp (commented):
@srdeotarse We discussed this in the stream quite a bit, but I left some feedback here inline of a few more things that stand out.

The restructuring into fix+lint commits makes this a *lot* easier to read and review, simply since each commit is a lot smaller already 👍 Distinct commit text that describes each part helps to explain what is happening in each one, which one can't really do in a larger commit without saying eg. 'line 50: some change' multiple times.

The ordering of functions helps slightly too, though the smaller commits are perhaps a larger factor, at least for me.

I guess your editor is doing type-checking for you? You might also consider adding limited checking of `tools/` to `run-mypy`, as I mention in the stream, as that will enforce things in CI - and there are certainly complaints from mypy right now from a quick test of it :)

tools/lint-docstring
mounilKshah: Okay.

Actually, I tried being overly cautious regarding the type and hence kept the cast function. I tested the code without the cast and it works well. I'll remove it in the updated commit(s).

neiljp: By using `reveal_type` with `mypy`, this is apparently `Optional[str]` - this may be important since some modules may end up having no docstring :)

neiljp (commented):
@mounilKshah This is looking much cleaner now :+1:

There may be more feedback, but time for bed :)

Let's squash the generate commit with the improvements commit - the generate tool doesn't work after the first of those :)

srdeotarse:
@mounilKshah I have reviewed your code and found that it does not display a success message in one case.

mounilKshah:
> it does not display a success message in one case.

I'll add it, thanks for pointing that out.

zulipterminal/server_url.py
```diff
@@ -1,3 +1,7 @@
+"""
+Constructs and encodes server_url of messages.
```
plugyawn, line 2: Could be more detailed?

In a few of these, seems like there's no new info added other than what could be assumed from the filename.

zulipterminal/ui_tools/views.py
```diff
@@ -1,3 +1,7 @@
+"""
+UI views for larger elements such as Streams, Messages, Topics, Help, etc
```
plugyawn, line 2: Maybe some contrast with the definition for ui.py would help.

zulipterminal/config/regexes.py
```diff
@@ -1,3 +1,6 @@
+"""
+Regular expression constants
+"""
 # All regexes with (*) are inspired from webapp's
```
plugyawn, line 4: Missing newline?

plugyawn (commented):
Personally think everything's fine, just a few things I noticed that could be discussed, but I don't really think need to be changed.

tools/lint-docstring
```diff
@@ -33,9 +34,47 @@ DOCSTRINGS_FOR_NO_FILE_FOLDERS = {
...
+        lint_file_overview()
+        print("Successful")
+
+
+def lint_file_overview() -> None:
+    """
+    Used for linting of the developer-file-overview.md document
+    """
+    folder_file_docstring = generate_folder_file_docstrings_dict()
+    computed_doc_dict = extract_docstrings_from_file_overview()
```
Rohitth007, line 51: What exactly is "computed" here? Does `existing_doc_dict` sound better. I'm not sure though. What do others think here?

tools/lint-docstring
```diff
@@ -138,6 +177,32 @@ def extract_folder_file_structure() -> Dict[str, List[str]]:
...
+    containing files and their respective descriptons
+    """
+    with open(DEVELOPER_DOC, "r") as file:
+        doc_data = file.readlines()
+
+    computed_doc_dict: Dict[str, Dict[str, str]] = {}
+
+    # table is present from line 6 to (length - 1)
+    for i in range(TABLE_OFFSET - 1, len(doc_data)):
+        _, folder, filename, docstring, _ = str(doc_data[i]).split("|")
```
Rohitth007, line 193: Use the same slicing iterator you used in the other function here to keep the same style.
`for column in doc_data[TABLE_OFFSET-1:]`

tools/lint-docstring
```diff
@@ -157,6 +222,12 @@ def column_text_with_spaces(
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
-        description="Lint and fix docstrings in all ZulipTerminal files"
+        description="Lint and fix docstrings in all zulip-terminal files"
+    )
+    parser.add_argument(
+        "--fix",
+        action="store_true",
+        help="Update developer-file-overview with docstrings of all files",
```
Rohitth007, line 230: Here the description and help text sound opposite to each other. The description sounds like we are linting and fixing the docstrings based on the markdown file but we are doing the opposite.

Rohitth007: Even the name of the file sounds that way.

tools/lint-docstring
```diff
@@ -33,9 +34,47 @@ DOCSTRINGS_FOR_NO_FILE_FOLDERS = {
...
+        for folder_name in DOCSTRINGS_FOR_NO_FILE_FOLDERS:
+            missing_entity = folder_name
+            assert (
+                computed_doc_dict[folder_name]["no_file_present"]
+                == DOCSTRINGS_FOR_NO_FILE_FOLDERS[folder_name]
+            )
+    except AssertionError:
+        # in case the docstrings do not match
+        print(f"{missing_entity}: docstrings have been changed")
+        print("Run './tools/lint-docstring --fix' to fix")
```
Rohitth007, line 71: Suppose someone changes the markdown file and then runs fix seeing this error, they loose everything they wrote right?

I believe the markdown file is not to be edited after this PR. Should we add a `Generated by code. DO NOT EDIT` warning at the top of the file?

Rohitth007 (changes requested):
Went through your PR. Most of the comments are regarding variable names.
Hope we can merge this soon :+1: .

tools/lint-docstring
```diff
@@ -0,0 +1,162 @@
...
+
+
+def create_file_overview_doc() -> None:
+    """
+    Creates the entire developer-file-overview.md document again by retaining the
+    lines 1 to TABLE_OFFSET and the rest is written using the docstrings in the files
+    """
+    folder_file_docstring = generate_folder_file_docstrings_dict()
+    with open(DEVELOPER_DOC, "r+") as developer_file_read:
+        doc_data = developer_file_read.readlines()
```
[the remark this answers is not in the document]
Rohitth007: read next to a read and write next to a write looks weird, so we can remove at least that?

tools/lint-docstring
```diff
@@ -0,0 +1,150 @@
...
+        dictionary_of_files = folder_file_docstring[folder]
+
+        folder_text = folder
+        for file in sorted(dictionary_of_files):
+            docstring = folder_file_docstring[folder][file]
+            text_tuple = (folder_text, file, docstring)
+            padded_folder_cell, padded_file_cell, padded_desc_cell = pad_column_cells(
+                text_tuple, COLUMN_WIDTHS
+            )
+            new_line = (
```
Rohitth007, line 110: Do you want to use `line` here or `col(umn)`? Both are technically correct but cell, column and table fit a pattern.

tools/lint-docstring
```diff
@@ -0,0 +1,162 @@
...
+        for file in files:
+            imported_file = importlib.import_module(
+                f'{folder.replace("/",".")}.{file[:-3]}'
+            )
+            extracted_docstring = str(imported_file.__doc__)
+            docstring = extracted_docstring.strip().replace("\n", " ")
+            if len(docstring) >= COLUMN_WIDTHS[2] + 1:
+                raise ValueError(
+                    f"{file} has docstring longer than {int(COLUMN_WIDTHS[2])+1}"
+                )
```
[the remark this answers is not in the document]
Rohitth007: Since space isn't included in the docstring, shouldn't we just say `"{file} has docstring longer than to {int(COLUMN_WIDTHS[2])}"`?
Similarily we can remove the = and +1 from the if condition?

tools/lint-docstring
```diff
@@ -0,0 +1,150 @@
...
+    Returns dictionary containing folders and respective python files within them
+    """
+    folders_and_files = {}
+    for path_to_folder in glob.glob(f"{ZULIPTERMINAL}/**/", recursive=True):
+        complete_directory_path = Path(path_to_folder)
+        relative_directory_path = complete_directory_path.relative_to(ROOT_DIRECTORY)
+        if str(relative_directory_path) not in DOCSTRINGS_FOR_NO_FILE_FOLDERS:
+            files_in_directory = [
+                Path(file_name).name
+                for file_name in glob.glob(f"{complete_directory_path}/*.py")
```
Rohitth007, line 171: Probably using `complete_directory_path.glob('*.py')` makes the code simpler here. You can then directly do `file.name`

Rohitth007 (changes requested):
@mounilKshah Left a few more comments following up on the previous review.

tools/lint-docstring
mounilKshah: Considering the pattern, maybe call it `new_row`? As we are adding new _rows_ in each iteration?

Rohitth007:
@mounilKshah Most of the PR looks good to me :+1: .
It would be easier for other people to go through open conversations if you can resolve those have been addressed.
I think there are only a few things left. I'll do some testing but most of the review from my side is done.
Hope we can merge this soon.

tools/lint-docstring
```diff
@@ -0,0 +1,231 @@
...
+                existing_doc_dict[folder_name][filename] = docstring.strip()
+            else:
+                existing_doc_dict[folder_name]["no_file_present"] = docstring.strip()
+    return existing_doc_dict
+
+
+def pad_column_cells(
+    column_text: Tuple[str, str, str],
+    columnwise_widths: Tuple[int, int, int] = COLUMN_WIDTHS,
+) -> Tuple[str, ...]:
```
andersk, line 207: This function is unnecessary. Python f-strings have this built in.

```pycon
>>> folder = "zulipterminal"
>>> file = "api_types.py"
>>> docstring = "Preliminary Zulip API types defined in python, to allow type checking"
>>> print(f"| {folder:23}| {file:20}| {docstring:103}|\n")
| zulipterminal          | api_types.py        | Preliminary Zulip API types defined in python, to allow type checking                                  |
```

(The widths can be computed expressions too. [Documentation](https://docs.python.org/3/reference/lexical_analysis.html#f-strings).)

Rohitth007: Aah, I remember using this a while back but it didn't cross my mind. Let's use this.

tools/lint-docstring
```diff
@@ -0,0 +1,162 @@
...
+    """
+    Returns the text parsed alongwith the adequate amount of spaces
+    after the text
+    """
+    final_text_list = []
+    for iterator in range(0, len(column_text)):
+        given_text = column_text[iterator]
```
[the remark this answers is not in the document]
Rohitth007: This can be closed.

tools/lint-docstring
```diff
@@ -0,0 +1,162 @@
...
+    return folder_file_docstring
+
+
+def extract_folder_file_structure() -> Dict[str, List[str]]:
+    """
+    Returns dictionary containing folders and respective python files within them
+    """
+    # contains path to the folders to be parsed to search for files
+    folder_paths = {}
+    for path_to_folder in glob.glob(f"{ZULIPTERMINAL}/**", recursive=True):
```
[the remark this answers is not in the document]
Rohitth007: This also can be closed.

tools/lint-docstring
mounilKshah: Thank you for this suggestion @andersk 💯

tools/lint-docstring
```diff
@@ -0,0 +1,117 @@
...
+# absolute path to zulip-terminal/zulipterminal to be passed as parameter
+ZULIPTERMINAL = ROOT_DIRECTORY / "zulipterminal"
+
+# new doc file has been created for the time being to compare it with the original
+DEVELOPER_DOC = ROOT_DIRECTORY / "docs" / "developer-file-overview.md"
+
+# zulipterminal/scripts and zulipterminal/themes do not have any files
+# mentioned in the file-overview. Hence, their description has been
+# added separately.
+DOCSTRINGS_WITHOUT_FILES = {
```
Rohitth007, line 32: Was this name change to ensure everything fits on 1 line?
Maybe `DESC_FOR_NO_FILE_FOLDERS` is a better name?

neiljp (commented):
@mounilKshah Latest round of feedback here. There's still a minor issue with updating the file when it shouldn't, that I thought we cleared up before. Did you drop a change which was fixing that? I do have a patch to fix it if necessary, since I was experimenting with avoiding __pycache__ folders.

neiljp:
@mounilKshah I see you didn't change the labels, but since you pushed, I wanted to check if you meant this to be ready for review? Did you intend to push a different version than this? This looks like the previous version and doesn't address the comments, so just wanted to check.

zulipterminal/config/color.py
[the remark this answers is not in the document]
neiljp: Resolving this comment now I've opened #1246.

tools/lint-docstring
```diff
@@ -33,9 +34,47 @@ DOCSTRINGS_FOR_NO_FILE_FOLDERS = {
 BLANK_LINE = f"| {COLUMN_WIDTHS[0]*SPACEBAR}| {COLUMN_WIDTHS[1]*SPACEBAR}| {COLUMN_WIDTHS[2]*SPACEBAR}|\n"
 
 
-def main() -> None:
-    create_file_overview_doc()
-    print("Generated developer-file-overview.md successfully.")
+def main(fix_file: bool = False) -> None:
+    if fix_file:
+        create_file_overview_doc()
+        print("Generated developer-file-overview.md successfully.")
```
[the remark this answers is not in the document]
neiljp: There were some references like this left behind just now, so I rolled those into my partial merge.

tools/lint-docstring
neiljp: @mounilKshah We didn't address this before I did the partial merge, but an extra commit updating the overview file and adjusting the fixing code to take that into account should be doable?

tools/lint-docstring
```diff
@@ -7,6 +7,8 @@ import sys
 from pathlib import Path
 from typing import Dict, List, Tuple
 
+from zulipterminal.cli.run import in_color
```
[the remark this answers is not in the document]
neiljp: While color is nice, it doesn't always render well depending on the background, or in different situations like `tools/lint-docstring|less`. This would be cleanly solved if we used something like the `crayons` library.

Weren't the color changes previously in another commit? I think it'd be cleaner to have that separately, if we do end up doing so.

In terms of use in the main app, we could move to crayons there too, though ultimately I hope we can move to a loading screen instead, ie. load urwid very early, as explored in at least one PR.

tools/lint-docstring
```diff
@@ -120,9 +177,40 @@ def extract_folder_file_structure() -> Dict[str, List[str]]:
...
+    # table is present from line 6 to (length - 1)
+    for doc_row in doc_data[TABLE_OFFSET - 1 :]:
+        _, folder, filename, docstring, _ = doc_row.split("|")
+        folder, filename = folder.strip(), filename.strip()
+        if folder:
+            folder_name = folder
+            existing_doc_dict[str(folder_name)] = {}
+        if docstring.strip():
+            if filename:
+                existing_doc_dict[folder_name][filename] = docstring.strip()
```
neiljp, line 200: Am I understanding correctly that folder_name is a variable retained through the loop to cope with no folder value when there are files? (eg. the core.py row)
If so, that deserves a comment.

This code combined *could* lead to folder_name being undefined here, if the table was not formatted correctly. It works now as there's a folder at the top of the file.

tools/lint-docstring
```diff
@@ -35,9 +37,64 @@ DESC_FOR_NO_FILE_FOLDERS = {
...
+        print("Run './tools/lint-docstring --fix' to update " + DEVELOPER_DOC_NAME)
+        sys.exit(1)
+    except KeyError:
+        # in case a new file is added and its docstring is not present in the document
+        print(
+            in_color(
+                "red",
+                missing_entity
+                + ": docstrings have been added but not listed in "
+                + DEVELOPER_DOC_NAME,
```
neiljp, line 90: This is also triggered if a docstring/file is **removed**, eg. removing the scripts or themes entry.

tools/lint-docstring
```diff
@@ -35,9 +37,64 @@ DESC_FOR_NO_FILE_FOLDERS = {
...
+                == DESC_FOR_NO_FILE_FOLDERS[folder_name]
+            )
+    except AssertionError:
+        # in case the docstrings do not match
+        print(
+            in_color(
+                "red",
+                missing_entity
+                + ": docstrings do not matched those listed in "
+                + DEVELOPER_DOC_NAME,
```
neiljp, line 78: Is there a reason you're not using fstrings here and elsewhere in this commit?

tools/lint-docstring
```diff
@@ -35,9 +37,64 @@ DESC_FOR_NO_FILE_FOLDERS = {
...
+                    computed_doc_dict[folder][file]
+                    == folder_file_docstring[folder][file]
+                )
+        # zulipterminal/themes and zulipterminal/scripts are linted using DESC_FOR_NO_FILE_FOLDERS
+        for folder_name in DESC_FOR_NO_FILE_FOLDERS:
+            missing_entity = folder_name
+            assert (
+                computed_doc_dict[folder_name]["no_file_present"]
+                == DESC_FOR_NO_FILE_FOLDERS[folder_name]
+            )
```
neiljp, line 70: These assert statements do work, with the downside that they only show one error result, even if there are more than one.

neiljp:
@mounilKshah This is at a milestone where I'm tempted to merge, but we discussed a few things in the stream after your push, and there are some points above. To summarize, I think the outstanding things are:
1. ideally, it would be beneficial to show all issues, rather than just the first, by accumulating the issues and outputting at the end, rather than exiting on assert or failing to index the array
2. As part of (1) it likely would fall out naturally that we would use conditionals instead of exceptions, which would be a pleasant side effect
3. Re the coloring, the dependency needs adding with care to one of our dev dependencies (typically separate commit to setup.py), and then I think it's worth applying color across various tools in one go. That's not so important, so we can take the last commit or other working code you have and push into another PR - it'd be good to have to build on later (by you or someone else)
4. Add a 'do not modify manually' section, as per discussion here https://github.com/zulip/zulip-terminal/pull/1208#discussion_r894330639 This would be similar to the section at the top of hotkeys.md
5. I just discovered that adding a file triggers the linter, is fixed by the fixer, but removing the file doesn't trigger the linter - the document is left as it was!

If you can split out (3) into a separate PR, and fix (5), I'd be happy to merge.

I expect (4) might be simple as an extra commit after the linting commit; it's not a functional change, but useful.

While (1) and (2) could improve the linting feedback, it does fail to lint when there's an error, and says what to do to fix it, so I'm less concerned about fixing that - unless you already have worked further on it.

neiljp:
@mounilKshah Thanks for this last update - this seems to function just fine, including with multiple file differences :+1:

I've adjusted some text, switched some parts to fstrings, and updated the variable name as per the comment at https://github.com/zulip/zulip-terminal/pull/1208#discussion_r894316045
The full differences are available via `range-diff` as always.

I'm merging now the tests are done :tada:

PR #1210, opened 2022-04-19
hotkeys: Add linting for hotkeys.

### LINKED ISSUE #1209: hotkeys: Add linting for hotkeys.
Currently, there is no linting for hotkeys.

For e.g. if ``help_text`` in ``keys.py`` ends with ``.``, ``./tools/generate_hotkeys.py`` updates ``help_text`` without linting it as 
``key description`` in help menu doesn't contain ``.``

srdeotarse:
Fixes #1209

Adds GitHub Actions and changes to ``lint-all`` for linting ``generate_hotkeys.py``

**What does this PR do?**  

**Tested?** 
- [x] Manually

**Commit flow** 

**Notes & Questions** 

**Interactions** 

**Visual changes**

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -83,6 +83,23 @@ jobs:
       - name: Check code & tests meet black standards
         run: black --check zulipterminal/ tests/ setup.py `tools/python_tools.py`
 
+  hotkeys:
+    runs-on: ubuntu-latest
+    name: Lint - Hotkeys linting & docs sync check
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.6
+          cache: 'pip'
+          cache-dependency-path: 'setup.py'
+      - name: Minimal install
+        run: pip install .
+      - name: Run generate_hotkeys
+        run: ./tools/generate_hotkeys.py --check-only
+
   pytest:
     strategy:
       # Not failing fast allows all matrix jobs to try & finish even if one fails early
--- a/tools/generate_hotkeys.py
+++ b/tools/generate_hotkeys.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python3
+import re
+import sys
 from collections import defaultdict
 from pathlib import Path, PurePath
 from typing import Dict, List, Tuple
@@ -13,10 +15,40 @@ KEYS_FILE_NAME = KEYS_FILE.name
 OUTPUT_FILE = Path(__file__).resolve().parent.parent / "docs" / "hotkeys.md"
 OUTPUT_FILE_NAME = OUTPUT_FILE.name
 SCRIPT_NAME = PurePath(__file__).name
+HELP_TEXT_STYLE = re.compile(r"^[a-zA-Z /()',&@#:_-]*$")
 
 
 def main() -> None:
     generate_hotkeys_file()
+
+
+def lint_hotkeys_file() -> None:
+    """
+    Lint KEYS_FILE for key description, then compare if in sync with
+    existing OUTPUT_FILE
+    """
+    hotkeys_file_string = get_hotkeys_file_string()
+    # To lint keys description
+    error_flag = 0
+    categories = read_help_categories()
+    for action in HELP_CATEGORIES.keys():
+        for help_text, key_combinations_list in categories[action]:
+            various_key_combinations = " / ".join(key_combinations_list)
+            # Check description style
+            if not re.match(HELP_TEXT_STYLE, help_text):
+                print(
+                    f"Description - ({help_text}) for key combination - [{various_key_combinations}]\n"
+                    "It should contain only alphabets, spaces and special characters except ."
+                )
+                error_flag = 1
+    if error_flag == 1:
+        print(f"Rerun this command after resolving errors in config/{KEYS_FILE_NAME}")
+    else:
+        print("No hotkeys linting errors")
+        if not output_file_matches_string(hotkeys_file_string):
+            print(f"Run './tools/{SCRIPT_NAME}' to update {OUTPUT_FILE_NAME} file")
+            error_flag = 1
+    sys.exit(error_flag)
 
 
 def generate_hotkeys_file() -> None:
--- a/tools/generate_hotkeys.py
+++ b/tools/generate_hotkeys.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import argparse
 import re
 import sys
 from collections import defaultdict
@@ -18,8 +19,11 @@ SCRIPT_NAME = PurePath(__file__).name
 HELP_TEXT_STYLE = re.compile(r"^[a-zA-Z /()',&@#:_-]*$")
 
 
-def main() -> None:
-    generate_hotkeys_file()
+def main(check_only: bool) -> None:
+    if check_only:
+        lint_hotkeys_file()
+    else:
+        generate_hotkeys_file()
 
 
 def lint_hotkeys_file() -> None:
@@ -118,4 +122,15 @@ def write_hotkeys_file(hotkeys_file_string: str) -> None:
 
 
 if __name__ == "__main__":
-    main()
+    parser = argparse.ArgumentParser(
+        description=f"Generate {OUTPUT_FILE_NAME} file by extracting key description and key "
+        f"combination from config/{KEYS_FILE_NAME} file"
+    )
+    parser.add_argument(
+        "--check-only",
+        action="store_true",
+        help=f"Lint hotkeys by checking extracted key description style and key "
+        f"duplication from config/{KEYS_FILE_NAME} file",
+    )
+    args = parser.parse_args()
+    main(args.check_only)
--- a/tools/generate_hotkeys.py
+++ b/tools/generate_hotkeys.py
@@ -6,18 +6,12 @@ from typing import Dict, List, Tuple
 from zulipterminal.config.keys import HELP_CATEGORIES, KEY_BINDINGS
 
 
-def read_help_categories() -> Dict[str, List[Tuple[str, List[str]]]]:
-    """
-    Get all help categories from keys.py
-    """
-    categories = defaultdict(list)
-    for item in KEY_BINDINGS.values():
-        categories[item["key_category"]].append((item["help_text"], item["keys"]))
-    return categories
-
-
 OUTPUT_FILE = Path(__file__).resolve().parent.parent / "docs" / "hotkeys.md"
 SCRIPT_NAME = PurePath(__file__).name
+
+
+def main() -> None:
+    generate_hotkeys_file()
 
 
 def generate_hotkeys_file() -> None:
@@ -58,16 +52,22 @@ def get_hotkeys_file_string() -> str:
     return hotkeys_file_string
 
 
+def read_help_categories() -> Dict[str, List[Tuple[str, List[str]]]]:
+    """
+    Get all help categories from keys.py
+    """
+    categories = defaultdict(list)
+    for item in KEY_BINDINGS.values():
+        categories[item["key_category"]].append((item["help_text"], item["keys"]))
+    return categories
+
+
 def write_hotkeys_file(hotkeys_file_string: str) -> None:
     """
     Write hotkeys_file_string variable once to hotkeys.md file
     """
     with open(OUTPUT_FILE, "w") as mdFile:
         mdFile.write(hotkeys_file_string)
-
-
-def main() -> None:
-    generate_hotkeys_file()
 
 
 if __name__ == "__main__":
--- a/tools/generate_hotkeys.py
+++ b/tools/generate_hotkeys.py
@@ -18,6 +18,9 @@ OUTPUT_FILE_NAME = OUTPUT_FILE.name
 SCRIPT_NAME = PurePath(__file__).name
 HELP_TEXT_STYLE = re.compile(r"^[a-zA-Z /()',&@#:_-]*$")
 
+# Exclude keys from duplicate keys checking
+KEYS_TO_EXCLUDE = ["q", "e", "m", "r"]
+
 
 def main(check_only: bool) -> None:
     if check_only:
@@ -36,7 +39,9 @@ def lint_hotkeys_file() -> None:
     error_flag = 0
     categories = read_help_categories()
     for action in HELP_CATEGORIES.keys():
+        check_duplicate_keys_list: List[str] = []
         for help_text, key_combinations_list in categories[action]:
+            check_duplicate_keys_list.extend(key_combinations_list)
             various_key_combinations = " / ".join(key_combinations_list)
             # Check description style
             if not re.match(HELP_TEXT_STYLE, help_text):
@@ -45,6 +50,20 @@ def lint_hotkeys_file() -> None:
                     "It should contain only alphabets, spaces and special characters except ."
                 )
                 error_flag = 1
+        # Check key combination duplication
+        check_duplicate_keys_list = [
+            key for key in check_duplicate_keys_list if key not in KEYS_TO_EXCLUDE
+        ]
+        duplicate_keys = [
+            key
+            for key in check_duplicate_keys_list
+            if check_duplicate_keys_list.count(key) > 1
+        ]
+        if len(duplicate_keys) != 0:
+            print(
+                f"Duplicate key combination for keys {duplicate_keys} for category ({HELP_CATEGORIES[action]}) detected"
+            )
+            error_flag = 1
     if error_flag == 1:
         print(f"Rerun this command after resolving errors in config/{KEYS_FILE_NAME}")
     else:
--- a/tools/generate_hotkeys.py
+++ b/tools/generate_hotkeys.py
@@ -1,37 +1,74 @@
 #!/usr/bin/env python3
 from collections import defaultdict
 from pathlib import Path, PurePath
+from typing import Dict, List, Tuple
 
 from zulipterminal.config.keys import HELP_CATEGORIES, KEY_BINDINGS
 
 
-categories = defaultdict(list)
-for item in KEY_BINDINGS.values():
-    categories[item["key_category"]].append((item["help_text"], item["keys"]))
+def read_help_categories() -> Dict[str, List[Tuple[str, List[str]]]]:
+    """
+    Get all help categories from keys.py
+    """
+    categories = defaultdict(list)
+    for item in KEY_BINDINGS.values():
+        categories[item["key_category"]].append((item["help_text"], item["keys"]))
+    return categories
+
 
 OUTPUT_FILE = Path(__file__).resolve().parent.parent / "docs" / "hotkeys.md"
 SCRIPT_NAME = PurePath(__file__).name
 
-hotkeys_file_string = (
-    f"<!--- Generated automatically by tools/{SCRIPT_NAME} -->\n"
-    "<!--- Do not modify -->\n\n# Hot Keys\n"
-)
-for action in HELP_CATEGORIES.keys():
-    hotkeys_file_string += (
-        f"## {HELP_CATEGORIES[action]}\n"
-        "|Command|Key Combination|\n"
-        "| :--- | :---: |\n"
-    )
-    for help_text, key_combinations_list in categories[action]:
-        various_key_combinations = " / ".join(
-            [
-                " + ".join([f"<kbd>{key}</kbd>" for key in key_combination.split()])
-                for key_combination in key_combinations_list
-            ]
-        )
-        hotkeys_file_string += f"|{help_text}|{various_key_combinations}|\n"
-    hotkeys_file_string += "\n"
-with open(OUTPUT_FILE, "w") as mdFile:
-    mdFile.write(hotkeys_file_string)
 
-print(f"Hot Keys list saved in {OUTPUT_FILE}")
+def generate_hotkeys_file() -> None:
+    """
+    Generate hotkeys.md in docs folder based on help text description and
+    shortcut key combinations in config/keys.py file
+    """
+    hotkeys_file_string = get_hotkeys_file_string()
+    write_hotkeys_file(hotkeys_file_string)
+    print(f"Hot Keys list saved in {OUTPUT_FILE}")
+
+
+def get_hotkeys_file_string() -> str:
+    """
+    Construct string in form for output to hotkeys.md in docs folder based on help text
+    description and shortcut key combinations in config/keys.py file
+    """
+    categories = read_help_categories()
+    hotkeys_file_string = (
+        f"<!--- Generated automatically by tools/{SCRIPT_NAME} -->\n"
+        "<!--- Do not modify -->\n\n# Hot Keys\n"
+    )
+    for action in HELP_CATEGORIES.keys():
+        hotkeys_file_string += (
+            f"## {HELP_CATEGORIES[action]}\n"
+            "|Command|Key Combination|\n"
+            "| :--- | :---: |\n"
+        )
+        for help_text, key_combinations_list in categories[action]:
+            various_key_combinations = " / ".join(
+                [
+                    " + ".join([f"<kbd>{key}</kbd>" for key in key_combination.split()])
+                    for key_combination in key_combinations_list
+                ]
+            )
+            hotkeys_file_string += f"|{help_text}|{various_key_combinations}|\n"
+        hotkeys_file_string += "\n"
+    return hotkeys_file_string
+
+
+def write_hotkeys_file(hotkeys_file_string: str) -> None:
+    """
+    Write hotkeys_file_string variable once to hotkeys.md file
+    """
+    with open(OUTPUT_FILE, "w") as mdFile:
+        mdFile.write(hotkeys_file_string)
+
+
+def main() -> None:
+    generate_hotkeys_file()
+
+
+if __name__ == "__main__":
+    main()
--- a/tools/generate_hotkeys.py
+++ b/tools/generate_hotkeys.py
@@ -12,25 +12,26 @@ for item in KEY_BINDINGS.values():
 OUTPUT_FILE = Path(__file__).resolve().parent.parent / "docs" / "hotkeys.md"
 SCRIPT_NAME = PurePath(__file__).name
 
-with open(OUTPUT_FILE, "w") as mdFile:
-    mdFile.write(
-        f"<!--- Generated automatically by tools/{SCRIPT_NAME} -->\n"
-        "<!--- Do not modify -->\n\n# Hot Keys\n"
+hotkeys_file_string = (
+    f"<!--- Generated automatically by tools/{SCRIPT_NAME} -->\n"
+    "<!--- Do not modify -->\n\n# Hot Keys\n"
+)
+for action in HELP_CATEGORIES.keys():
+    hotkeys_file_string += (
+        f"## {HELP_CATEGORIES[action]}\n"
+        "|Command|Key Combination|\n"
+        "| :--- | :---: |\n"
     )
-    for action in HELP_CATEGORIES.keys():
-        mdFile.write(
-            f"## {HELP_CATEGORIES[action]}\n"
-            "|Command|Key Combination|\n"
-            "| :--- | :---: |\n"
+    for help_text, key_combinations_list in categories[action]:
+        various_key_combinations = " / ".join(
+            [
+                " + ".join([f"<kbd>{key}</kbd>" for key in key_combination.split()])
+                for key_combination in key_combinations_list
+            ]
         )
-        for help_text, key_combinations_list in categories[action]:
-            various_key_combinations = " / ".join(
-                [
-                    " + ".join([f"<kbd>{key}</kbd>" for key in key_combination.split()])
-                    for key_combination in key_combinations_list
-                ]
-            )
-            mdFile.write(f"|{help_text}|{various_key_combinations}|\n")
-        mdFile.write("\n")
+        hotkeys_file_string += f"|{help_text}|{various_key_combinations}|\n"
+    hotkeys_file_string += "\n"
+with open(OUTPUT_FILE, "w") as mdFile:
+    mdFile.write(hotkeys_file_string)
 
 print(f"Hot Keys list saved in {OUTPUT_FILE}")
--- a/tools/generate_hotkeys.py
+++ b/tools/generate_hotkeys.py
@@ -21,8 +21,8 @@ def main() -> None:
 
 def generate_hotkeys_file() -> None:
     """
-    Generate hotkeys.md in docs folder based on help text description and
-    shortcut key combinations in config/keys.py file
+    Generate OUTPUT_FILE based on help text description and
+    shortcut key combinations in KEYS_FILE
     """
     hotkeys_file_string = get_hotkeys_file_string()
     output_file_matches_string(hotkeys_file_string)
@@ -32,8 +32,8 @@ def generate_hotkeys_file() -> None:
 
 def get_hotkeys_file_string() -> str:
     """
-    Construct string in form for output to hotkeys.md in docs folder based on help text
-    description and shortcut key combinations in config/keys.py file
+    Construct string in form for output to OUTPUT_FILE based on help text
+    description and shortcut key combinations in KEYS_FILE
     """
     categories = read_help_categories()
     hotkeys_file_string = (
@@ -69,7 +69,7 @@ def output_file_matches_string(hotkeys_file_string: str) -> bool:
 
 def read_help_categories() -> Dict[str, List[Tuple[str, List[str]]]]:
     """
-    Get all help categories from keys.py
+    Get all help categories from KEYS_FILE
     """
     categories = defaultdict(list)
     for item in KEY_BINDINGS.values():
@@ -79,7 +79,7 @@ def read_help_categories() -> Dict[str, List[Tuple[str, List[str]]]]:
 
 def write_hotkeys_file(hotkeys_file_string: str) -> None:
     """
-    Write hotkeys_file_string variable once to hotkeys.md file
+    Write hotkeys_file_string variable once to OUTPUT_FILE
     """
     with open(OUTPUT_FILE, "w") as mdFile:
         mdFile.write(hotkeys_file_string)
--- a/tools/generate_hotkeys.py
+++ b/tools/generate_hotkeys.py
@@ -1,17 +1,14 @@
 #!/usr/bin/env python3
+from collections import defaultdict
 from pathlib import Path, PurePath
 
 from zulipterminal.config.keys import HELP_CATEGORIES, KEY_BINDINGS
 
 
-categories = {
-    category: {
-        item["help_text"]: item["keys"]
-        for item in KEY_BINDINGS.values()
-        if item["key_category"] == category
-    }
-    for category in HELP_CATEGORIES.keys()
-}
+categories = defaultdict(list)
+for item in KEY_BINDINGS.values():
+    categories[item["key_category"]].append((item["help_text"], item["keys"]))
+
 OUTPUT_FILE = Path(__file__).resolve().parent.parent / "docs" / "hotkeys.md"
 SCRIPT_NAME = PurePath(__file__).name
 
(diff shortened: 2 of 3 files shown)


neiljp:
@srdeotarse This contains quite a few different types of changes in one commit; it would be more useful to split each change into its own commit - they can always be squashed together later, and often reordered too, if independent.

Please check out my comment in #1209, but this PR currently contains elements to:
- `make fix` one additional tools/ file itself; this doesn't seem so useful, and we might consider multiple files (eg. more of `tools/` in general). I suspect this was confusion over my thought of *running* the script in `make fix` and my third point in #1209 
- parse the help text and lint it for errors; this was my second comment in #1209 
- run mypy on one file (fits with the first bullet here again)

This doesn't actually test whether the hotkeys are in sync in CI yet (first point in #1209).

I hope this makes it clearer! Please discuss in the stream further if still unclear :)

srdeotarse:

tools/generate_hotkeys.py
```diff
@@ -12,28 +21,51 @@
...
+            for key in keys_list:
+                if key in KEYS_TO_EXCLUDE:
+                    keys_list.remove(key)
+            # Check key combination duplication
+            duplicate_keys = [key for key in keys_list if keys_list.count(key) > 1]
+            if len(duplicate_keys) != 0:
+                print(
+                    f"Duplicate key combination for keys {duplicate_keys} for category ({HELP_CATEGORIES[action]})"
+                )
+                sys.exit(1)
```
Rohitth007, line 59: I suppose duplicate checking need not be done for each `help_text` inside the for loop?

tools/generate_hotkeys.py
```diff
@@ -12,28 +21,51 @@
...
+    # print("keys_list -", keys_list)
     for action in categories.keys():
         mdFile.write(
             f"## {HELP_CATEGORIES[action]}\n"
             "|Command|Key Combination|\n"
             "| :--- | :---: |\n"
         )
+        keys_list = []
+        for help_text, key_combinations_list in categories[action].items():
+            keys_list.append(key_combinations_list)
```
Rohitth007, line 42: Continuing the previous comment, if `for key in keys_list` can be pulled out of the for loop then I suppose the above can also be combined into the for loop below?

tools/generate_hotkeys.py
```diff
@@ -12,28 +21,51 @@
...
+OUTPUT_FILE = ROOT_DIRECTORY / "docs" / "hotkeys.md"
+SCRIPT_NAME = pathlib.PurePath(__file__).name
 
 with open(OUTPUT_FILE, "w") as mdFile:
     mdFile.write(
         f"<!--- Generated automatically by tools/{SCRIPT_NAME} -->\n"
         "<!--- Do not modify -->\n\n# Hot Keys\n"
     )
+    # keys_list = [item["keys"] for item in KEY_BINDINGS.values()]
+    # print("keys_list -", keys_list)
```
Rohitth007, line 33: Any reason for keeping the above comments?

Rohitth007 (commented):
Did a quick review and left a few comments below. I haven't test it out yet though.

tools/generate_hotkeys.py
Rohitth007: I meant to say, if we can move duplicate checking to the end, then we can do everything in 1 for loop itself.

tools/generate_hotkeys.py
```diff
@@ -26,14 +35,34 @@
             "|Command|Key Combination|\n"
             "| :--- | :---: |\n"
         )
+        keys_list = []
+        for help_text, key_combinations_list in categories[action].items():
+            keys_list.append(key_combinations_list)
```
Rohitth007, line 40: Here, do you mean to `extend` instead of `append`? As appending would not check if, `q` for example is also in ['ctrl a', 'q']

srdeotarse: I will change it to extend

Rohitth007: Not that you have to change the rest of the `key_list` logic as well.

Rohitth007:
@srdeotarse It would be easier to review your PRs, if you start resolving any of the open comments there might be if they are resolved, unless you are waiting for a response or want a second opinion. It can get confusing to see many open comments, especially if only partly addressed. (Not specific to this PR)

Apart from that, I wanted to ask if we want to do duplicate checking per category or over all keys. Currently it seems to be for per category.

srdeotarse:
@Rohitth007 I have received no confirmation from @neiljp about duplicate checking per category or over all keys. I have suggested for per category because the same shortcut may work based on where the user is currently in ZT (for e.g. ``esc`` to exit from compose area, stream messages or stream search)

Thank you for reviewing my code and the point noted about resolving the conversation.

tools/generate_hotkeys.py
```diff
@@ -1,8 +1,17 @@
...
+from importlib.machinery import SourceFileLoader
 
-from zulipterminal.config.keys import HELP_CATEGORIES, KEY_BINDINGS
 
+ROOT_DIRECTORY = pathlib.Path(__file__).resolve().parent.parent
+keys_file = SourceFileLoader(
+    "keys", f"{ROOT_DIRECTORY}/zulipterminal/config/keys.py"
+).load_module()
+HELP_CATEGORIES = keys_file.HELP_CATEGORIES
+KEY_BINDINGS = keys_file.KEY_BINDINGS
```
neiljp, line 13: What's the motivation for this refactoring?

tools/generate_hotkeys.py
```diff
@@ -26,14 +35,33 @@
             "|Command|Key Combination|\n"
             "| :--- | :---: |\n"
         )
+        keys_list = []
```
neiljp, line 38: If this is just for checking duplicates, please name it more specifically than this.

tools/generate_hotkeys.py
```diff
@@ -26,14 +35,33 @@
...
         for help_text, key_combinations_list in categories[action].items():
+            keys_list.extend(key_combinations_list)
+            for key in key_combinations_list:
+                if key in KEYS_TO_EXCLUDE:
+                    keys_list.remove(key)
```
neiljp, line 43: This may be simpler with a comprehension?

tools/generate_hotkeys.py
```diff
@@ -26,14 +35,33 @@
...
-            mdFile.write(f"|{help_text}|{various_key_combinations}|\n")
+            # Check description style
+            if re.match(r"^[a-zA-Z\s/()',&@#:_-]*$", help_text):
+                mdFile.write(f"|{help_text}|{various_key_combinations}|\n")
+            else:
+                print(
+                    f"Description - ({help_text}) for key combination - [{various_key_combinations}] should contain only alphabets, spaces and special characters except ."
+                )
+                print(f"Hot Keys list not saved in {OUTPUT_FILE}")
+                exit()
```
neiljp, line 58: Exit with what value?

neiljp (commented):
@srdeotarse This has potential, but is a little confusing still. This is partially since in the first commit you're adding the linting code, but also changing other unrelated things - if you think things can be tidied, place this into an earlier commit, and we can consider that separately.

The first commit appears to both try to (re)generate the file but also error out based on linting for description and duplication. Then the second hooks that into lint-all, and the third adds it into CI.

Most importantly, I think it's important to separate the current regeneration of the hotkeys.md file from the linting. If just linting, we shouldn't make any changes. It's still OK if we give errors during linting or regeneration of the file. I'd suggest something like adding a `--check-only` parameter, sort of like the opposite of what is in #1208, and only check for errors and not write the file in that case.

I tried to express in #1209 that I wasn't just interested in checking for duplication and description text, though this has use, but also in particular whether the keys.py and hotkeys.md were in sync. For example, if running the generate script creates a file which is different from the current one, the script wasn't run in advance. To enable this, perhaps with another option, you might consider writing to a different specified location and then doing a compare after that.

tools/generate_hotkeys.py
srdeotarse: GitHub Actions cannot import from ``zulipterminal.config.keys`` thus the refactoring of imports.

tools/generate_hotkeys.py
Rohitth007: You can even bring it out of the for loop to do it just the once.

tools/generate_hotkeys.py
Rohitth007: Not sure I like the new name. Can't think of a better one right now though.

tools/generate_hotkeys.py
srdeotarse: Sure, I will bring that comprehension out of the for loop.

tools/generate_hotkeys.py
neiljp: This kind of explanation belongs in the commit text, and possibly a comment in the file.

Unless I missed it, you didn't discuss what error you may have had and why you took this approach.

It's good you separated this out into another commit, since it's clearer it's a refactor.

tools/generate_hotkeys.py
```diff
@@ -13,34 +17,135 @@
...
+    """
+    Get all help categories from keys.py
+    """
+    categories = {
+        category: {
+            item["help_text"]: item["keys"]
+            for item in KEY_BINDINGS.values()
+            if item["key_category"] == category
+        }
+        for category in HELP_CATEGORIES.keys()
```
Rohitth007, line 41: What is the reasoning here for using the second `dict` inside the main `dict`?
Later `categories[action]` is anyway converted into a list of tuples using `items()`.
So why not just store it as a list of tuples in the first place?

Another point here is, you are looping through the large ~70 `KEY_BINDINGS` for every category (6). Instead why not try using a `defaultdict(list)` and run through `KEY_BINDINGS` only once?

tools/generate_hotkeys.py
```diff
@@ -13,34 +17,135 @@
...
+def compare_temp_hotkeys_file(generate_flag: bool = False) -> None:
+    """
+    To compare temp_hotkeys and hotkeys file.
+    """
+    if filecmp.cmp(ORIGINAL_OUTPUT_FILE, TEMP_OUTPUT_FILE, shallow=False):
+        print("Hot Keys already in sync with config/keys.py")
+    else:
+        print("Hot Keys not in sync with config/keys.py")
+        if not generate_flag:
+            exit(1)
```
Rohitth007, line 68: Here why do we want to generate the `ORIGINAL_OUTPUT_FILE`, i.e., not exit, even though it's not in sync?

tools/generate_hotkeys.py
```diff
@@ -13,34 +17,135 @@
...
+    for action in categories.keys():
+        check_duplicate_keys_list: List[str] = []
         for help_text, key_combinations_list in categories[action].items():
-            various_key_combinations = " / ".join(
-                [
-                    " + ".join([f"<kbd>{key}</kbd>" for key in key_combination.split()])
-                    for key_combination in key_combinations_list
-                ]
+            check_duplicate_keys_list.extend(key_combinations_list)
+            check_duplicate_keys_list = [
```
Rohitth007, line 119: I believe this can be pulled out of the inner for loop?

tools/generate_hotkeys.py
```diff
@@ -13,34 +17,135 @@
...
+                key for key in check_duplicate_keys_list if key not in KEYS_TO_EXCLUDE
+            ]
+            various_key_combinations = various_key_combination(key_combinations_list)
+            # Check description style
+            if not re.match(r"^[a-zA-Z\s/()',&@#:_-]*$", help_text):
```
Rohitth007, line 124: I'm not aware of the style here. Is it basically everything except numbers and full stop?
(as the above doesn't have all special characters?)

Rohitth007: You can also use `re.compile` once before the for loop to speed up the regex matching.

tools/generate_hotkeys.py
```diff
@@ -13,34 +17,135 @@
...
+                error_flag = 1
+        # Check key combination duplication
+        duplicate_keys = [
+            key
+            for key in check_duplicate_keys_list
+            if check_duplicate_keys_list.count(key) > 1
+        ]
+        if len(duplicate_keys) != 0:
+            print(
+                f"Duplicate key combination for keys {duplicate_keys} for category ({HELP_CATEGORIES[action]})"
```
Rohitth007, line 137: We might want to add "detected." at the end of this print statement?

tools/generate_hotkeys.py
```diff
@@ -13,34 +17,135 @@
...
+    exit(error_flag)
 
-print(f"Hot Keys list saved in {ORIGINAL_OUTPUT_FILE}")
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Lint and generate hotkeys")
+    parser.add_argument(
+        "--check-only",
+        action="store_true",
+        help="Check hotkeys file with key description and key duplication from keys file",
```
Rohitth007, line 148: This help text is a bit hard to read and understand. Maybe something like `"Lint hotkeys.md for key description style and key duplication"`?

tools/generate_hotkeys.py
```diff
@@ -13,34 +17,135 @@
...
-        mdFile.write("\n")
+            error_flag = 1
+    exit(error_flag)
 
-print(f"Hot Keys list saved in {ORIGINAL_OUTPUT_FILE}")
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Lint and generate hotkeys")
+    parser.add_argument(
+        "--check-only",
```
Rohitth007, line 146: Does `"--lint-only"` sound better here than `"--check-only"`?

Rohitth007 (changes requested):
Seems like a lot has changed since the last time I looked at it. The PR is starting to look good :+1: 
Apart from the comments below, there were a few points I wanted to make:
* It would improve readability, if you could order the functions in the order in which they are discovered, i.e., first `main()` then `lint_hotkeys_file()` then `generate_hotkeys_file()` and so on.
* Most of your commit texts are empty. It helps in reviewing, if there is a little more detail.
* You also might want to update your docstrings as some are outdated. (Eg. `check_hotkeys.md`)

tools/generate_hotkeys.py
neiljp: `isort` uses `--check-only`, which is where I heard it before I expect. I'm not sure what `zulint` uses in the main project, if there's a similar flag.

tools/generate_hotkeys.py
```diff
@@ -13,34 +17,135 @@
...
+                original_mdFile.write(line)
+        TEMP_OUTPUT_FILE.unlink()
+        print(f"Hot Keys list saved in {ORIGINAL_OUTPUT_FILE}")
+
+
+def lint_hotkeys_file() -> None:
+    """
+    First compare temp_hotkeys and hotkeys file, if in sync then lint for key description and key duplication
+    """
+    generate_hotkeys_file(True)
```
neiljp, line 109: Best to use a named argument here for clarity.

.github/workflows/lint-and-test.yml
```diff
@@ -97,3 +97,22 @@ jobs:
...
+        run: python --version
+      - name: Upgrade pip
+        run: python -m pip install --upgrade pip
+      - name: Ensure libxml-related libraries are installed (some pypy versions don't have wheels)
+        if: startsWith(matrix.env.PYTHON, 'pypy-')
+        run: sudo apt install libxml2-dev libxslt1-dev
+      - name: Ensure regular package installs from checkout
+        run: pip install .
+      - name: Run generate_hotkeys
+        run: ./tools/generate_hotkeys.py --check-only
```
neiljp, line 118: This symbol indicates there's a missing newline at the end of the file.

neiljp (commented):
@srdeotarse I've left some comments inline, but my biggest concern here is that the main (second) commit is very large, making a lot of additions (and deletions) according to the GitHub diff.

The larger a commit is, the harder it is to reason about. To more clearly express to readers, reviewers, and those who come later (reading the PR or git log), it's useful to break large changes into smaller commits which add specific minimal features - or else restructure the code but make no behavioral changes (ie. refactors).

tools/generate_hotkeys.py
```diff
@@ -1,36 +1,69 @@
...
+from typing import Dict, List, Tuple
 
 from zulipterminal.config.keys import HELP_CATEGORIES, KEY_BINDINGS
 
 
-categories = defaultdict(list)
-for item in KEY_BINDINGS.values():
-    categories[item["key_category"]].append((item["help_text"], item["keys"]))
-
+# Path to original docs/hotkeys.md
```
Rohitth007, line 11: Let's try an see if we can simplify the diffs here.
Why not first create the functions in their original place in one commit. Then change the ordering on those functions in another commit. We can squash them later while merging, if needed.

Also, I don't think we need this and the below comment now that they are pretty obvious?

tools/generate_hotkeys.py
```diff
@@ -21,25 +21,26 @@ def generate_hotkeys_file() -> None:
...
+
+
+def get_hotkeys_file_string() -> str:
+    """
+    Get existing hotkeys.md in docs folder as string based on help text description and shortcut key combinations in config/keys.py file
+    """
+    categories = read_help_categories()
+    hotkeys_file_string = f"<!--- Generated automatically by tools/{SCRIPT_NAME} -->\n<!--- Do not modify -->\n\n# Hot Keys\n"
+    for action in HELP_CATEGORIES.keys():
+        hotkeys_file_string += f"## {HELP_CATEGORIES[action]}\n|Command|Key Combination|\n| :--- | :---: |\n"
```
Rohitth007, line 93: Here as well, we can keep the original style of:
```python
hotkeys_file_string += (
    f"## {HELP_CATEGORIES[action]}\n"
    "|Command|Key Combination|\n"
    "| :--- | :---: |\n"
)
```
which I'm hoping would improve the diffs, if `categories = read_help_categories()` was inserted inside `with open` in the previous commit.

Rohitth007:
@srdeotarse I've left a few comments, mostly related to diffs being more readable. This makes reviewing much easier.

Apart from that, please acknowledge the reviews with either a :+1:  or saying some version of "I've fixed this" or answering the questions before resolving them or whenever you push a new update, so that the reviewer knows what's addressed and what is not. It even makes it easier for a new person going through the conversation.

tools/generate_hotkeys.py
```diff
@@ -20,13 +17,13 @@
         f"<!--- Generated automatically by tools/{SCRIPT_NAME} -->\n"
         "<!--- Do not modify -->\n\n# Hot Keys\n"
     )
-    for action in categories.keys():
+    for action in HELP_CATEGORIES.keys():
```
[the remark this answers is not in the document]
srdeotarse: According to your suggestion of using a defaultdict(list) for categories, and loping through 70 ``KEY_BINDINGS`` at once, order of categories in defaultdict is according to order of keys in ``config/keys`` which don't follow the order of ``HELP_CATEGORIES``. Therefore, I have used the ``HELP_CATEGORIES`` ordering.

tools/generate_hotkeys.py
srdeotarse: Sure, I will remove the comments and split the ordering of functions into 2 commits.

tools/generate_hotkeys.py
```diff
@@ -1,39 +1,141 @@
...
+            check_duplicate_keys_list: List[str] = []
+            for help_text, key_combinations_list in categories[action]:
+                check_duplicate_keys_list.extend(key_combinations_list)
+                various_key_combinations = various_key_combination(
+                    key_combinations_list
+                )
+                # Check description style
+                if not re.match(help_text_style, help_text):
+                    print(
+                        f"Description - ({help_text}) for key combination - [{various_key_combinations}] should contain only alphabets, spaces and special characters except ."
```
mounilKshah, line 48: Here, the `various_key_combinations` is printed with the `kbd` tags. I think it would be more readable if it is written without the tags.

mounilKshah (commented):
The PR worked fine for me and under all the cases that I tested it against – hotkeys not matching/missing, duplication of keys and description in unexpected format.

tools/generate_hotkeys.py
Rohitth007: The question above was to see if a regex like `"^[^\d.]*$"` would be more suitable.
I'll wait for @neiljp to clarify the description style here.

srdeotarse: For the ``help_text`` description style, text string except ``.`` and some ``special characters`` are working for now. But, exactly what ``special_characters`` should be allowed that needs some discussion.

tools/generate_hotkeys.py
srdeotarse: Earlier also listing ``categories`` into dict was outside ``with open``. So I think diff is clearer if ``categories = read_help_categories`` is outside ``with open``

tools/generate_hotkeys.py
Rohitth007: Here don't move `OUTPUT_FILE` and `SCRIPT_NAME` then look at the diffs.
You can also move the creation of `various_key_combination()` into another commit. That will make the diffs even more easier to read.

tools/generate_hotkeys.py
Rohitth007: Here, I meant all such multi-line `write()`'s. You missed atleast one at the top.
Re categories, I meant change the position in the previous commit. Try with an without this and see the difference.

Rohitth007 (changes requested):
Left a review below but regarding linting workflow:
* Suppose I change keys.py, in the process I also add a full stop.
* Now I want to lint so I run `--check-only`.
* But it tells me to sync up (Note: It doesn't tell me to lint again).
* I sync up the file.
* I run linting again. It tells me there's a style error.
* I have to start this loop all over again until no errors throw up.
* Instead if linting was done before comparing, I wouldn't have to do all this juggling.
* Also one expects to see something like `No linting errors.` when it's successful rather than `already in sync`.

tools/generate_hotkeys.py
```diff
@@ -1,39 +1,148 @@
...
+        for help_text, key_combinations_list in categories[action]:
+            various_key_combinations = various_key_combination(key_combinations_list)
+            hotkeys_file_string += f"|{help_text}|{various_key_combinations}|\n"
+        hotkeys_file_string += "\n"
+    return hotkeys_file_string
+
+
+def compare_hotkeys_file(hotkeys_file_string: str) -> bool:
+    if hotkeys_file_string == open(OUTPUT_FILE).read():
+        print("Hot Keys already in sync with config/keys.py")
```
Rohitth007, line 110: Using `hotkeys.md` instead of `Hot Keys` makes sense in both this and the below print statements as well.

tools/generate_hotkeys.py
```diff
@@ -6,20 +6,14 @@
 from zulipterminal.config.keys import HELP_CATEGORIES, KEY_BINDINGS
```
Rohitth007, line 8: Re this commit text, you shouldn't be to refer to the previous git history because once it's merged it's like it never existed so it wouldn't make sense unless it's for the reviewer. You should mostly talk about what this commit is doing.

tools/generate_hotkeys.py
```diff
@@ -38,6 +38,10 @@ def lint_hotkeys_file() -> None:
             various_key_combinations = various_key_combination(key_combinations_list)
             # Check description style
             if not re.match(help_text_style, help_text):
+                various_key_combinations = various_key_combinations.replace("<kbd>", "")
+                various_key_combinations = various_key_combinations.replace(
+                    "</kbd>", ""
+                )
```
Rohitth007, line 44: Here you can do both replaces one after the other, like so.
`str.replace().replace()`

Rohitth007 (changes requested):
I've left a few suggestion below but apart from that I would suggest to improve your commit messages to give more clarity about everything that is happening in that commit.

tools/generate_hotkeys.py
```diff
@@ -9,11 +10,62 @@
...
+    if error_flag == 1:
+        print(
+            "After resolving above errors, Run './tools/generate_hotkeys' to update hotkeys.md file."
+        )
+        print(
+            "Again run './tools/generate_hotkeys --check-only' to lint hotkeys.md file"
+        )
+    else:
+        print("No linting errors for hotkeys.md file")
+        compare_hotkeys_file(hotkeys_file_string)
```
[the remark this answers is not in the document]
srdeotarse: I should add ``Run generate_hotkeys....`` in ``compare_hotkeys_file`` function if files are not in sync, as we need to instruct running the command every time hot keys are not in sync.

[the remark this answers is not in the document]
Rohitth007: But then when you run `./tootls/generate_hotkeys` it will show `Run ./tootls/generate_hotkeys`.

tools/generate_hotkeys.py
```diff
@@ -38,6 +38,9 @@ def lint_hotkeys_file() -> None:
             various_key_combinations = various_key_combination(key_combinations_list)
             # Check description style
             if not re.match(help_text_style, help_text):
+                various_key_combinations = various_key_combinations.replace(
+                    "<kbd>", ""
+                ).replace("</kbd>", "")
```
Rohitth007, line 43: I think you undid whatever you did here?

tools/generate_hotkeys.py
```diff
@@ -9,11 +10,58 @@
...
+            if check_duplicate_keys_list.count(key) > 1
+        ]
+        if len(duplicate_keys) != 0:
+            print(
+                f"Duplicate key combination for keys {duplicate_keys} for category ({HELP_CATEGORIES[action]}) detected"
+            )
+            error_flag = 1
+    if error_flag == 1:
+        print(
+            "After resolving above errors, Run './tools/generate_hotkeys --check-only' to lint hotkeys.md file."
```
Rohitth007, line 64: Maybe just say, "Rerun this command after resolving errors in config/keys.py".

Also the thing is we are effectively linting the markdown file but we are doing it my linting keys.py. So it might confuse the user as to what they should edit if says "to lint hotkeys.md"

Rohitth007 (changes requested):
Gave a quick look at your changes. Didn't look at your new commit messages though.

tools/generate_hotkeys.py
Rohitth007: Oh black took effect. Why not just use `key_combinations_list` in the print statement instead of undoing whatever was done in the `various_key_combinations()` function.  (What does `" / ".join(key_combinations_list)` look like?)

Rohitth007: Also I don't like that fact that `various_key_combinations` is both a variable name and a function name.

srdeotarse: I have used ``" / ".join(key_combinations_list)`` in my recent commit, it shows key combinations like ``ctrl _`` instead of earlier ``ctrl + _`` and also for two key combinations it shows something like ``up / ctrl p``.

Rohitth007: I think it makes sense to show it this way as we are anyway asking them to edit config/keys.py (`ctlr _`) and not the markdown file (`ctrl + _`). The other one might be confusing if we show markdown and expect them to edit python.

tools/generate_hotkeys.py
```diff
@@ -35,7 +35,7 @@ def lint_hotkeys_file() -> None:
         check_duplicate_keys_list: List[str] = []
         for help_text, key_combinations_list in categories[action]:
             check_duplicate_keys_list.extend(key_combinations_list)
-            various_key_combinations = various_key_combination(key_combinations_list)
+            various_key_combinations = " / ".join(key_combinations_list)
```
Rohitth007, line 38: you could probably squash this with the linting commit.

tools/generate_hotkeys.py
```diff
@@ -9,11 +10,56 @@
...
+            check_duplicate_keys_list.extend(key_combinations_list)
+            various_key_combinations = various_key_combination(key_combinations_list)
+            # Check description style
+            if not re.match(help_text_style, help_text):
+                print(
+                    f"Description - ({help_text}) for key combination - [{various_key_combinations}] should contain only alphabets, spaces and special characters except ."
+                )
+                error_flag = 1
+        # Check key combination duplication
+        check_duplicate_keys_list = [
```
Rohitth007, line 46: You can probably split this commit into 2. One for checking description style and the other for checking key combination duplication.

Rohitth007 (changes requested):
Left a few notes. They're all small changes. After that maybe Neil can take a look.

tools/generate_hotkeys.py
```diff
@@ -9,11 +10,43 @@
 OUTPUT_FILE = Path(__file__).resolve().parent.parent / "docs" / "hotkeys.md"
 SCRIPT_NAME = PurePath(__file__).name
 
+# Exclude keys from duplicate keys checking
+KEYS_TO_EXCLUDE = ["q", "e", "m", "r"]
```
Rohitth007, line 15: This probably belongs in the next commit?

tools/generate_hotkeys.py
```diff
@@ -9,11 +10,43 @@
...
+            # Check description style
+            if not re.match(help_text_style, help_text):
+                print(
+                    f"Description - ({help_text}) for key combination - [{various_key_combinations}] should contain only alphabets, spaces and special characters except ."
+                )
+                error_flag = 1
+    if error_flag == 1:
+        print("Rerun this command after resolving errors in config/keys.py")
+    else:
+        print("No linting errors for hotkeys.md file")
```
Rohitth007, line 62: Lets just say `No hotkey linting errors`? As it maybe confusing if there are `no errors in hotkeys.md` but you are still asking me to `update hotkeys.md`

Rohitth007 (changes requested):
Except the below the rest of the code seems good. It's better to give your commit messages a once over to make it more descriptive wherever necessary before Neil takes a look.

tools/generate_hotkeys.py
```diff
@@ -14,6 +15,35 @@ def main() -> None:
...
+                print(
+                    f"Description - ({help_text}) for key combination - [{various_key_combinations}] should contain only alphabets, spaces and special characters except ."
+                )
+                error_flag = 1
+    if error_flag == 1:
+        print("Rerun this command after resolving errors in config/keys.py")
+    else:
+        print("No hotkeys linting errors")
+        if not compare_hotkeys_file(hotkeys_file_string):
+            print("Run './tools/generate_hotkeys' to update hotkeys.md file")
```
neiljp, line 64: It would be good to have hotkeys.md in a variable, to avoid the duplication.

tools/generate_hotkeys.py
```diff
@@ -14,6 +15,35 @@ def main() -> None:
...
+    exit(error_flag)
```
neiljp, line 65: This should be `sys.exit`.

neiljp (commented):
@srdeotarse Thanks for keeping this up :+1:

This looks good; I left some inline comments to keep things moving - let me know if you have queries.

tools/generate_hotkeys.py
```diff
@@ -127,4 +131,11 @@ def various_key_combination(key_combinations_list: List[str]) -> str:
 
 
 if __name__ == "__main__":
-    main()
+    parser = argparse.ArgumentParser(description="Lint and generate hotkeys")
```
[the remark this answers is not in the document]
srdeotarse: description=" Generate or update hotkeys.md file based on changes in config/keys.py by default. To lint hotkeys add '--check-only' option."
How about this description?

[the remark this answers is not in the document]
neiljp: The first sentence is all that's necessary - I think this is only shown in the help text, and in that case the option will be listed below. So if the option is well-described, we don't need to include it here.

The 'by default' could be in parentheses, perhaps? As I mentioned elsewhere, it would be cleaner to use string substitution for the file names which are used throughout, and likely refer to the variables in docstrings instead.
(alternatively, some of those functions could take in file parameters, but let's leave that change for this PR)

tools/generate_hotkeys.py
neiljp: Thanks for pointing this out @Rohitth007, though I think this was in the original version from early 2021?

It's a good refactor to have, though being outside of the main application I'm not convinced the optimization is required :) However, it's good this filtering pattern, as this might be more warranted where we partition a larger data-structure there, eg. users based on presence.

tools/generate_hotkeys.py
```diff
@@ -14,6 +15,35 @@ def main() -> None:
...
+    categories = read_help_categories()
+    for action in HELP_CATEGORIES.keys():
+        check_duplicate_keys_list: List[str] = []
+        for help_text, key_combinations_list in categories[action]:
+            check_duplicate_keys_list.extend(key_combinations_list)
+            various_key_combinations = " / ".join(key_combinations_list)
+            # Check description style
+            if not re.match(help_text_style, help_text):
+                print(
+                    f"Description - ({help_text}) for key combination - [{various_key_combinations}] should contain only alphabets, spaces and special characters except ."
```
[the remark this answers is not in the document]
neiljp: Having tested this today (via adjusting the regex), the "should..." part runs into the end of the key output - so maybe actually output it on the next line with a slight indent?

tools/generate_hotkeys.py
neiljp: I added a separate comment for this just now; the original aim was eg. detect `.`, but something like the original regex above would benefit consistency.

.github/workflows/lint-and-test.yml
neiljp: This still seems to be the case?

tools/generate_hotkeys.py
```diff
@@ -1,24 +1,36 @@
...
 
-with open(OUTPUT_FILE, "w") as mdFile:
-    mdFile.write(
+
+def generate_hotkeys_file() -> None:
+    """
+    Generate hotkeys.md in docs folder based on help text description and shortcut key combinations in config/keys.py file
+    """
+    categories = read_help_categories()
+    hotkeys_file_string = (
```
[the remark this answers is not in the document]
srdeotarse: I have extracted writing of ``hotkeys_file_string`` variable to ``hotkeys.md`` file in a new function ``write_hotkeys_file`` if thats what you meant?

tools/generate_hotkeys.py
```diff
@@ -1,39 +1,141 @@
...
+        if len(duplicate_keys) != 0:
+            print(
+                f"Duplicate key combination for keys {duplicate_keys} for category ({HELP_CATEGORIES[action]}) detected"
+            )
+            error_flag = 1
+    if error_flag == 1:
+        print("Rerun this command after resolving errors in config/keys.py")
+    else:
+        print("No hotkeys linting errors")
+        if not compare_hotkeys_file(hotkeys_file_string):
```
srdeotarse, line 63: I have used return value for ``compare_hotkeys_file`` to output different statement based on return value.

tools/generate_hotkeys.py
```diff
@@ -19,6 +19,7 @@ def generate_hotkeys_file() -> None:
     Generate hotkeys.md in docs folder based on help text description and shortcut key combinations in config/keys.py file
     """
     hotkeys_file_string = get_hotkeys_file_string()
+    compare_hotkeys_file(hotkeys_file_string)
```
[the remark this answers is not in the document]
srdeotarse: @neiljp It is intentional as I have used return value [here](https://github.com/zulip/zulip-terminal/pull/1210/files#r949848327) for linting.

neiljp:
@srdeotarse Thanks for all your work on this - this was close enough that I worked through a few minor issues to get it merged, which I'll do after pushing to here :tada:

Functionally, the only issue seemed to be not failing on being out of sync, and the GH job being out of date and based on pytest rather than simpler linting. Commit-wise, a little code was also in the wrong commit.

Other than fixing those functional issues I made some commit ordering changes, partly to explain the commit organization in earlier comments; these weren't strictly necessary, but I hope you can see how the flow is a little clearer.

An overview of the changes per-commit:
- first two commits as-is except slight commit text change
- 3rd commit combined with other function extraction commit (6th) and a docstring reword, also retaining order of code from before to minimize diff
- 4th commit (reorder) now smaller due to reordering in 3rd commit
[- 5th commit adding extra function dropped (as per discussion)]
[- 6th commit integrated into 3rd (as above)]
- 7th commit slightly reordered and reworded (it adds keys/output variables)
- added extra refactor to refer to those variables in docstrings (later commits adjusted too)
- 8th commit minor commit text change, add extra error_flag=1 if files don't match
- 9th commit as-is, plus code from 10th commit that was specific to duplication linting
- 10th commit tool slightly renamed, extra code split into previous commit (duplication related)
- 11th commit moved job below other linting jobs & adapted for recent changes (+lint rename)

I'd suggest you verify the changes by using `tools/fetch-pull-request 1210` and doing a `git range-diff main hotkey-linting review-original-1210`, though the output format can be confusing at times. That may help you understand changes I've made in detail.

Returning to the issue of the order of the linting commits, in principle it might have been better if the 8th commit with the linting function had been split into just linting for differences (ie. needs sync or not), and the description linting, since the former could then be combined with the changes to main() to make a minimal working linting option. The linting for description and duplication could then be extensions built on top in later commits. The linting code would then be testable from the start. However, given the commits seem functionally fine, I didn't want either of us to spend more time on that - just something to bear in mind for the future :)

PR #1213, opened 2022-04-21
Display names of reacting users when few users have reacted.

### LINKED ISSUE #1212: Display names of reacting users when few users have reacted in ZT
To display reacting users when few users have reacted, as shown below -

image

srdeotarse:
Fixes #1212

**What does this PR do?**  
This PR displays reacting usernames besides reacting emoji in messages.

CZO - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Feature.20discussion.20-.20Show.20usernames.20adding.20emoji.20to.20message

**Tested?** 
- [x] Manually

**Commit flow** 

**Notes & Questions** 

**Interactions** 

**Visual changes** 
In Zulip web app -
image

In ZT -
image

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -2736,55 +2736,71 @@ class TestMessageBox:
 
     # FIXME This is the same parametrize as MsgInfoView:test_height_reactions
     @pytest.mark.parametrize(
-        "to_vary_in_each_message",
+        "to_vary_in_each_message, expected_text, expected_attributes",
         [
-            {
-                "reactions": [
-                    {
-                        "emoji_name": "thumbs_up",
-                        "emoji_code": "1f44d",
-                        "user": {
-                            "email": "iago@zulip.com",
-                            "full_name": "Iago",
-                            "id": 5,
+            case(
+                {
+                    "reactions": [
+                        {
+                            "emoji_name": "thumbs_up",
+                            "emoji_code": "1f44d",
+                            "user": {
+                                "email": "iago@zulip.com",
+                                "full_name": "Iago",
+                                "id": 5,
+                            },
+                            "reaction_type": "unicode_emoji",
                         },
-                        "reaction_type": "unicode_emoji",
-                    },
-                    {
-                        "emoji_name": "zulip",
-                        "emoji_code": "zulip",
-                        "user": {
-                            "email": "iago@zulip.com",
-                            "full_name": "Iago",
-                            "id": 5,
+                        {
+                            "emoji_name": "zulip",
+                            "emoji_code": "zulip",
+                            "user": {
+                                "email": "iago@zulip.com",
+                                "full_name": "Iago",
+                                "id": 5,
+                            },
+                            "reaction_type": "zulip_extra_emoji",
                         },
-                        "reaction_type": "zulip_extra_emoji",
-                    },
-                    {
-                        "emoji_name": "zulip",
-                        "emoji_code": "zulip",
-                        "user": {
-                            "email": "AARON@zulip.com",
-                            "full_name": "aaron",
-                            "id": 1,
+                        {
+                            "emoji_name": "zulip",
+                            "emoji_code": "zulip",
+                            "user": {
+                                "email": "AARON@zulip.com",
+                                "full_name": "aaron",
+                                "id": 1,
+                            },
+                            "reaction_type": "zulip_extra_emoji",
                         },
-                        "reaction_type": "zulip_extra_emoji",
-                    },
-                    {
-                        "emoji_name": "heart",
-                        "emoji_code": "2764",
-                        "user": {
-                            "email": "iago@zulip.com",
-                            "full_name": "Iago",
-                            "id": 5,
+                        {
+                            "emoji_name": "heart",
+                            "emoji_code": "2764",
+                            "user": {
+                                "email": "iago@zulip.com",
+                                "full_name": "Iago",
+                                "id": 5,
+                            },
+                            "reaction_type": "unicode_emoji",
                         },
-                        "reaction_type": "unicode_emoji",
-                    },
-                ]
-            }
+                    ],
+                },
+                " :thumbs_up: 1   :zulip: 2   :heart: 1  ",
+                [
+                    ("reaction", 15),
+                    (None, 1),
+                    ("reaction_mine", 11),
+                    (None, 1),
+                    ("reaction", 11),
+                ],
+            ),
         ],
     )
-    def test_reactions_view(self, message_fixture, to_vary_in_each_message):
+    def test_reactions_view(
+        self,
+        message_fixture,
+        to_vary_in_each_message,
+        expected_text,
+        expected_attributes,
+    ):
         self.model.user_id = 1
         varied_message = dict(message_fixture, **to_vary_in_each_message)
         msg_box = MessageBox(varied_message, self.model, None)
@@ -2792,16 +2808,8 @@ class TestMessageBox:
 
         reactions_view = msg_box.reactions_view(reactions)
 
-        assert reactions_view.original_widget.text == (
-            " :thumbs_up: 1   :zulip: 2   :heart: 1  "
-        )
-        assert reactions_view.original_widget.attrib == [
-            ("reaction", 15),
-            (None, 1),
-            ("reaction_mine", 11),
-            (None, 1),
-            ("reaction", 11),
-        ]
+        assert reactions_view.original_widget.text == expected_text
+        assert reactions_view.original_widget.attrib == expected_attributes
 
     @pytest.mark.parametrize(
         "message_links, expected_text, expected_attrib, expected_footlinks_width",
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -2792,6 +2792,162 @@ class TestMessageBox:
                     ("reaction", 11),
                 ],
             ),
+            case(
+                {
+                    "reactions": [
+                        {
+                            "emoji_name": "thumbs_up",
+                            "emoji_code": "1f44d",
+                            "user": {
+                                "email": "iago@zulip.com",
+                                "full_name": "Iago",
+                                "id": 5,
+                            },
+                            "reaction_type": "unicode_emoji",
+                        },
+                        {
+                            "emoji_name": "zulip",
+                            "emoji_code": "zulip",
+                            "user": {
+                                "email": "AARON@zulip.com",
+                                "full_name": "aaron",
+                                "id": 1,
+                            },
+                            "reaction_type": "zulip_extra_emoji",
+                        },
+                        {
+                            "emoji_name": "heart",
+                            "emoji_code": "2764",
+                            "user": {
+                                "email": "iago@zulip.com",
+                                "full_name": "Iago",
+                                "id": 5,
+                            },
+                            "reaction_type": "unicode_emoji",
+                        },
+                    ],
+                },
+                " :thumbs_up: Iago   :zulip: You   :heart: Iago  ",
+                [
+                    ("reaction", 18),
+                    (None, 1),
+                    ("reaction_mine", 13),
+                    (None, 1),
+                    ("reaction", 14),
+                ],
+            ),
+            case(
+                {
+                    "reactions": [
+                        {
+                            "emoji_name": "zulip",
+                            "emoji_code": "zulip",
+                            "user": {
+                                "email": "iago@zulip.com",
+                                "full_name": "Iago",
+                                "id": 5,
+                            },
+                            "reaction_type": "unicode_emoji",
+                        },
+                        {
+                            "emoji_name": "zulip",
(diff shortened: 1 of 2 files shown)


srdeotarse:

zulipterminal/ui_tools/boxes.py
```diff
@@ -1112,26 +1112,39 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
             reaction_texts = [
                 (
-                    "reaction_mine" if my_user_id in ids else "reaction",
-                    f":{reaction}: {len(ids)}",
+                    "reaction_mine"
+                    if my_user_id in [id[0] for id in ids]
+                    else "reaction",
+                    f":{reaction}: {len(ids)}"
+                    if len(ids) > 3
+                    else f":{reaction}: {str([id[1] for id in ids]).strip('[]').translate(translation)}",
```
Rohitth007, line 1144: Try using `str.join()` here instead. It'll be much simpler to read and also avoid magic numbers like `39`

zulipterminal/ui_tools/boxes.py
```diff
@@ -1112,26 +1112,39 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
+                reaction_stats[reaction["emoji_name"]].add((user_id, user_name))
 
             sorted_stats = sorted(
-                (reaction, count) for reaction, count in reaction_stats.items()
+                (reaction, list(count)) for reaction, count in reaction_stats.items()
             )
 
-            my_user_id = self.model.user_id
+            for stats in sorted_stats:
+                stats[1].sort(key=lambda id: id[0])
```
Rohitth007, line 1131: Whats the need of sorting based on id's here? Does that mean users who have a smaller id are preferred? From [this](https://chat.zulip.org/#narrow/stream/277-discussions/topic/Next.20discussion.20topic/near/1319047) message, it looks like your ordering and webapps ordering differ.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1112,26 +1112,39 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
                     user_id = int(reaction["user"]["user_id"])
-                reaction_stats[reaction["emoji_name"]].add(user_id)
+                user_name = reaction["user"]["full_name"]
+                if user_id == my_user_id:
+                    user_name = "You"
+                reaction_stats[reaction["emoji_name"]].add((user_id, user_name))
 
             sorted_stats = sorted(
-                (reaction, count) for reaction, count in reaction_stats.items()
+                (reaction, list(count)) for reaction, count in reaction_stats.items()
```
Rohitth007, line 1127: I'm not sure why the reactions were sorted here in the first place. From [this](https://chat.zulip.org/#narrow/stream/1-announce/topic/server.20releases/near/1355228) message, the ordering of reactions seem to differ from the webapp.

Also what's the reason for converting the set into a list?

Rohitth007 (changes requested):
I had a few doubts and suggestions which I left below. Apart from that I'm wondering why we used `defaultdict(set)` here if we are converting to a list right after. Is it avoiding any duplicates?

zulipterminal/ui_tools/boxes.py
srdeotarse: - Every time ``reacting user_ids`` are fetched, I noticed that ordering was even if there was no change in ``reacting users`` for a reaction. Therefore, I sorted the ``ids``.
- Next, in ``test_reaction_views`` because of random ordering, tests were failing.

zulipterminal/ui_tools/boxes.py
srdeotarse: To rearrange ``You`` reaction ``user_name`` to the end of users reacted.

neiljp (commented):
@srdeotarse Interested to see the next iterations; guessing the code may simplify with the max overall, but will save further review until we're more settled.

srdeotarse:

zulipterminal/ui_tools/boxes.py
Rohitth007: I think now with the updated PR, the tests are also not a problem. So maybe it's better to display the reactions in the order in which they were reacted instead of alphabetically?

tests/ui/test_ui_tools.py
```diff
@@ -2780,28 +2780,177 @@ def test_transform_content(self, mocker, raw_html, expected_content):
...
         ],
     )
     def test_reactions_view(self, message_fixture, to_vary_in_each_message):
         self.model.user_id = 1
         varied_message = dict(message_fixture, **to_vary_in_each_message)
         msg_box = MessageBox(varied_message, self.model, None)
         reactions = to_vary_in_each_message["reactions"]
-
+        expected_result = to_vary_in_each_message["expected_result"]
+        reaction_attributes = to_vary_in_each_message["reaction_attributes"]
```
neiljp, line 2949: This is what made me concerned about the test, as here we're changing the message we're testing with. That doesn't cause any issues here as it doesn't overwrite anything we expect to be there (and the Message is badly defined type-wise right now), but effectively we're generating a message with extra fields which the code could 'cheat' with, to pass the test ;)

tests/ui/test_ui_tools.py
```diff
@@ -2780,28 +2780,177 @@ def test_transform_content(self, mocker, raw_html, expected_content):
...
-            " :heart: 1   :thumbs_up: 1   :zulip: 2  "
-        )
-        assert reactions_view.original_widget.attrib == [
-            ("reaction", 11),
-            (None, 1),
-            ("reaction", 15),
-            (None, 1),
-            ("reaction_mine", 11),
-        ]
+        assert reactions_view.original_widget.text == (expected_result)
```
neiljp, line 2952: Is this a left-over parenthesis? Black reformats code, but please check the code for readability afterwards :)

neiljp (commented):
@srdeotarse Thanks for persevering; this is looking close to clean and ready :+1:

The tests need a little adjustment as noted, and I'm not sure the first commit has anything to do with #1212?

While the tests aren't final, I'm happy the code works OK.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1112,22 +1112,34 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
+
             reaction_texts = [
                 (
-                    "reaction_mine" if my_user_id in ids else "reaction",
-                    f" :{reaction}: {len(ids)} ",
+                    "reaction_mine"
+                    if my_user_id in [id[0] for id in ids]
+                    else "reaction",
+                    f" :{reaction}: {len(ids)} "
+                    if len(ids) > 3 or len(reactions) > 3
```
Rohitth007, line 1141: Do we actually need the `len(ids) > 3` condition here?

Also lets define variables for hardcoded constants. Eg: `MAX_USERNAMES_PER_REACTIONS = 3` or something similar.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1112,22 +1112,34 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
                 user_id = int(reaction["user"].get("id", -1))
                 if user_id == -1:
                     user_id = int(reaction["user"]["user_id"])
-                reaction_stats[reaction["emoji_name"]].add(user_id)
+                user_name = reaction["user"]["full_name"]
+                if user_id == my_user_id:
+                    user_name = "You"
+                reaction_stats[reaction["emoji_name"]].append((user_id, user_name))
 
             sorted_stats = sorted(
```
Rohitth007, line 1126: Do we need this, now that the tests also pass and given that ordering is also preserved? This would show emojis in the order in which they were reacted. This can be a separate commit.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1112,22 +1112,34 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
         if not reactions:
             return ""
         try:
-            reaction_stats = defaultdict(set)
+            my_user_id = self.model.user_id
+            reaction_stats = defaultdict(list)
```
Rohitth007, line 1116: Let's maybe do the `set` to `list` change in a refactor commit before everything else, so that the diffs look cleaner?

Rohitth007 (changes requested):
@srdeotarse Nice work on the feature, it's visually more pleasant and the PR looks to be in it's final stages as well. Left a few comments below.

srdeotarse:

zulipterminal/ui_tools/boxes.py
```diff
@@ -1112,18 +1112,30 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
-            my_user_id = self.model.user_id
             reaction_texts = [
                 (
-                    "reaction_mine" if my_user_id in ids else "reaction",
-                    f" :{reaction}: {len(ids)} ",
+                    "reaction_mine"
+                    if my_user_id in [id[0] for id in ids]
+                    else "reaction",
+                    f" :{reaction}: {len(ids)} "
+                    if len(reactions) > 3
```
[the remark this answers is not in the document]
neiljp: While I wouldn't say that it's necessary to use upper case, you haven't addressed Rohitth's point and extracted the number into a variable for clarity (though you may need to adjust the name slightly now).

tests/ui/test_ui_tools.py
```diff
@@ -2736,72 +2736,235 @@ def test_transform_content(self, mocker, raw_html, expected_content):
...
+                    ],
+                },
+                " :thumbs_up: 1   :zulip: 2   :heart: 1  ",
+                [
+                    ("reaction", 15),
+                    (None, 1),
+                    ("reaction_mine", 11),
+                    (None, 1),
+                    ("reaction", 11),
+                ],
```
neiljp, line 2793: That I need to scroll around to check that this matches the original expected values (must further down) emphasizes the point I made in the previous review that you didn't respond to: add a small refactor commit before this which moves these values from the body of the test to use parameters.

zulipterminal/ui_tools/boxes.py
neiljp: This has turned into a very small commit, but we can always squash it later, and generally small is good for reviews too. However, let me address the ordering :)

I know Rohitth said "before everything else", but note that the original 'story' of commits was to adjust the spacing and then add the user names. Putting this first feels strange to a reviewer as this seems unrelated to the next commit (spacing), and maybe part of a new 'feature' change, ie. adjusting the ordering?

If there are purely refactoring commits these generally come before the related feature. This is currently a pure refactor since it doesn't change the behavior, but it seems unrelated to the following commit. This is easy to reorder in this case since the changes don't intersect or depend on each other (you can move the commits around using an interactive rebase).

The ordering change is arguably a new (visible) feature, so you might consider this part of that block of commits.

Other than making it easier to review, this also helps with merging earlier commits (or groups of commits) from a PR first, if PRs become unexpectedly large, or if earlier commits are mergeable and later ones need further work.

neiljp (commented):
@srdeotarse Tests look better. I don't think you addressed all the prior points, but I left some feedback since the structure has changed.

The implementation of reaction-names was valid, but particularly if making the conditional change that Rohitth noted, then I feel we should express it more cleanly given the complexity in at least the last few modified blocks.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1124,22 +1126,32 @@ def reactions_view(self, reactions: List[Dict[str, Any]]) -> Any:
...
+                                "reaction_mine",
+                                f" :{reaction}: {', '.join([id[1] for id in ids])} ",
+                            )
+                        )
+                    else:
+                        reaction_texts.append(
+                            (
+                                "reaction",
+                                f" :{reaction}: {', '.join([id[1] for id in ids])} ",
+                            )
```
Rohitth007, line 1145: Note that you can simplify the above by bringing the `append` statements out of the conditionals using `reaction_style` for `"reaction"` and`"reaction-mine"`.

This can be added when the feature itself was introduced in the previous commit but I'd wait for @neiljp to confirm if this was what he imagined.

srdeotarse:

neiljp:
@srdeotarse I left some notes in the stream. This is essentially fine now, just a few minor points outstanding which we can discuss, though it'd be fairly easy to adjust slightly before merging. The code can be adjusted, but it has essentially the same behavior as before so we're mostly just tidying/optimizing at this point.

neiljp:
@srdeotarse I'm going to merge the last two commits now, with a few minor changes, thanks! :tada:

I didn't expect you to strip the last commit back out, but the code did work before as far as i could tell, so as I said I'm happy to merge the previous state, particularly with the extracted constant added.

I slightly adjusted some commit text and test parameter names, and (re)moved a few lines between commits where the formatting fitted better. Please do check the difference using `git range-diff`. See the recent topic about that if you need to review the command. If you have questions, please do follow-up in the appropriate topic :)

PR #1214, opened 2022-04-22
tools: Add check-branch to run 'make check' on all branch commits

neiljp:
**What does this PR do?**  

This is a first pass at a script to ensure a branch has commits which individually pass `make check`.

**Tested?** 
- [x] Manually

^ checked with this script!

**Notes & Questions** 

This may become integrated into CI at some point, but an initial try at this turned out to be more complex than expected, so working with something we can easily run locally first.

I've not added documentation yet, since I'd like to let this get some simple testing first.

This script passes if it leaves you back at your branch; if it fails it leaves you at the failed commit, at which point you can `git rebase --abort` and explore it more yourself, or use the rebase interactively.

### DIFF
--- a/tools/check-branch
+++ b/tools/check-branch
@@ -11,4 +11,22 @@ fi
 
 main_branch=${1:-"upstream/main"}
 
+if [ ! "$1" ]; then
+    echo "Using default main branch '$main_branch'"
+    if ! git diff --quiet "$main_branch" main; then
+        set +x
+        echo "Branch '$main_branch' appears to differ from 'main'; please specify one!"
+        exit 1
+    fi
+else
+    echo "Specified main branch as '$main_branch'"
+fi
+
+if git diff --quiet "$main_branch" HEAD; then
+    set +x
+    echo "No changes! Nothing to do!"
+    exit 1
+fi
+
+echo "Checking commits..."
 git rebase $(git merge-base "$main_branch" @) --exec 'git --no-pager log -1 && make check'
--- a/tools/check-branch
+++ b/tools/check-branch
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -e
+
+if ! git diff-index --quiet HEAD; then
+    set +x
+    echo "There are uncommitted changes:"
+    git status --short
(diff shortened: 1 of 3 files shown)


tools/check-branch
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+if git diff --quiet "$main_branch" HEAD; then
+    set +x
+    echo "No changes! Nothing to do!"
+    exit 1
+fi
+
+echo "Checking commits..."
+git rebase $(git merge-base "$main_branch" @) --exec 'git log -1 && make check'
```
Rohitth007, line 32: `git log -1` make the process stop at each commit until I press `q`. Was that intended? What about using `cat <(git log -1)` instead?

tools/check-branch
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e
+
+if ! git diff-index --quiet HEAD; then
+    set +x
+    echo "There are uncommitted changes:"
```
Rohitth007, line 6: I'm not sure this works because I had uncommitted changes and it still started testing.

Rohitth007 (changes requested):
I tested this out and it seems to work fine except a few issues as stated below.
It also works with the Makefile. I tested with the below code and it works like as if you had run it directly:
```bash
check-branch:
    @ ./tools/check-branch
```

tools/check-branch
neiljp: I tried using `--no-pager` instead; does that resolve this issue for you?
(this avoids extra commands/shells, though I agree various options are possible)

tools/check-branch
neiljp: Could you give more information about this situation? `tools/fetch-pull-request` does something similar and I took the test from there. It seems to work for me with simple testing.

neiljp:
@Rohitth007 Thanks for the review and testing!

Good to hear that this slightly revised version does work for you with the `--no-pager` change :tada:

We can consider adding to the makefile, though by default this would limit being able to pass through an alternative merge-base.

Issue #1211, opened 2022-04-20
TRACKING: Consider alternative date rendering


neiljp:
Currently we show the full date/time for each message, and in user 'last active' times.

The only case we currently don't show the full detail is with years, where we don't show the current year, adding it for messages from previous years. I suspect this doesn't update in real-time as time switches to the next year, but that's fairly rare.

In contrast, the web app shows dates in various formats:
- messages have just a time, with a header for the date part, which can sometimes be eg. `yesterday`, and the year if appropriate; this header floats, so this can save space in a long discussion (we don't handle this floating header idea in the same way, so it may not)
- messages have dividers between 'significant' times, indicating previous day and next day; not necessarily consecutive, and for long conversations on the previous day helps avoid the need for the date header being visible
- it doesn't actually show 'last active' times?
- in the recent-topics view (which ZT doesn't currently have an equivalent) the most-recent message time is listed, and it is sorted that way by default, with a relative time (eg. n minutes ago, an hour ago, n hours ago, Yesterday, n days ago)

I believe the web app used to have more instances of relative dates/times, but this is the current state.

---
The most implementable initial feature might be to explore extracting the date part of a message into date dividers for above/below. The downside is that this may add extra space requirements.

As per the current label, please discuss this feature in chat.zulip.org first.

Issue #1086, opened 2021-07-17
Enhance UX for User Info View (including custom profile fields)


Ezio-Sarthak:
ZT currently supports a basic version of the user information popup (press `i` in users list to open), with some trivial details of a user. However, we can improve its user experience further by:
 - ~~Displaying human-friendly last active timestamp. e.g., `5 hours ago`, instead of the formatted date string.~~
 - Displaying additional user information (e.g., favorite editor, status message, Github profile) as per a user's API response.

**UPDATE** Following brief discussion of this issue, it would seem that date rendering is best handled more uniformly, for which I've opened #1211 - and the web app doesn't currently show the `n hours ago` for users at the moment in any case? -- **neiljp**

neiljp:
Closing this in favor of #1338, now that we have split the other issue out.

PR #1218, opened 2022-04-29
User-friendly date and time formatting.

plugyawn:
**What does this PR do?**  
Adds user friendly times to the user and message info views. The current view just includes the timestamp, which is hard to decipher.

Partial fix for issues #1211  and addresses #1086

**Tested?** 
- [x] Manually

**Commit flow** 

- first commit adds a helper function to friendly-fy the timestamps.
- second commit implements the helper function in views.py for message and user info popups.

**Notes & Questions** 

**Interactions** 

**Visual changes** 
For message info:
Screenshot 2022-04-29 at 1 11 14 PM

For user info:
Screenshot 2022-04-29 at 1 11 01 PM

### DIFF
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -3,6 +3,7 @@ import subprocess
 import time
 from collections import OrderedDict, defaultdict
 from contextlib import contextmanager
+from datetime import datetime
 from functools import partial, wraps
 from itertools import chain, combinations
 from re import ASCII, MULTILINE, findall, match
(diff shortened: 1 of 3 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1162,6 +1163,7 @@ def _fetch_user_data(controller: Any, user_id: int) -> Dict[str, Any]:
             }
             return display_data
 
+        pretty_date = user_friendly_time(data["last_active"])
```
mounilKshah, line 1166: For users who have been inactive for longer than two weeks, the `data["last_active"]` is empty. This is causing your code to break for such users when I tried running it.

neiljp:
@plugyawn You saw my feedback in the stream, but essentially:
- this is unstable right now, at least with the issue @mounilKshah pointed out, as well as for me
- actual timestamps rather than formatted dates would likely be more reliable for comparison
- tests would make it easier to provide potential buggy inputs and get them working
- git commit style could be improved, as we discussed via gitlint.

plugyawn:
> @plugyawn You saw my feedback in the stream, but essentially:
> 
> * this is unstable right now, at least with the issue @mounilKshah pointed out, as well as for me
> * actual timestamps rather than formatted dates would likely be more reliable for comparison
> * tests would make it easier to provide potential buggy inputs and get them working
> * git commit style could be improved, as we discussed via gitlint.

Sorry for the late reply! 

> @mounilKshah thanks for pointing out the error! I've fixed it. 
> Agreed, I'll check into that. In this case, though, do you think it'll make much of a difference?
> On it! I'll get started with adding more tests.
> I set up gitlint, hopefully it's working now : )

plugyawn:
@neiljp could you check the commit flow now?

neiljp:
@plugyawn As per discussion in the stream, you really need a rebase (or other adjustment) since you seem to have squashed commits from upstream/main or another PR into one of these commits.

zulipterminal/helper.py
```diff
@@ -725,16 +725,17 @@ def suppress_output() -> Iterator[None]:
...
-    
-    if len(time_stamp) > 22: # for message_info, year is included
+
+    if len(time_stamp) > 22:  # for message_info, year is included
         datetime_object = datetime.strptime(
             time_stamp, "%a %b %d %Y %I:%M:%S %p"
         )  # Fri Apr 29 2022 05:15:12 AM is an example
-    else: # for user_info, year not included
-        if time_stamp[-1] == "M": # AM or PM
+    else:  # for user_info, year not included
```
neiljp, line 737: We'll want to keep these lint fixes in the commit in which they were introduced. If you do an interactive rebase and edit the previous commit and run `make fix`, then when you continue the rebase, these changes should likely drop out of this commit (or similarly in later commits, in other complex PRs).

plugyawn: Oh, okay! Didn't know, I'll try it out.

plugyawn:
@mounilKshah @neiljp Could you check this now? I think the changes I made in test_popups.py need to be changed -- but I'm unsure as to what those changes should be.

Also, I've made sure the commits only contain changes in their respective files, as you said. Editing the commits and running commit --amend was the solution, as you said.

plugyawn:
@zulipbot add PR awaiting review.

zulipterminal/helper.py
```diff
@@ -816,3 +817,61 @@ def open_media(controller: Any, tool: str, media_path: str) -> None:
 
     if error:
         controller.report_error(error)
+
+
+def user_friendly_time(time_stamp: str) -> str:  # message_info has a different format
```
mounilKshah, line 822: Instead of writing what the function does in the commit message, I think it would be more convenient to write it as comments within the function body, so that the next person working on it doesn't have to search for the respective commit(s).

zulipterminal/helper.py
```diff
@@ -816,3 +817,61 @@ def open_media(controller: Any, tool: str, media_path: str) -> None:
...
+        if seconds < 120:
+            return "A minute ago"
+        if seconds < 3600:
+            return str(seconds // 60) + " minutes ago"
+        if seconds < 7200:
+            return "An hour ago"
+        if seconds < 86400:  # 24 hours
+            return str(seconds // 3600) + " hours ago"
+
+    time_string = ""
```
mounilKshah, line 864: The value of `time_string` is assigned only in the first conditional, but not returned.

mounilKshah (commented):
The PR works well, without any glitches and the commit structure is better than when I had last seen. I have left a few comments, 

Also, maybe change the name of the branch on your forked repo? It can create confusion while working with multiple PRs.

plugyawn:
Reviews incorporated into the PR.

PR #1221, opened 2022-05-06
Show detected platform on startup and in About popup

### LINKED ISSUE #1216: Show detected platform on startup and in About popup
This is an element of #791.

We now have a clear unique platform determination via code in `platform_code.py` which runs on startup.

For various reasons it would be useful to show this information:
- on the terminal, when the application starts up, possibly with a warning if unsupported
- in the About popup, possibly in a new section at the bottom

Each of these could be separate commits or PRs.

plugyawn:
**What does this PR do?**  
Shows the detected platform on startup and in the About pop-up.

Tries to fix #1216. 

**Tested?** 
- [x] Manually

**Commit flow** 

- Add platform information on startup
- Add platform information to About menu

**Notes & Questions** 

**Interactions** 

**Visual changes** 
image
image

### DIFF
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -17,6 +17,7 @@ from zulipterminal.cli.run import (
     parse_args,
 )
 from zulipterminal.model import ServerConnectionFailure
+from zulipterminal.platform_code import PLATFORM
 from zulipterminal.version import ZT_VERSION
 
 
@@ -132,6 +133,7 @@ def test_valid_zuliprc_but_no_connection(
 
     lines = captured.out.strip().split("\n")
     expected_lines = [
+        "Detected Platform: " + PLATFORM,
         "Loading with:",
         "   theme 'zt_dark' specified with no config.",
         "   autohide setting 'no_autohide' specified with no config.",
@@ -182,6 +184,7 @@ def test_warning_regarding_incomplete_theme(
 
     lines = captured.out.strip().split("\n")
     expected_lines = [
+        "Detected Platform: " + PLATFORM,
         "Loading with:",
         f"   theme '{bad_theme}' specified on command line.",
         "\x1b[93m   WARNING: Incomplete theme; results may vary!",
@@ -395,6 +398,7 @@ def test_successful_main_function_with_config(
     captured = capsys.readouterr()
     lines = captured.out.strip().split("\n")
     expected_lines = [
+        f"Detected Platform: {PLATFORM}",
         "Loading with:",
         "   theme 'zt_dark' specified in zuliprc file (by alias 'default').",
         "   autohide setting 'autohide' specified in zuliprc file.",
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -20,6 +20,7 @@ from zulipterminal.config.themes import (
 )
 from zulipterminal.core import Controller
 from zulipterminal.model import ServerConnectionFailure
+from zulipterminal.platform_code import PLATFORM
 from zulipterminal.version import ZT_VERSION
 
 
@@ -439,6 +440,10 @@ def main(options: Optional[List[str]] = None) -> None:
         if args.notify:
             zterm["notify"] = (args.notify, "on command line")
 
+        if PLATFORM:
+            print("Detected Platform: " + PLATFORM)
+        else:
+            print("Running on undetected platform.")
         print("Loading with:")
         print("   theme '{}' specified {}.".format(*theme_to_use))
         complete, incomplete = complete_and_incomplete_themes()
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -38,6 +38,7 @@ from zulipterminal.helper import (
     match_stream,
     match_user,
 )
+from zulipterminal.platform_code import PLATFORM
 from zulipterminal.server_url import near_message_url
 from zulipterminal.ui_tools.boxes import MessageBox, PanelSearchBox
 from zulipterminal.ui_tools.buttons import (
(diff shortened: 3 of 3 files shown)


zulipterminal/core.py
```diff
@@ -145,6 +145,7 @@ def spinning_cursor() -> Any:
 
         spinner = spinning_cursor()
         sys.stdout.write("\033[92mWelcome to Zulip.\033[0m\n")
+        sys.stdout.write(f"Running on {PLATFORM}.\n")
```
neiljp, line 148: This location does work, but note that most of the rest of the startup code is otherwise in `run.py`, for which we have tests, and for which having a small test for this would be reasonable too.

This approach is also supported by the fact that it'd likely be useful to see the platform at the top of the loading process? It doesn't necessarily have to be in the 'loading with' section, as that's historically been used just for controllable configuration, rather than an auto-detected platform.

zulipterminal/ui_tools/views.py
```diff
@@ -1113,7 +1114,7 @@ def __init__(
             else []
         )
         contents = [
-            ("Application", [("Zulip Terminal", zt_version)]),
+            ("Application", [("Zulip Terminal", zt_version), ("Platform", PLATFORM)]),
```
neiljp, line 1117: This looks OK, though I had originally envisaged it being in a separate section below the others, rather than in the `Application` section.

neiljp (commented):
@plugyawn Thanks for adding this, when ready it should be helpful for debugging in future :+1:

Generally it may be useful to use something along the lines of `Detected Platform` instead of the 'Running' and 'Platform' you currently have, both for consistency, and since this is what ZT determines is the platform - we can't know for sure.

This doesn't currently address the idea of showing extra text if the platform is unsupported, from the original issue; you didn't mention skipping it for now in the PR description, so I assume you missed that.

I think I mentioned this in another PR review, but I'd recommend reading your PRs in GitHub and maybe also something like tig or another local viewer, to ensure that the commit text and contents looks like what you'd expect.

zulipterminal/core.py
plugyawn: Okay, makes sense. I'll rellocate it to run.py.

zulipterminal/ui_tools/views.py
plugyawn: I assumed the version and the platform would go together? I'll put it in a separate section, that does sound better.

plugyawn:
> @plugyawn Thanks for adding this, when ready it should be helpful for debugging in future 👍
> 
> Generally it may be useful to use something along the lines of `Detected Platform` instead of the 'Running' and 'Platform' you currently have, both for consistency, and since this is what ZT determines is the platform - we can't know for sure.
> 
> This doesn't currently address the idea of showing extra text if the platform is unsupported, from the original issue; you didn't mention skipping it for now in the PR description, so I assume you missed that.
> 
> I think I mentioned this in another PR review, but I'd recommend reading your PRs in GitHub and maybe also something like tig or another local viewer, to ensure that the commit text and contents looks like what you'd expect.

I missed the part about checking if the platform is unsupported, sorry -- I'm adding it right now.

zulipterminal/ui_tools/views.py
```diff
@@ -1113,7 +1114,13 @@ def __init__(
             else []
         )
         contents = [
-            ("Application", [("Zulip Terminal", zt_version)]),
+            (
+                "Application",
+                [
+                    ("Zulip Terminal", zt_version),
+                    ("Platform detected", PLATFORM if PLATFORM else "Invalid platform"),
```
neiljp, line 1121: This condition shouldn't arise - again, check the values of PLATFORM.

tests/cli/test_run.py
```diff
@@ -395,6 +398,7 @@ def test_successful_main_function_with_config(
     captured = capsys.readouterr()
     lines = captured.out.strip().split("\n")
     expected_lines = [
+        f"Detected Platform: {PLATFORM}",
```
neiljp, line 401: Is there a reason you used different code here?

tests/cli/test_run.py
```diff
@@ -132,6 +133,7 @@ def test_valid_zuliprc_but_no_connection(
 
     lines = captured.out.strip().split("\n")
     expected_lines = [
+        "Detected Platform: " + PLATFORM,
```
neiljp, line 136: Note that this will be the platform that the tests are running on. We may want to mock this instead to ensure we get some known output (and avoid the running vs testing confusion for readers).

It's good to keep tests in the same commit as the code that changes, as it's a unit of change, so here combined with one of the other commits.

We may want a specific test to handle how this output should vary on different platforms, including unsupported.

neiljp (commented):
@plugyawn You didn't mark this for review, but here's some feedback since you mentioned updating it in the stream. Please do toggle the labels if you're seeking review.

tests/cli/test_run.py
plugyawn: None, I'll change it to the same code as the others.

zulipterminal/cli/run.py
```diff
@@ -439,6 +440,10 @@ def main(options: Optional[List[str]] = None) -> None:
         if args.notify:
             zterm["notify"] = (args.notify, "on command line")
 
+        if PLATFORM:
+            print("Detected Platform: " + PLATFORM)
+        else:
+            print("Running on undetected platform.")
```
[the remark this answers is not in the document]
plugyawn: > Is this intended to be for unsupported platforms, or something else?
> 
> Please make sure you're aware what values this can take :)
> 
> You're not testing this result, in any case.

I'll look into that; I'm still not exactly sure how to check for unsupported platforms, though.

neiljp:
@plugyawn Thanks for your work on this :+1: I recently merged #1435, which includes the About popup element, so I'm closing this now.

Issue #1075, opened 2021-07-10
Support (un)resolving topics


neiljp:
This feature is likely to be in Zulip 5.0 and while still under development appears fairly settled and ready to be implemented.

This feature involves toggling a prefix+space marker, though we should check the web app to confirm what is required.

We don't yet have one, but this is likely most easily presented through an option in a TopicsPopUp.

PR #1235, opened 2022-06-20
Add support for (un)resolving topics in ZT.

srdeotarse:
**What does this PR do?**  
This PR adds support for (un)resolving topics in ZT via ``TopicInfoView`` popup menu when topic is highlighted in ``left_stream_bar`` and ``i`` key is pressed to toggle ``topic settings``.

[WIP] #1075

**Tested?** 
- [ ] Manually

**Commit flow** 

First commit adds ``TOPIC_DESC`` shortcut to ``keys.py``
Second commit adds above shorcut to ``TopicButton`` in ``buttons.py``
Third commit still figuring out a way to update topic name with ``RESOLVED_TOPIC_PREFIX`` in ``model.py``
Fourth commit  adds ``TopicInfoView`` to ``views.py``
Last commit adds ``show_topic_info`` function to ``core.py`` 

**Notes & Questions** 

How to update the topic name by adding or removing ``RESOLVED_TOPIC_PREFIX`` when ``Topic Resolved`` checkbox is toggled?

**Interactions** 

**Visual changes** 
image

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1097,11 +1097,11 @@ class TestModel:
                 },
                 {
                     "_": ["User not found", False],
-                    "_owner": ["Editing messages is disabled", False],
-                    "_admin": ["Editing messages is disabled", False],
-                    "_moderator": ["Editing messages is disabled", False],
-                    "_member": ["Editing messages is disabled", False],
-                    "_guest": ["Editing messages is disabled", False],
+                    "_owner": [" Editing messages is disabled", False],
+                    "_admin": [" Editing messages is disabled", False],
+                    "_moderator": [" Editing messages is disabled", False],
+                    "_member": [" Editing messages is disabled", False],
+                    "_guest": [" Editing messages is disabled", False],
                 },
                 id="editing_msg_disabled:PreZulip4.0",
             ),
@@ -1114,9 +1114,9 @@ class TestModel:
                     "_": ["User not found", False],
                     "_owner": ["", True],
                     "_admin": ["", True],
-                    "_moderator": ["Editing topic is disabled", False],
-                    "_member": ["Editing topic is disabled", False],
-                    "_guest": ["Editing topic is disabled", False],
+                    "_moderator": [" Editing topic is disabled", False],
+                    "_member": [" Editing topic is disabled", False],
+                    "_guest": [" Editing topic is disabled", False],
                 },
                 id="editing_topic_disabled:PreZulip4.0",
             ),
@@ -1136,18 +1136,6 @@ class TestModel:
                 id="editing_topic_and_msg_enabled:PreZulip4.0",
             ),
             case(
-                {"allow_message_editing": False, "edit_topic_policy": 1},
-                {
-                    "_": ["User not found", False],
-                    "_owner": ["Editing messages is disabled", False],
-                    "_admin": ["Editing messages is disabled", False],
-                    "_moderator": ["Editing messages is disabled", False],
-                    "_member": ["Editing messages is disabled", False],
-                    "_guest": ["Editing messages is disabled", False],
-                },
-                id="all_but_no_editing:Zulip4.0+:ZFL75",
-            ),
-            case(
                 {"allow_message_editing": True, "edit_topic_policy": 1},
                 {
                     "_": ["User not found", False],
@@ -1156,8 +1144,8 @@ class TestModel:
                     "_moderator": ["", True],
                     "_member": ["", True],
                     "_guest": [
-                        "Only organization administrators, moderators, full members and"
-                        " members can edit topic",
+                        "Only organization administrators, moderators, full members"
+                        " and members can edit topic",
                         False,
                     ],
                 },
@@ -1208,13 +1196,13 @@ class TestModel:
                     "_admin": ["", True],
                     "_moderator": ["", True],
                     "_member": [
-                        "Only organization administrators and moderators"
-                        " can edit topic",
+                        "Only organization administrators and moderators can edit"
+                        " topic",
                         False,
                     ],
                     "_guest": [
-                        "Only organization administrators and moderators"
-                        " can edit topic",
+                        "Only organization administrators and moderators can edit"
+                        " topic",
                         False,
                     ],
                 },
@@ -1246,9 +1234,9 @@ class TestModel:
     ):
         model.get_user_info = mocker.Mock(return_value=user_role)
         model.initial_data = initial_data
-        initial_data["realm_allow_message_editing"] = realm_editing_settings[
+        initial_data["realm_allow_message_editing"] = realm_editing_settings.get(
             "allow_message_editing"
-        ]
+        )
         allow_community_topic_editing = realm_editing_settings.get(
             "allow_community_topic_editing", None
         )
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -7,6 +7,7 @@ import pytest
 from pytest import param as case
 from zulip import Client, ZulipError
 
+from zulipterminal.api_types import RESOLVED_TOPIC_PREFIX
 from zulipterminal.config.symbols import STREAM_TOPIC_SEPARATOR
 from zulipterminal.helper import initial_index, powerset
 from zulipterminal.model import (
@@ -1258,6 +1259,87 @@ class TestModel:
             report_error.assert_not_called()
         else:
             report_error.assert_called_once_with(expected_response[user_type][0])
+
+    @pytest.mark.parametrize(
+        "topic_name, msg_response, server_feature_level, topic_editing_limit_seconds,"
+        " expected_new_topic_name, expected_footer_error",
+        [
+            case(
+                "hi!",
+                {
+                    "subject": "hi!",
+                    "timestamp": 11662271397,
+                    "id": 1,
+                },
+                12,
+                259200,
+                RESOLVED_TOPIC_PREFIX + "hi!",
+                None,
+                id="topic_resolved:Zulip2.1+:ZFL12",
+            ),
+            case(
+                "hi!",
+                {
+                    "subject": "hi!",
+                    "timestamp": 0,
+                    "id": 1,
+                },
+                None,
+                None,
+                RESOLVED_TOPIC_PREFIX + "hi!",
+                None,
+                id="no_time_limit:Zulip2.1+:None",
+            ),
+            case(
+                RESOLVED_TOPIC_PREFIX + "hi!",
+                {
+                    "subject": RESOLVED_TOPIC_PREFIX + "hi!",
+                    "timestamp": 11662271397,
+                    "id": 1,
+                },
+                10,
+                86400,
+                "hi!",
+                None,
+                id="topic_unresolved:Zulip2.1+:ZFL10",
+            ),
+        ],
+    )
+    def test_toggle_topic_resolve_status(
+        self,
+        mocker,
+        model,
+        initial_data,
+        topic_name,
+        msg_response,
+        server_feature_level,
+        topic_editing_limit_seconds,
+        expected_new_topic_name,
+        expected_footer_error,
+        stream_id=1,
+    ):
+        model.initial_data = initial_data
+        model.server_feature_level = server_feature_level
+        initial_data[
+            "realm_community_topic_editing_limit_seconds"
+        ] = topic_editing_limit_seconds
+        # If user can't edit topic, topic (un)resolve is disabled. Therefore,
+        # default return_value=True
+        model.can_user_edit_topic = mocker.Mock(return_value=True)
+        model.get_latest_message_in_topic = mocker.Mock(return_value=msg_response)
+        model.update_stream_message = mocker.Mock(return_value={"result": "success"})
+        report_error = model.controller.report_error
+
+        model.toggle_topic_resolve_status(stream_id, topic_name)
+
+        if not expected_footer_error:
+            model.update_stream_message.assert_called_once_with(
+                message_id=msg_response["id"],
+                topic=expected_new_topic_name,
+                propagate_mode="change_all",
+            )
+        else:
+            report_error.assert_called_once_with(expected_footer_error)
 
     # NOTE: This tests only getting next-unread, not a fixed anchor
     def test_success_get_messages(
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -310,6 +310,11 @@ KEY_BINDINGS: Dict[str, KeyBinding] = {
         'help_text': 'View user information (From Users list)',
         'key_category': 'general',
     },
+    'TOPIC_INFO': {
+        'keys': ['i'],
+        'help_text': 'Show/hide topic information & modify settings',
+        'key_category': 'general',
+    },
     'BEGINNING_OF_LINE': {
         'keys': ['ctrl a'],
         'help_text': 'Jump to the beginning of line',
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -43,6 +43,7 @@ from zulipterminal.ui_tools.views import (
     PopUpConfirmationView,
     StreamInfoView,
     StreamMembersView,
+    TopicInfoView,
     UserInfoView,
 )
 from zulipterminal.version import ZT_VERSION
@@ -304,6 +305,10 @@ class Controller:
     def show_stream_info(self, stream_id: int) -> None:
         show_stream_view = StreamInfoView(self, stream_id)
         self.show_pop_up(show_stream_view, "area:stream")
+
+    def show_topic_info(self, stream_id: int, topic_name: str) -> None:
+        show_topic_view = TopicInfoView(self, stream_id, topic_name)
+        self.show_pop_up(show_topic_view, "area:topic")
 
     def show_stream_members(self, stream_id: int) -> None:
         stream_members_view = StreamMembersView(self, stream_id)
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -648,7 +648,7 @@ class Model:
         user_info = self.get_user_info(self.user_id)
         if user_info is not None:
             if not self.initial_data.get("realm_allow_message_editing"):
-                self.controller.report_error("Editing messages is disabled")
+                self.controller.report_error(" Editing messages is disabled")
                 return False
             role = user_info["role"]
             if role <= 200:
@@ -661,7 +661,7 @@ class Model:
             if allow_community_topic_editing is True:
                 return True
             elif allow_community_topic_editing is False:
-                self.controller.report_error("Editing topic is disabled")
+                self.controller.report_error(" Editing topic is disabled")
                 return False
             else:
                 edit_topic_policy = self.initial_data.get("realm_edit_topic_policy")
@@ -693,7 +693,7 @@ class Model:
                     else:
                         self.controller.report_error(EDIT_TOPIC_POLICY[1])
                         return False
-                else:  # edit_topic_policy == 5 (or None)
+                else:
                     # All users including guests
                     return True
         self.controller.report_error("User not found")
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -30,6 +30,7 @@ from typing_extensions import Literal, TypedDict
 
 from zulipterminal import unicode_emojis
 from zulipterminal.api_types import (
+    RESOLVED_TOPIC_PREFIX,
     Composition,
     EditPropagateMode,
     Event,
@@ -698,6 +699,38 @@ class Model:
                     return True
         self.controller.report_error("User not found")
         return False
+
+    def toggle_topic_resolve_status(self, stream_id: int, topic_name: str) -> None:
+        if self.can_user_edit_topic():
+            latest_msg = self.get_latest_message_in_topic(stream_id, topic_name)
+            if latest_msg:
+                time_since_msg_sent = time.time() - latest_msg["timestamp"]
+                # ZFL < 11, community_topic_editing_limit_seconds
+                # was hardcoded as int value in secs eg. 86400s (1 day) or None
+                if self.server_feature_level is None or self.server_feature_level >= 11:
+                    edit_time_limit = self.initial_data.get(
+                        "realm_community_topic_editing_limit_seconds", None
+                    )
+                else:
+                    edit_time_limit = 86400
+                # Don't allow editing topic if time-limit exceeded.
+                if (
+                    edit_time_limit is not None
+                    and time_since_msg_sent >= edit_time_limit
+                ):
+                    self.controller.report_error(
+                        " Time limit for editing topic has been exceeded."
+                    )
+                else:
+                    if topic_name.startswith(RESOLVED_TOPIC_PREFIX):
+                        topic_name = topic_name[2:]
+                    else:
+                        topic_name = RESOLVED_TOPIC_PREFIX + topic_name
+                    self.update_stream_message(
+                        message_id=latest_msg["id"],
+                        topic=topic_name,
+                        propagate_mode="change_all",
+                    )
 
     def generate_all_emoji_data(
         self, custom_emoji: Dict[str, RealmEmojiData]
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -349,6 +349,8 @@ class TopicButton(TopButton):
         if is_command_key("TOGGLE_TOPIC", key):
             # Exit topic view
             self.view.left_panel.show_stream_view()
+        elif is_command_key("TOPIC_INFO", key):
+            self.model.controller.show_topic_info(self.stream_id, self.topic_name)
         return super().keypress(size, key)
 
 
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -10,7 +10,7 @@ import pytz
 import urwid
 from typing_extensions import Literal
 
-from zulipterminal.api_types import EditPropagateMode
+from zulipterminal.api_types import RESOLVED_TOPIC_PREFIX, EditPropagateMode
 from zulipterminal.config.keys import (
     HELP_CATEGORIES,
     KEY_BINDINGS,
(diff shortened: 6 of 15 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1492,6 +1493,55 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+        super().__init__(controller, self.widgets, "TOPIC_DESC", popup_width, title)
+
+    def toggle_resolve_status(self, button: Any, new_state: bool) -> None:
+        if self.topic_name.startswith(RESOLVED_TOPIC_PREFIX):
+            self.controller.model.toggle_topic_resolved_status(
+                self.stream_id, self.topic_name
+            )
+        else:
+            self.controller.model.toggle_topic_resolved_status(
+                self.stream_id, RESOLVED_TOPIC_PREFIX + self.topic_name
```
neiljp, line 1541: It seems strange to have the model function called `toggle_...` which you then pass a different parameter to depending upon the topic name, and which is the *new* name - other than relying on the called function to do the toggling. That is, I'd expect a toggle function to be passed in fixed parameters. In other words, a toggle is a 'move to the other position', not a 'move to *this* position'.

For at least some similar cases we have logic in the model function instead. Take a look at the muting and pinning logic, for example.

zulipterminal/core.py
```diff
@@ -300,6 +301,10 @@ def show_stream_info(self, stream_id: int) -> None:
         show_stream_view = StreamInfoView(self, stream_id)
         self.show_pop_up(show_stream_view, "area:stream")
 
+    def show_topic_info(self, stream_id: int, topic_name: str) -> None:
+        show_topic_view = TopicInfoView(self, stream_id, topic_name)
+        self.show_pop_up(show_topic_view, "area:stream")
```
neiljp, line 306: Similarly to the help, we may want to update the label here, or add a new one if want a new area - though we risk running out of styles :)

neiljp (commented):
@srdeotarse This is a good almost-functional start with good commit separation, but possibly missing some consideration of the commit ordering. Looking forward to some tests later too, at least initially for the model.

zulipterminal/config/keys.py
```diff
@@ -279,6 +279,11 @@ class KeyBinding(TypedDict):
         'help_text': 'Show/hide stream information & modify settings',
         'key_category': 'stream_list',
     }),
+    ('TOPIC_DESC', {
+        'keys': ['i'],
+        'help_text': 'Show/hide topic information & modify settings',
+        'key_category': 'stream_list',
```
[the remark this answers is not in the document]
srdeotarse: How about ``topic_list``?

Rohitth007:
I think it's better to remove `Topic settings` from the popup because when we add things like `Mark all as read`, it's not exactly a setting. Also maybe lets use a button, that says either `Resolve Topic` or `Unresolve Topic`, instead of a checkbox because when we add things like `Edit Topic`, it won't exactly be a checkbox.

zulipterminal/config/keys.py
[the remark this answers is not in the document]
neiljp: Looking at similar popups/sections, we have one in the message-actions section (MSG_INFO in msg_actions), one in stream list actions section (STREAM_DESC in stream_list), but also user information in general (USER_INFO in general).

Based on that, it seems like it might be worth tidying the existing STREAM_DESC into an _INFO (so we can use a similar suffix in this commit). Until we restructure the help sections, to be consistent for now, we could either add the new hotkey to general (less tidy) or move it to a new topic_list section (and also move the existing user one into one two).

The latter option would possibly also allow adding an entry for ENTER in the user list for replying to a user, which we could possibly do in other cases too in some way, though we might want to hold off on that until we have more of a plan for that without adding too many options.

neiljp (commented):
@srdeotarse The commit ordering here is cleaner and I've tested manually and confirmed it works for me - at least on topics I've written to :+1:

As per the note inline, if the server has editing disabled then I believe topics cannot currently be resolved (and this will lead to this PR erroring?), and at the other end if 'community editing' is enabled (or various other options) then anyone can resolve at any time (perhaps excluding guests?) (which will lead to this PR not working when it should?). That may motivate a refactor of the existing topic-editing code to extract this logic, and connect to adding the community-editing feature (ie. not needing to have posted to a topic to edit it). I've also possibly seen some discussion elsewhere on czo about having separate topic resolution settings, but I believe these are currently still based on topic editing, though this would be good to confirm - and of course for now this remains the case.

I'm unsure if the updated UI is improved, particularly since multiple ENTER/SPACE/CLICK doesn't toggle (or show it), and there's no way of either avoiding it toggling on exiting the popup, or knowing if there will be an action on exiting the popup.

While there may be updates to the model logic, now that we're past WIP in that it works, it would be useful to get some preliminary tests for that logic - and that may help think about the error cases we can get and what messages the user may see and when. You may find that extracting the logic for knowing if a user can edit the topic separately (and testing that) will make this function less tangled, if you address that first.

zulipterminal/ui_tools/views.py
```diff
@@ -1492,6 +1493,58 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+
+        self.widgets = self.make_table_with_categories(
+            topic_info_content, column_widths
+        )
+
+        self.widgets.append(resolve_topic_setting)
+        super().__init__(controller, self.widgets, "TOPIC_DESC", popup_width, title)
+
+    def toggle_resolve_status(self, args: Any) -> None:
+        self.resolve_topic_btn_lbl = "Button clicked"
```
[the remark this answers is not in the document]
Rohitth007: Probably exiting out of the popup (after executing current functionality) as soon as you press the button would help with this.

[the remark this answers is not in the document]
srdeotarse: So, like if we press the button, the popup will exit and the topic will be (un)resolved.
But, we will be also adding different settings to the topic popup in the future. So if a user wants to (un)resolve topic and change any other setting, user again has to toggle the popup?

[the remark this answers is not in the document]
Rohitth007: Since we'll place 1 time use features here like `Mute Topic`, `Mark topic as read`, `Copy Topic Link`, `Resolve Topic`, it make sense to close it when you activate the button.

zulipterminal/model.py
```diff
@@ -559,6 +560,38 @@ def update_private_message(self, msg_id: int, content: str) -> bool:
         display_error_if_present(response, self.controller)
         return response["result"] == "success"
 
+    def toggle_topic_resolved_status(self, stream_id: int, topic_name: str) -> bool:
+        request: Dict[str, Any] = {
+            "anchor": "newest",
+            "num_before": 1,
+            "num_after": 0,
+            "narrow": [
+                {"operator": "sender", "operand": self.user_email},
```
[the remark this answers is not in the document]
Rohitth007: Agreed, I think removing this might make it possible to edit topic names even if you haven't sent any messages there.
Re, for topic edit permissions, I think we should do that before calling the toggle function. We should ideally not show them the button in the first place.

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
neiljp: We can revisit this, but a button like Rohitth indicates may be simplest for now.

Note that for the stream info, we have a check-box which updates, but as I think I mentioned before, that's a user setting so isn't as big an issues if eg. updated repeatedly via a check-box.

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: True, if we can't resolve the topic, it would be confusing to bring it up. That said, in future we may want the topic info popup to be visible even if actions aren't possible, so we could take an approach a little like with the visual notifications checkbox in the stream popup when external notifications are not configured - that checkbox gets grayed out and we show red text to indicate why.

zulipterminal/model.py
```diff
@@ -559,6 +560,50 @@ def update_private_message(self, msg_id: int, content: str) -> bool:
...
+                            "topic": topic_name[2:],
+                            "propagate_mode": "change_all",
+                        }
+                    else:
+                        request = {
+                            "message_id": topic_msg_id,
+                            "topic": RESOLVED_TOPIC_PREFIX + topic_name,
+                            "propagate_mode": "change_all",
+                        }
+                    response = self.client.update_message(request)
```
Rohitth007, line 600: Can `model.update_stream_message()` be used here? unless you don't want to display `"You changed all messages' topic from #stream>topic to #stream>✔ topic ."`

Rohitth007: If you instead want to show `"#stream>topic has been resolved"` then do include `display_error_if_present(response)` and report_success("#stream>topic has been resolved") here.

zulipterminal/model.py
```diff
@@ -559,6 +560,50 @@ def update_private_message(self, msg_id: int, content: str) -> bool:
...
+                        "operand": self.stream_dict[stream_id]["name"],
+                    },
+                    {"operator": "topic", "operand": topic_name},
+                ],
+            }
+            response = self.client.get_messages(request)
+            if response["messages"] != []:
+                time_since_msg_sent = time.time() - response["messages"][0]["timestamp"]
+                edit_time_limit = self.initial_data[
+                    "realm_community_topic_editing_limit_seconds"
```
Rohitth007, line 583: Didn't know topics had a time limit. Then why do we show `"Only topic editing allowed. Time limit for editing the message body has been exceeded."` when `realm_message_content_edit_limit_seconds` is crossed?

Rohitth007: Who can resolve such a topic then, only admins? I assume it would be a future feature for ZT then? Probably comment it?

zulipterminal/model.py
```diff
@@ -559,6 +560,50 @@ def update_private_message(self, msg_id: int, content: str) -> bool:
...
+                "num_after": 0,
+                "narrow": [
+                    {
+                        "operator": "stream",
+                        "operand": self.stream_dict[stream_id]["name"],
+                    },
+                    {"operator": "topic", "operand": topic_name},
+                ],
+            }
+            response = self.client.get_messages(request)
```
Rohitth007, line 579: Can `model.get_messages()` be used here? We may have to force users to narrow to a topic to resolve it so that `self.narrow` is not a problem.

If that is not ideal, would it be more readable if this whole request (along with error handling which you haven't done) was extracted into a function like`get_latest_message_in_topic`?

zulipterminal/ui_tools/buttons.py
```diff
@@ -322,6 +322,8 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         if is_command_key("TOGGLE_TOPIC", key):
             # Exit topic view
             self.view.left_panel.show_stream_view()
+        elif is_command_key("TOPIC_INFO", key):
```
Rohitth007, line 325: Probably better to move the hotkey changes here since `i` is not usable until this point.

Rohitth007 (commented):
Your code is looking better than before, especially after handling the API better.
I've mostly reviewed the API calls and not the Popup. I'll give that a look soon. Maybe 
@neiljp should also comment on the new API changes introduced as he's more familiar with them.

zulipterminal/model.py
```diff
@@ -584,13 +588,45 @@ def get_latest_message_in_topic(
...
             self.controller.report_error(" Editing messages is disabled")
             return
         elif self.initial_data.get("realm_allow_community_topic_editing", None):
             self.controller.report_error(" Editing topic is disabled")
             return
+        elif user_topic_permission is not None:
+            self.controller.report_error(user_topic_permission)
+            return
         else:
             response = self.get_latest_message_in_topic(stream_id, topic_name)
```
neiljp, line 631: You introduce `can_user_edit_topic` in this commit, but it's only handling a feature which I think is new (check the zulip API changelog). That it's looking to extend the base functionality to handle a new feature is great! The reason I've highlighted this section is since all the error outputs due to permissions in toggle_topic_resolved_status are also relevant to this, so I'd suggest adding can_user_edit_topic in the first related commit, then extending it here when you add the functionality only available in updated servers.

neiljp (commented):
@srdeotarse I'm coming back to this after Rohitth did some reviews, but I think the basic UI will be fine for now, and I note you've updated the logic handling as per some of our discussion so I've mostly focused on reading the code :+1:

I'm going to merge the first commit separately to shorten the PR and since this is independent and looks fine.

As some of the inline comments explain, I think the PR will benefit from introducing the can_user_edit_topic function at the start with all the error strings and checks you have in the toggle function, then simply going ahead in the toggle function otherwise.

I don't see a large reason to separate the first and second commits (currently 2nd and 3rd) changing the model, which would also make the can_user_edit_topic function more meaningful initially. The additional functionality for more featureful servers is perfectly fine to come later :+1:
(if it modifies a mostly working can_user_edit_topic, then the upgraded functionality commits could potentially even come after the UI, and likely the commits will move fairly easily like that if those need further work, to enable merging of the basic+UI commits)

Note that the issue won't really be fixed until we have a UI element to access it, so that fixed line belongs in the last commit currently.

Along with my overall feedback here, note that we need tests, certainly initially for can_user_edit_topic and toggle_topic_resolved_status (which will then adjust in later commits with the extra server feature). I expect when you write the tests, you'll see how making the function split early simplifies them.

zulipterminal/model.py
```diff
@@ -564,6 +565,53 @@ def update_private_message(self, msg_id: int, content: str) -> bool:
...
+                {"operator": "topic", "operand": topic_name},
+            ],
+        }
+        response = self.client.get_messages(request)
+        display_error_if_present(response, self.controller)
+        return response
+
+    def toggle_topic_resolved_status(self, stream_id: int, topic_name: str) -> bool:
+        response = self.get_latest_message_in_topic(stream_id, topic_name)
+        if response["messages"] != []:
```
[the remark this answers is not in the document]
srdeotarse: result["messages"] contains empty list if get_messages request fails. Therefore, the condition to check it before (un)resolving topics.

zulipterminal/model.py
```diff
@@ -564,6 +565,60 @@ def update_private_message(self, msg_id: int, content: str) -> bool:
...
+        else:
+            return True
+
+    def toggle_topic_resolved_status(self, stream_id: int, topic_name: str) -> None:
+        if self.can_user_edit_topic():
+            response = self.get_latest_message_in_topic(stream_id, topic_name)
+            if response["messages"] != []:
+                time_since_msg_sent = time.time() - response["messages"][0]["timestamp"]
+                edit_time_limit = self.initial_data[
+                    "realm_community_topic_editing_limit_seconds"
```
neiljp, line 635: Note that as per the API changelog, this parameter was new in feature level 11, so we need a backup value for this.

In an ideal world we could put all of the error conditions in one function, but let's keep them here for now.

zulipterminal/model.py
neiljp: We haven't added support for 'community' topic editing in the regular edit feature, so we give a notice there which represents the feature we support.

zulipterminal/model.py
```diff
@@ -564,6 +565,60 @@ def update_private_message(self, msg_id: int, content: str) -> bool:
...
+                    else:
+                        topic_name = RESOLVED_TOPIC_PREFIX + topic_name
+                    args = dict(
+                        message_id=response["messages"][0]["id"],
+                        topic=topic_name,
+                        propagate_mode="change_all",
+                    )
+                    self.update_stream_message(**args)
+
     def update_stream_message(
```
neiljp, line 654: This is outside your changes, but I noticed that the code splits this function from the private equivalent, and it would be good to keep them together.

neiljp (commented):
@srdeotarse This is looking cleaner :+1:

I would have posted the review sooner, but I have concerns over how we should best handle the logic generally. That is not down to your implementation in particular, but rather how it is documented and wider issues observed elsewhere. That seems related to how topic editing is dependent upon the message id in particular and if it belongs to the active user or not, as well as user roles.
eg.
- #issues > topic resolution order
- #issues > if marked as resolved it duplicates the argument
- #design > time restrictions on topic editing
- #21837

For now I think we should proceed as we are, and with the logic isolated and with tests we can consider if it matches expectations and adapt later.

We may want to use the first message in the narrow, as one possible example change.

zulipterminal/model.py
```diff
@@ -585,14 +589,42 @@ def get_latest_message_in_topic(
...
             self.controller.report_error(" Editing messages is disabled")
             return False
         elif self.initial_data.get("realm_allow_community_topic_editing", None):
             self.controller.report_error(" Editing topic is disabled")
             return False
-        else:
-            return True
+        elif user_info is not None:
+            role = user_info["role"]
+            if role is not None:
```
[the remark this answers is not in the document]
srdeotarse: Because mypy test is failing as we can't compare role(None) and int.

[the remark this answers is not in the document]
neiljp: Ah, so this seems to be due to the way in which that dict is constructed in get_user_info - approximately, then updated. I think we should be able to reorder that and make the dict element not be Optional. Do you want to try that as a refactor? It'd be a reasonable other PR or early commit that we could merge first. I could do that if you want to focus on this.

neiljp (commented):
@srdeotarse The reordering of commits is much easier to read now along with the tests. With extra tests I'd be comfortable starting to merge this, though as I've tried to clarify, we need to cover the different server versions properly, and the tests appear to describe strange editing behavior?

One other UI element to explore would be to lose the flashing cursor on the button. We do that elsewhere, so it's certainly achievable :)

I was testing this manually today and did trigger some thread tracebacks, but that's likely to be from the event handling instead.

neiljp:
@srdeotarse It'd also be good to get the hotkeys linting resolved somehow.

zulipterminal/model.py
```diff
@@ -564,6 +564,28 @@ def update_private_message(self, msg_id: int, content: str) -> bool:
         display_error_if_present(response, self.controller)
         return response["result"] == "success"
 
+    def get_latest_message_in_topic(
+        self, stream_id: int, topic_name: str
+    ) -> Optional[List[Dict[str, Any]]]:
+        request: Dict[str, Any] = {
+            "anchor": "newest",
+            "num_before": 1,
+            "num_after": 0,
```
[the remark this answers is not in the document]
srdeotarse: But if we fetch the oldest message then time_limit for editing topic will exceed the set threshold.

zulipterminal/model.py
```diff
@@ -564,6 +564,28 @@ def update_private_message(self, msg_id: int, content: str) -> bool:
         display_error_if_present(response, self.controller)
         return response["result"] == "success"
 
+    def get_latest_message_in_topic(
+        self, stream_id: int, topic_name: str
+    ) -> Optional[List[Dict[str, Any]]]:
```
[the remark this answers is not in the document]
srdeotarse: I didn't understand this point @neiljp

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: Hmm, yes this is confusing. Let's leave it at last for now.

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: I meant for the return type. The exact format would depend on if you adjust the data that's returned though.

neiljp (commented):
@srdeotarse Another good update :+1:

Lots/most of the inline comments are about the tests, due to the last update adding or adjusting them, or where we're missing coverage due to new code. Some are also outstanding issues, some being clearer now due to the revised tests.

I've resolved a lot of older comments which are fixed or superseded by these, and it feels like good progress :)

neiljp:
The lack of blinking cursor looks good too :+1:

tests/model/test_model.py
```diff
@@ -1019,6 +1019,185 @@ def test_get_latest_message_in_topic(
...
+                },
+                {
+                    "_": ["User not found", False],
+                    "_owner": ["", True],
+                    "_admin": ["", True],
+                    "_moderator": ["", True],
+                    "_member": ["", True],
+                    "_guest": ["", True],
+                },
+                id="editing_topic_disabled:PreZulip4.0",
```
neiljp, line 1080: Duplicate ID?

tests/model/test_model.py
```diff
@@ -1019,6 +1019,185 @@ def test_get_latest_message_in_topic(
...
+                    "_owner": ["", True],
+                    "_admin": ["", True],
+                    "_moderator": ["", True],
+                    "_member": ["", True],
+                    "_guest": [
+                        "Only organization administrators, moderators and members can edit topic",
+                        False,
+                    ],
+                },
+                id="all_users:Zulip4.0+:ZFL75",
```
neiljp, line 1095: The edit topic policy value and the response appears consistent with the server documentation here (though not the test id?), so based on my other comment regarding TOPIC_EDIT_POLICY not matching documentation, does the code use it correctly?

tests/model/test_model.py
```diff
@@ -1019,6 +1019,185 @@ def test_get_latest_message_in_topic(
...
+                    "_owner": ["", True],
+                    "_admin": ["", True],
+                    "_moderator": ["", True],
+                    "_member": ["", True],
+                    "_guest": [
+                        "Only organization administrators, moderators and full members can edit topic",
+                        False,
+                    ],
+                },
+                id="moderators:Zulip4.0+:ZFL75",
```
neiljp, line 1131: The first part of these strings don't all match the edit topic policy meanings?

neiljp (commented):
@srdeotarse Just a partial review here, since I think you didn't update everything from before and we discussed a few other points, and I wanted to get some feedback to you now.

neiljp:
@srdeotarse The last push was a good tidy following my review :+1: The first 3 commits were very close, so I just merged those with adjustments to commit messages, removed spaces at the start of some error strings, added a simple test case for no-edits in ZFL75 and the indexing point I mentioned at https://github.com/zulip/zulip-terminal/pull/1235#discussion_r962376941

neiljp (commented):
@srdeotarse Here's some extra feedback after the partial merge. I followed up on some older comments too, but you may need to look further up the thread of comments for context.

Mostly it's regarding improved testing or fairly minor points.

In manual testing I did find one case where it claimed the topic/message was too old to be edited, but I could do so manually. I've not dug deeper to know if that's due to the logic in your remaining un-merged model function, since it's not as well-tested yet, or due to nuances in the server logic which we didn't incorporate in the code that's already merged. Once the function is better tested then that might be clearer, but we can treat fine-tuning to can_user_edit_topic after we've merged this PR.

tests/model/test_model.py
```diff
@@ -1259,6 +1260,87 @@ def test_can_user_edit_topic(
...
+        "topic_name, msg_response, server_feature_level, topic_editing_limit_seconds,"
+        " expected_new_topic_name, expected_footer_error",
+        [
+            case(
+                "hi!",
+                {
+                    "subject": "hi!",
+                    "timestamp": 11662271397,
+                    "id": 1,
+                },
```
neiljp, line 1273: This dict appears identical in each case, except for the time, so let's build it in the function to keep it simpler? You could test for the `None` return value case, but that may be simpler in a separate test.

tests/model/test_model.py
```diff
@@ -1259,6 +1260,87 @@ def test_can_user_edit_topic(
...
+                12,
+                259200,
```
neiljp, line 1275: Is the timestamp within this time difference? The test case doesn't make it clear that's why we expect it to resolve.

zulipterminal/model.py
```diff
@@ -699,6 +700,38 @@ def can_user_edit_topic(self) -> bool:
...
         return False
 
+    def toggle_topic_resolve_status(self, stream_id: int, topic_name: str) -> None:
+        if self.can_user_edit_topic():
+            latest_msg = self.get_latest_message_in_topic(stream_id, topic_name)
+            if latest_msg:
+                time_since_msg_sent = time.time() - latest_msg["timestamp"]
+                # ZFL < 11, community_topic_editing_limit_seconds
+                # was hardcoded as int value in secs eg. 86400s (1 day) or None
+                if self.server_feature_level is None or self.server_feature_level >= 11:
```
neiljp, line 710: See https://github.com/zulip/zulip-terminal/pull/1235#discussion_r962358962

For server 2.1, this will set edit_time_limit to None.
For server 3.0 (or ZFL 11 at least), this will set edit_time_limit to the value in the server (None or the value).
For servers between those (eg. ZFL 10), this will set edit_limit to the default, ie. 86400 - this is an edge case

My impression is that the expected server behavior was
- `old` 86400 <-- 2.1
- `<11` 86400 <-- edge case
- `>=11` 86400 or None (per server value) <-- 3.0+

This appears to generate
- `old` None
- `<11` 86400
- `>=11` 86400 or None (per server value)

Is this going to fail for 2.1? If so, could you update the test to fail first?

tests/model/test_model.py
```diff
@@ -1259,6 +1260,87 @@ def test_can_user_edit_topic(
...
+        model.initial_data = initial_data
+        model.server_feature_level = server_feature_level
+        initial_data[
+            "realm_community_topic_editing_limit_seconds"
+        ] = topic_editing_limit_seconds
+        # If user can't edit topic, topic (un)resolve is disabled. Therefore,
+        # default return_value=True
+        model.can_user_edit_topic = mocker.Mock(return_value=True)
+        model.get_latest_message_in_topic = mocker.Mock(return_value=msg_response)
+        model.update_stream_message = mocker.Mock(return_value={"result": "success"})
```
neiljp, line 1330: See https://github.com/zulip/zulip-terminal/pull/1235#discussion_r962360251

This does not need a return value.

tests/model/test_model.py
```diff
@@ -1259,6 +1260,87 @@ def test_can_user_edit_topic(
...
+                RESOLVED_TOPIC_PREFIX + "hi!",
+                {
+                    "subject": RESOLVED_TOPIC_PREFIX + "hi!",
+                    "timestamp": 11662271397,
+                    "id": 1,
+                },
+                10,
+                86400,
+                "hi!",
+                None,
```
neiljp, line 1303: This value is always `None`? Have we lost some test cases?

neiljp (commented):

I went through some old comments, there are a few outstanding points to address from that and a quick read.

PR #1236, opened 2022-06-23
[WIP] Migrate message['subject'] to message['topic']

mounilKshah:
**What does this PR do?**
This commit migrates the ZT to newer API elements. Here, the migration is from message['subject'] to message['topic'].
Partial fix for #965

Discussion:
[#zulip-terminal>Update to modern API elements (support v2.1) #T965](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Update.20to.20modern.20API.20elements.20.28support.20v2.2E1.29.20.23T965)

**Tested?**
- [x] Manually

**Commit flow**
- First commit updates message element from `subject` to `topic`
- Second commit refactors the tests to accommodate this migration

**Notes & Questions**
- Fixes item 4 on #965

### DIFF
--- a/zulipterminal/api_types.py
+++ b/zulipterminal/api_types.py
@@ -35,6 +35,8 @@ class Message(TypedDict, total=False):
     timestamp: int
     client: str
     subject: str  # Only for stream msgs.
+    # Changing subject -> topic as per the migration in Zulip 2.0
+    topic: str  # Only for stream msgs.
     # NOTE: new in Zulip 3.0 / ZFL 1, replacing `subject_links`
     # NOTE: API response format of `topic_links` changed in Zulip 4.0 / ZFL 46
     topic_links: List[Any]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -416,7 +416,7 @@ def msg_template_factory(
     msg_type: str,
     timestamp: int,
     *,
-    subject: str = "",
+    topic: str = "",
     stream_id: Optional[int] = None,
     recipients: Union[str, List[Dict[str, Any]]] = "PTEST",
 ) -> Message:
@@ -435,7 +435,8 @@ def msg_template_factory(
         flags=["read"],
         sender_id=5140,
         content_type="text/x-markdown",
-        subject=subject,
+        topic=topic,
+        subject=topic,
         reactions=[],
         subject_links=[],
         avatar_url="dummy_avatar_url",
@@ -459,7 +460,7 @@ def msg_template_factory(
 @pytest.fixture
 def stream_msg_template() -> Message:
     msg_template = msg_template_factory(
-        537286, "stream", 1520918722, subject="Test", stream_id=205
+        537286, "stream", 1520918722, topic="Test", stream_id=205
     )
     return msg_template
 
@@ -467,7 +468,7 @@ def stream_msg_template() -> Message:
 @pytest.fixture
 def extra_stream_msg_template() -> Message:
     msg_template = msg_template_factory(
-        537289, "stream", 1520918740, subject="Test", stream_id=205
+        537289, "stream", 1520918740, topic="Test", stream_id=205
     )
     return msg_template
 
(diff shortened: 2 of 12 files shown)


zulipterminal/api_types.py
```diff
@@ -21,8 +21,8 @@ class PrivateComposition(TypedDict):
 class StreamComposition(TypedDict):
     type: Literal["stream"]
     content: str
-    to: str  # stream name  # TODO: Migrate to using int (stream id)
-    subject: str  # TODO: Migrate to using topic
+    to: Optional[int]
```
neiljp, line 24: This is a separate migration. It may be OK in this PR, but certainly in another commit.

zulipterminal/api_types.py
```diff
@@ -21,8 +21,8 @@ class PrivateComposition(TypedDict):
...
+    topic: str
```
neiljp, line 25: As per a comment in the stream, it may be cleaner to separate the compose and message parts of the topic migration into separate commits (with associated test changes in each).

zulipterminal/api_types.py
```diff
@@ -34,7 +34,9 @@ class Message(TypedDict, total=False):
     content: str
     timestamp: int
     client: str
-    subject: str  # Only for stream msgs.
+    subject: str
+    # Changing subject -> topic as per the migration in Zulip 2.0
+    topic: str  # Only for stream msgs.
```
neiljp, line 39: I'd suggest we comment out the subject, with a trailing comment referring to the migration.
In an ideal world that would mean using subject would cause a type error, but currently this TypedDict has total=False, so that won't happen (yet).

zulipterminal/api_types.py
```diff
@@ -53,7 +55,7 @@ class Message(TypedDict, total=False):
     avatar_url: str
     content_type: str
     match_content: str  # If keyword search specified in narrow params.
-    match_subject: str  # If keyword search specified in narrow params.
+    match_topic: str  # If keyword search specified in narrow params.
```
neiljp, line 58: Do you have support for this migration?

zulipterminal/api_types.py
```diff
@@ -145,8 +147,8 @@ class UpdateMessageEvent(TypedDict):
     rendered_content: str
     # B: Subject of these message ids needs updating?
     message_ids: List[int]
-    orig_subject: str
-    subject: str
+    orig_topic: str
+    topic: str
```
neiljp, line 151: Do you have support for this migration in the API?

neiljp (commented):
@mounilKshah You have the right ideas here, but as per my stream feedback, it's good to keep code changes and associated test changes in the same commit.

In addition, you've got multiple changes here, which ideally would be in different commits. Part of that is that the main subject->topic migration is split into Message and Composition parts, ie. what we receive from the server vs what a user sends. However, the other aspect are API changes that have similar names but haven't necessarily changed, or that are adjacent. It's fine to make these changes, but it would be cleaner to see each change in a separate commit.

Each API change may be in a different version, so that opens up the scope for writing differently about the change, when it occurred, etc. in the commit text body.

zulipterminal/api_types.py
mounilKshah: No. Actually, as `subject` was changed to `topic`, I felt it would be good to rename this accordingly. But as the API is still the same, I'll retain the original names for this and `orig_topic`

neiljp:
@mounilKshah I left some answers in the stream; do you need further feedback before the tests are updated?

neiljp:
@mounilKshah Thanks for investigating this!

I left a message in the stream, but for other readers also:
- The API relating to migration of the subject field to topic is the only (or almost only) renaming of this kind which it seems has *not* yet occurred.
- Since the only current commit in this PR relates to this, I propose we hold off on this migration, since this changes our internal data structures unnecessarily.
- Another part related to this migration is when sending messages via Composition objects, which supported topic since 2.0.0; let's redirect that effort into another PR.

PR #1239, opened 2022-06-26
Skip muted streams and topics using `n` (from #442)

neiljp:
**What does this PR do?**  

This is a WIP taking the last commit of #442, rebasing, reformatting and correcting for how that PR treats is_muted_topic as also incorporating if the stream is muted (which the model method we have now doesn't do).

This was brought up again recently in #**zulip-terminal > ZT impressions** by @rht but I'm continuing the discussion in #**zulip-terminal > Skip muted topics and streams on 'n' #T442 #T1239**.

I tested this in explore mode and toggled muted streams and topics. For example, where it starts on #general (id=2) for me, and will skip #backend (id=3) next if muted, and then also skips #test here just fine.

**Tested?** 
- [x] Manually

**Commit flow** 

The flow is very WIP; this is just for functional testing and broad feedback. I've reformatted and it should be functional, but I've not adapted the tests from #442 so far.

**Notes & Questions** 

Would appreciate feedback from anyone who uses the `n` shortcut regularly.

Refactor in last commit improves clarity?

This doesn't fix various other factors regarding this feature, including:
- order of unreads is stream via stream id, and topic alphabetically, neither of which match the left panel list order (which is likely desirable)
- I've not dug into if the unreads get updated cleanly when messages get read or new messages are received

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -832,24 +832,58 @@ class TestMiddleColumnView:
             "MSG_LIST", header=self.search_box, footer=self.write_box
         )
 
-    def test_get_next_unread_topic(self, mid_col_view):
-        mid_col_view.model.unread_counts = {"unread_topics": {1: 1, 2: 1}}
-        return_value = mid_col_view.get_next_unread_topic()
-        assert return_value == 1
-        assert mid_col_view.last_unread_topic == 1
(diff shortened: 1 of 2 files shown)


rht:
I have tested this and it works great.

neiljp:
Closing in favor of #1245, which is now merged 🎉

Issue #1238, opened 2022-06-24
Move message links above topic links in message information popup


neiljp:
Since images and other media can now be opened from here, and vary more per-message than topic links, this would improve the user experience.

See https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Order.20of.20message.20info.20links.20.23T1238
(#**zulip-terminal > Order of message links #T1238**)

To see this, go to a message in a topic which has a link in the topic name (including eg. a linkifier such as #T1238) and an attachment such as an image. While the link to this topic may be useful, it's not currently a cursor-related action, whereas moving down to the image and pressing enter will download and prompt to open the image. The aim of this issue is to show the group of message links above the group of topic links, to reduce the need to navigate down.

Kelp710:
@zulipbot claim

zulipbot:
Hello @Kelp710!

Thanks for your interest in Zulip! You have attempted to claim an issue without the label "help wanted". You can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/zulip-terminal/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) label.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

Kelp710:
@zulipbot claim.

zulipbot:
Welcome to Zulip, @Kelp710! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

zulipbot:
@Kelp710 We noticed that you have not made any updates to this issue or linked PRs for 10 days. Please comment here if you are still actively working on it. Otherwise, we'd appreciate a quick `@zulipbot abandon` comment so that someone else can claim this issue and continue from where you left off.

[...]

PR #1241, opened 2022-07-01
Changing the order of message links and topic links in popup(from #1238)

Kelp710:
**What does this PR do?**  
Changed the order of message links and topic links in popup. in order to achieve this, switched if statements in MsgInfoView class in views.py 

**Tested?** 
- [x] Manually

**Visual changes** <image>

### DIFF
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -876,7 +876,7 @@ class TestMsgInfoView:
         assert self.msg_info_view.message_links == OrderedDict()
         assert self.msg_info_view.time_mentions == list()
 
-    def test_pop_up_info(self, message_fixture: Message) -> None:
+    def test_pop_up_info_order(self, message_fixture: Message) -> None:
         msg_test = MsgInfoView(
             self.controller,
             message_fixture,
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -876,18 +876,20 @@ class TestMsgInfoView:
         assert self.msg_info_view.message_links == OrderedDict()
         assert self.msg_info_view.time_mentions == list()
 
-    def test_pop_up_info(self, message_fixture):
+    def test_pop_up_info(self, message_fixture: Message) -> None:
         msg_test = MsgInfoView(
             self.controller,
             message_fixture,
             "Message Information",
             OrderedDict([("https://bar.com", ("topic", 1, True))]),
-            OrderedDict([("image.jpg", ("image", 1, True))]),
+            OrderedDict([("image", ("image", 1, True))]),
             list(),
         )
         meg_button = msg_test.button_widgets
-        assert meg_button == [OrderedDict([("https://bar.com", ("topic", 1, True))]),OrderedDict([('image.jpg', ('image', 1, True))])]
-
+        assert meg_button == [
+            OrderedDict([("image", ("image", 1, True))]),
+            OrderedDict([("https://bar.com", ("topic", 1, True))]),
+        ]
 
     def test_keypress_any_key(
         self, widget_size: Callable[[Widget], urwid_Size]
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -876,6 +876,19 @@ class TestMsgInfoView:
         assert self.msg_info_view.message_links == OrderedDict()
         assert self.msg_info_view.time_mentions == list()
 
+    def test_pop_up_info(self, message_fixture):
+        msg_test = MsgInfoView(
+            self.controller,
+            message_fixture,
+            "Message Information",
+            OrderedDict([("https://bar.com", ("topic", 1, True))]),
+            OrderedDict([("image.jpg", ("image", 1, True))]),
+            list(),
+        )
+        meg_button = msg_test.button_widgets
+        assert meg_button == [OrderedDict([("https://bar.com", ("topic", 1, True))]),OrderedDict([('image.jpg', ('image', 1, True))])]
+
+
     def test_keypress_any_key(
         self, widget_size: Callable[[Widget], urwid_Size]
     ) -> None:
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1604,26 +1604,12 @@ def __init__(
...
-            widgets = widgets[:slice_index] + topic_link_widgets + widgets[slice_index:]
-            popup_width = max(popup_width, topic_link_width)
-
         if message_links:
             message_link_widgets, message_link_width = self.create_link_buttons(
                 controller, message_links
             )
 
-            # slice_index = Number of labels before message links + 1 newline
+            # slice_index = Number of labels before message links + 1 newlineSS
```
neiljp, line 1612: It's useful to review the commit using your own local tools (I use `tig` on the command line) or here on GitHub to find things like this which sneak in by accident :)

neiljp (commented):
@Kelp710 Thanks for looking at this! This looks good except for the extra line I commented on 👍

If you're open to explore further, we could add a test for the existing order first, and then change the expected behavior and update the test in the current commit you have now.

zulipterminal/ui_tools/views.py
Kelp710: Thank you for the excellent advice! I`ll fix the problem and look into testing.

zulipterminal/ui_tools/views.py
```diff
@@ -1609,7 +1609,7 @@ def __init__(
                 controller, message_links
             )
 
-            # slice_index = Number of labels before message links + 1 newlineSS
+            # slice_index = Number of labels before message links + 1 newline
```
Rohitth007, line 1612: @Kelp710 Usually when you want to change something that you did in a previous commit, you use git interactive rebase (`git rebase -i`) and change git history such that that error never happened in the first place!

In this case you can squash your 2 commits while doing an interactive rebase and it would fix this.

Rohitth007 (changes requested):
Learning to getting used to interactive rebase would lead to cleaner commits and faster merges.

zulipterminal/ui_tools/views.py
Kelp710: Thank you for the great advice! I`ll fit this asap

Kelp710:
Hi! I`m trying to do the testing part. However, I couldn`t understand what "we could add a test for the existing order first, and then change the expected behaviour and update the test in the current commit you have now" mean.
If you could, could you give me a little insight into it?

zulipterminal/ui_tools/views.py
```diff
@@ -1525,13 +1525,13 @@ def __init__(
         controller: Any,
         msg: Message,
         title: str,
-        topic_links: "OrderedDict[str, Tuple[str, int, bool]]",
         message_links: "OrderedDict[str, Tuple[str, int, bool]]",
+        topic_links: "OrderedDict[str, Tuple[str, int, bool]]",
```
neiljp, line 1529: This change isn't strictly necessary, and right now will cause an issue since you've not change the call sites.

neiljp:
> Hi! I'm trying to do the testing part. However, I couldn't understand what "we could add a test for the existing order first, and then change the expected behaviour and update the test in the current commit you have now" mean. If you could, could you give me a little insight into it?

I meant to add a separate first commit which would add a test for the object we're displaying, which would check the order of the sections matches what we expect. That new test should be expected to pass right now.

Then in the existing commit you have, you would update that new test you just added to ensure the new expected order instead.

If you need more detail or advice I'd suggest discussing in the stream topic on chat.zulip.org

Rohitth007:
@Kelp710 Were you meaning to squash the last 2 "new" commits into the first test commit?

Kelp710:
@Rohitth007 
Thank you for mentioning the problem!

neiljp:
@Kelp710 Are you ready for further feedback on this?

tests/ui_tools/test_popups.py
```diff
@@ -876,6 +876,19 @@ def test_init(self, message_fixture: Message) -> None:
...
+    def test_pop_up_info(self, message_fixture):
+        msg_test = MsgInfoView(
+            self.controller,
+            message_fixture,
+            "Message Information",
+            OrderedDict([("https://bar.com", ("topic", 1, True))]),
+            OrderedDict([("image.jpg", ("image", 1, True))]),
+            list(),
+        )
+        meg_button = msg_test.button_widgets
```
Rohitth007, line 888: meg?

tests/ui_tools/test_popups.py
```diff
@@ -876,6 +876,19 @@ def test_init(self, message_fixture: Message) -> None:
...
+        assert meg_button == [OrderedDict([("https://bar.com", ("topic", 1, True))]),OrderedDict([('image.jpg', ('image', 1, True))])]
```
Rohitth007: The spacing edits you did in the next commit can be done here itself.

Rohitth007: Also if you intend to not have `.jpg` then remove it from both the commits.

zulipterminal/ui_tools/views.py
```diff
@@ -1602,7 +1602,7 @@ def __init__(
 
         # To keep track of buttons (e.g., button links) and to facilitate
         # computing their slice indexes
-        button_widgets = []  # type: List[Any]
+        self.button_widgets = []  # type: List[Any]
```
Rohitth007, line 1605: Was `button_widget` made into an attribute only so that it can be accessed from tests? If so `MsgInfoView.log` or `MsgInfoView.body` would be better candidates for doing the test instead of `button_widget`

Rohitth007 (changes requested):
I left a few suggestions below. Apart from that be mindful of our max line length
in our commit style guide as it would lead to sentences getting chopped like in your 1st commit.

Kelp710:
@neiljp @Rohitth007 
Thank you for the message and feedback, I`ll try to fix it soon!

neiljp:
@Kelp710 Thanks for taking a look at this one previously :+1:

This was followed up in #1265, work I just split into the 2-commit style I requested here, ie. add a test for the current condition, then adjust the test and the code. I've adjusted both commits from there a little (particularly commit text), and primarily attributed the first commit to that author for following through with the PR and tidying the test, and left the second attributed to you.

This is now merged in the two commits ending 77218c1a646b431a463d3ab84ff61d92ceaf5c76 :tada:
(check the parent commit to see the previous one, or update your main branch)

PR #1234, opened 2022-06-20
[WIP] Highlight active stream in side panel.

plugyawn:
**What does this PR do?**  

In the side panel, once narrowed, the active stream is highlighted.

Partial fix for #516.

https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Highlight.20currently.20narrowed.20stream.2Ftopic.20.23T516

**Tested?** 
- [x] Manually
- [ ] Existing tests (adapted, if necessary)
- [ ] New tests added (for any new behavior)
- [ ] Passed linting & tests (each commit)

**Commit flow** 

**Notes & Questions** 

```buttons.by``` >> ```TopButton( )``` has an attribute named text_color which is required to highlight the text in the button, but I'm unsure what format the input needs to be in. [ FIXED ]

**Interactions** 

**Visual changes** 
image

There used to be no visual indicator on the side panel showing which stream was selected.

Issue #516, opened 2020-02-25
Indicate active stream


blenderskool:
The currently selected stream can be indicated a little differently from the rest of the streams. Few ideas that I have:
- Maybe a specific highlight color
- Making the active stream bold

neiljp:
@blenderskool I agree we could do this, though we already do show the active stream (or more broadly, 'narrow') at the top of the message list. Do you mean the active narrow being highlighted in eg. the stream list, user list, or whichever other place matches the narrow being shown?

blenderskool:
@neiljp Yes, the active stream is shown at the top of message list. But I think it's more natural to see the active stream on the sidebar. So that while choosing another stream, you don't have to look away from the sidebar to remind yourself of the active stream.

In the screenshot, we could have something highlighting the `#frontend` while still keeping it at the top of the messages list.

![zt-active-stream](https://user-images.githubusercontent.com/21107799/75334534-0afb0c00-58ae-11ea-810a-34a8a4b2ed08.jpg)

neiljp:
@blenderskool Right, that's what I suspected you meant, but also note that we'd likely want to do it for all other 'special' narrows that there is a user-interface for - ie. if a topic list is shown, highlight the topic, if in the top-left narrows, highlight those, if a single-user conversation, highlight the user, etc.

blenderskool:
@neiljp Yeah makes sense. There is already highlighting for the selected message, online users. Maybe a consistent pattern across the layouts could be built based on that.

SGeetansh:
@neiljp @amanagr 
I think this would be interesting to implement, should/can I work on this?

amanagr:
Yes, of course! (You can use `@zulipbot claim` to assign yourself on issues)

SGeetansh:
@zulipbot claim

zulipbot:
Hello @SGeetansh, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

SGeetansh:
@zulipbot claim

zulipbot:
[...]

PR #1242, opened 2022-07-01
[WIP] Highlighting selected stream.

plugyawn:
**What does this PR do?**  

In the side panel, once narrowed, the active stream is highlighted.

Partial fix for #516.
Supersedes #1234 

https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Highlight.20currently.20narrowed.20stream.2Ftopic.20.23T516

**Tested?** 
- [x] Manually

**Commit flow** 

**Notes & Questions** 

**Interactions** 

**Visual changes** 
image

There used to be no visual indicator on the side panel showing which stream was selected.


neiljp (commented):
@plugyawn This almost works for me right now - I added a line to set the style, as we do in update_widget to use set_attr_map, and it performs rather well :+1:

There are a few cases where the previous narrow remains marked, eg. stream->all-messages (press `a` while on a message). In those cases the previous stream stays selected, even though one is in the all-messages narrow. As soon as one selects a different narrow, this is updated though :+1:

While I do think we should consider storing the UI state outside of the model, before we approach that I think it would be beneficial to first get this code into a mergeable set of commits and fix the issue above (and any important others we find along the way). For example, add the new style in a separate commit first.

As part of getting the commits mergeable, also note that the additional style confuses ZT on startup; if you git-grep for existing styles then you'll see there is an existing location where they are specified for mono mode support and also for testing which styles a complete theme should have.

While this PR does interact with @Rohitth007 's #1088, the logic is separate and it's only the style setting that should be affected.

neiljp:
Any feedback on this by others would be great; the line I added was this, to the end of `update_widget_highlight`:
```python3
self._w.set_attr_map({None: highlight_style})

File: zulipterminal/ui_tools/buttons.py [hunk 1]
```diff
@@ -82,6 +83,9 @@ def update_widget(
         self.button_suffix.set_text(suffix)
         self._w.set_attr_map({None: text_color})
 
+    def update_widget_highlight(self, highlight_style: str) -> None:
+        self.original_color = highlight_style
```
srdeotarse, line 87: In WSL, it did not highlight active stream until I added line ``self._w.set_attr_map({None: highlight_style})`` as suggested by ``Neil``.

srdeotarse (changes requested):
* @plugyawn The selected stream highlight works fine and looks fantastic for ZT but, I also faced the same glitch that selected stream remained highlighted if switched to ``All messages``.
* Also, the selected stream highlight is working as expected when switched to ``PMs``, ``Starred messages`` and ``Mentions``.

Some suggestions to extend this selected item highlighting -
* We can also extend it to ``All messages``, ``PMs``, ``Starred messages`` and ``Mentions`` similar to selected stream highlighting. (This could be accomplished in a different PR)
* Also, we can implement highlighting ``user name`` or ``stream`` if we toggle ``information popup`` using ``i``

Rohitth007:
@plugyawn, apart from the topic panel doubt which I left in the stream:
* There's no need to create the `update_widget_highlight()` function. We can do everything directly from inside `mark_(in)active()`.
* Also don't use `self.original_color` as:
  * It will be removed by #1088 and
  * You would have to go through `update_count()` to use it, especially when we are not actually updating count.

neiljp:
Closing in favor of rebased/readjusted version of later version of this PR, ie. #1334.

PR #1244, opened 2022-07-28
Highlight narrowed stream/topic.

plugyawn:
**What does this PR do?**  
Solves #516. Highlights the current narrow. 

Discussed <a href=https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Highlight.20currently.20narrowed.20stream.2Ftopic.20.23T516.20.23T1242> here </a>.
**Tested?** 
- [x] Manually

**Commit flow** 

- first commit sets theme for zt_dark. Will rebase and change for other themes.
- second commit sets functions for highlighting button
- third commit is for model.py highlight whenever narrow_to is called.

**Notes & Questions** 

- Crashes when non-stream narrows are triggered (will fix, is fixed in last WIP PR)
- topics are not highlighted yet.
- Is the issue on linux systems fixed (related to second commit).

**Visual changes** 
image

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -71,6 +71,7 @@ REQUIRED_STYLES = {
     'area:error'      : 'standout',
     'area:user'       : 'standout',
     'search_error'    : 'standout',
+    'active_narrow'   : 'standout',
     'task:success'    : 'standout',
     'task:error'      : 'standout',
     'task:warning'    : 'standout',
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -98,6 +98,7 @@ class Model:
         self.stream_id: Optional[int] = None
         self.recipients: FrozenSet[Any] = frozenset()
         self.index = initial_index
+        self.active_button: Any = None
 
         self.user_id = -1
         self.user_email = ""
@@ -302,8 +303,36 @@ class Model:
         else:
             raise RuntimeError("Model.set_narrow parameters used incorrectly.")
 
+        try:
+            if self.stream_id:
+                self.active_button = self.controller.view.stream_id_to_button[
+                    self.stream_id
+                ]
+        except AttributeError:
+            pass
+
         if new_narrow != self.narrow:
+
+            try:
+                if self.stream_id and new_narrow:
+                    self.active_button.is_active = False
+                    self.active_button._w.set_attr_map({None: None})
+                if not self.narrow and self.active_button:
+                    self.active_button.is_active = False
+                    self.active_button._w.set_attr_map({None: None})
+
+            except AttributeError:
+                pass
+
             self.narrow = new_narrow
+
+            if stream and hasattr(Model, "controller"):
+                stream_button = self.controller.view.stream_id_to_button[
+                    self.stream_id_from_name(stream)
+                ]
+                self.active_button = stream_button
+                self.active_button.is_active = True
+                self.active_button._w.set_attr_map({None: "active_narrow"})
 
             if pm_with is not None and new_narrow[0][0] == "pm_with":
                 users = pm_with.split(", ")
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -26,14 +26,17 @@ class TopButton(urwid.Button):
         text_color: Optional[str] = None,
         count: int = 0,
         count_style: Optional[str] = None,
+        is_active: bool = False,
     ) -> None:
         self.controller = controller
+        self.model = controller.model
         self._caption = caption
         self.show_function = show_function
         self.prefix_character = prefix_character
         self.original_color = text_color
         self.count = count
         self.count_style = count_style
+        self.is_active = is_active
 
         super().__init__("")
 
@@ -80,7 +83,8 @@ class TopButton(urwid.Button):
         self.button_prefix.set_text(prefix)
         self.set_label(self._caption)
         self.button_suffix.set_text(suffix)
-        self._w.set_attr_map({None: text_color})
+        if not self.is_active:
+            self._w.set_attr_map({None: text_color})
 
     def activate(self, key: Any) -> None:
         self.controller.view.show_left_panel(visible=False)
@@ -160,6 +164,7 @@ class StreamButton(TopButton):
         controller: Any,
         view: Any,
         count: int,
+        is_active: bool = False,
     ) -> None:
         # FIXME Is having self.stream_id the best way to do this?
         # (self.stream_id is used elsewhere)
@@ -172,6 +177,7 @@ class StreamButton(TopButton):
         self.model = controller.model
         self.count = count
         self.view = view
+        self.is_active = is_active
 
         for entry in view.palette:
             if entry[0] is None:
--- a/zulipterminal/themes/gruvbox_dark.py
+++ b/zulipterminal/themes/gruvbox_dark.py
@@ -70,6 +70,7 @@ STYLES = {
     'area:error'       : (Color.DARK0_HARD,            Color.BRIGHT_RED),
     'area:user'        : (Color.DARK0_HARD,            Color.BRIGHT_YELLOW),
     'search_error'     : (Color.BRIGHT_RED,            Color.DARK0_HARD),
+    'active_narrow'    : (Color.DARK0_HARD,            Color.BRIGHT_GREEN),
     'task:success'     : (Color.DARK0_HARD,            Color.BRIGHT_GREEN),
     'task:error'       : (Color.DARK0_HARD,            Color.BRIGHT_RED),
     'task:warning'     : (Color.DARK0_HARD,            Color.NEUTRAL_PURPLE),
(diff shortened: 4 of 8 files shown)


zulipterminal/ui_tools/buttons.py
```diff
@@ -82,6 +82,10 @@ def update_widget(
         self.button_suffix.set_text(suffix)
         self._w.set_attr_map({None: text_color})
 
+    def set_highlight(self, style:str) -> None:
```
Rohitth007, line 85: This function isn't needed at all. Primarily because we don't want every child of TopButton to be highlightable. (Eg. EmojiButton)

zulipterminal/ui_tools/buttons.py
```diff
@@ -223,6 +227,12 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             self.model.controller.show_stream_info(self.stream_id)
         return super().keypress(size, key)
 
+    def set_status(self, is_active:bool) -> None:
+        if is_active:
+            self.set_highlight("active_narrow")
```
Rohitth007, line 232: What you can instead do is instead of this line use:
```python
        self._w.set_attr_map({None: "active_narrow"})
```

zulipterminal/ui_tools/buttons.py
```diff
@@ -223,6 +227,12 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+        else:
+            self.set_highlight(None)
+            self.update_count(self.count)
```
Rohitth007, line 235: Instead of these 2 use:
```python
        self._w.set_attr_map({None: None})
```

Rohitth007 (commented):
Left a review on your 2nd commit. I'm assuming you'll be moving your code in the 3rd commit to core.py

plugyawn:
Formatting changes.

zulipterminal/config/themes.py
```diff
@@ -71,6 +71,7 @@
     'area:error'      : 'standout',
     'area:user'       : 'standout',
     'search_error'    : 'standout',
+    'active_narrow'   : 'standout',
```
neiljp, line 74: This commit is fine now, but the commit title seems inaccurate now?

Note that Zulip (and ZT) aims to use a text style which essentially means saying what the commit does rather than what it did, so eg "Add styles..." here, which also saves a few characters here and there :)

zulipterminal/ui_tools/buttons.py
```diff
@@ -86,6 +87,10 @@ def activate(self, key: Any) -> None:
         self.controller.view.show_left_panel(visible=False)
         self.controller.view.show_right_panel(visible=False)
         self.controller.view.body.focus_col = 1
+        self.set_status(is_active=True)
+        if self.model.active_button:
+            self.model.active_button.set_status(is_active=False)
+        self.model.active_button = self
```
neiljp, line 93: My understanding is that this code is only triggering when activating a focused button via eg. space or enter (or clicking), not when using keys to otherwise change the narrow - when the highlights also need to change. So, while this logic works (with some issues right now?), it will also end up being a duplicate of the logic we really want.

zulipterminal/ui_tools/buttons.py
```diff
@@ -95,6 +100,12 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         else:  # This is in the else clause, to avoid multiple activation
             return super().keypress(size, key)
 
+    def set_status(self, is_active: bool) -> None:
+        if is_active:
+            self._w.set_attr_map({None: "active_narrow"})
+        else:
+            self._w.set_attr_map({None: None})
```
neiljp, line 107: This works, but one of the issues we discussed in the stream is how this interacts with other changes where the button is updated, due to line 84. As I think Rohitth mentioned, this means we likely need state in the button, so it knows the correct behavior to do there.

zulipterminal/model.py
```diff
@@ -284,6 +285,7 @@ def set_narrow(
         pm_with: Optional[str] = None,
         starred: bool = False,
         mentioned: bool = False,
+        active_button: Any = None,
```
neiljp, line 288: This parameter isn't being used right now?

zulipterminal/model.py
```diff
@@ -301,10 +303,34 @@ def set_narrow(
                 break
         else:
             raise RuntimeError("Model.set_narrow parameters used incorrectly.")
+        
+        try:
+            if self.stream_id:
+                self.active_button = self.controller.view.stream_id_to_button[
+                    self.stream_id
+                ]
```
neiljp, line 311: If the narrow changes to a non-stream one, this seems like we're setting the button to the old stream one?

zulipterminal/model.py
```diff
@@ -301,10 +303,34 @@ def set_narrow(
...
+        except AttributeError:
+            pass
 
         if new_narrow != self.narrow:
+            
+            try:
+                if self.stream_id and new_narrow:
+                    self.active_button.set_status(is_active=False)
+            except AttributeError:
+                pass
```
neiljp, line 324: Why is everything now wrapped in try/except blocks?

neiljp (commented):
@plugyawn I wanted to give a specific review, to back up some of the points I made in the stream.

This looks on the surface like it does more than before, but the code you added to activate doesn't solve the full problem (just for streams even).

I assume the last commit is in-progress right now - it's normally worth keeping those local unless you want specific feedback.

zulipterminal/model.py
plugyawn: Yep, sorry, removed this.

zulipterminal/config/themes.py
plugyawn: Oh, okay! I'll change it. What do you think will be a better commit title?

zulipterminal/ui_tools/buttons.py
plugyawn: Right now it's changed to work only for streams, so I think this is no longer a concern? I might be wrong.

zulipterminal/model.py
plugyawn: Perhaps the active_button should rather be last_active_stream?

zulipterminal/model.py
plugyawn: I was going to ask for why these errors were popping up, but until then I decided to wrap them in try/except. I'll have to see why it's happening.

zulipterminal/model.py
neiljp: Perhaps the larger point here is really that we're changing the button, before we know if the narrow changed.

zulipterminal/model.py
neiljp: I think this can overall be much simpler if you think about the logic in the following sections.

zulipterminal/ui_tools/buttons.py
```diff
@@ -80,12 +83,16 @@ def update_widget(
...
+        if self.is_active:
+            self._w.set_attr_map({None: "active_narrow"})
+        else:
+            self._w.set_attr_map({None: text_color})
 
     def activate(self, key: Any) -> None:
         self.controller.view.show_left_panel(visible=False)
         self.controller.view.show_right_panel(visible=False)
         self.controller.view.body.focus_col = 1
+        self.model.active_button = self
```
neiljp, line 95: This should still not be necessary.

zulipterminal/ui_tools/buttons.py
```diff
@@ -221,6 +238,10 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
             )
         elif is_command_key("STREAM_DESC", key):
             self.model.controller.show_stream_info(self.stream_id)
+        elif is_command_key("ENTER", key):
+            self.set_status(is_active=True)
+            if self.model.active_button:
+                self.model.active_button.set_status(is_active=False)
```
neiljp, line 244: This is still duplicate functionality, to that taking place in the model which should cover all stream narrowing behaviors.

neiljp (commented):
@plugyawn I'm glad to see the status in the updated version, but you still have duplicated code.

I think @Rohitth007 may have mentioned keeping things in the stream-button, but the core functionality as you have it in the topbutton looks reasonable to me right now. That said, you can trim the remaining activate changes there, and we aren't using the is_active parameter to `__init__`.

You should also be able to dramatically simplify the model commit based on the existing logic in the function you're modifying.

neiljp:
@plugyawn Also as I mentioned elsewhere, please do your best to keep the linters and tests passing. The linting is normally easily fixed with black or `make fix`.

Were the `AttributeError`s you were trying to fix coming from the tests?

zulipterminal/config/themes.py
plugyawn: It's just one style, right? ```active_narrow```?

plugyawn: Changed the commit message.

plugyawn:
Removed the redundant code.

plugyawn:
> @plugyawn Also as I mentioned elsewhere, please do your best to keep the linters and tests passing. The linting is normally easily fixed with black or `make fix`.
> 
> Were the `AttributeError`s you were trying to fix coming from the tests?

Did the linting just now, sorry!

Yep, the AttributeErrrors are coming from the tests, not sure why.

plugyawn:
> > @plugyawn Also as I mentioned elsewhere, please do your best to keep the linters and tests passing. The linting is normally easily fixed with black or `make fix`.
> > Were the `AttributeError`s you were trying to fix coming from the tests?
> 
> Did the linting just now, sorry!
> 
> Yep, the AttributeErrrors are coming from the tests, not sure why.

Do you think shifting code to core.py would resolve these errors?

neiljp:
Closing in favor of rebased/readjusted version of later version of this PR, ie. #1334.

PR #1248, opened 2022-08-20
[WIP] User list section/filters

### LINKED ISSUE #575: Filter/emphasize users who are in the present narrow
I feel this has been discussed in the past, but couldn't find a reference to an issue.

When in a narrow other than all-messages, we should limit the display of users to show those in the narrow, much like the webapp makes the display of out-of-narrow users.

We could take the approach as in the webapp, or simply filter the users.

For private streams, we definitely should just filter to users - this is happening in the webapp and can be tracked here https://github.com/zulip/zulip/pull/12846

srdeotarse:
**What does this PR do?**  
This WIP PR migrates earlier struture of ``RightColumnView`` in ``views.py`` to structure similar to ``LeftColumnView`` with toggling between ``StreamsView`` and ``TopicsView``.
This PR mainly intends to add new ``RecipientsView`` that displays recipients of the current narrow(``All_messages``, ``PMs``, ``Streams``, ``Topics`` etc) in the right column.

This fixes #575

CZO - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/User.20list.20sections.2Ffilters.20.23T575
**Tested?** 
- [ ] Manually

**Commit flow** 

**Notes & Questions** 

This WIP PR only includes transition from earlier structure of ``RightColumnView`` to new structure based on ``LeftColumnView`` structure. It includes modifications to ``UserView``
**Interactions** 

This PR is motivated by PR #459
**Visual changes**

### DIFF
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -13,6 +13,7 @@
 |Redraw screen|<kbd>ctrl</kbd> + <kbd>l</kbd>|
 |Quit|<kbd>ctrl</kbd> + <kbd>c</kbd>|
 |View user information (From Users list)|<kbd>i</kbd>|
+|Toggle Users/Recipients View (From Users list)|<kbd>t</kbd>|
 
 ## Navigation
 |Command|Key Combination|
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -311,6 +311,11 @@ KEY_BINDINGS: 'OrderedDict[str, KeyBinding]' = OrderedDict([
         'help_text': 'View user information (From Users list)',
         'key_category': 'general',
     }),
+    ('TOGGLE_RECIPIENTS_VIEW', {
+        'keys': ['t'],
+        'help_text': 'Toggle Users/Recipients View (From Users list)',
+        'key_category': 'general',
+    }),
     ('BEGINNING_OF_LINE', {
         'keys': ['ctrl a'],
         'help_text': 'Jump to the beginning of line',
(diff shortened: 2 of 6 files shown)


neiljp:
@srdeotarse We discussed this in the stream, so I'm marking it as awaiting update on the assumption that we've clarified how to move forward with this. Let me know if you need more feedback/review.

neiljp (commented):
@srdeotarse Thanks for looking at this! This is so large, and promises to be more commits, since it's really a big tidy/refactor first, followed by the feature itself :)

I seem to have written an essay of notes to help with the preliminary refactor ;) Whenever adjusting working code, it's useful to break it into almost too many independent working commits, as it tends to make each commit smaller and easier to read, and so take less energy and time to review. `git` can tell me if code has just been moved in a commit, but when it's interspersed with other changes then it's more difficult to validate.

While you could adapt tests for your giant code move in one go, it would likely be a lot of changes there too, which are also then difficult to associate with a given code change without breaking into smaller commits.

I almost suggested adding tests for the new model method, but like I said there, I'm not sure if you want to use it to provide the appropriate list/set of user ids only instead. That may be more obvious as you make the PR slightly more functional.

While you hopefully make sense of the refactoring splitting, for the functionality of the recipients part, there are various elements to address:
- toggling-back from recipients (we discussed); currently the list is only updated to the filtered view on toggling, so it would be useful to know it can be explored/tested manually by toggling back and forth in different narrows :)
- updating reliably from regular refreshes; this currently does an all-users view, since the user list assumes on updating, that the user list is the list to display. Maybe handle the filtering in the UsersView?
- updating on changing narrow; this is a similar problem to the 'highlight button representing active narrow' PR

PR #1249, opened 2022-08-21
Add handling for User update events

### LINKED ISSUE #988: Handle user update events & consequential UI changes
Currently we regularly get new presence information, but I believe the user information is fixed as per the start of the session.

We should implement event handling to update the user data, as well as UI changes which may be necessary promptly, ie which aren't refreshed from the updated model data on a regular basis or as needed, though we could consider updating more frequently for better representation.

mounilKshah:
**What does this PR do?**
The PR is for handling any changes made with respect to the users' meta data ie, meta data of the current user and that of other users.

Fixes #988 

**Tested?**
- [x] Manually

**Commit flow**
- Only one commit currently: Updates data structures in the codebase as per the events received.

**Notes & Questions**
- Yet to imbibe UI changes for the user events.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -237,6 +237,7 @@ class TestModel:
             "update_display_settings",
             "user_settings",
             "realm_emoji",
+            "realm_user",
         ]
         fetch_event_types = [
             "realm",
@@ -3345,6 +3346,48 @@ class TestModel:
         for stream_id in stream_ids:
             new_subscribers = model.stream_dict[stream_id]["subscribers"]
             assert new_subscribers == expected_subscribers
+
+    @pytest.mark.parametrize(
+        "person, changed_details",
+        [
+            (
+                {"full_name": "New Full Name"},
+                "full_name",
+            ),
+            ({"timezone": "New Timezone"}, "timezone"),
+            ({"is_billing_admin": False}, "is_billing_admin"),
+            ({"role": 10}, "role"),
+            (
+                {"avatar_url": "new_avatar_url", "avatar_version": 21},
+                "avatar_url",
+            ),
+            ({"new_email": "new_display@email.com"}, "email"),
+            (
+                {"delivery_email": "new_delivery@email.com"},
+                "delivery_email",
+            ),
+        ],
+        ids=[
+            "full_name",
+            "timezone",
+            "billing_admin_role",
+            "role",
+            "avatar",
+            "display_email",
+            "delivery_email",
+        ],
+    )
+    def test__handle_realm_user_event(
+        self, person, changed_details, model, initial_data
+    ):
+        # For testing purposes, initial_data["realm_users"][1] from the initial_data fixture is used here.
+        person["user_id"] = 11
+        event = {"type": "realm_user", "op": "update", "id": 1000, "person": person}
+        model._handle_realm_user_event(event)
+        assert (
+            initial_data["realm_users"][1][changed_details]
+            == event["person"][changed_details]
+        )
 
     @pytest.mark.parametrize("value", [True, False])
     def test__handle_user_settings_event(self, mocker, model, value):
--- a/zulipterminal/api_types.py
+++ b/zulipterminal/api_types.py
@@ -167,6 +167,37 @@ class ReactionEvent(TypedDict):
     message_id: int
 
 
+class RealmUserEventPerson(TypedDict, total=False):
+    user_id: int
+
+    full_name: str
+
+    avatar_url: str
+    avatar_source: str
+    avatar_url_medium: str
+    avatar_version: int
+
+    # NOTE: This field will be removed in future as it is redundant with the user_id
+    # email: str
+    timezone: str
+
+    role: int
+
+    email: str
+
+    new_email: str
+
+    delivery_email: str
+
+    is_billing_admin: bool
+
+
+class RealmUserEvent(TypedDict):
+    type: Literal["realm_user"]
+    op: Literal["update"]
+    person: RealmUserEventPerson
+
+
 class SubscriptionEvent(TypedDict):
     type: Literal["subscription"]
     op: str
@@ -249,4 +280,5 @@ Event = Union[
     UpdateRealmEmojiEvent,
     UpdateUserSettingsEvent,
     UpdateGlobalNotificationsEvent,
+    RealmUserEvent,
 ]
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -154,6 +154,7 @@ class Model:
                 ("update_display_settings", self._handle_update_display_settings_event),
                 ("user_settings", self._handle_user_settings_event),
                 ("realm_emoji", self._handle_update_emoji_event),
+                ("realm_user", self._handle_realm_user_event),
             ]
         )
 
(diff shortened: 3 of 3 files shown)


neiljp (commented):
@mounilKshah Thanks for looking at this - this looks functional, at least in some ways :+1:

While not strictly necessary, it'd be clearer to have the api_type changes in a separate commit since it documents the API first (you could even link to it in the commit text). I would also consider separate commits for first updating realm_users and then the active user for reasons below.

As per my inline comments I think you can simplify the update logic you already have, which may mean the additional function is unnecessary. It is called in every branch of the conditional, and the function itself seems like it could become 3 lines of code.

Currently your update of realm_users benefits from the regular calling of get_all_users and subsequently the user list update, and the former propagating that to other user datastructures that are either loosely (eg. by get_user_info()) or more tightly (by eg. user_dict) coupled. However, the logged-in user info updates are not currently updated from the initial_data by other code, so you'll need to add this manually, which I suspect won't be as simple as a dict update.

So, if you split this into 3 commits and address the inline comments, and add/adjust a test for the handler method, I think at least the first 2 commits could be quickly mergeable if you focus on those first. If the first two commits are clean and you re-push then I could re-review while you consider the third more. If the 3rd is also ready that would be fine too of course - just emphasizing that we could merge incrementally, which would help decrease turnaround time :)

zulipterminal/model.py
```diff
@@ -8,6 +8,7 @@
 from concurrent.futures import Future, ThreadPoolExecutor, wait
 from copy import deepcopy
 from datetime import datetime
+from pprint import pprint
```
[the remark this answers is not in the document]
mounilKshah: Yeah, it was to print the events containing realm user update. As this is still in draft stage, I thought of keeping it for reference for anybody testing this.

zulipterminal/model.py
```diff
@@ -1747,6 +1748,18 @@ def _handle_update_display_settings_event(self, event: Event) -> None:
                 view.message_view.log[msg_pos] = msg_w_list[0]
         self.controller.update_screen()
 
+    def _handle_realm_user_event_update(self, event: Event) -> None:
```
neiljp, line 1751: It was fine without the _update suffix here too - we want to maintain consistency with the similar functions, just as with the other event types in api_types.py.

neiljp (commented):
@mounilKshah This is nicely split and tidier, and I appreciate you pushing the changes, but as per inline notes the api types don't seem to match the docs, and I think there was confusion over how to handle what I meant by only 'update' ops.

The above and a test for the new function in the 2nd commit are necessary to get those merged ahead of the 3rd commti.

For the part related to the third commit, this changes initial_data but doesn't go further as per comments in my last review. Look for where the various 'active' user data that could get updated by these events is first extracted from initial_data, and think how we might go from there.

zulipterminal/api_types.py
```diff
@@ -167,6 +167,35 @@ class ReactionEvent(TypedDict):
     message_id: int
 
 
+class RealmUserEventPerson(TypedDict):
```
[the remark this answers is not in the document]
mounilKshah: `display_email` and `profile_data` are actually present in the commit. Did you mean to say `delivery_email` instead of `display_email`?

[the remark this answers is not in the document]
neiljp: I don't see `display_email` in the API, or `profile_data`, but they were in the commit.

zulipterminal/api_types.py
```diff
@@ -167,6 +167,37 @@ class ReactionEvent(TypedDict):
...
+    avatar_url_medium: str
+    avatar_version: int
+
+    # NOTE: This field will be removed in future as it is redundant with the user_id
+    # email: str
+    timezone: str
+
+    role: int
+
+    email: str
```
neiljp, line 186: In comparing the fields that can be updated, this one stands out as different from the remaining one field possible (custom_profile_field), unless I'm missing another use of email?

tests/model/test_model.py
```diff
@@ -3346,6 +3347,46 @@ def test__handle_subscription_event_subscribers_one_user_multiple_streams(
             new_subscribers = model.stream_dict[stream_id]["subscribers"]
             assert new_subscribers == expected_subscribers
 
+    @pytest.mark.parametrize(
+        "person, changed_details",
+        [
+            (
+                {"user_id": 11, "full_name": "New Full Name"},
```
neiljp, line 3354: The user_id is common in all cases, so let's specify and build it in the test function.

Rather than setting the variable in the test function, you can alternatively list it as a default argument value, as long as it's a non-mutable value.

tests/model/test_model.py
```diff
@@ -3346,6 +3347,46 @@ def test__handle_subscription_event_subscribers_one_user_multiple_streams(
...
+                {"user_id": 11, "avatar_url": "new_avatar_url", "avatar_version": 21},
+                "avatar_url",
+            ),
+            ({"user_id": 11, "new_email": "new_display@email.com"}, "email"),
+            (
+                {"user_id": 11, "delivery_email": "new_delivery@email.com"},
+                "delivery_email",
+            ),
+        ],
+        ids=[
```
neiljp, line 3370: There a lot of test cases & ids, so it's simpler to keep them inline - then later we wouldn't have to check where in each list to add/remove/edit them, or to move them around.

tests/model/test_model.py
```diff
@@ -3346,6 +3347,46 @@ def test__handle_subscription_event_subscribers_one_user_multiple_streams(
...
+            "delivery_email",
+        ],
+    )
+    def test__handle_realm_user_event(
+        self, person, changed_details, model, initial_data
+    ):
+        event = {"type": "realm_user", "op": "update", "id": 1000, "person": person}
+        model._handle_realm_user_event(event)
+        assert (
+            initial_data["realm_users"][1][changed_details]
```
neiljp, line 3386: I'm guessing the `1` is the index into initial_data that corresponds to `user_id=11`?

If that's the case, it would be more reliable to make that explicit in the test code, eg. if the initial data gets reordered at a later stage, this would make this test fail even though it's functionally correct?

neiljp (commented):
@mounilKshah Tests look good, left a few thoughts to make them clearer :+1:

API spec isn't critical, but unsure if there's something which has confused your aligning these values with those in the docs.

zulipterminal/api_types.py
mounilKshah: The `email` field does not exist under `RealmUserEventPerson`, rather we get `new_email` from the API. However, in `model.initial_data["realm_users"]`, we use `email` and not `new_email`. Hence, in the updated code, where we now have a conditional for replacing the `new_email` key with the `email` key, mypy returns an error.

tests/model/test_model.py
mounilKshah: Okay, I'll add a comment for this 👍

zulipterminal/api_types.py
neiljp: Ah, I see your point. Can we avoid that by moving the conditional into the loop where we've found the user-id?

neiljp: OK, let me just read the code again, I may just merge these first two commits roughly as they are.

neiljp:
@mounilKshah Thanks for the updates here! I just merged the first two commits of this manually with some minor adjustments to avoid the need to add the "email" field :tada:
(last commit was https://github.com/zulip/zulip-terminal/commit/01b3853c45260bcc5dc0174d6a028e8d9c03fc50)

I think we'll likely need to do quite a lot of refactoring before other updates of this form, so the user updates aren't as critical right now.

PR #1255, opened 2022-09-15
Add id to users with same name

### LINKED ISSUE #1151: Two users with the the same name appear as the same person
On zulip-term version 0.6.0, if two users have the same name and send two messages in sequence within the same topic, the messages will be displayed as if coming from the same user

Sushmey:
**What does this PR do?**  
On the Zulip terminal, users with same name weren't discernable because they were considered the same users. This PR compares users based on their user id and then displays the user id beside their user name if they have the same user name.

Fixes #1151 

**Tested?** 
- [x] Manually

**Commit flow** 
**first commit** -> comparing users based on user_id instead of name so that a separate header can be displayed

**second commit** -> displaying user_id beside all the users who have same name as more than one user in the list of all users. Adding a new style class so that the appended user_id doesn't look like a part of the username

**third commit** -> changing label from `names` to `msg_sender` in all colour themes

**Notes & Questions** 
This makes users with same name have different headers by appending the id to the author name and shows it as a visually separate addition.

**Interactions** 
- Waiting on PR review 

**Visual changes** 
Before:
Screenshot 2022-09-13 at 7 44 25 PM

After:
Screenshot 2022-10-21 at 11 48 24 AM

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -4263,6 +4263,21 @@ class TestModel:
         with pytest.raises(RuntimeError, match="Invalid user ID."):
             model.user_name_from_id(user_id)
 
+    @pytest.mark.parametrize(
+        "user_name, user_name_counter, is_duplicate",
+        [
+            case("Alice", {"Greg": 2}, False, id="username_does_not_exist"),
+            case("Alice", {"Alice": 1}, False, id="username_is_not_duplicate"),
+            case("Greg", {"Greg": 2}, True, id="username_is_duplicate"),
+        ],
+    )
+    def test_is_user_name_duplicate(
+        self, model, user_name, user_name_counter, is_duplicate
+    ):
+        model.user_name_counter = user_name_counter
+        return_value = model.is_user_name_duplicate(user_name)
+        assert return_value == is_duplicate
+
     def test_generate_all_emoji_data(
         self,
         mocker,
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -5,7 +5,7 @@ Defines the `Model`, fetching and storing data retrieved from the Zulip server
 import itertools
 import json
 import time
-from collections import defaultdict
+from collections import Counter, defaultdict
 from concurrent.futures import Future, ThreadPoolExecutor, wait
 from copy import deepcopy
 from datetime import datetime
@@ -174,6 +174,7 @@ class Model:
 
         self.user_dict: Dict[str, MinimalUserData] = {}
         self.user_id_email_dict: Dict[int, str] = {}
+        self.user_name_counter: Dict[str, int] = {}
         self._all_users_by_id: Dict[int, RealmUser] = {}
         self._cross_realm_bots_by_id: Dict[int, RealmUser] = {}
         self.users: List[MinimalUserData] = []
@@ -1171,6 +1172,7 @@ class Model:
         # and a user-id to email mapping
         self.user_dict = dict()
         self.user_id_email_dict = dict()
+        self.user_name_counter = Counter()
         for user in self.initial_data["realm_users"]:
             if self.user_id == user["user_id"]:
                 self._all_users_by_id[self.user_id] = user
@@ -1278,6 +1280,7 @@ class Model:
         # NOTE: Do this after generating user_list to avoid current_user duplication
         self.user_dict[current_user["email"]] = current_user
         self.user_id_email_dict[self.user_id] = current_user["email"]
+        self.user_name_counter = Counter(user["full_name"] for user in user_list)
 
         self.users = user_list
 
@@ -1291,6 +1294,12 @@ class Model:
             raise RuntimeError("Invalid user ID.")
 
         return self.user_dict[user_email]["full_name"]
+
+    def is_user_name_duplicate(self, user_name: str) -> bool:
+        """
+        Returns if the count of the users with the same name is more than 1.
+        """
+        return self.user_name_counter.get(user_name, -1) > 1
 
     def _subscribe_to_streams(self, subscriptions: List[Subscription]) -> None:
         def make_reduced_stream_data(stream: Subscription) -> StreamData:
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1079,7 +1079,7 @@ class TestMessageBox:
             ([STATUS_INACTIVE, "alice", " ", "DAYDATETIME"], {"timestamp": 0}),
         ],
         ids=[
-            "show_author_as_authors_different",
+            "show_authors_with_different_name_as_different",
             "merge_messages_as_only_slightly_earlier_message",
             "dont_merge_messages_as_much_earlier_message",
         ],
@@ -1115,12 +1115,67 @@ class TestMessageBox:
         all_to_vary = dict(to_vary_in_last_message, **stars["last"])
         last_msg = dict(message, **all_to_vary)
 
+        # Marking this as false so as to not fail
+        # under the condition of same username which needs user id appended to it
+        # since that is not what we are testing here.
+        self.model.is_user_name_duplicate = mocker.Mock(return_value=False)
+
         msg_box = MessageBox(this_msg, self.model, last_msg)
 
         expected_header[2] = output_date_time
         if current_year > 2018:
             expected_header[2] = "2018 - " + expected_header[2]
         expected_header[3] = "*" if starred_msg == "this" else " "
+
+        view_components = msg_box.main_view()
+
+        assert len(view_components) == 2
+        assert isinstance(view_components[0], Columns)
+        assert [w.text for w in view_components[0].widget_list] == expected_header
+        assert isinstance(view_components[1], Padding)
+
+    @pytest.mark.parametrize(
+        "message",
+        [
+            {
+                "id": 4,
+                "sender_id": 4,
+                "type": "stream",
+                "display_recipient": "Verona",
+                "stream_id": 5,
+                "subject": "Test topic",
+                "flags": [],
+                "is_me_message": False,
+                "content": "<p>what are you planning to do this week</p>",
+                "reactions": [],
+                "sender_full_name": "alice",
+                "timestamp": 1532103879,
+            }
+        ],
+    )
+    @pytest.mark.parametrize(
+        "expected_header, to_vary_in_last_message",
+        [
+            (
+                [" ", "alice (4)", " ", " "],
+                {"sender_id": 5},
+            ),
+        ],
+        ids=[
+            "show_authors_with_same_name_different_ids_as_different",
+        ],
+    )
+    def test_main_view_content_header_without_header_for_same_name_different_user_id(
+        self,
+        mocker,
+        message,
+        expected_header,
+        to_vary_in_last_message,
+    ):
+        this_msg = dict(message)
+        last_msg = dict(message, **to_vary_in_last_message)
+
+        msg_box = MessageBox(this_msg, self.model, last_msg)
 
         view_components = msg_box.main_view()
 
--- a/zulipterminal/ui_tools/messages.py
+++ b/zulipterminal/ui_tools/messages.py
@@ -650,6 +650,7 @@ class MessageBox(urwid.Pile):
                 "author": (
                     msg["sender_full_name"] if "sender_full_name" in msg else None
                 ),
+                "author_id": (msg["sender_id"] if "sender_id" in msg else None),
                 "time": (
                     self.model.formatted_local_time(
                         msg["timestamp"], show_seconds=False
@@ -668,6 +669,7 @@ class MessageBox(urwid.Pile):
         different = {  # How this message differs from the previous one
             "recipients": recipient_header is not None,
             "author": message["this"]["author"] != message["last"]["author"],
+            "author_id": message["this"]["author_id"] != message["last"]["author_id"],
             "24h": (
                 message["last"]["datetime"] is not None
                 and ((message["this"]["datetime"] - message["last"]["datetime"]).days)
@@ -686,7 +688,9 @@ class MessageBox(urwid.Pile):
             text_keys = ("author", "star", "time", "status")
             text: Dict[str, urwid_MarkupTuple] = {key: (None, " ") for key in text_keys}
 
-            if any(different[key] for key in ("recipients", "author", "24h")):
+            if any(
+                different[key] for key in ("recipients", "author", "author_id", "24h")
+            ):
                 text["author"] = ("msg_sender", message["this"]["author"])
 
                 # TODO: Refactor to use user ids for look up instead of emails.
--- a/zulipterminal/ui_tools/messages.py
+++ b/zulipterminal/ui_tools/messages.py
@@ -686,13 +686,23 @@ class MessageBox(urwid.Pile):
 
         if any_differences:  # Construct content_header, if needed
             text_keys = ("author", "star", "time", "status")
-            text: Dict[str, urwid_MarkupTuple] = {key: (None, " ") for key in text_keys}
+            text: Dict[str, List[urwid_MarkupTuple]] = {
+                key: [(None, " ")] for key in text_keys
+            }
 
             if any(
                 different[key] for key in ("recipients", "author", "author_id", "24h")
             ):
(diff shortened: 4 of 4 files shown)


zulipterminal/model.py
```diff
@@ -1144,6 +1147,14 @@ def user_name_from_id(self, user_id: int) -> str:
 
         return self.user_dict[user_email]["full_name"]
 
+    def user_name_count(self, user_name: str) -> int:
+        """
+        Returns the count of the users with the same name as the given name.
+        """
+        if not self.user_name_dict_count[user_name]:
+            return -1
+        return self.user_name_dict_count[user_name]
```
mounilKshah, line 1156: I think it would be better/safer if you used `.get()` to access the count, instead of using the index.

zulipterminal/ui_tools/boxes.py
```diff
@@ -76,7 +76,12 @@ def __init__(self, view: Any) -> None:
         super().__init__(self.main_view(True))
         self.model = view.model
         self.view = view
-
+        # global users_list  # Making it global so it can be accessed in the MessageBox class
+        # users_list = self.view.users
+        # global user_names_counter  # Making it global so it needs to be calculated only once
+        # user_names_counter = Counter(
+        #     user["full_name"] for user in users_list
+        # )  # Counting number of users having same name
```
mounilKshah, line 84: This commented code seems redundant. If you don't think it is necessary to the PR, then you may remove it.

mounilKshah (changes requested):
I ran your code manually and it worked fine for me. I tried going through the files to give an in-depth review, but it is difficult to review due to the current commit structure. It would be great if you could squash the commits using `git rebase`.
Also, upon running `pytest` on your PR, there were errors. All those can be looked into deeper once the commit structure is streamlined.

neiljp:
@Sushmey I just merged the last commit you had and merged that first with slight adjustments, since it does the independent rename of name->msg_sender.

Some things to note:
- this change was quite separate, so having it first in the sequence of commits would mean it'd have been closer to the original code branch (now main), so easier to partially merge first if it's a small preparation for others and fits in the same PR
- since it was the last commit, I adjusted the UI code to work without the changes - not much effort this time :)
- I updated the commit title & description:
  - in ZT we include the file names as an 'area' if they can be made to fit
  - generally it's good to use the present tense, so 'Rename' here vs 'Renaming'
  - this particular commit doesn't fix 1151 :)
  - it's useful to put the 'why' in the commit description

I'll take a look at the other commits.

zulipterminal/ui_tools/boxes.py
```diff
@@ -1566,12 +1565,19 @@ def main_view(self) -> List[Any]:
         any_differences = any(different.values())
 
         if any_differences:  # Construct content_header, if needed
-            TextType = Dict[str, urwid_MarkupTuple]
-            text_keys = ("author", "star", "time", "status")
-            text: TextType = {key: (None, " ") for key in text_keys}
-
-            if any(different[key] for key in ("recipients", "author", "24h")):
-                text["author"] = ("name", message["this"]["author"])
+            TextType = Dict[str, List[urwid_MarkupTuple]]
```
neiljp, line 1568: This change, with accompanying adjustments, is a perfectly fine separate refactor commit.

Generally more commits are better than fewer, since they're more difficult to separate later. However, note that these are ideally structured commits rather than commits that one does from just adding fixes to previous code in a branch.

tests/ui/test_ui_tools.py
```diff
@@ -2221,11 +2224,11 @@ def test_msg_generates_search_and_header_bar(
         "expected_header, to_vary_in_last_message",
         [
             (
-                [STATUS_INACTIVE, "alice", " ", "DAYDATETIME"],
+                [STATUS_INACTIVE, "alice(4)", " ", "DAYDATETIME"],
```
neiljp, line 2227: I think we'll probably want a space between the name and user id.

neiljp (commented):
@Sushmey This looks like it works well :+1:

I hope you can see after the refactor merge I did, that it's possible to do partial merges step by step of larger PRs. So we can do the same with chunks of this PR that do the first stage and then the second. As I mentioned some time ago in the stream, ideally they'd have tests for the new/changed functionality, if you feel comfortable with that. Also each stage (previously just two commits) can be split into smaller changes as separate commits, notably the second stage where you add the function to the model, and also you refactor the TextType, then you can add the changes which enact the feature after that.

While I'd normally have been glad to see the commit titles and description tidied (see previous comment), it was useful to see here where you'd presumably squashed commits, since some of that is fine as separate commits to explain the process - what I think of as the 'story' of the sequence of commits making up the PR.

Note that each commit should pass `make check` (running linting and pytest), not just the overall result, and `tools/check-branch` is useful for ensuring that applies for all commits in a branch.

zulipterminal/model.py
```diff
@@ -1130,7 +1131,9 @@ def get_all_users(self) -> List[Dict[str, Any]]:
         user_list.insert(0, current_user)
         self.user_dict[current_user["email"]] = current_user
         self.user_id_email_dict[self.user_id] = current_user["email"]
-
+        self.user_name_dict_count = Counter(
```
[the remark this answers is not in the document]
Sushmey: Okay, refactoring name to `user_name_counter` instead of `user_name_dict_count`

zulipterminal/model.py
```diff
@@ -1144,6 +1147,12 @@ def user_name_from_id(self, user_id: int) -> str:
 
         return self.user_dict[user_email]["full_name"]
 
+    def user_name_count_compare(self, user_name: str) -> int:
+        """
+        Returns if the count of the users with the same name is more than 1.
+        """
+        return True if self.user_name_dict_count.get(user_name, -1) > 1 else False
```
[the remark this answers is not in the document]
Sushmey: Sorry missed changing the return type to bool. No, mypy didn't note this. 
Refactoring the name to `is_user_name_duplicate`

Yes, I don't need inline condition, thank you!

zulipterminal/model.py
```diff
@@ -1018,6 +1018,7 @@ def get_all_users(self) -> List[Dict[str, Any]]:
         # and a user-id to email mapping
         self.user_dict: Dict[str, Dict[str, Any]] = dict()
         self.user_id_email_dict: Dict[int, str] = dict()
+        self.user_name_dict_count: Dict[str, int] = dict()
```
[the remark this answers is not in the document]
Sushmey: Yes, initialized it as `Counter()` instead of `dict()`

zulipterminal/ui_tools/boxes.py
```diff
@@ -1550,7 +1548,8 @@ def main_view(self) -> List[Any]:
         }
         different = {  # How this message differs from the previous one
             "recipients": recipient_header is not None,
-            "author": message["this"]["author_id"] != message["last"]["author_id"],
+            "author": message["this"]["author"] != message["last"]["author"],
+            "author_id": message["this"]["author_id"] != message["last"]["author_id"],
```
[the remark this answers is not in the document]
Sushmey: Nope, both are not needed. Will remove it in a separate commit since I might to change test cases since it's failing some assertion test

tests/ui/test_ui_tools.py
Sushmey: Okay, making that change in both test cases, and boxes.py

zulipterminal/ui_tools/boxes.py
Sushmey: So this wasn't purely a refactor since adding a new style class to the urwid Text widget needed me to change the TextType also so that it would pass all `make check` checks. 

But I understand what you meant by keeping commits structured, hopefully the newer ones I made are more suitable

neiljp (commented):
@Sushmey Thanks for your work on this! Aside from some minor suggestions, the test you added looks fine 👍 After restructuring, you may see other opportunities for where more/adjusted tests may make sense (see below).

Your intentions in the PR would be easier to read if you could group code into commits differently, ie. as per https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-discipline

One example, to prepare for the second change, would be to group the new method you introduced (`is_user_name_duplicate`), as well as the related model data & setup, and the test you have in a separate commit right now.

Similarly keep only the TextType changes in one commit.

tests/ui/test_ui_tools.py
```diff
@@ -2224,11 +2224,11 @@ def test_msg_generates_search_and_header_bar(
         "expected_header, to_vary_in_last_message",
         [
             (
-                [STATUS_INACTIVE, "alice(4)", " ", "DAYDATETIME"],
+                [STATUS_INACTIVE, "alice (4)", " ", "DAYDATETIME"],
                 {"sender_full_name": "bob"},
             ),
             ([" ", " ", " ", "DAYDATETIME"], {"timestamp": 1532103779}),
-            ([STATUS_INACTIVE, "alice(4)", " ", "DAYDATETIME"], {"timestamp": 0}),
+            ([STATUS_INACTIVE, "alice (4)", " ", "DAYDATETIME"], {"timestamp": 0}),
```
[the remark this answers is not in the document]
Sushmey: I had thought at that time that it made sense for them to be in a different commit with other little refactors but now I've squashed it with other previous commits

zulipterminal/model.py
```diff
@@ -1019,7 +1019,7 @@ def get_all_users(self) -> List[Dict[str, Any]]:
         # and a user-id to email mapping
         self.user_dict: Dict[str, Dict[str, Any]] = dict()
         self.user_id_email_dict: Dict[int, str] = dict()
-        self.user_name_dict_count: Dict[str, int] = dict()
+        self.user_name_counter: Dict[str, int] = Counter()
```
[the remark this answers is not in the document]
Sushmey: Yep, fair. Done!

tests/model/test_model.py
```diff
@@ -3774,6 +3774,21 @@ def test_user_name_from_id_invalid(self, model, user_id):
         with pytest.raises(RuntimeError, match="Invalid user ID."):
             model.user_name_from_id(user_id)
 
+    @pytest.mark.parametrize(
+        "user_name, user_name_counter, is_duplicate",
+        [
+            ("Alice", {"Greg": 2}, False),
+            ("Alice", {"Alice": 1}, False),
+            ("Greg", {"Greg": 2}, True),
```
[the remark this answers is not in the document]
Sushmey: I didn't know this. Thank you for pointing it out to me. I've added the cases with relevant messages.

mounilKshah (commented):
Thank you for your effort @Sushmey!
I tried the code and it worked pretty well for me. The commit structure looks much better than the last time I saw it. Only thing that I feel might need to be changed is the commit messages as they do not follow the [gitlint rules](https://github.com/mounilKshah/zulip-terminal#gitlint). Except for that, everything looks good to me.

tests/ui/test_ui_tools.py
```diff
@@ -2321,7 +2324,7 @@ def test_main_view_generates_EDITED_label(
     @pytest.mark.parametrize(
         "to_vary_in_last_message, update_required",
         [
-            ({"sender_full_name": "Unique name (won't be in next message)"}, True),
+            ({"sender_id": "Unique id (won't be in next message)"}, True),
```
neiljp, line 2327: Why is this updated? Does this behavior need updating too, as a follow-up?

tests/ui/test_ui_tools.py
```diff
@@ -2021,6 +2022,7 @@ def test_main_view_renders_slash_me(self, mocker, message, content, is_me_messag
         [
             {
                 "id": 4,
+                "sender_id": 4,
```
neiljp, line 2025: Nor here.

tests/ui/test_ui_tools.py
```diff
@@ -1983,6 +1983,7 @@ def test_main_view(self, mocker, message, last_message):
...
```
neiljp, line 1986: This does not appear to need this update?

I understand the desire to keep this message data up to date. We should likely be using something like a `message_fixture` from `conftest.py`. These tests were initially copy and pasted quite a lot, so a good refactor in a different commit/PR would be to simplify these by using our standard fixtures. One possible point to check is whether these are fixtures representing messages we receive from the server, and whether we update them later in the model (eg. modernize message response).

zulipterminal/ui_tools/boxes.py
```diff
@@ -1567,7 +1569,9 @@ def main_view(self) -> List[Any]:
             text_keys = ("author", "star", "time", "status")
             text: TextType = {key: (None, " ") for key in text_keys}
 
-            if any(different[key] for key in ("recipients", "author", "24h")):
+            if any(
+                different[key] for key in ("recipients", "author", "24h", "author_id")
+            ):
```
neiljp, line 1574: This is now new behavior, so while it passes existing tests, it would be good to extend or add a test to illustrate this, and include it in this commit.

neiljp (commented):
@Sushmey Thanks for your work on this - I'm trying to catch up on PRs!

This was an early PR, and unfortunately before I extracted the message box code into a separate file, so I've just migrated your changes across while rebasing. That should have been done correctly, but please check that the diffs in your local branch and the version I just pushed looks OK, even if they are in different files in some cases.

In doing the migration, I also:
- split the model changes out into a separate commit from the UI update
- updated the commit text a little

Some of my inline comments may not migrate cleanly since I'm pushing different files, so they may not appear in the commit-by-commit view, but will in the main conversation page.


The code definitely looks reasonable, but as per my inline comments, this would benefit from some work on the tests, including adding specific tests for the new behavior separately.

neiljp:
@Sushmey Ah, I'm also finding the tests are failing locally, so that may have been a migration issue on my part.

I also meant to say that the last commit doesn't have test changes and clearly changes behavior, which is one reason I'd like to see a new test for previous commit, so that the code change can match a test change.

tests/ui/test_ui_tools.py
```diff
@@ -2221,11 +2224,11 @@ def test_msg_generates_search_and_header_bar(
...
         [
             (
-                [STATUS_INACTIVE, "alice", " ", "DAYDATETIME"],
-                {"sender_full_name": "bob"},
+                [STATUS_INACTIVE, "alice (4)", " ", "DAYDATETIME"],
+                {"sender_id": "4"},
             ),
             ([" ", " ", " ", "DAYDATETIME"], {"timestamp": 1532103779}),
-            ([STATUS_INACTIVE, "alice", " ", "DAYDATETIME"], {"timestamp": 0}),
+            ([STATUS_INACTIVE, "alice (4)", " ", "DAYDATETIME"], {"timestamp": 0}),
```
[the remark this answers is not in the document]
Sushmey: So I tried changing it back to normal (i.e. name without user id) but it was still passing the duplicate name condition and appending `None`. At first I thought the test user_list has `alice` as a common name but changing it to a unique name also gave the same results. Then I realised the `this_msg` and `last_msg` in the test function are derived from the same parameters and they have the same sender name (although different `sender_id`) and that's why its always appending the `sender_id` to the name.

tests/ui/test_ui_tools.py
Sushmey: The function is merely checking update of author status so might not need that change. At that time I thought it'd need it since we were shifting from comparing based on `sender_full_name` to `sender_id`. Changing it back now

tests/ui/test_ui_tools.py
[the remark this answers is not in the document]
Sushmey: I looked into this a bit more and realised there's no way to include the duplicate username case in the earlier function where we test headers  without appending the userid to every name in the parameters because we are keeping the same message and varying just a few attributes. Also, added a mock response as `false` for the `is_user_name_duplicate` function so that it doesn't fail for the reasons mentioned earlier, and since duplicate usernames aren't what we are testing in that test anyway.

Finally, I decided to make a new test for testing the specific case where username is a duplicate where it would need userid appended to it.

Sushmey:
@neiljp Thank you so much for the review! 🤗  

I've made changes based on your comments. 

I added a separate test for duplicate usernames but different author ids to account for the new behaviour and updated a test case for the same.  I tried including it in the same test but it doesn't seem possible without having to change ever test case parameter. Also, added a mock response as false for the is_user_name_duplicate function so that it doesn't fail for the reasons mentioned earlier, and since duplicate usernames aren't what we are testing in that test anyway.

PR #1258, opened 2022-09-18
Add support for keywords in searchbar

### LINKED ISSUE #638: Expand search support beyond narrow+keywords
The full search feature available from a Zulip server is rather general, as listed here:
  https://zulipchat.com/help/search-for-messages

Exploratory searches using eg. `sender:me` shows that currently these do not just get passed through and work right now, and are just treated as keywords. However, if in a narrow, we *can* search for keywords within it - this corresponds to eg. `stream:design keyword`.

So, aspects we could explore are:
* Support special search phrases just typed in (note that some will conflict)
* Levels of search hiding/exploration (`#design narrow` <-> `stream:design`)
* Add a typeahead system for this, like the webapp

mounilKshah:
**What does this PR do?**
This PR is to add support for keywords in search bar.

Fixes #638 

Channel discussion:
[Expand search support #T638 #T1258](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Expand.20search.20support.20.23T638.20.23T1258)

**Tested?**
- [x] Manually

**Commit flow**
- One commit: Support only for stream and topic keywords

**Notes & Questions**
- This is a WIP PR. It currently only has support for stream and topic name keywords.
- As per the discussions, the plan ahead would be to:
  - Extend support to other keywords for narrows supported by ZT
  - Provide auto-populated text for the current narrow in the search bar
  - Write tests

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -472,15 +472,18 @@ class TestController:
         [
             ([], [["search", "FOO"]]),
             ([["search", "BOO"]], [["search", "FOO"]]),
-            ([["stream", "PTEST"]], [["stream", "PTEST"], ["search", "FOO"]]),
+            ([["stream", "Stream 1"]], [["stream", "Stream 1"], ["search", "FOO"]]),
             (
-                [["pm_with", "foo@zulip.com"], ["search", "BOO"]],
-                [["pm_with", "foo@zulip.com"], ["search", "FOO"]],
+                [["pm_with", "person1@example.com"], ["search", "BOO"]],
+                [["pm_with", "person1@example.com"], ["search", "FOO"]],
             ),
             (
-                [["stream", "PTEST"], ["topic", "RDS"]],
-                [["stream", "PTEST"], ["topic", "RDS"], ["search", "FOO"]],
+                [["stream", "Stream 1"], ["topic", "RDS"]],
+                [["stream", "Stream 1"], ["topic", "RDS"], ["search", "FOO"]],
             ),
+            ([["is", "starred"]], [["is", "starred"], ["search", "FOO"]]),
+            ([["is", "mentioned"]], [["is", "mentioned"], ["search", "FOO"]]),
+            ([["is", "private"]], [["is", "private"], ["search", "FOO"]]),
         ],
         ids=[
             "Default_all_msg_search",
@@ -488,6 +491,9 @@ class TestController:
             "search_within_stream",
             "pm_search_again",
             "search_within_topic_narrow",
+            "search_in_is_starred_narrow",
+            "search_in_is_mentioned_narrow",
+            "search_in_is_private_narrow",
         ],
     )
     @pytest.mark.parametrize("msg_ids", [({200, 300, 400}), (set()), ({100})])
@@ -498,12 +504,17 @@ class TestController:
         controller: Controller,
         mocker: MockerFixture,
         msg_ids: Set[int],
+        user_dict: Dict[str, Dict[str, Any]],
+        stream_dict: Dict[int, Dict[str, Any]],
         index_search_messages: Index,
     ) -> None:
         get_message = mocker.patch(MODEL + ".get_messages")
         create_msg = mocker.patch(MODULE + ".create_msg_box_list")
         mocker.patch(MODEL + ".get_message_ids_in_current_narrow", return_value=msg_ids)
         controller.model.index = index_search_messages  # Any initial search index
+        controller.model.user_id = 10
+        controller.model.user_dict = user_dict
+        controller.model.stream_dict = stream_dict
         controller.view.message_view = mocker.patch("urwid.ListBox")
         controller.model.narrow = initial_narrow
 
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -390,6 +390,48 @@ class TestModel:
         assert model.is_search_narrow() == is_search_narrow
 
     @pytest.mark.parametrize(
+        "search_text, expected_search_text, expected_search_narrow",
+        [
+            case("FOO", "FOO", [], id="all_messages"),
+            case(
+                "FOO stream:Stream+1",
+                "FOO",
+                [["stream", "Stream 1"]],
+                id="stream_without_spaces",
+            ),
+            case(
+                "FOO stream:Stream+1 topic:some+topic1%2Bsome+topic2",
+                "FOO",
+                [["stream", "Stream 1"], ["topic", "some topic1+some topic2"]],
+                id="stream_and_topic_without_spaces",
+            ),
+            case("FOO is:private", "FOO", [["is", "private"]], id="is_private"),
+            case("FOO is:mentioned", "FOO", [["is", "mentioned"]], id="is_mentioned"),
+            case("FOO is:starred", "FOO", [["is", "starred"]], id="is_starred"),
+            case(
+                "FOO pm_with:person1@example.com,person2@example.com",
+                "FOO",
+                [["pm_with", "person1@example.com, person2@example.com"]],
+            ),
+        ],
+    )
+    def test_extract_keywords_and_set_narrow(
+        self,
+        search_text: str,
+        model,
+        stream_dict,
+        user_dict,
+        expected_search_text: str,
+        expected_search_narrow: List[List[str]],
+    ) -> None:
+        model.stream_dict = stream_dict
+        model.narrow = []
+        model.user_dict = user_dict
+        result_query = model.extract_keywords_and_set_narrow(search_text)
+        assert model.narrow == expected_search_narrow
+        assert result_query == expected_search_text
+
+    @pytest.mark.parametrize(
         "bad_args",
         [
             dict(topic="some topic"),
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -370,6 +370,48 @@ class TestModel:
         assert model.index["pointer"][str(narrow)] == msg_id
 
     @pytest.mark.parametrize(
+        "narrow, expected_additional_text",
+        (
+            case([], "", id="all_messages"),
+            case([["stream", "PTEST"]], "stream:PTEST", id="stream_without_spaces"),
+            case(
+                [["stream", "PTEST1 PTEST2+PTEST3"]],
+                "stream:PTEST1+PTEST2%2BPTEST3",
+                id="stream_with_spaces_and_plus_symbol",
+            ),
+            case(
+                [["stream", "PTEST"], ["topic", "Test"]],
+                "stream:PTEST topic:Test",
+                id="stream_and_topic_no_spaces",
+            ),
+            case(
+                [["stream", "PTEST1 PTEST2+PTEST3"], ["topic", "Test1 Test2+Test3"]],
+                "stream:PTEST1+PTEST2%2BPTEST3 topic:Test1+Test2%2BTest3",
+                id="stream_and_topic_with_space_and_plus",
+            ),
+            case([["is", "private"]], "is:private", id="is_private"),
+            case([["is", "mentioned"]], "is:mentioned", id="is_mentioned"),
+            case([["is", "starred"]], "is:starred", id="is_starred"),
+            case(
+                [["pm_with", "some1@email.com, some2@email.com"]],
+                "pm_with:some1@email.com,some2@email.com",
+                id="group_pms",
+            ),
+        ),
+    )
+    def test_additional_search_text_for_current_narrow(
+        self,
+        model,
+        narrow: List[List[str]],
+        expected_additional_text: str,
+    ) -> None:
+        model.narrow = narrow
+        assert (
+            model.additional_search_text_for_current_narrow()
+            == expected_additional_text
+        )
+
+    @pytest.mark.parametrize(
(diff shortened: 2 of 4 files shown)


zulipterminal/core.py
```diff
@@ -492,6 +492,21 @@ def show_media_confirmation_popup(
...
+        # then the spaces in stream and topic names are converted to '+'
+        if self.model.narrow[0][0] == "stream":
+            search_query += f" {self.model.narrow[0][0]}:{self.model.narrow[0][1].replace('+', '%2B').replace(' ', '+')}"
+        if len(self.model.narrow) > 1:
+            if self.model.narrow[1][0] == "topic":
+                self.model.narrow[1][1].replace(" ", "+")
+                search_query += f" {self.model.narrow[1][0]}:{self.model.narrow[1][1].replace('+', '%2B').replace(' ', '+')}"
+        return search_query
+
     def check_for_prefixes(self, query: str) -> Tuple[str, List[List[str]]]:
```
neiljp, line 510: The name while WIP doesn't matter too much, but now we're firming up the logic, let's make this name more expressive. If I see `check_for_prefixes(something)` in code, it doesn't tell me what it does (I can think of various meanings), or why it returns what it does do - I can look it up in an editor or IDE, but better to read right there :)

So, maybe think what goes in the blank:
```python3
search_query, search_narrow = self.<BLANK>(final_search_query)
```

While some/many ZT functions don't have docstrings (or comments) since they're older, some are very simple and ideally well-named that it may be sufficient to understand what's happening when they're used without them.

You might also rename the other function, if you change the logic of what it does.

zulipterminal/core.py
```diff
@@ -501,22 +516,33 @@ def check_for_prefixes(self, query: str) -> Tuple[str, List[List[str]]]:
...
         final_search_text = ""
         new_narrow = []
-        # In the current version, the stream name and the topic name
-        # cannot have spaces within them
         for word in query.split(" "):
             if word.startswith("stream:"):
-                new_narrow.append(word.split(":"))
+                new_narrow.append(
+                    [
+                        word.split(":")[0],
```
neiljp, line 523: We're doing the same word.split operation a lot here, potentially repeatedly, and it likely will appear in other branches I suspect :)

neiljp (commented):
@mounilKshah This does work better :+1:

I've left feedback mainly on the second commit since it represents better the final state you have right now.

For tests on the current-narrow -> search-text function, the inputs for cases should be achievable from browsing how we handle non-search narrows.

I'd suggest you first split into commits to add each function, then use them together in core.py for 3 commits. Then some tests for each new function in the relevant commit. Covering streams/topics with tests would be good first. More narrow cases may flow quicker once you have the general code/test format :)

zulipterminal/core.py
```diff
@@ -492,6 +492,31 @@ def show_media_confirmation_popup(
             self, question, callback, location="center"
         )
 
+    def update_search_text_for_narrow(self, search_query: str) -> str:
```
neiljp, line 495: This looks more complete than the previous version :+1:

Note that all the internal code refers to the model data. This suggests the function belongs in the model.

Also note that we adapt the passed-in string as a return value. It's good to keep each function for one purpose, so leaving the string handling (eg. appending) for the query as something outside of this function.

tests/core/test_core.py
```diff
@@ -467,6 +467,50 @@ def test_stream_muting_confirmation_popup(
...
         pop_up.assert_called_once_with(controller, text(), partial())
 
+    @pytest.mark.parametrize(
+        "narrow, search_query, expected_search_text",
+        (
+            ([["stream", "PTEST"]], "FOO", "FOO stream:PTEST"),
+            (
+                [["stream", "PTEST PTEST2"]],
+                "FOO",
+                "FOO stream:PTEST+PTEST2",
```
neiljp, line 477: Using inline 'case' ids for the tests would make it clear what you're testing.

tests/core/test_core.py
```diff
@@ -467,6 +467,50 @@ def test_stream_muting_confirmation_popup(
...
+                "FOO",
+                "FOO is:private",
+            ),
+            (
+                [["is", "mentioned"]],
+                "FOO",
+                "FOO is:mentioned",
+            ),
+            (
+                [["pm_with", "some1@email.com", "some2@email.com"]],
```
neiljp, line 495: Is this the correct format? My memory is that each sublist is 2 elements long.

tests/core/test_core.py
```diff
@@ -467,6 +467,50 @@ def test_stream_muting_confirmation_popup(
...
+                "FOO",
+                "FOO pm_with:some1@email.com,some2@email.com",
+            ),
+        ),
+    )
+    def test_update_search_text_for_narrow(
```
neiljp, line 501: Other narrows: all messages, starred.

tests/core/test_core.py
```diff
@@ -467,6 +467,34 @@ def test_stream_muting_confirmation_popup(
         )
         pop_up.assert_called_once_with(controller, text(), partial())
 
+    @pytest.mark.parametrize(
+        "search_text, expected_search_text, expected_search_narrow",
+        [
+            ("FOO", "FOO ", []),
```
neiljp, line 517: I wouldn't expect to see spaces around the output search text.

tests/core/test_core.py
```diff
@@ -467,6 +467,34 @@ def test_stream_muting_confirmation_popup(
...
+            ("FOO stream:PTEST", "FOO ", [["stream", "PTEST"]]),
+            (
+                "FOO stream:PTEST topic:TEST",
+                "FOO ",
+                [["stream", "PTEST"], ["topic", "TEST"]],
+            ),
+            ("FOO is:private", "FOO ", [["is", "private"]]),
+            ("FOO is:mentioned", "FOO ", [["is", "mentioned"]]),
+            ("FOO is:starred", "FOO ", [["is", "starred"]]),
```
neiljp, line 526: It'd be great to see some 'space-filled' examples here, like with the other function tests.

zulipterminal/core.py
```diff
@@ -492,6 +492,37 @@ def show_media_confirmation_popup(
...
+        new_narrow = []
+        for word in query.split(" "):
+            if word.startswith("stream:"):
+                new_narrow = [
+                    ["stream", word.split(":")[1].replace("+", " ").replace("%2B", "+")]
+                ]
+                stream_prefix_present_flag = True
+            elif word.startswith("topic:") and stream_prefix_present_flag:
+                new_narrow.append(
+                    ["topic", word.split(":")[1].replace("+", " ").replace("%2B", "+")]
```
neiljp, line 539: It would be good to see various of these replacement operations tested - other than testing the code, it also demonstrates the input->output one might expect, on reading the tests.

neiljp (commented):
@mounilKshah This looks more readable and complete, including with the tests :+1:

Once you finish the code and tests for the few existing narrows, along with the final (refactor) commit to replace the existing functions with the new functions, this will be mergeable. The functionality may not seem changed, but this refactor will enable supporting more search options in future :)

Note for tests, if you note the changes to tests (separate line) in the git log, there's no need to include it in the commit title.

tests/core/test_core.py
mounilKshah: Usually it is, but in case of `pm_with`, I have always observed the length of the sublist to be 1 + no. of recipients.

neiljp: See https://github.com/zulip/zulip-terminal/blob/main/zulipterminal/model.py#L288, for example.

mounilKshah: Yes, I got it now. I actually didn't realise that all the users under `pm_with` are mentioned under the same string.

zulipterminal/core.py
```diff
@@ -531,8 +531,14 @@ def extract_keywords_from_search_query(
...
         self.model.index["search"].clear()
-        self.model.set_search_narrow(text)
-
+        modified_search_text = (
+            text + self.model.addtional_search_text_for_current_narrow()
+        )
+        text_without_keyword, search_narrow = self.extract_keywords_from_search_query(
+            modified_search_text
+        )
+        self.model.narrow = search_narrow
```
neiljp, line 540: My primary remaining concern, given we're potentially taking a pause on this after your work, is this direct setting of the narrow here. Doing so means we potentially don't update other data which would otherwise be set, via eg. model.set_narrow or _narrow_to or something which also updated the UI.

Right now the positive is that the narrow *shouldn't* change, unless a user enters extra search text - and not all combinations work. For now I'd suggest we alert the user with a simple warning popup (popup_with_message), if the search looks like a complicated search which would require setting the narrow, ie. if the old and new narrows wouldn't match.

Examples of cases where this can happen is being in all messages, and adding eg. `is:starred some search text`, in which case the narrow changes. That 'works', but then adding `stream:general` or similar does not. So it'd be good to be consistent for now.

zulipterminal/model.py
```diff
@@ -278,6 +278,32 @@ def get_focus_in_current_narrow(self) -> Union[int, Set[None]]:
     def set_focus_in_current_narrow(self, focus_message: int) -> None:
         self.index["pointer"][repr(self.narrow)] = focus_message
 
+    def addtional_search_text_for_current_narrow(self) -> str:
+        """
+        This function is to check if the user is currently inside a narrow or not
+        and if they are, then the additional_search text is created accordingly
```
neiljp, line 284: Can this be simplified/updated?

zulipterminal/model.py
```diff
@@ -278,6 +278,32 @@ def get_focus_in_current_narrow(self) -> Union[int, Set[None]]:
...
+        """
+        # In cases where '+' is already present in stream and/or topic name, it is converted to '%2B'
+        # then the spaces in stream and topic names are converted to '+'
+        additional_search_query_text = ""
```
neiljp, line 288: Minor: The logic here is fairly simple and sets just one primary name, so a much shorter variable name should be fine, within reason - and make it easier to read :) The text gets returned as the name of the function, so the function name can explain what the 'additional_search_query_text` means on it's own.

tests/model/test_model.py
```diff
@@ -369,6 +369,47 @@ def test_set_focus_in_current_narrow(self, mocker, model, narrow, msg_id):
         model.set_focus_in_current_narrow(msg_id)
         assert model.index["pointer"][str(narrow)] == msg_id
 
+    @pytest.mark.parametrize(
+        "narrow, expected_additional_text",
+        (
+            case([], "", id="all_messages"),
+            case([["stream", "PTEST"]], " stream:PTEST", id="stream_without_spaces"),
```
neiljp, line 376: As before, I'd expect the output text to not have any surrounding spaces.

neiljp (commented):
@mounilKshah The extra commit means I can test this, and it appears to work fairly well from manual testing - and the test cases are reassuring too :+1:

I left some final comments. This is essentially mergeable with some textual adjustments, it's only my concern re unexpectedly safely/unsafely changing the narrow or not, due to this likely being the end of the Summer work on this.

zulipterminal/model.py
```diff
@@ -285,6 +285,38 @@ def is_search_narrow(self) -> bool:
         """
         return "search" in [subnarrow[0] for subnarrow in self.narrow]
 
+    def extract_keywords_from_search_query(self, query: str) -> str:
```
neiljp, line 313: This function no longer just extracts keywords, it also sets the base narrow.

zulipterminal/model.py
```diff
@@ -278,6 +278,31 @@ def get_focus_in_current_narrow(self) -> Union[int, Set[None]]:
     def set_focus_in_current_narrow(self, focus_message: int) -> None:
         self.index["pointer"][repr(self.narrow)] = focus_message
 
+    def addtional_search_text_for_current_narrow(self) -> str:
```
neiljp, line 281: Spelling?

tests/core/test_core.py
```diff
@@ -472,22 +472,28 @@ def test_stream_muting_confirmation_popup(
...
-                [["pm_with", "foo@zulip.com"], ["search", "BOO"]],
-                [["pm_with", "foo@zulip.com"], ["search", "FOO"]],
+                [["pm_with", "person1@example.com"], ["search", "BOO"]],
+                [["pm_with", "person1@example.com"], ["search", "FOO"]],
             ),
             (
-                [["stream", "PTEST"], ["topic", "RDS"]],
-                [["stream", "PTEST"], ["topic", "RDS"], ["search", "FOO"]],
+                [["stream", "Stream 1"], ["topic", "RDS"]],
+                [["stream", "Stream 1"], ["topic", "RDS"], ["search", "FOO"]],
```
neiljp, line 482: Are these changes just tidying, or required? If tidying, better to keep in a separate commit.

neiljp (commented):
@mounilKshah I'd apparently started a review for this but not finished, so just did so. Looks close to a safe refactor.

I thought we might be able to progress further with just search-specific narrow details. We may be able to and you're welcome to investigate, but we may need to refactor the narrow handling more first. Right now that's spread over many locations, possibly with duplicate logic.

tests/core/test_core.py
mounilKshah: They are required changes. The `final_narrow` is created only when the narrow exists in the org, else the test case fails.

PR #1265, opened 2022-10-12
refactor: views: Moved message links above topic links.

### LINKED ISSUE #1238: Move message links above topic links in message information popup

Sushmey:
Fixes #1238

**What does this PR do?**  
refactored code such as topic_links are now below message_links so that when displayed in the information popup, message link is shown above topic link. 

**Tested?** 
- [x] Manually

**Commit flow** 
Single commit- Moved message links above topic links in message information popup

**Notes & Questions** 

**Interactions** 

**Visual changes** 
Before:
Screenshot 2022-10-12 at 6 21 17 PM

After: 
Screenshot 2022-10-12 at 6 07 19 PM

### DIFF
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -884,6 +884,32 @@ class TestMsgInfoView:
         self.msg_info_view.keypress(size, key)
         assert not self.controller.exit_popup.called
 
+    @pytest.mark.parametrize(
+        "topic_links, message_links",
+        [
+            (
+                OrderedDict([("image", ("image", 1, True))]),
+                OrderedDict([("https://bar.com", ("topic", 1, True))]),
+            ),
+        ],
+    )
+    def test_pop_up_info_order(
+        self,
+        message_fixture: Message,
+        topic_links: "OrderedDict[str, Tuple[str, int, bool]]",
+        message_links: "OrderedDict[str, Tuple[str, int, bool]]",
+    ) -> None:
+        msg_info_view = MsgInfoView(
+            self.controller,
+            message_fixture,
+            title="Message Information",
+            topic_links=topic_links,
+            message_links=message_links,
+            time_mentions=list(),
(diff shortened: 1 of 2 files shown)


SRV332003:
Hi @Sushmey i want to work on this issue, can you please assign it to me?
Also tell a bit what's left to be done in this issue?

Sushmey:
Hey @SRV332003 I'm still working on this issue. 
Feel free to look for other issues, there are plenty other you could work on :)

neiljp:
@Sushmey I expect this should be OK, though we don't tend to mention test files in the commit titles in general. I would like to see how much of this is different to the other author, since while small it is good to share credit if that's the case :)

neiljp:
@Sushmey Thanks for pushing this through and simplifying the previous iteration :+1:

I just split this back up into the 2-commit style I requested in #1241, ie. add a test for the current condition (since we didn't have one), then adjust the test and the code. I've adjusted both commits a little (particularly commit text), and left the second attributed to the original author in #1241, since you primarily improved the test readability which is now in the first commit (leaving the previous author as a co-author).

This is now merged in the two commits ending 77218c1a646b431a463d3ab84ff61d92ceaf5c76 :tada:
(check the parent commit to see the previous one, or update your main branch)

Issue #873, opened 2021-01-13
Add FAQ regarding the meta key and alternatives in README


Abhirup-99:
There is no specific key called meta on the keyboard and it would be better if we add additional info about the meta and alt key relation in the README.
Relevant discussion:
 https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Alt.20.2F.20Meta.20keys

avats-dev:
This chat reference is giving 404. @Abhirup-99 can you guide me what needs to be done here? Thanks.

Abhirup-99:
Hi @avats-dev, thank you for expressing interest in the issue.
There is no "meta" key on the keyboard, but `alt` serves as the meta key. We would like so documentation entry in the faqs section related to this. Research what's the meta key on mac/windows and make a small entry in the faqs section.

zulipbot:
Hello @zulip/server-hotkeys members, this issue was labeled with the "area: hotkeys" label, so you may want to check it out!

Issue #680, opened 2020-06-08
make it possible to rebind keys


alfonsrv:
In Terminal on macOS `Ctrl + d` closes processes, while everything including the `Meta` key (aka `Alt`) is translated to a special character instead, hence making the `zulip-terminal` client a read-only client on macOS if the native Terminal configuration is not altered.  

This could also help prevent `tmux` collisions by giving users the freedom to choose their key-bindings freely.

neiljp:
This is actually possible by editing the `keys.py` file in the source distribution, though granted this is not straightforward or sustainable with upgrades. We have discussed the issue in the context of an appropriate key for sending messages previously, including in #294 . Do you have any thoughts given that discussion?

alfonsrv:
Editing the `keys.py` file indeed does not seem sustainable over the long run. If you ask for my opinion, I think a seperate `[keybindings]` section within the config would be desirable.  

We could either work with a `prefix` – such as tmux, e.g. `C-z` (`^Z`) (is not used by default for any application I know of) – and then `s` for send. I think that behavior while indeed having a steeper learning curve would be desirable as it would give way more flexibility instead of having to battle with previously existing shortcuts. Or we just work with static keyboard bindings, however I think the collisions are too likely given what terminal power users such as the one using `zulip-terminal` use on a regular basis.  

As for a default shortcut for sending messages, I think `C-Space` (`^@`) could work?

neiljp:
I agree that making keybindings configurable outside of the source is the long term solution, and we could reserve this issue for that.

Given the ongoing debate over #294, setting sending keys in particular seems like the priority for now, whether as a configuration option or resolving the problem in another way. I recommend we discuss this aspect in that issue, and also in #**zulip-terminal**. There are past topics in that stream on chat.zulip.org such as "Key for sending msg" and "Contributing & Send Message shortcut". Looking back in the history I'm wondering if we want to re-try 

Re MacOS I believe other known ZT users use alternative emulators due to issues with eg. color support, as per #662, which could be a workaround to the keyboard issue for you.

sumanthvrao:
@alfonsrv A hack which worked for me with Mac OS was to toggle this setting under `Terminal > preferences > Profiles > Keys` from `Meta` to `ESC+`

Note: I use iTerm2 terminal emulator and not the default one.

![image](https://user-images.githubusercontent.com/30312566/84096690-2c78a180-aa20-11ea-95f5-093a705cb53d.png)

[...]

PR #1267, opened 2022-10-19
Change MOUSE_SELECTION_KEY  to Fn + cmd

Khushiyant:
**What does this PR do?**  

_MOUSE_SELECTION_KEY_ is changed to _**Fn + cmd**_ in MacOS platform

**Testing** 
- [x] Manually

Commit changes
```diff
- MOUSE_SELECTION_KEY = "Fn + Alt" if PLATFORM == "MacOS" else "Shift"
+ MOUSE_SELECTION_KEY = "Fn + cmd" if PLATFORM == "MacOS" else "Shift"
```

**Visual changes** 

Before
Screenshot 2022-10-19 at 1 28 09 PM

After
Screenshot 2022-10-19 at 1 29 29 PM
> Changed to Fn+cmd is used to avoid confusion due to Mac keyboard layout

### DIFF
--- a/zulipterminal/platform_code.py
+++ b/zulipterminal/platform_code.py
@@ -25,7 +25,7 @@ else:
 
 
 # PLATFORM DEPENDENT HELPERS
-MOUSE_SELECTION_KEY = "Fn + Alt" if PLATFORM == "MacOS" else "Shift"
+MOUSE_SELECTION_KEY = "Fn + cmd" if PLATFORM == "MacOS" else "Shift"
 
 
 def notify(title: str, text: str) -> str:


Khushiyant:
@neiljp Please check

neiljp:
I don't use a mac so this is difficult for me to confirm personally.

Which terminal emulators have you tested this on?

My understanding is that depending on the terminal emulator, eg. Terminal (built-in) or iTerm (app that some use), Alt/cmd may work differently, see for example https://github.com/zulip/zulip-terminal/issues/680#issuecomment-640977268

Are you about on chat.zulip.org?

---

If this does resolve this one particular case across terminals, more generally I'd like to combine this fix with a translation of the `Meta` which is necessary for urwid to understand, into something like `physical_key_representing_meta()` for user output (in platform_code.py), since `Meta` is not an available physical key in most cases.

Note that resolving the Mac (and cross-platform) situation properly would involve documenting the situation via #873, as well as ensuring we address points that have come up in topics that have come up in
- https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Alt.20.2F.20Meta.20keys
- https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/About.20Popup.20.28alt.2Fmeta.29

Khushiyant:
@neiljp
* I have used the default ```zsh``` terminal for testing
* I feel that key mapping for above problem is correct but instruction text is displayed wrong according to Mac keyboard layout since there is no ```Alt``` in it.

<hr>

In case of documentation, I'll change the things accordingly

neiljp:
@Khushiyant I was referring to the application you're using to run `zsh` (which itself is a shell), so that doesn't quite answer my question.

Khushiyant:
@neiljp I have tested the change on Terminal(built-in, ver 2.12.7) and also on iTerm

PR #1274, opened 2022-12-06
model: checking 'code' field instead of 'msg' field

### LINKED ISSUE #1272: Fix incorrect check for bad event queue
Currently, zulip-terminal is checking for the string "Bad event queue id:" to check whether an error is due to a bad event queue ID. This is not the correct algorithm; as documented on https://zulip.com/api/register-queue, it should be checking the `code` on the response object, not the `msg` field. 

In https://github.com/zulip/zulip/pull/23711, we will be changing that string, though the logic would have also failed for users with a non-English language configured (as then the error message would come back in that language).

```
    @asynch                                                                                                                              
    def poll_for_events(self) -> None:                                                                                                   
        reregister_timeout = 10                                                                                                          
        queue_id = self.queue_id                                                                                                         
        last_event_id = self.last_event_id                                                                                               
        while True:                                                                                                                      
            if queue_id is None:                                                                                                         
                while True:                                                                                                              
                    if not self._register_desired_events():                                                                              
                        queue_id = self.queue_id                                                                                         
                        last_event_id = self.last_event_id                                                                               
                        break                                                                                                            
                    time.sleep(reregister_timeout)                                                                                       
                                                                                                                                         
            response = self.client.get_events(                                                                                           
                queue_id=queue_id, last_event_id=last_event_id                                                                           
            )                                                                                                                            
                                                                                                                                         
            if "error" in response["result"]:                                                                                            
                if response["msg"].startswith("Bad event queue id:"):                                                                    
                    # Our event queue went away, probably because                                                                        
                    # we were asleep or the server restarted                                                                             
                    # abnormally.  We may have missed some                                                                               
                    # events while the network was down or                                                                               
                    # something, but there's not really anything                                                                         
                    # we can do about it other than resuming                                                                             
                    # getting new ones.                                                                                                  
                    #                                                                                                                    
                    # Reset queue_id to register a new event queue.                                                                      
                    queue_id = None                                                                                                      
                time.sleep(1)                                                                                                            
                continue                                                                                                                 
```

Sushmey:
**What does this PR do?**  
Currently, zulip-terminal is checking for the string "Bad event queue id:" to check whether an error is due to a bad event queue ID. This is not the correct algorithm; as documented on https://zulip.com/api/register-queue, it should be checking the code on the response object, not the msg field.

Fixes #1272

https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/events.20error.20handling

**Tested?** 
- [ ] Manually

**Commit flow** 
- First commit, changed `msg` field to `code` field in response object and changed matching text accordingly

**Notes & Questions** 

- Not sure how I'd test the working of this manually. 
**Interactions** 

**Visual changes**

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -1903,7 +1903,7 @@ class Model:
             )
 
             if "error" in response["result"]:
-                if response["msg"].startswith("Bad event queue id:"):
+                if response.get("code") == "BAD_EVENT_QUEUE_ID":
                     # Our event queue went away, probably because
                     # we were asleep or the server restarted
                     # abnormally.  We may have missed some


neiljp:
@Sushmey Thanks for looking at this - good to come back to this, and I'm merging it now!

I made some changes to the commit title and text to meet the ZT style guide, and include the 'why' of the change in the commit body. I'd suggest looking at the guides we have, and setting up gitlint.

PR #1275, opened 2022-12-09
Maintain position in topic list when toggling topic list (issue #617)

Subhasish-Behera:
With the current changes, this PR solves the issue partially.
What it solves?
**->** it solves the maintaining of focus of the topic list of a stream( say _stream 1_) from where the `t` button was pressed.
But if the user presses `t` for some stream button (_stream 2_) and once again presses `t` for _stream 1_ then the focus of the topic list resets to the top.
**->** In other words, pressing 't' multiple times on the same stream button would maintain the focus.
**->** This PR is based on the idea that just like a topic is always
coupled with a stream, a stream can also be coupled with a particular
a topic that was last focussed.
**The code changes done are**
**views.py**
_TopicsView:_
-> Introducing a parameter `curr_topic_name` to `TopicsView` that stores the topic name of the last focussed topic in a stream. (This would make the implementation work even if the topic list is rearranged as it depends on topic_name not the pos on the list)
_LeftColumnView:_
->adding parameters `topic_name` and  `stream_id` to def `stream_view` which is used to create `stream_btn_list` according to the changes made in the `StreamButton` class in buttons.py.
**buttons.py**
_StreamButton_
->Adding a parameter called `curr_topic_name` to class `StreamButton`
->def `keypress` of `StreamButton`:  passes the curr_topic_name to TopicView so that, curr_topic_name can be focussed while rendering that stream's TopicView
_TopicButton_
->def `keypress` of `TopicButton`: passes the topic_name and stream_id of the Topicbutton from where `t` was pressed to show_stream_view.
Further suggestions would be required to make changes outside of UI files to store the state(which could solve the remaining part of the issue).

### DIFF
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -176,7 +176,7 @@ class StreamButton(TopButton):
         controller: Any,
         view: Any,
         count: int,
-        curr_topic_name=None,
+        curr_topic_name: Any =None,
     ) -> None:
         # FIXME Is having self.stream_id the best way to do this?
         # (self.stream_id is used elsewhere)
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -21,14 +21,14 @@ from zulipterminal.urwid_types import urwid_MarkupTuple, urwid_Size
 
 class TopButton(urwid.Button):
     def __init__(
-        self,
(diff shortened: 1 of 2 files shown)


Subhasish-Behera:
This PR is opened to discuss #617

Subhasish-Behera:
@neiljp Can you please give some feedback on the changes I have made?

PR #1276, opened 2022-12-12
ci: Provide initial Python 3.11 testing support.

klardotsh:
Initial local testing with Pipenv on x86_64 glibc seems to indicate the application runs on Python 3.11, so let's enable CI to (start to) prove it.

Follow-up from [CZO thread](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Python.203.2E11.20support/near/1478433)


neiljp:
@klardotsh As per the discussion on czo, pushed this simplified version, now merging before #1278.

PR #1278, opened 2022-12-29
lint-and-test: Use latest default CPython on macos.

neiljp:
**What does this PR do?**  

Explore change to do tests on the macos platform against the latest available CPython.

This follows the discussion on adding support for Python 3.11 here:
https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Python.203.2E11.20support/near/1478470

**Tested?** 
- [x] Manually
(CI change)

**Notes & Questions** 
Depends on #1276 since the latest version on macos is automatically installed properly and is actually 3.11; the last commit here can be removed after that is merged.


klardotsh (approved):

neiljp:
Removed the extra commit, and after confirmed it passes CI I am now merging.

PR #1280, opened 2023-01-04
[WIP] Disable cycling while editing a PM

Subhasish-Behera:
**What does this PR do?**  
-> Disables cycling during editing a PM by `tab` key .
->  Shows a warning if one uses tab during editing a PM.

Partial fix for #774 

**Tested?** 
- [X] Manually

**Commit flow** 

commit 1: changes made to boxes.py
commit 2: changes made to test_boxes.py
**Notes & Questions** 

**Interactions** 

**Visual changes** 
zt1

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -1706,6 +1706,39 @@ class TestWriteBox:
                 expected_focus_col_name
             )
 
+    def test_private_box_view(
+        self,
+        mocker: MockerFixture,
+        user_dict: List[Dict[str, Any]],
+        user_id_email_dict: Dict[int, str],
+    ) -> None:
+        recipient_user_ids = [11]
+
+        write_box = WriteBox(self.view)
+        write_box.to_write_box = mocker.MagicMock(spec=ReadlineEdit)
+        enable_autocomplete = mocker.patch.object(ReadlineEdit, "enable_autocomplete")
+        write_box.model.user_id_email_dict = user_id_email_dict
+        write_box.model.user_dict = user_dict
+        connect_signal_mock = mocker.patch.object(urwid, "connect_signal")
+        write_box.private_box_view(recipient_user_ids=recipient_user_ids)
+        enable_autocomplete.assert_has_calls(
+            [
+                mock.call(
+                    func=write_box._to_box_autocomplete,
+                    key=primary_key_for_command("AUTOCOMPLETE"),
+                    key_reverse=primary_key_for_command("AUTOCOMPLETE_REVERSE"),
+                ),
+                mock.call(
+                    func=write_box.generic_autocomplete,
+                    key=primary_key_for_command("AUTOCOMPLETE"),
+                    key_reverse=primary_key_for_command("AUTOCOMPLETE_REVERSE"),
+                ),
+            ]
+        )
+        connect_signal_mock.assert_called_once()
+        assert isinstance(write_box.to_write_box, ReadlineEdit)
+        assert write_box.to_write_box.text == "To: Human 1 <person1@example.com>"
+
     def test__setup_common_private_compose(self, mocker: MockerFixture) -> None:
         write_box = WriteBox(self.view)
         write_box.to_write_box = mocker.MagicMock()
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -1739,6 +1739,34 @@ class TestWriteBox:
         assert isinstance(write_box.to_write_box, ReadlineEdit)
         assert write_box.to_write_box.text == "To: Human 1 <person1@example.com>"
 
+    def test_private_box_edit_view(
+        self,
+        mocker: MockerFixture,
+        user_dict: List[Dict[str, Any]],
+        user_id_email_dict: Dict[int, str],
+    ) -> None:
+        recipient_user_ids = [11]
+        write_box = WriteBox(self.view)
+        write_box.model.user_id_email_dict = user_id_email_dict
+        write_box.model.user_dict = user_dict
+        connect_signal_mock = mocker.patch.object(urwid, "connect_signal")
+        enable_autocomplete_mock = mocker.patch.object(
+            ReadlineEdit, "enable_autocomplete"
+        )
+        write_box.private_box_edit_view(recipient_user_ids=recipient_user_ids)
+        enable_autocomplete_mock.assert_has_calls(
+            [
+                mock.call(
+                    func=write_box.generic_autocomplete,
+                    key=primary_key_for_command("AUTOCOMPLETE"),
+                    key_reverse=primary_key_for_command("AUTOCOMPLETE_REVERSE"),
+                ),
+            ]
+        )
+        connect_signal_mock.assert_not_called()
+        assert isinstance(write_box.to_write_box, urwid.Text)
+        assert write_box.to_write_box.text == "To: Human 1 <person1@example.com>"
+
     def test__setup_common_private_compose(self, mocker: MockerFixture) -> None:
         write_box = WriteBox(self.view)
         write_box.to_write_box = mocker.MagicMock()
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -1,11 +1,14 @@
 import datetime
 from collections import OrderedDict
 from typing import Any, Callable, Dict, List, Optional
+from unittest import mock
 
 import pytest
+import urwid
 from pytest import param as case
 from pytest_mock import MockerFixture
 from urwid import Widget
+from urwid_readline import ReadlineEdit
 
 from zulipterminal.config.keys import keys_for_command, primary_key_for_command
 from zulipterminal.config.symbols import (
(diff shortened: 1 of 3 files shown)


Subhasish-Behera:

zulipterminal/ui_tools/boxes.py
```diff
@@ -97,6 +97,15 @@ def __init__(self, view: Any) -> None:
         # Prioritizes autocomplete in message body
         self.recipient_user_ids: List[int]
 
+        # Marks if the write box formed is involves [EDIT_MESSAGE: e] key
+        # Only used for private messages(for now)
+        self.edit_keypressed: bool
+        self.edit_keypressed = False
```
neiljp, line 103: The edit state is encoded in `msg_edit_state` - does this give more information?

zulipterminal/ui_tools/boxes.py
```diff
@@ -97,6 +97,15 @@ def __init__(self, view: Any) -> None:
...
+
+        # Marks if the write box formed is involves [REPLY_MESSAGE: r] key
+        # Only used for private messages(for now)
+        self.reply_keypressed: bool
+        self.reply_keypressed = False
```
neiljp, line 108: Why are we tracking if it's a reply? Ah, this is since private edits are triggered by reusing the reply logic?

neiljp (commented):
@Subhasish-Behera Thanks for working on this too :+1:

I've tested this manually and it seems to function well, for use of cycling using <kbd>TAB</kbd> or similar.

In the version we merged for streams, it benefits from the stream being a non-editable text object, which also limits clicking into the box or other accidental cases where it receives focus. That is a small issue with the solution in this PR right now, since we have only one private compose setup function.

Taking an approach like for streams but for private messages, seems like it would also avoid the possible need for tracking the reply state. The branch where we reuse the reply mode would then use a dedicated setup function for private edit, like we already have for stream_box_edit_view.

We could still maintain an error message for tab (or up) when editing a PM, as a separate commit/feature, though I don't think it's as important as making the recipient(s) unchangeable.

How does that sound?

zulipterminal/ui_tools/boxes.py
Subhasish-Behera: Yes, the keypress of `EDIT_MESSAGE` on private messages works on calling the keypress for `REPLY_MESSAGE` using `self.keypress(size, primary_key_for_command("REPLY_MESSAGE"))`.That's why reply_keypressed is used to differentiate between cases .

zulipterminal/ui_tools/boxes.py
Subhasish-Behera: I didn't think of that...nope, it does not give any other information. It seems that the same work can be done using `msg_edit_state`(I ran the code replacing `edit_keypressed` with `msg_edit_state` and it worked fine.)
So I will remove the `edit_keypressed` part and use `msg_edit_state`  instead :)

Subhasish-Behera:
The new code dosen't need `self.reply_keypressed` or `self.edit_keypressed` to track which key triggered the `private_box_view` as you suggested.

Subhasish-Behera:

zulipterminal/ui_tools/boxes.py
```diff
@@ -205,18 +205,16 @@ def _setup_common_private_compose(self) -> None:
         ]
         self.focus_position = self.FOCUS_CONTAINER_MESSAGE
 
-    def private_box_view(
-        self,
-        *,
-        recipient_user_ids: Optional[List[int]] = None,
+    def convert_id_to_info(
+        self, recipient_user_ids: Optional[List[int]] = None
```
neiljp, line 209: Would this be better incorporated into `_set_regular_and_typing_recipient_user_ids`?

zulipterminal/ui_tools/messages.py
```diff
@@ -1082,7 +1082,9 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                     )
 
             if self.message["type"] == "private":
-                self.keypress(size, primary_key_for_command("REPLY_MESSAGE"))
+                self.model.controller.view.write_box.private_box_edit_view(
+                    recipient_user_ids=self.recipient_ids,
+                )
```
neiljp, line 1087: This is what enables the new edit-private-message functionality, correct?

zulipterminal/ui_tools/boxes.py
```diff
@@ -898,22 +898,29 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
-                    all_valid = self._tidy_valid_recipients_and_notify_invalid_ones(
-                        self.to_write_box
-                    )
-                    if not all_valid:
-                        return key
-                    # We extract recipients' user_ids and emails only once we know
-                    # that all the recipients are valid, to avoid including any
-                    # invalid ones.
-                    self.update_recipients(self.to_write_box)
+                    if isinstance(self.to_write_box, ReadlineEdit):
```
neiljp, line 901: Can we use the same conditional here as below? I expect isinstance to work, but it's not as descriptive as the current state.

zulipterminal/ui_tools/boxes.py
```diff
@@ -927,6 +927,10 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                 else:
                     header.focus_col = self.FOCUS_HEADER_BOX_STREAM
             else:
+                if self.msg_edit_state:
+                    self.model.controller.report_error(
+                        [" Recipient(s) can not be edited while editing a PM."]
```
neiljp, line 932: This is now DM, but let's use a fuller version too.

tests/ui_tools/test_boxes.py
```diff
@@ -1642,7 +1677,10 @@ def test_keypress_CYCLE_COMPOSE_FOCUS(
             else:
                 write_box.stream_box_view(stream_id)
         else:
-            write_box.private_box_view()
+            if recipient_box_editable:
+                write_box.private_box_view()
+            else:
+                write_box.private_box_edit_view()
```
neiljp, line 1686: We can adjust this test as soon as we have these functions, ie. sooner in the sequence of commits and potentially in the commit that adds this method (can squash later, but keep it adjacent).

neiljp (commented):
@Subhasish-Behera This looks much different - and more detailed - and certainly seems to work in practice :+1:

Your commit style is broadly OK, but there are a few elements that are throwing the style off:
- there's no blank line between the title and summary (so tig is showing it all on one line for me!)
- it's challenging, but it works best, including for github breaking the title into multiple lines, if you can keep the title to 72 characters
- using gitlint as per the README should help prompt for these issues; if not then we can adjust the rules it uses perhaps
- for referring to future work, it's useful to refer to the principle of why you're doing it, rather than eg. specific function names that don't yet exist :)

The tests are generally good to see, with a few points to bear in mind:
- if adding a test for existing code, do it before modifying; you can then be more sure that you don't break anything
- if adding a function, generally add the test in that commit too
- it's not always necessary to test every underlying helper function
- similarly, it's good to aim for the big picture behavior, though sometimes it can be challenging not to test the implementation

I like how you extracted the function to handle the common code, in addition to the one for the base PM/DM setup, though I do wonder if it would be simpler to incorporate it into the existing similar function.

I'll review the tests in more detail later, since we may wish to restructure again slightly before we're done, as per my comments - this is another reason that testing every little helper function can be a challenge, since it's testing the implementation in some sense!

zulipterminal/ui_tools/boxes.py
Subhasish-Behera: If I have understood correctly you are referring to `msg_edit_state`. Yep, that should be a better choice as it seems more basic. Probably used `isinstance(self.to_write_box, ReadlineEdit)` because I was thinking along the lines of `to_write_box` being editable or not as the variable between two cases.

zulipterminal/ui_tools/boxes.py
Subhasish-Behera: It would be possible if we were only talking about `convert_id_to_info` but `_set_regular_and_typing_recipient_user_ids` is also used by `update_recipients` function .

zulipterminal/ui_tools/boxes.py
```diff
@@ -172,41 +172,11 @@ def send_stop_typing_status(self) -> None:
...
-            recipient_info = ""
-
-        self.send_next_typing_update = datetime.now()
-        self.to_write_box = ReadlineEdit("To: ", edit_text=recipient_info)
-        self.to_write_box.enable_autocomplete(
-            func=self._to_box_autocomplete,
-            key=primary_key_for_command("AUTOCOMPLETE"),
-            key_reverse=primary_key_for_command("AUTOCOMPLETE_REVERSE"),
-        )
-        self.to_write_box.set_completer_delims("")
```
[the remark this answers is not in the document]
neiljp: If we shouldn't be sending typing updates for editing, I suspect that could be clearer in the API documentation. If you're sure this is the behavior we want, posting to that stream that you think that's the case would be useful, or even a quick PR for it.

Since we're splitting the private-compose and private-edit UI it certainly makes it simpler to separate this behavior in this way.

zulipterminal/ui_tools/boxes.py
neiljp: OK, so
- `_set_regular_and_typing_recipient_user_ids` does the following, dependent upon passed user_ids
  - sets `recipient_user_ids` (autocompletion)
  - sets `typing_recipient_user_ids` (not including user)
- `update_recipients` (which appears to effectively be `update_recipients_from_emails_in_widget` or similar):
  - sets `recipient_emails`
  - calls `_set_regular_and_typing_recipient_user_ids` with calculated ids from recipient_emails
- `convert_id_to_info` does the following, dependent upon the passed user_ids
  - calls `_set_regular_and_typing_recipient_user_ids`
  - sets `recipient_emails`
  - sets `recipient_info`

This pattern suggests that the last two methods have a similar intent and so should share a name prefix, and generally could benefit from renaming.

The latter sets an extra attribute, which seems to be for initialization only - can we make that a return value instead?

zulipterminal/ui_tools/boxes.py
neiljp: Right, `msg_edit_state` is used in the logic below.

zulipterminal/ui_tools/boxes.py
neiljp: OK, so
- `_set_regular_and_typing_recipient_user_ids` does the following, dependent upon passed user_ids
  - sets `recipient_user_ids` (autocompletion)
  - sets `typing_recipient_user_ids` (not including user)
- `update_recipients` (which appears to effectively be `update_recipients_from_emails_in_widget` or similar):
  - sets `recipient_emails`
  - calls `_set_regular_and_typing_recipient_user_ids` with calculated ids from recipient_emails
- `convert_id_to_info` does the following, dependent upon the passed user_ids
  - calls `_set_regular_and_typing_recipient_user_ids`
  - sets `recipient_emails`
  - sets `recipient_info`

This pattern suggests that the last two methods have a similar intent and so should share a name prefix, and generally could benefit from renaming.

The latter sets an extra attribute, which seems to be for initialization only - can we make that a return value instead?

neiljp:
@Subhasish-Behera Is this ready for another review?

Subhasish-Behera:
> @Subhasish-Behera Is this ready for another review?

Yep.

neiljp (commented):
@Subhasish-Behera This looks like it will be really useful! :+1:

I did leave a lot of comments, but some relate to commit structure, commit text, and smaller changes or suggestions.

It seems like the functionality is essentially done, so this should be a case of moving things around and some final tidying, with some optional extensions.

Other than fixing the issue fully, this will be a good refactor to the compose code to have in place for future work! :tada:

PR #1282, opened 2023-01-06
Create FAQ section explaining footlinks and how to access the links (attachments, internal links, external links)

neiljp:
**What does this PR do?**  

This was motivated by a realization that we don't explain footlinks except in the CHANGELOG, and a query regarding how to open images and similar.

(see https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Q.3A.20Opening.20images.3F
ie. #**zulip-terminal > Q: Opening images?**)

This PR introduces a single section, then adds subsections for how footlinks are constrained in the message list, and how the full link data is available in the message information popup.

This expands upon a pre-existing FAQ entry regarding how footlinks are shown, which is removed.

**Tested?** 
- [X] Manually

**Notes & Questions** 

@Sushmey pointed out as feedback that the reference to images is not as clear as it could be, though is mentioned in the last subsection, which doesn't stand out that well, which I agree with.

It's important to note that 'link' as used here refers to any non-text object that we can't (or potentially choose not to) display inline. It may be clearer to express the main point in that way.

A good direct follow-up to this would be to consider ways to restructure this information to make the definition of 'link' more clear (as well as use consistent terms for the 'text' and 'location', to make them easier to refer to?), as well as expose the 'I want to open an attachment' type questions, which may be better expressed as separate FAQ entries which redirect to a link subsection?

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -66,6 +66,14 @@ to show links at the end of a given message.
 
 **NOTE** Footlinks were added in version 0.5.2.
 
+### Why can I only see 3 footlinks when there are more links in my message?
+
+This behavior was introduced in version 0.7.0, to avoid a message with many
+links having a very long footlinks list below it.
+
+This can be controlled by giving a different value to the new
+`maximum-footlinks` setting in the `zuliprc` file, which defaults to 3.
+
 ## When are messages marked as having been read?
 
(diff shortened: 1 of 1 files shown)


neiljp:
Minor adjustments made, plan to merge soon.

PR #1291, opened 2023-02-07
[WIP] Handle typing events from multiple users in a PM narrow 

### LINKED ISSUE #992: Handle typing events from multiple users in a PM narrow
We currently don't support displaying multiple users' typing statuses in a PM narrow. 
In a huddle/group PM, we only handle typing events from the first email in the narrow specification. 
We should be handling typing statuses from all the senders in the huddle.

The web-app handles multiple users typing this way:
```
Person 1 is typing...
Person 2 is typing...
```
with Person 1 being the user that sent the most recent `start` update. Since we're limited on footer area, we would want it shown in the same line, i.e., `Person 1, Person 2 are typing...` for the first few users, and on crossing a threshold, resort to displaying `Multiple people are typing...`.

Each user's typing status in the huddle should be displayed accurately in the footer, without overlap/change in the handling of the typing status received from the other users, i.e., a `stop` event from `Person 1` in the above scenario should not affect the display of `Person 2`'s typing status.

This can be further discussed here, or at [#zulip-terminal > Multiple users typing](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Multiple.20users.20typing).

Subhasish-Behera:
**What does this PR do?**  

Fixes #992 

**Tested?** 
- [X] Manually

**Commit flow** 

**Notes & Questions** 

Basically now the `controller.active_conversation_info` is a `set` which adds a sender if the event is "start" and removes if the event is "stop". 
Reasons for change :-
1. There seems to be no requirement for Dict as It had only one key -"sender"
2. in dict , discarding/removing by element name is not possible (but possible in `set`)

**Interactions** 

**Visual changes**


tests/core/test_core.py
```diff
@@ -553,22 +553,22 @@ def test_maximum_popup_dimensions(
...
-        active_conversation_info: Dict[str, str],
+        active_conversation_info: Set[Any],
     ) -> None:
         set_footer_text = mocker.patch(VIEW + ".set_footer_text")
         sleep = mocker.patch(MODULE + ".time.sleep")
         controller.active_conversation_info = active_conversation_info
 
         def mock_typing() -> None:
-            controller.active_conversation_info = {}
+            controller.active_conversation_info = set()
```
Subhasish-Behera, line 571: In Line 571 and Line 589 ,The `{}` works for test whereas `set()` works for Lint

tests/model/test_model.py
```diff
@@ -2882,7 +2882,7 @@ def test_toggle_stream_visual_notifications(
                     "id": 0,
                 },
                 False,
-                {"sender_name": "hamlet"},
+                {"hamlet"},
```
Subhasish-Behera, line 2885: In the cases of 1. `in_pm_narrow_with_sender_typing:start`, 
2. `in_pm_narrow_with_sender_typing:start_while_animation_in_progress`,
3 `in_pm_narrow_with_sender_typing:stop`,
when the test function calls `model._handle_typing_event(event)`, there is a AttributeError where `self.controller.active_conversation_info` is considered as `Dict` where as it is actually `Set`

zulipterminal/model.py
```diff
@@ -1374,7 +1374,7 @@ def _handle_typing_event(self, event: Event) -> None:
 
         narrow = self.narrow
         controller = self.controller
-        active_conversation_info = controller.active_conversation_info
+        active_conversation_info = self.controller.active_conversation_info
```
neiljp, line 1377: It's unclear to me why this change is necessary. The controller local variable is defined above this.

zulipterminal/core.py
```diff
@@ -79,7 +79,7 @@ def __init__(
 
         self._editor: Optional[Any] = None
 
-        self.active_conversation_info: Dict[str, Any] = {}
+        self.active_conversation_info: Set[Any] = set()
```
neiljp, line 82: Whatever data structure we use here, we should move to a situation where we are not using `Any`, but rather the specific type being used.

zulipterminal/core.py
```diff
@@ -433,13 +433,31 @@ def show_typing_notification(self) -> None:
...
         # Until conversation becomes "inactive" like when a `stop` event is sent
         while self.active_conversation_info:
-            sender_name = self.active_conversation_info["sender_name"]
-            self.view.set_footer_text(
-                [
-                    ("footer_contrast", " " + sender_name + " "),
-                    " is typing" + next(dots),
-                ]
+            active_conversation_info = ", ".join(
+                map(str, self.active_conversation_info)
```
neiljp, line 437: If these are already strings - try changing that `Any` I mention elsewhere - then this is misleading, since we don't need to apply that mapping.

zulipterminal/core.py
```diff
@@ -433,13 +433,31 @@ def show_typing_notification(self) -> None:
...
             )
+            if len(self.active_conversation_info) == 1:
```
neiljp, line 439: In the following if/elif chain, there is a lot of common structure which is repeated. Try setting the footer text once, with variables that are set in the if/elif chain.

The logic will likely read clearer if you also give a name to the `len` before that logic.

zulipterminal/model.py
```diff
@@ -1383,18 +1383,19 @@ def _handle_typing_event(self, event: Event) -> None:
         if (
             len(narrow) == 1
             and narrow[0][0] == "pm_with"
-            and sender_email in narrow[0][1].split(",")
+            and sender_email in [email.strip() for email in narrow[0][1].split(",")]
```
neiljp, line 1386: Is this change a bugfix? Why are we stripping?

neiljp (commented):
@Subhasish-Behera Thanks for working on this :+1:

I've left a little feedback, but I'll follow up in the stream regarding the datastructure point (dict vs set).

zulipterminal/model.py
Subhasish-Behera: Yep. Earlier only the first user was detected as there was a space after the comma. This problem was mentioned in the issue as well ("In a huddle/group PM, we only handle typing events from the first email in the narrow specification.")

zulipterminal/model.py
Subhasish-Behera: My bad . I thought one way to clear the attribute error(in tests) could be to directly assign it (using self.controller.active_convesation_info rather than passing it from a variable). But that didn't work. So, this change is not necessary and I will fix it.

zulipterminal/core.py
Subhasish-Behera: Will keep this in mind and change `Any` to `str`

zulipterminal/core.py
mounilKshah: I think you can modify it even further. If you notice, only the string containing `active_conversation_info` and *is/are typing* changes across all conditions. So you can to set only those strings in the switch case, and while passing the text to `set_footer_text()`, you can provide remaining elements.

mounilKshah (commented):
I tried running this PR on my local machine, and it worked well for PMs having 3 users and 2 typing simultaneously.
I had one query. All the tests currently present seem to have only one other user in the `pm_with` narrow, so test cases for multiple users typing simultaneously are yet to be implemented right?

zulipterminal/core.py
Subhasish-Behera: Yep. can be better. This version should address your suggestions
```
        while self.active_conversation_info:
           active_conversation_info = ", ".join(
                map(lambda x: self.model.user_dict[x]["full_name"], self.active_conversation_info)
            )
            no_of_typing_users = len(self.active_conversation_info)
            if no_of_typing_users == 1:
                typing_phrase = " is typing"
            elif no_of_typing_users < 4:
                typing_phrase = " are typing"
            else:
                active_conversation_info = "Multiple people"
                typing_phrase = " are typing"
            typing_text = [("footer_contrast",active_conversation_info+" ")]
            typing_text.append(typing_phrase+next(dots))
            self.view.set_footer_text(typing_text)
```
the whole initial active_conversation_info part just addresses Neil's indexing through the user_id(rather than full name) suggestion.

Subhasish-Behera:

tests/core/test_core.py
```diff
@@ -580,10 +580,30 @@ def mock_typing() -> None:
...
-                    mocker.call([("footer_contrast", " hamlet "), " is typing"]),
-                    mocker.call([("footer_contrast", " hamlet "), " is typing."]),
-                    mocker.call([("footer_contrast", " hamlet "), " is typing.."]),
-                    mocker.call([("footer_contrast", " hamlet "), " is typing..."]),
+                    mocker.call(
+                        [
+                            ("footer_contrast", "hamlet "),
+                            ("footer", " is typing"),
+                        ]
+                    ),
```
neiljp: I'm a bit confused why this reformatting was necessary? If you drop the comma after the second list element, does black shrink it back again?

zulipterminal/core.py
```diff
@@ -439,12 +439,17 @@ def show_typing_notification(self) -> None:
...
+            no_of_typing_users = len(self.active_conversation_info)
+            if no_of_typing_users == 1:
+                typing_phrase = " is typing"
+            elif no_of_typing_users < 4:
+                typing_phrase = " are typing"
+            else:
+                active_conversation_info = "Multiple people"
+                typing_phrase = " are typing"
+            typing_text = [("footer_contrast", active_conversation_info + " ")]
+            typing_text.append(("footer", typing_phrase + next(dots)))
```
neiljp, line 451: This part is definitely better with less duplication :+1:

zulipterminal/model.py
```diff
@@ -1392,7 +1392,8 @@ def _handle_typing_event(self, event: Event) -> None:
                     controller.show_typing_notification()
 
             elif event["op"] == "stop":
-                controller.active_conversation_info = {}
+                sender_email = self.user_dict[sender_email]["email"]
+                active_conversation_info.discard(sender_email)
```
neiljp, line 1396: This commit doesn't yet add tests for this case, or the output the user sees.

Does the model change along show reasonable improvements? If so, you could add a test change for the model in this commit, and move the core changes into the next commit with the test you added.

tests/core/test_core.py
```diff
@@ -578,35 +599,94 @@ def mock_typing() -> None:
...
+                        ),
+                        mocker.call(
+                            [
+                                ("footer_contrast", f"{active_conversation_info} "),
+                                ("footer", " is typing..."),
+                            ]
+                        ),
+                    ]
+                )
+            elif length < 4:
```
neiljp, line 631: This conditional should not be present in the test. Instead add a different parameter to the test for what you expect in each case.

tests/model/test_model.py
```diff
@@ -3021,10 +3030,131 @@ def test_toggle_stream_visual_notifications(
...
+                        {"user_id": 4, "email": "hamlet@zulip.com"},
+                        {"user_id": 6, "email": "claudius@zulip.com"},
+                    ],
+                    "id": 0,
+                },
+                True,
+                {"claudius@zulip.com", "hamlet@zulip.com"},
+                {"claudius@zulip.com", "hamlet@zulip.com"},
+                False,
+                id="in_group_pm_narrow,usertyping:stop_" "while animation in progress",
```
neiljp, line 3156: These ids look confusing.

tests/core/test_core.py
```diff
@@ -551,22 +551,28 @@ def test_maximum_popup_dimensions(
...
+            case({"hamlet@zulip.com"}, id="in_pm_narrow_with_sender_typing:start"),
+            case(set(), id="in_pm_narrow_with_sender_typing:stop"),
         ],
     )
     def test_show_typing_notification(
         self,
         mocker: MockerFixture,
         controller: Controller,
-        active_conversation_info: Dict[str, str],
+        active_conversation_info: Set[Any],
```
neiljp: Check the types across commits - whether we settle on Set[str] as now or Set[int] for user ids, we shouldn't need to use Any now.

neiljp (commented):
@subhasish-Behera This looks to be working fine from a manual test :+1:

I left feedback inline. My main thought at this point, having not looked at the topic recently until again now just now, is that the logic would certainly be simpler if we tracked user ids instead of emails. We then have user_name_from_id to avoid diving into the internal model data, which should minimize some of the mocking.

This PR also currently has the commit text directly under the title as you had before, but that should be straightforward since you fixed other PRs already.

prah23 (changes requested):
Thanks for the updates @Subhasish-Behera!

My feedback would be to let `active_conversation_info` remain a `Dict` with emails as the keys and full names as the values. This way, the lookup for names in the controller's `show_typing_notifications` can be pushed to the model and can be performed in place of the line I marked in the comment below.

zulipterminal/model.py
```diff
@@ -1381,18 +1381,19 @@ def _handle_typing_event(self, event: Event) -> None:
...
             len(narrow) == 1
             and narrow[0][0] == "pm_with"
-            and sender_email in narrow[0][1].split(",")
+            and sender_email in list(narrow[0][1].split(", "))
             and sender_id != self.user_id
         ):
             if event["op"] == "start":
-                sender_name = self.user_dict[sender_email]["full_name"]
-                active_conversation_info["sender_name"] = sender_name
+                sender_email = self.user_dict[sender_email]["email"]
```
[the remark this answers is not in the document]
Subhasish-Behera: Yeah agreed. When I wrote the next line `active_conversation_info.add(sender_email)`, I was thinking that sender_email is the variable that we are talking about now to remove and not the previously declared `sender_email = event["sender"]["email"]`.
I think this can also be done in the `event["op"] == "stop"`

Issue #588, opened 2020-04-11
Alert word handling


neiljp:
We can discuss options for this, but possible ideas could include:
* [ ] Notify when an alert word is used (if narrow is not muted) [as in webapp]
* [ ] Button for alert words? Showing unread alerts? Hotkey? [not in webapp]
* [ ] Ability to configure alert words, possibly via using `i` on the button? [webapp uses settings]

preetmishra:
@neiljp We don't markup the alert words differently (like the web app) currently, right?

I can see my alert words in 'green' in the web app but not in ZT.

neiljp:
Agreed - added to list; we basically don't do anything with alert words right now.

zee-bit:
This looks interesting, didn't know there was such a feature in Zulip. :sweat_smile: 
I would love to work on this. :smile:

PR #1301, opened 2023-02-20
[WIP] Notify when a message with alert word is received

Subhasish-Behera:
model/test_model/conftest : Notify when a message with alert word is received

**What does this PR do?**  

Partial fix for #588

**Tested?** 
- [X] Manually

**Commit flow** 

**Notes & Questions** 

->No changes have been made to for private messages(not required as alert_words would be a subset case for this) 
and in stream added the condition `not is_muted_topic` and `not is_muted_stream` when a message has `flag:'has_alert_word`'
-> Added tests specifically for stream with 3 parameters  `visual_notification_status`,`is_stream_muted`,`is_topic_muted`
-> Added a new `stream_msg_fixture` in `conftest.py` as msg_fixture was present for private messages but not for stream. And the same fixture has been used in added tests.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1978,6 +1978,8 @@ class TestModel:
             "vary_each_msg",
             "visual_notification_status",
             "types_when_notify_called",
+            "is_muted_stream",
+            "is_muted_topic",
         ],
         [
             (
@@ -1985,11 +1987,52 @@ class TestModel:
                 {"flags": ["mentioned", "wildcard_mentioned"]},
                 True,
                 [],
+                None,
+                None,
             ),  # message_fixture sender_id is 5140
-            (5179, {"flags": ["mentioned"]}, False, ["stream", "private"]),
-            (5179, {"flags": ["wildcard_mentioned"]}, False, ["stream", "private"]),
-            (5179, {"flags": []}, True, ["stream", "private"]),
-            (5179, {"flags": []}, False, ["private"]),
+            (5179, {"flags": ["mentioned"]}, False, ["stream", "private"], None, None),
+            (
+                5179,
+                {"flags": ["wildcard_mentioned"]},
+                False,
+                ["stream", "private"],
+                None,
+                None,
+            ),
+            (5179, {"flags": []}, True, ["stream", "private"], None, None),
+            (5179, {"flags": []}, False, ["private"], None, None),
+            (
+                5179,
+                {"flags": ["has_alert_word"]},
+                False,
+                ["stream", "private"],
+                True,
+                True,
+            ),
+            (
+                5179,
+                {"flags": ["has_alert_word"]},
+                False,
+                ["stream", "private"],
+                True,
+                False,
+            ),
+            (
+                5179,
+                {"flags": ["has_alert_word"]},
+                False,
+                ["stream", "private"],
+                False,
+                True,
+            ),
+            (
+                5179,
+                {"flags": ["has_alert_word"]},
+                False,
+                ["stream", "private"],
+                False,
+                False,
+            ),
         ],
         ids=[
             "not_notified_since_self_message",
@@ -1997,6 +2040,10 @@ class TestModel:
             "notified_stream_and_private_since_wildcard_mentioned",
             "notified_stream_since_stream_has_desktop_notifications",
             "notified_private_since_private_message",
+            "not_notified_for_stream_since_topic/stream_both_muted",
+            "not_notified_for_stream_since_stream_muted",
+            "not_notified_for_stream_since_topic_muted",
+            "notified_for_stream_since_topic/stream_both_not_muted",
         ],
     )
     def test_notify_users_calling_msg_type(
@@ -2008,6 +2055,8 @@ class TestModel:
         vary_each_msg,
         visual_notification_status,
         types_when_notify_called,
+        is_muted_stream: Optional[bool],
+        is_muted_topic: Optional[bool],
     ):
         message_fixture.update(vary_each_msg)
         model.user_id = user_id
@@ -2015,6 +2064,8 @@ class TestModel:
             MODEL + ".is_visual_notifications_enabled",
             return_value=visual_notification_status,
         )
+        mocker.patch.object(model, "is_muted_stream", return_value=is_muted_stream)
+        mocker.patch.object(model, "is_muted_topic", return_value=is_muted_topic)
         notify = mocker.patch(MODULE + ".notify")
 
         model.notify_user(message_fixture)
@@ -2031,7 +2082,17 @@ class TestModel:
         if target is not None:
             title = f"Test Organization Name:\nFoo Foo (to {target})"
             # TODO: Test message content too?
-            notify.assert_called_once_with(title, mocker.ANY)
+            if message_fixture["type"] == "private":
+                notify.assert_called_once_with(title, mocker.ANY)
+            elif len(
+                vary_each_msg["flags"]
+            ) == 1 and "has_alert_word" in vary_each_msg.get("flags"):
+                if not is_muted_topic and not is_muted_stream:
+                    notify.assert_called_once_with(title, mocker.ANY)
+                else:
+                    notify.assert_not_called()
+            else:
+                notify.assert_called_once_with(title, mocker.ANY)
         else:
             notify.assert_not_called()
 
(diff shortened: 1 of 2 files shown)


neiljp (commented):
@Subhasish-Behera Thanks for looking at this :+1:

I left some feedback inline, which I think mostly matches the feedback I gave in the stream, but of course more specific in nature :)

Certainly while this is under development it's fine to keep it as a separate test. As the PR develops you may find it's increasingly similar to what the other test does and can be combined - or not. Regarding specific points:
- I'm aware the muted aspect is distinct from the other test right now, but it's possible that we need to incorporate handling muted streams/topics there too, as a separate improvement.
- As with this PR, it's good to focus on the specific change rather than 'in the future' - unless you're going to include it in a commit straight after! So considering muted users to be an obstacle right now is not important.

As I noted, you should be able to simply use the stream_msg_template fixture. I'd also be interested as to what happened with using the general fixture; the similar test function has logic in the test that translates from `types_when_notify_called` to a variable, to enable testing the title, but I don't think that should affect the logic - and should probably be in another test.

mounilKshah (commented):
This works just as per expectations on my local machine. I haven't gone through the automated tests yet, but upon manually testing it, everything seemed to work fine 👍

tests/model/test_model.py
```diff
@@ -2048,6 +2048,99 @@ def test_notify_users_hides_PM_content_based_on_user_setting(
...
+            model,
+            "is_visual_notifications_enabled",
+            return_value=visual_notification_status,
+        )
+        mocker.patch.object(model, "is_muted_stream", return_value=is_stream_muted)
+        mocker.patch.object(model, "is_muted_topic", return_value=is_topic_muted)
+        notify = mocker.patch(MODULE + ".notify")
+        model.notify_user(stream_msg_fixture)
+        if no_of_times_notify_call_expected == 1:
+            assert notify.called
```
theViz343, line 2139: Maybe, the parameters with which the `notify` method would also benefit from testing? If you decide to do that, you can use `notify.assert_called_once_with(param1,..)` and `notify.assert_not_called()`.

theViz343 (commented):
This looks good! @Subhasish-Behera I've tested it manually and it seems to work great :)

tests/model/test_model.py
Subhasish-Behera: Yep! agreed.

tests/model/test_model.py
```diff
@@ -2048,6 +2048,99 @@ def test_notify_users_hides_PM_content_based_on_user_setting(
             f"Test Organization Name:\nFoo Foo (to you{others})", expected_content
         )
 
+    @pytest.mark.parametrize(
+        [
+            "user_id",
+            "flags",
+            "visual_notification_status",
+            "display_recipient",
+            "subject",
```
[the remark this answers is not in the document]
Subhasish-Behera: I included `user_id` by seeing the  `model.user_id = user_id` of `test_notify_users_calling_msg_type` but I will be removing `user_id` totally(as it is not used).
 And set the rest of them as you suggested.

zulipterminal/model.py
```diff
@@ -1447,7 +1447,13 @@ def notify_user(self, message: Message) -> str:
...
             ) or self.is_visual_notifications_enabled(stream_id):
                 recipient = "{display_recipient} -> {subject}".format(**message)
-
+            if "has_alert_word" in message["flags"]:
+                # Check if stream or topic is muted
+                topic = message.get("subject", "")
+                if not self.is_muted_stream(stream_id) and not self.is_muted_topic(
+                    stream_id, topic
+                ):
+                    recipient = "{display_recipient} -> {subject}".format(**message)
```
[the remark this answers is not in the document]
neiljp: Could you address this, to avoid the duplication? The same action is done in each case, so we can combine the conditionals, though I'd suggest some intermediate variables to make the conditional more readable.

tests/model/test_model.py
```diff
@@ -1978,25 +1978,72 @@ def test__update_topic_index(
             "vary_each_msg",
             "visual_notification_status",
             "types_when_notify_called",
+            "is_muted_stream",
+            "is_muted_topic",
```
neiljp, line 1982: Nit: These are setup values, so it would be clearer to have them before the `types_when_notify_called`.

tests/model/test_model.py
```diff
@@ -2008,13 +2055,17 @@ def test_notify_users_calling_msg_type(
         vary_each_msg,
         visual_notification_status,
         types_when_notify_called,
+        is_muted_stream: Optional[bool],
+        is_muted_topic: Optional[bool],
```
neiljp, line 2059: The return value for these functions cannot be None, so using it is confusing.

Consider a real bool value in each case which will strengthen a test case - eg. if mentions work even if a stream or topic are muted, show that. The ids may need adjusting.

tests/model/test_model.py
```diff
@@ -1978,25 +1978,72 @@ def test__update_topic_index(
...
         ids=[
             "not_notified_since_self_message",
             "notified_stream_and_private_since_directly_mentioned",
             "notified_stream_and_private_since_wildcard_mentioned",
             "notified_stream_since_stream_has_desktop_notifications",
             "notified_private_since_private_message",
+            "not_notified_for_stream_since_topic/stream_both_muted",
+            "not_notified_for_stream_since_stream_muted",
+            "not_notified_for_stream_since_topic_muted",
+            "notified_for_stream_since_topic/stream_both_not_muted",
```
neiljp, line 2046: This is a lot of ids to manually map to the cases above them. Given there are already 5 and we're going to have many more, it would be great to have an initial refactor commit which inlines the existing ones into the test cases.

Also see my comment on an earlier iteration of the PR regarding this.

tests/model/test_model.py
```diff
@@ -2031,7 +2082,17 @@ def test_notify_users_calling_msg_type(
...
+                notify.assert_called_once_with(title, mocker.ANY)
+            elif len(
+                vary_each_msg["flags"]
+            ) == 1 and "has_alert_word" in vary_each_msg.get("flags"):
+                if not is_muted_topic and not is_muted_stream:
+                    notify.assert_called_once_with(title, mocker.ANY)
+                else:
+                    notify.assert_not_called()
+            else:
+                notify.assert_called_once_with(title, mocker.ANY)
```
neiljp, line 2095: It's not clear to me why we need this change?

zulipterminal/model.py
```diff
@@ -1517,6 +1517,13 @@ def notify_user(self, message: Message) -> str:
             ) or self.is_visual_notifications_enabled(stream_id):
                 recipient = "{display_recipient} -> {subject}".format(**message)
 
+            if "has_alert_word" in message["flags"]:
+                # Check if stream or topic is muted
+                topic = message.get("subject", "")
+                if not self.is_muted_stream(stream_id) and not self.is_muted_topic(
+                    stream_id, topic
+                ):
+                    recipient = "{display_recipient} -> {subject}".format(**message)
```
neiljp, line 1526: See my response to a comment on an earlier iteration of the PR regarding simplifying this.

neiljp (commented):
@Subhasish-Behera Thanks for updating this - I left some feedback now you have extended/combined the tests :+1:

zulipterminal/model.py
mounilKshah: How about creating a new function for providing value of `recipient`  in the notification?
We have 2 conditionals: private and stream message. Under stream message, the code for assigning value to `recipient` requires us to check across multiple conditions. One way to simplify this entire block can be to create a separate function which returns the value of `recipient`. This can also be used for assigning value to recipient under the private message conditional as well.

We can have the `message` as a parameter passed to the new function. Also, can we have this as an nested function?

mounilKshah (commented):
I tried this on my local machine and it worked as per expectations. Looks good to me. I have left feedback for the code written to assign value to `recipient` variable. Apart from Neil's comments, this is something that can be looked into.

zulipterminal/model.py
neiljp: The original simplification I saw possible here was the equivalent of changing
```python
if big_long_complicated_expression_for_one_case:
  x = value
if another_big_long_complicated_expression_for_second_case:
  x = value
```
into
```python
is_one_case = big_long_complicated_expression_for_one_case
is_second_case = big_long_complicated_expression_for_second_case
if is_one_case or is_second_case:
  x = value
```
This has the benefit of easily naming the two conditions one is testing for, via the intermediate variables, and deduplicating the assignment.

mounilKshah: Oh alright. This is a simpler approach than what I had suggested.

neiljp:
I'm following up on this now.

PR #1310, opened 2023-02-28
Explicitly specify supported Python versions (3.7-3.11) & upgrade dependencies

### LINKED ISSUE #984: Consider changes needed/possible when Python 3.6 reaches end of life
This is the next iteration of #567, and is mainly for tracking. See that issue for potentially useful tools.

3.6 is due to end 23 Dec 2021 according to https://endoflife.date/python or more officially https://devguide.python.org/#status-of-python-branches

neiljp:
**What does this PR do?**  

This makes the remaining changes to explicitly state we support Python 3.7-3.11, before migrating some settings and dependencies which this now enables.

Fixes #984.

**Tested?** 
- [x] Manually

**Notes & Questions** 

Some of these elements may migrate when moving to pyproject.toml, but this migration does not have a fixed timeline as yet.


neiljp:
Merging now after bugfix in #1311 to allow isolated commits which update dependencies to pass CI.

PR #1314, opened 2023-03-02
[WIP]model/messages/api_types: alert_words mark up

Subhasish-Behera:
**What does this PR do?**  

Partial fix for #588

**Tested?** 
- [X] Manually

**Commit flow** 

first commit:Fetches alert words and registers for any `alert_word` event and updates the alerts words.
second commit: adds transform_content_alert_words which adds `<span class=alert-word>` around alert-words

**Notes** 

-> This PR(specifically the second commit) is based on the alert word implementation of the Zulip web client which was later adapted by other clients. 
link for the web implementation code: [link](https://github.com/zulip/zulip/blob/main/web/src/alert_words.js)
-> Other helpful documentations for contributors:
     1. [regex wrt python](https://docs.python.org/3/howto/regex.html#regex-howto)
     2. [re module](https://docs.python.org/3/library/re.html#functions)
-> Whats Done?
1. fetching and registering of alert_words
2. finding the alert_word patterns and applying the <span> 

-> What needs to be done?

1. Even though applying span logic is ready but the logic for applying markup for alert words inside of multiple elements is not ready. So basically the markup works for the alert-words without any other `tag`s indie `soup2markup`. Before that we have to decide whether we want to apply different markup when multiple tags are present for a alert_word or we have to choose the markup from any one of them.

2. Currently the markup applied for alert-words is same as `code`  tag i.e `msg_code` as the markup for different `zt_themes` are yet to be discussed.

3. Tests are not added as the non-test code is yet to reach a stable state.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -255,6 +255,7 @@ class TestModel:
             "user_settings",
             "realm_emoji",
             "zulip_version",
+            "alert_words",
         ]
         model.client.register.assert_called_once_with(
             event_types=event_types,
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -228,6 +228,7 @@ class TestModel:
         model = Model(self.controller)
 
         event_types = [
+            "alert_words",
             "message",
             "update_message",
             "reaction",
@@ -2705,6 +2706,49 @@ class TestModel:
     )
     def update_message_flags_operation(self, request):
         return request.param
+
+    @pytest.mark.parametrize(
+        "event,initial_alerted_words, expected_alert_words",
+        [
+            case(
+                {"type": "alert_words", "alert_words": ["word1", "word2", "word3"]},
+                ["word1"],
+                ["word1", "word2", "word3"],
+                id="Add_multiple_alert_words",
+            ),
+            case(
+                {"type": "alert_words", "alert_words": []},
+                ["word1"],
+                [],
+                id="Empty_alert_words",
+            ),
+            case(
+                {"type": "alert_words", "alert_words": ["word1", "word4"]},
+                ["word1"],
+                ["word1", "word4"],
+                id="Add_single_new_alert_words",
+            ),
+            case(
+                {"type": "alert_words", "alert_words": ["word1", "word2"]},
+                ["word1", "word2", "word3"],
+                ["word1", "word2"],
+                id="Delete_an_alert_word",
+            ),
+            case(
+                {"type": "alert_words", "alert_words": ["word1", "word3"]},
+                ["word1", "word2"],
+                ["word1", "word3"],
+                id="Add_and_delete_alert_words",
+            ),
+        ],
+    )
+    def test__handle_alert_words_event__single_user(
+        self, mocker, model, event, initial_alerted_words, expected_alert_words
+    ):
+        model._handle_alert_words_event(event)
+
+        assert model._alert_words != initial_alerted_words
+        assert model._alert_words == expected_alert_words
 
     def test_update_star_status_no_index(
         self, mocker, model, update_message_flags_operation
--- a/zulipterminal/api_types.py
+++ b/zulipterminal/api_types.py
@@ -517,6 +517,11 @@ class UpdateDisplaySettingsEvent(TypedDict):
 
 
 # -----------------------------------------------------------------------------
+class AlertWordEvent(TypedDict):
+    type: Literal["alert_words"]
+    alert_words: List[str]
+
+
 Event = Union[
     MessageEvent,
     UpdateMessageContentEvent,
@@ -531,6 +536,7 @@ Event = Union[
     UpdateUserSettingsEvent,
     UpdateGlobalNotificationsEvent,
     RealmUserEvent,
+    AlertWordEvent,
 ]
 
 ###############################################################################
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -147,6 +147,7 @@ class Model:
 
         # Events desired with their corresponding callback
         self.event_actions: Dict[str, Callable[[Event], None]] = {
+            "alert_words": self._handle_alert_words_event,
             "message": self._handle_message_event,
             "update_message": self._handle_update_message_event,
             "reaction": self._handle_reaction_event,
@@ -1379,6 +1380,13 @@ class Model:
                     else:
                         for user_id in user_ids:
                             subscribers.remove(user_id)
+
+    def _handle_alert_words_event(self, event: Event) -> None:
+        """
+        Handle alert_words events
+        """
+        assert event["type"] == "alert_words"
+        self._alert_words = event["alert_words"]
 
     def _handle_typing_event(self, event: Event) -> None:
         """
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -142,6 +142,7 @@ class Model:
             # zulip_version and zulip_feature_level are always returned in
             # POST /register from Feature level 3.
             "zulip_version",
+            "alert_words",
         ]
 
         # Events desired with their corresponding callback
@@ -180,6 +181,8 @@ class Model:
         self.visual_notified_streams: Set[int] = set()
 
         self._subscribe_to_streams(self.initial_data["subscriptions"])
+
+        self._alert_words: List[str] = self.initial_data["alert_words"]
 
         # NOTE: The date_created field of stream has been added in feature
         # level 30, server version 4. For consistency we add this field
@@ -236,6 +239,9 @@ class Model:
 
     def user_settings(self) -> UserSettings:
         return deepcopy(self._user_settings)
+
+    def get_alert_words(self) -> List[str]:
+        return self._alert_words.copy()
 
     def message_retention_days_response(self, days: int, org_default: bool) -> str:
         suffix = " [Organization default]" if org_default else ""
--- a/zulipterminal/ui_tools/messages.py
+++ b/zulipterminal/ui_tools/messages.py
@@ -824,10 +824,8 @@ class MessageBox(urwid.Pile):
 
         clean: str
 
-        def replace_callback(match: Match[Union[str, str, str]]) -> str:
-            before = match.group(1)
-            word = match.group(2)
-            after = match.group(3)
+        def replace_callback(match: Union[Match[str]]) -> str:
+            before,word,after = match.group(1,2,3)
             offset = match.start()
             matched_content = match.string
             pre_match = matched_content[:offset]
@@ -840,8 +838,7 @@ class MessageBox(urwid.Pile):
         for word in alerted_list:
             clean = "".join(alert_regex_replacements.get(c, c) for c in word)
             regex = f"({before_punctuation})({clean})({after_punctuation})"
-            regex1 = re.compile(regex, re.IGNORECASE)
-            content = re.sub(regex1, replace_callback, content)
+            content = re.sub(regex, replace_callback, content, flags=re.IGNORECASE)
 
         return content
 
@@ -857,6 +854,8 @@ class MessageBox(urwid.Pile):
         Dict[str, Tuple[str, int, bool]],
         List[Tuple[str, str]],
     ]:
+        #This method is translated from web/src/aler_tword.js of zulip/zulip
+
         if alert_word_present:
             content = cls.transform_content_alert_words(content, alerted_list)
 
--- a/zulipterminal/ui_tools/messages.py
+++ b/zulipterminal/ui_tools/messages.py
@@ -2,11 +2,12 @@
 UI to render a Zulip message for display, and respond contextually to actions
 """
 
+import re
 import typing
 from collections import defaultdict
 from datetime import date, datetime
 from time import time
-from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Union
+from typing import Any, Dict, List, Match, NamedTuple, Optional, Tuple, Union
 from urllib.parse import urljoin, urlparse
 
 import dateutil.parser
@@ -392,6 +393,9 @@ class MessageBox(urwid.Pile):
                     metadata["bq_len"] -= 1
                     continue
                 markup.append(element)
+            elif tag == "span" and "alert-word" in element.get("class", []):
+                if tag_text:
+                    markup.append(("msg_code", tag_text))
             elif tag == "div" and (set(tag_classes) & set(unrendered_div_classes)):
                 # UNRENDERED DIV CLASSES
                 # NOTE: Though `matches` is generalized for multiple
@@ -628,6 +632,8 @@ class MessageBox(urwid.Pile):
         else:
             recipient_header = None
 
+        self.alerted_words: Optional[Any] = self.model.get_alert_words()
+
         # Content Header
         message = {
             key: {
@@ -716,7 +722,11 @@ class MessageBox(urwid.Pile):
 
         # Transform raw message content into markup (As needed by urwid.Text)
         content, self.message_links, self.time_mentions = self.transform_content(
-            self.message["content"], self.model.server_url
+            self.message["content"],
+            self.model.server_url,
+            self.alerted_words,
+            "has_alert_word" in self.message["flags"]
+            # self.message["flags"],
         )
         self.content.set_text(content)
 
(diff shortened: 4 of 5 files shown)


neiljp (commented):
@Subhasish-Behera Thanks for looking at this other piece of the alert words puzzle :+1:

I've focused on the first commit, since we need the alert words first in order to make progress, at least in the application itself.

It'd be good to split the first commit into two parts:
- getting the initial data
- updating the data

While forgetting to include the updating wouldn't be great (and we have that with other data right now), alert words aren't often updated, so you should be able to manually test a new first commit easily - and even the highlighting if you wanted. That first commit could be merged sooner on it's own once it looks good.

Please note my point about the alert words being 'private' to the model; while other parts of the code simply access inside the model, ideally we want to use a method for this (and migrate the others too). That would sit in the new first commit, and could form part of a test that the data from the server comes out via that function. For that, I'd suggest you check out the API link I gave, as well as the fixtures in `tests/conftest.py`.

The updating part does look very simple, and in a data sense it should be, based on the event data, but note that a change in the data should affect how messages are rendered. That is, not just ones that are constructed based on the message content and alert words at the time, but they need updating too. We could ignore that initially, but we shouldn't forget it :)

As a final quick point, if you've not, I'd suggest installing gitlint as per the README. It should complain about your long commit title ;)
That section should also give you pointers on the general format.
If you find you need to say too much, or include too many files, that can indicate the commit can be split to be simpler.

zulipterminal/api_types.py
```diff
@@ -273,6 +273,12 @@ class UpdateGlobalNotificationsEvent(TypedDict):
     setting: Any
 
 
+class AlertWordEvent(TypedDict):
+    type: Literal["alert_words"]
+    alert_words: List[str]
+    id: int
```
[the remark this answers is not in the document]
Subhasish-Behera: Yep,` id` is not present for all the Events. I probably added it by looking at some non-Event class. So I will remove it.

zulipterminal/model.py
```diff
@@ -136,11 +136,15 @@ def __init__(self, controller: Any) -> None:
             # zulip_version and zulip_feature_level are always returned in
             # POST /register from Feature level 3.
             "zulip_version",
+            "alert_words",  # added alert_words
         ]
 
+        self.alert_words = self.fetch_alert_words()
```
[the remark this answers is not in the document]
Subhasish-Behera: True. So I will change it to `self._alert_words` and add this function
```
  def get_alert_words(self):
        return self._alert_words
```
in `model.py`
and
`self.alerted_words = self. model.get_alert_words()`
in messages.py

zulipterminal/model.py
```diff
@@ -813,6 +817,18 @@ def modernize_message_response(message: Message) -> Message:
 
         return message
 
+    def fetch_alert_words(self) -> List[str]:
+        """
+        Fetches alert words.
+        """
+        response = self.client.get_alert_words()
```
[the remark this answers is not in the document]
Subhasish-Behera: I think I am getting what you are saying. So I will totally remove `fetch_alert_words(self)` function and add `self.alert_words = self.initial_data['alert_words']` in the `_fetch_initial_data(self)` function in the `if not any(results.values()):` conditional.
This along with the the line `"alert_words"` inside `self.initial_data_to_fetch` will go in the first commit . 

And the line `("alert_words", self._handle_alert_words_event)` along with the `_handle_alert_words_event` function and change made to `api_types.py` will go in a second commit.

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: This would be the general idea, yes. Note that if the data is not immutable, or not copied, this still allows the caller to modify the alert words (even accidentally).

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: We currently have a hybrid approach there, it's true. I'm not sure there's a restriction on having the data initialized there, but for now I'd suggest putting it at an appropriate place in `__init__` after the data is fetched.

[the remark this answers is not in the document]
Subhasish-Behera: cool, so I will make a change:
`self.alert_words = self.initial_data['alert_words']` in `__init__`

zulipterminal/ui_tools/messages.py
```diff
@@ -715,7 +721,11 @@ def main_view(self) -> List[Any]:
 
         # Transform raw message content into markup (As needed by urwid.Text)
         content, self.message_links, self.time_mentions = self.transform_content(
-            self.message["content"], self.model.server_url
+            self.message["content"],
+            self.model.server_url,
+            self.alerted_words,
+            "has_alert_word" in self.message["flags"]
```
neiljp, line 727: These new parameters could be combined into one. If the message doesn't have alert words flagged, then the function doesn't need to know/process them.

zulipterminal/ui_tools/messages.py
```diff
@@ -796,14 +806,58 @@ def update_message_author_status(self) -> bool:
 
         return author_is_present
 
+    def transform_content_alert_words(content: str, alerted_list: List[str]) -> Any:
```
neiljp, line 809: If this is translated from a server implementation, it would be useful to add a comment to note that, and in the commit text too.

That could refer to the file as it is now, but also a commit ref, so that if code changes later then it is easier to track what has changed in the original code.

Having this function in a commit on it's own first, with detailed tests to cover various cases, would be good before we hook it into the processing in a subsequent commit.

neiljp (commented):
@Subhasish-Behera Let's still address the first commit first, but I just took a look at the second commit as it stands.

I can see how this mirrors web-app code, but we should test it ourselves too, to ensure that it works how we expect in this translated form.

If we can demonstrate this manually and with tests, this should be a good v1 implementation of this feature.

Beyond that, if we have more tests, and can compare what the web app highlights (and tests), we should be able to determine whether we can do the same using this current approach. This will affect if we need to think differently to cover more cases.

We may want to consider an alternative way of formatting this kind of text, and maybe search text too, if it crosses tag/style boundaries.

I expect the web app applies tags dynamically since it uses CSS to style the text, and I'm guessing the spans can overlap? That's complex with urwid - we process the HTML into urwid-styled Text instead.

If so, we might end up considering processing the content into text only, finding matches, then mapping those into adjacent pieces in the HTML or urwid-processed text.

However, let's focus on testing this v1 first :)
(and the v0, ie. the data handling, before that!)

Subhasish-Behera:

zulipterminal/model.py
```diff
@@ -182,6 +183,8 @@ def __init__(self, controller: Any) -> None:
 
         self._subscribe_to_streams(self.initial_data["subscriptions"])
 
+        self._alert_words = self.initial_data.get("alert_words", None)
```
neiljp, line 187: While using `get` is sort of helpful here, in case it's not in the data, this would cause issues later if it were to be None. We don't validate structure but rely on the expected data being present, for the most part.

In this case, mypy doesn't complain since initial_data is a `Dict[str, Any]`, and it doesn't care what happens with the `Any` 👎.

This means we need to give the type for `self._alert_words`, and only from that point onwards is it guaranteed to be non-None, so a `.get` here makes little sense, unless we're going to do a separate assert before this - or simply let a KeyError drop out in case of an error.

zulipterminal/model.py
```diff
@@ -339,6 +342,9 @@ def set_search_narrow(self, search_query: str) -> None:
         self.unset_search_narrow()
         self.narrow.append(["search", search_query])
 
+    def get_alert_words(self) -> List[str]:
+        return self._alert_words
```
neiljp, line 347: We need to return a copy here, to ensure the data isn't modified by a caller of this function.

zulipterminal/model.py
```diff
@@ -1567,6 +1568,13 @@ def _handle_message_event(self, event: Event) -> None:
 
             self.controller.update_screen()
 
+    def _handle_alert_words_event(self, event: Event) -> None:
```
neiljp, line 1571: Is there a better location for this function to be added? :)

I know model.py is long, but the function above uses the function below, it seems.

tests/model/test_model.py
```diff
@@ -2624,6 +2624,51 @@ def test__handle_reaction_event_for_msg_in_index(
...
+        ],
+        ids=[
+            "Add multiple alert words",
+            "Empty alert words",
+            "Add single new alert words",
+            "Delete an alert word",
+            "Add and delete alert words",
+        ],
+    )
+    def test__handle_alert_words_event(
```
neiljp, line 2664: I would expect to see this test in the same commit as the new `_handle_*_event` function.

You might like to split the function+test, then connect it up with the event-handler system.

tests/model/test_model.py
```diff
@@ -2624,6 +2624,51 @@ def test__handle_reaction_event_for_msg_in_index(
...
+                ["word1", "word3"],
+            ),
+        ],
+        ids=[
+            "Add multiple alert words",
+            "Empty alert words",
+            "Add single new alert words",
+            "Delete an alert word",
+            "Add and delete alert words",
+        ],
```
neiljp, line 2662: Can you please place these inline using `case`?

tests/model/test_model.py
```diff
@@ -2624,6 +2624,51 @@ def test__handle_reaction_event_for_msg_in_index(
...
+    )
+    def test__handle_alert_words_event(
+        self, mocker, model, event, initial_alerted_words, expected_alert_words
+    ):
+        model._handle_alert_words_event(event)
+
+        assert model._alert_words != initial_alerted_words
```
neiljp, line 2669: Is this always true?

neiljp (commented):
@Subhasish-Behera I focused on the first 3 commits, since I think that was the recent focus?

The current overall PR fails when I run it right now, at line 843 of messages.py, compiling a regex (multiple repeat at position 27).

Otherwise, improving your commit text would improve the speed of merging, and the first three seem close, if needing some adjustments.

tests/model/test_model.py
Subhasish-Behera: Yep, this is always true. But it is more context-dependent. Basically, currently, this test is designed to test the most minimal of changes in the `active_conversation_info` i.e change(`start/stop`) of a single user. The `inital_alerted_words` means just before that change and not at the start of the whole typing event series.
So it is true because of the values that are passed in `mark.parametrize` i.e it is an approach followed in the test.

zulipterminal/model.py
Subhasish-Behera: Moved the position in between `_handle_subscription_event` and `_handle_typing_event`

tests/model/test_model.py
Subhasish-Behera: I have squashed the 2nd and 3rd commit into one.

tests/model/test_model.py
neiljp: If that is true, then it is good to add a suffix to the test function name to indicate that.

zulipterminal/model.py
Subhasish-Behera: Upon looking at the `self._alert_words = self.initial_data.get("alert_words", None)` part, I saw that when the user has no alert_words, it returns an empty list(I was probably thinking `None` earlier,not sure). So changed it to `self._alert_words: List[str] = self.initial_data["alert_words"]`.

tests/model/test_model.py
Subhasish-Behera: I think the name `test__handle_alert_words_event_minimal_change` would do the job.

neiljp: Hmm. Do we get an event if the changes aren't made?

I'd suggest a double underscore from the function name to any 'meaning'.

zulipterminal/model.py
```diff
@@ -334,6 +337,9 @@ def set_search_narrow(self, search_query: str) -> None:
         self.unset_search_narrow()
         self.narrow.append(["search", search_query])
 
+    def get_alert_words(self) -> List[str]:
+        return self._alert_words.copy()
+
```
neiljp, line 343: I didn't comment on this before, but this function placement in the file seems strange too - between two related functions?

tests/model/test_model.py
```diff
@@ -2623,6 +2624,49 @@ def test__handle_reaction_event_for_msg_in_index(
...
         return request.param
 
+    @pytest.mark.parametrize(
+        "event,initial_alerted_words, expected_alert_words",
+        [
+            case(
+                {"type": "alert_words", "alert_words": ["word1", "word2", "word3"]},
+                ["word1"],
+                ["word1", "word2", "word3"],
+                id="Add multiple alert words",
```
neiljp, line 2634: minor, but to keep consistent: we tend to use underlines instead of spaces, in case of multiple parametrize.

zulipterminal/model.py
Subhasish-Behera: will fix it asap

neiljp: Note that strictly this *could* go into a separate commit with a test to eg. ensure that one can't modify the data externally, but this is fine.

tests/model/test_model.py
Subhasish-Behera: I believe by changes you mean no typing in the writebox(of any client).
That would be a event with `op`=`stop`. So yes.

Subhasish-Behera: `test__handle_alert_words_event_minimal_change` sounds bit subjective,so i will be changing it to `test__handle_alert_words_event__single_user`. Though it's natural as a single event can't inform about more than 1 user in this case.

zulipterminal/ui_tools/messages.py
```diff
@@ -798,14 +808,59 @@ def update_message_author_status(self) -> bool:
...
+            in_tag = check_string.rfind("<") > check_string.rfind(">")
+            if in_tag:
+                return f"{before}{word}{after}"
+            return f"{before}<span class='alert-word'>{word}</span>{after}"
+
+        for word in alerted_list:
+            clean = "".join(alert_regex_replacements.get(c, c) for c in word)
+            regex = f"({before_punctuation})({clean})({after_punctuation})"
+            regex1 = re.compile(regex, re.IGNORECASE)
+            content = re.sub(regex1, replace_callback, content)
```
theViz343, line 845: The method `re.sub` has a parameter called `flags` in which you could pass the `IGNORECASE` flag, instead of defining a new regex. Ex: `re.sub(.., flags=re.IGNORECASE)`

theViz343 (commented):
@Subhasish-Behera Thanks for working on this :+1:! This seems to work well. There are a few cases I found during testing where it seems to be a bit unpredictable. 

- `\something`. Prefixing with a backslash seems to not get picked up by the regex if the alert word is `\something`. If the alert word is just `something`, it gets picked up when it shouldn't (according to the behavior on the webapp).
- `wqdnasj\` Suffixing with a backslash when the alert word is also `wqdnasj\` seems to crash ZT with the following error - `re.error: missing ), unterminated subpattern at position 23`. 

Apart from this, I've left some suggestions for minor code changes in the comments.

zulipterminal/ui_tools/messages.py
```diff
@@ -798,14 +808,59 @@ def update_message_author_status(self) -> bool:
...
+        before_punctuation = r"\s|^|>|[\\(\".,';\\[]"
+
+        after_punctuation = r"(?=\s|$|<|[\\)\"?!:.,';\]!])"
+
+        clean: str
+
+        def replace_callback(match: Match[Union[str, str, str]]) -> str:
+            before = match.group(1)
+            word = match.group(2)
+            after = match.group(3)
```
theViz343, line 831: Minor: You could use `match.group(1,2,3)` here, which returns a tuple consisting of the three values.

neiljp:
@Subhasish-Behera Is this awaiting an update from you, or have you addressed the feedback?
Just checking based on my earlier feedback and that from Vishwesh :)
(If so, you can add a 👍 emoji or a comment to specific comments to indicate this)

zulipterminal/ui_tools/messages.py
```diff
@@ -798,14 +808,59 @@ def update_message_author_status(self) -> bool:
...
+            # Accept quotes with or without HTML escaping
+            '"': r"(?:\"|&quot;)",
+            "'": r"(?:\'|&#39;)",
+        }
+
+        before_punctuation = r"\s|^|>|[\\(\".,';\\[]"
+
+        after_punctuation = r"(?=\s|$|<|[\\)\"?!:.,';\]!])"
+
+        clean: str
```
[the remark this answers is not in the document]
Subhasish-Behera: Yeah! `clean` here is unrequired.

zulipterminal/ui_tools/messages.py
Subhasish-Behera: The only use these 3 have is in here:
```
 if in_tag:
                return f"{before}{word}{after}"
            return f"{before}<span class='alert-word'>{word}</span>{after}"
```
 I could have directly used `match. group(n)` but didn't. I think separating them into 3 variables gives better readability
 By following this suggestion of `match.group(1,2,3)`, I need it to be stored in some var and break it using the index to use it in the above code which will not clarify any further of what is being done. 
So why group it when it can be accessed individually without creating any `tuple`

neiljp: If I follow the above correctly, would it help if you used tuple unpacking?
```python3
one, two, three = function_returning_tuple_of three_values()

Comment (Subhasish-Behera) [reply zulipterminal/ui_tools/messages.py:830]: Yep! Multiple assignments in one line can do the same thing.
```
before,word,after = match.group(1,2,3)
```

PR #1321, opened 2023-03-11
Improve URL error-handling prior to zuliprc generation

neiljp:
**What does this PR do?**  

This follows up on the bugfix of #1320, now introducing and applying types for the server-settings response.

Also following from #1320, this aims to give the user more accurate feedback and exit cleanly (cf the current confusing message(s) or tracebacks) if the server URL is not recognized as corresponding to a valid Zulip Organization.

Some of this is a little WIP, since the documentation on the server-settings endpoint is a little confusing, but this is close to mergeable.

This aims to work further towards separating the text input and the logic, as well as improving the testing.

**update**
- This is now tidied, and resolved an issue experienced in [**#zulip-terminal>Crash on entering bad server URL**](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Crash.20on.20entering.20bad.20server.20URL)
- This now also adds some documentation for api_types.py
- The last commit as referenced by alexmv was dropped for now, in order to get this merged

**Tested?** 
- [x] Manually

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -207,7 +207,7 @@ def styled_input(label: str) -> str:
     return input(in_color("blue", label))
 
 
-def get_login_id(server_properties: ServerSettings) -> str:
+def get_login_label(server_properties: ServerSettings) -> str:
     require_email_format_usernames = server_properties["require_email_format_usernames"]
     email_auth_enabled = server_properties["email_auth_enabled"]
 
@@ -219,7 +219,7 @@ def get_login_id(server_properties: ServerSettings) -> str:
         # TODO: Validate Email address
         label = "Email: "
 
-    return styled_input(label)
+    return label
 
 
 def get_api_key(realm_url: str) -> Optional[Tuple[str, str, str]]:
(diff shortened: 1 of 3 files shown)


zulipterminal/cli/run.py
```diff
@@ -206,31 +207,46 @@ def styled_input(label: str) -> str:
...
-    return styled_input(label)
+
+class NotAZulipOrganizationError(Exception):
+    pass
+
+
+def get_server_settings(realm_url: str) -> ServerSettings:
+    response = requests.get(url=f"{realm_url}/api/v1/server_settings")
+    unsuccessful_response = response.status_code != requests.codes.OK
+    response_realm_name = response.json().get("realm_name")
```
alexmv, line 230: Maybe only call this line if you have a successful response?  I can imagine a 500 error code with a non-JSON body, which would cause this line to throw an exception.

neiljp: Agreed. Reading this again, I suspect this last commit was a less thought-through exploration of that discussion, rather than the earlier tidying and refactoring.

Probably good to get the earlier commits in soon after the topic on czo, but maybe just dropping that line and replacing the conditional below with:
```python3
if unsuccessful_response or "realm_name" not in response.json():
    ...
```
(plus tests)

PR #1325, opened 2023-03-16
README: Add note regarding use of tools to synchronize source with docs.

neiljp:
**What does this PR do?**  

This documents the availability of `tools/lint-docstring` and `tools/generate-hotkeys.py`.

**Tested?** 
- [x] Manually


mounilKshah (commented):
Good idea to add this in the README. This will make sure people don't end up making changes at both locations, which would defeat the purpose of the linting scripts. This section looks good to me 👍

PR #1331, opened 2023-03-17
Add moved label for topic/stream changes

### LINKED ISSUE #1253: Distinguish EDITED vs MOVED in UI
When messages have been edited in any way (content, topic, stream) then we should be already showing an `EDITED` label to the left of the content, under the sender name.

At some point the web app added a `MOVED` label to distinguish the case where the content has not been modified, and it would be useful to also have this behavior.

Mobile is also adding/adjusting this feature, possibly with a redesign, which motivated this issue.
(eg. https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/UI.20redesign.3A.20edited.20and.20moved.20marks)

It would be worth investigating the approach taken in the other apps, but my take is that we likely want EDITED to take priority over MOVED if more is done to the message, but track both. Please read how we track and show EDITED first, and then think of how to implement MOVED. This would likely be good split into implementing the data-tracking first in the model, and adding tests for that, then extending to include the changes necessary to the UI.

---

**UPDATES**:

In feature level 365 (Zulip 10.0), the server introduced a `last_moved_timestamp` field on message objects, as well as narrowing the meaning of `last_edit_timestamp`. Together these are intended to let the client skip scanning through the `edit_history` object to determine whether a message has been edited or moved.

Therefore, implementing this feature should interpret the new fields, instead of doing that scanning. This will be more efficient, and also a step toward the server being able to stop sending the full edit history in the future.

**References**
- [#api design > edit history @ 💬](https://chat.zulip.org/#narrow/channel/378-api-design/topic/edit.20history/near/2114996)
- Server/web PR: https://github.com/zulip/zulip/pull/33872

Subhasish-Behera:
**What does this PR do?**  

Fixes #1253

**Tested?** 
- [X] Manually

**Commit flow** 

commit 1: 
1. add the moved_message field in class Index 
2. Add the logic of when to add message to index["moved_messages"] and index["edited_messages"].

Tried to achieve parity with the logic of Zulip webapp's handling of edits 
mainly from the `analyse_edit_history` function of `zulip/static/js/message_list_view.js` 
(web app's code doesn't use indexing like ZT though)

commit 2:

the second commit focusses on implementing the same logic in case of update_message_event. But in first commit the logic was related to the response values of `Message` but in this case its with `update_message` event.
does similar job.

In both of the commits , the code identifies a real topic change vs unresolve/resolve 
**Notes & Questions** 

**Interactions**

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -8,6 +8,7 @@ from zulipterminal.api_types import Composition
 from zulipterminal.config.keys import primary_key_for_command
 from zulipterminal.helper import (
     Index,
+    analyse_edit_history,
     canonicalize_color,
     classify_unread_counts,
     display_error_if_present,
@@ -110,6 +111,133 @@ def test_index_messages_narrow_user_multiple(
         "bar@zulip.com": {"user_id": 5180},
     }
     assert index_messages(messages, model, model.index) == index_user_multiple
+
+
+@pytest.mark.parametrize(
+    ("content_changed, stream_changed, old_topic, current_topic, expected_result"),
+    [
+        case(
+            False,
+            True,
+            None,
+            None,
+            {"edited_messages": {12345}, "moved_messages": set()},
+            id="only_stream_changed",
+        ),
+        case(
+            True,
+            False,
+            None,
+            None,
+            {"edited_messages": {12345}, "moved_messages": set()},
+            id="only_content_changed",
+        ),
+        case(
+            True,
+            True,
+            "Topic B",
+            "Topic C",
+            {"edited_messages": {12345}, "moved_messages": set()},
+            id="both_stream_and_content_changed",
+        ),
+        case(
+            True,
+            False,
+            "Topic B",
+            "Topic C",
+            {"edited_messages": {12345}, "moved_messages": set()},
+            id="both_content_and_topic_changed",
+        ),
+        case(
+            False,
+            False,
+            "Topic D",
+            "Topic E",
+            {"edited_messages": set(), "moved_messages": {12345}},
+            id="actual_topic_change",
+        ),
+        case(
+            False,
+            True,
+            "Topic F",
+            "Topic G",
+            {"edited_messages": {12345}, "moved_messages": set()},
+            id="topic_and_stream_changed",
+        ),
+        case(
+            False,
+            False,
+            "✔ Topic I",
+            "✔ Topic J",
+            {"edited_messages": set(), "moved_messages": {12345}},
+            id="topic_changed_from_resolved_to resolved_but_different_topic",
+        ),
+        case(
+            False,
+            False,
+            "Topic K",
+            "Topic L",
+            {"edited_messages": set(), "moved_messages": {12345}},
+            id="topic_changed_when_both_are_unresolved",
+        ),
+        case(
+            False,
+            False,
+            "✔ Topic M",
+            "Topic M",
+            {"edited_messages": set(), "moved_messages": set()},
+            id="only_resolved_topic_to_unresolved_topic",
+        ),
+        case(
+            False,
+            False,
+            "Topic M",
+            "✔ Topic M",
+            {"edited_messages": set(), "moved_messages": set()},
+            id="only_unresolved_topic_to_resolved_topic",
+        ),
+        case(
+            False,
+            False,
+            "✔ Topic N",
+            "Topic M",
+            {"edited_messages": set(), "moved_messages": {12345}},
+            id="resolved_topic_to_changed_unresolved_topic",
+        ),
+        case(
+            False,
+            False,
+            "Topic M",
+            "✔ Topic N",
+            {"edited_messages": set(), "moved_messages": {12345}},
+            id="unresolved_topic_to_changed_resolved_topic",
+        ),
+    ],
+)
+def test_analyse_edit_history(
+    content_changed: bool,
+    stream_changed: bool,
+    current_topic: str,
+    old_topic: str,
+    expected_result: Dict[str, Any],
+    initial_index: Index,
+) -> None:
+    msg_id = 12345
+    expected_index = dict(
+        initial_index,
+        edited_messages=expected_result["edited_messages"],
+        moved_messages=expected_result["moved_messages"],
+    )
+    analyse_edit_history(
+        msg_id,
+        initial_index,
+        content_changed,
+        stream_changed,
+        current_topic,
+        old_topic,
+    )
+
+    assert initial_index == expected_index
 
 
 @pytest.mark.parametrize(
--- a/zulipterminal/helper.py
+++ b/zulipterminal/helper.py
@@ -32,7 +32,12 @@ from urllib.parse import unquote
 import requests
 from typing_extensions import Literal, ParamSpec, TypedDict
 
-from zulipterminal.api_types import Composition, EmojiType, Message
+from zulipterminal.api_types import (
+    RESOLVED_TOPIC_PREFIX,
+    Composition,
+    EmojiType,
+    Message,
+)
 from zulipterminal.config.keys import primary_key_for_command
 from zulipterminal.config.regexes import (
     REGEX_COLOR_3_DIGIT,
@@ -110,6 +115,7 @@ class Index(TypedDict):
     topic_msg_ids: Dict[int, Dict[str, Set[int]]]
     # Extra cached information
     edited_messages: Set[int]  # {message_id, ...}
+    moved_messages: Set[int]
     topics: Dict[int, List[str]]  # {topic names, ...}
     search: Set[int]  # {message_id, ...}
     # Downloaded message data by message id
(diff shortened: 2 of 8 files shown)


Sushmey:
Thank you for your effort @Subhasish-Behera !
The code looks well written and seems to work like it's supposed to. Are these all the flags that need checking under message update event? Also, tests normally [don't need to be mentioned](https://github.com/zulip/zulip-terminal#commit-message-style) in the commit title/summary. 
Other than that everything looks good!

tests/helper/test_helper.py
```diff
@@ -144,7 +144,6 @@ def test_index_edited_message(
     for msg_id, msg in expected_index["messages"].items():
         if msg_id in edited_msgs:
             msg["edit_history"] = []
-
```
srdeotarse, line 147: if line omission is the only change in the file, then it is not required.

zulipterminal/helper.py
```diff
@@ -47,6 +47,9 @@
 )
 
 
+RESOLVED_TOPIC_PREFIX = "✔ "
+
+
```
srdeotarse, line 52: Adding new lines just increases the no. of changes. So, try minimizing your code changes by sticking to only code additions and omissions, not the spacing.

zulipterminal/model.py
```diff
@@ -64,6 +64,8 @@
 from zulipterminal.ui_tools.utils import create_msg_box_list
 
 
+RESOLVED_TOPIC_PREFIX = "✔ "
+
```
srdeotarse, line 68: As mentioned earlier about spacing.

zulipterminal/helper.py
```diff
@@ -399,8 +404,44 @@ def index_messages(messages: List[Message], model: Any, index: Index) -> Index:
...
+                            RESOLVED_TOPIC_PREFIX
+                        ):
+                            index["moved_messages"].add(msg["id"])
+                    else:
+                        if (
+                            edit_history_event["prev_topic"].startswith(
+                                RESOLVED_TOPIC_PREFIX
+                            )
+                            and edit_history_event["prev_topic"][2:]
+                            != edit_history_event["topic"][2:]
```
srdeotarse, line 438: I think you have not handled moved message case if message is edited to resolve/unresolve topic with same name.

tests/helper/test_helper.py
srdeotarse, line 147: if line omission is the only change in the file, then it is not required.

zulipterminal/helper.py
srdeotarse, line 52: Adding new lines just increases the no. of changes. So, try minimizing your code changes by sticking to only code additions and omissions, not the spacing.

zulipterminal/model.py
srdeotarse, line 68: As mentioned earlier about spacing.

Subhasish-Behera:
> Thank you for your effort @Subhasish-Behera ! The code looks well written and seems to work like it's supposed to. Are these all the flags that need checking under message update event? Also, tests normally [don't need to be mentioned](https://github.com/zulip/zulip-terminal#commit-message-style) in the commit title/summary. Other than that everything looks good!

Changed the commit messages. 
Re message update event. ZT treats update_message_event(content/stream/topic changes at least) almost like it's a message event but with a different `response` value. I thought that way and came up with the code for `update_message_event`. I am not solving any other bug/enhancement related to `edit_history` or `update_message_event` here i.e This code only applies the `MOVED` label if in the previous version of these functions  `EDITED` label can be applied to those messages.

zulipterminal/ui_tools/messages.py
```diff
@@ -723,14 +723,24 @@ def main_view(self) -> List[Any]:
         if self.message["id"] in self.model.index["edited_messages"]:
             edited_label_size = 7
             left_padding = 1
+            is_edited = True
```
srdeotarse, line 726: Instead of ``is_edited = True``, you can replace it directly with ``label_text = "EDITED`` and remove last conditional to reduce no. of code lines while keeping the code logic intact.
Same can be done with ``is_edited = False``

srdeotarse (commented):
Thanks for working on this issue 👍. You have understood the issue and have tried to accommodate all conditions that distinguish ``MOVED`` and ``EDITED`` messages in ZT.

Just a suggestion that it would be really good if you add comments for conditionals related to (un)resolve topics for better understanding.

zulipterminal/model.py
```diff
@@ -1591,7 +1593,31 @@ def _handle_update_message_event(self, event: Event) -> None:
         indexed_message = self.index["messages"].get(message_id, None)
 
         if indexed_message:
-            self.index["edited_messages"].add(message_id)
+            if "orig_content" in event:
+                self.index["edited_messages"].add(message_id)
+            if "prev_stream" in event:
+                self.index["moved_messages"].add(message_id)
+            if "subject" in event:
+                if not event["subject"].startswith(RESOLVED_TOPIC_PREFIX):
```
neiljp, line 1601: This looks very similar to the logic in the previous commit. Once you've simplified both, are they simple enough to combine into one function?

zulipterminal/ui_tools/messages.py
neiljp: The current changes don't look like they've resolved this issue yet?

I'd also suggest moving this change into its own commit.

neiljp (commented):
@Subhasish-Behera Thanks for looking into this :)

Practically speaking this looks to work OK after a quick test :+1:

However, the logic is rather difficult to read in parts, partly due to reuse of long variables, and I'm not sure how much has come from server or web app code.

It also looks like you could potentially implement the update part using similar functionality to the first commit, though that may only be clear once the code is easier to read.

There are effectively 3 changes here (at least), and while the last commit says tests added, it seems like it's only updated some tests to pass. We really need tests for all the new behavior, and the messages code doesn't require any test updates, which suggests we don't have a pre-existing test either.

If you end up extracting the common logic between the two parts that I pointed out, testing that function may simplify testing the other parts.

Subhasish-Behera:
Now, the `analyse_edit_history` is present in https://github.com/zulip/zulip/blob/main/web/src/message_edit_history.js

neiljp:
Ready to review? If so, please switch the PR labels :)

Subhasish-Behera:
still working on changing the code to be a single function and its tests.

Subhasish-Behera:

zulipterminal/helper.py
neiljp: Apologies, I should have said that this was already defined in api_types.py - it has to be consistent with the API for resolving to be recognized.

mounilKshah (commented):
Thank you for working on this @Subhasish-Behera !

I ran the code on my local machine and it worked as expected. The tests written also seem to cover all the possible cases. I manually tried testing for all the cases and it worked perfectly fine. As for the commit structure, we can discuss it on the ZT stream.

zulipterminal/helper.py
```diff
@@ -271,6 +274,44 @@ def set_count(id_list: List[int], controller: Any, new_count: int) -> None:
...
+    content_changed: bool,
+    stream_changed: bool,
+    current_topic: Any = None,
+    old_topic: Any = None,
+) -> None:
+    resolve_change = False
+    resolved_prefix = RESOLVED_TOPIC_PREFIX + " "
+    if content_changed or stream_changed:
+        index["edited_messages"].add(msg_id)
+    elif old_topic:
```
neiljp, line 289: This also depends upon current_topic being non-None, which as it stands is not guaranteed. In other similar cases we have an assert at the top of the function to ensure that 'coordinated' values are used correctly. The alternative is to define the parameter as a data structure.

neiljp (commented):
@Subhasish-Behera I intended to only remove the needs-review label, since this needs a rebase and may have outstanding comments to address, including from Mounil here and in the stream. However, I gave this another look and left some further notes.

It's great to see the test(s) :+1: I've not read all of the main function test cases this time, and that test may become simpler if my comments result in adjustment of eg. the function signature (parameters/return/side-effects).

A point I didn't make in an inline comment, is that limiting the UI from directly accessing the model (index) data is something we should aim for. So we could add a model accessor method which returns the 'edit status' as eg. a string (a limited Literal type, perhaps), with that method being the part that interprets the edited/moved sets - and would be easier to test. We could still test the UI, but it'd be a mapping from a string<->UI, so likely less important.

Keeping that interpretation of the sets in the model may make the way in which the sets are updated, eg. whether the sets are kept disjoint, less important - the UI just gets one of 3 results via the accessor method.

I do think it's important that we resolve the current/old topic issue I mention inline, since currently it's unclear what should happen with multiple topic renames in the message history, at least with the current logic.

neiljp:
@Subhasish-Behera I've just rebased and pushed a version of this with the commit text summary split by a blank line (which you didn't have early on while contributing) and rebased to skip adding the symbol and import it instead - since that was the main conflict.

I suspect this needs a little rework due to other changes - it didn't quite pass `check-branch` locally. That's partly related to the typing of the `update_message` event, but also a model test failure seems present now.

If you do have updates locally, then feel free to push over this version, but please do fix conflicts and compare it to the changes I just pushed first.

To get this merged, I'd suggest moving the tests into the commit with the related changes, and leaving the update_message commit at the end since that's the update-during-session extension - and the one that may be failing a test now? Some of my points above were comments, but we should certainly look at the server version dependency, and ensure the tests are clear.

PR #1334, opened 2023-03-18
Highlight current narrow (rebased/simplified/extended) #1244

### LINKED ISSUE #516: Indicate active stream

neiljp:
**What does this PR do?**  

This is a rebase and minor adjustment to get #1244 working against current `main`.

This seems to work well, highlighting the stream in the left panel when in the associated narrow, and with some simple extensions also shows the top-left buttons highlighted when changing to the appropriate narrow.

This still has no tests and may benefit from refactoring - digging into the view to toggle button highlighting is not great.

A final version of something like this would fix #516, like the original.

**Tested?** 
- [x] Manually

**Notes & Questions** 

I'm particularly interested in the style per theme - this can be changed later, but we can get some reasonable values to begin with.

**Visual changes**

### DIFF
--- a/zulipterminal/config/themes.py
+++ b/zulipterminal/config/themes.py
@@ -75,6 +75,7 @@ REQUIRED_STYLES = {
     'area:error'      : 'standout',
     'area:user'       : 'standout',
     'search_error'    : 'standout',
+    'active_narrow'   : 'standout',
     'task:success'    : 'standout',
     'task:error'      : 'standout',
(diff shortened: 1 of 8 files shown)


neiljp:
I've closed the PRs leading up to this work, but note there may be useful comments in those PRs for anyone taking this forward.

PR #1350, opened 2023-03-22
refactor: seperate displaybox and simplyfing text for hotkeys(Fixed #1339)

### LINKED ISSUE #1339: Message information popup: Tidy shortcut key text & place into a group
The current shortcut/hotkey list in the message information popup (`i` on a message) could be improved by:
- moving the current hotkey lines into a separate display-group, maybe titled 'Viewing Actions'
- simplifying the text in each entry, including replacing the 'Press <key> to...' text by `[<key>]` (like we use in other parts of the UI)

All 'popups' are currently defined in `views.py`, though may be split out in future since their tests are already extracted into `test_popups.py`. See `MsgInfoView` for the relevant class.

There are currently existing display-groups for links in messages, links in topics that a message is in, time mentions, and reactions - though these are conditional upon those elements being present. Explore those in existing messages, or create some in **#test here** on chat.zulip.org (or your own Zulip server), which should give a guide of how groups are shown.

axrav:
**What does this PR do?**
This fixes #1339 

**Tested?** 
- [x] Manually

**Visual changes**

image

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -1520,13 +1520,15 @@ class MsgInfoView(PopUpView):
         date_and_time = controller.model.formatted_local_time(
             msg["timestamp"], show_seconds=True, show_year=True
         )
-        view_in_browser_keys = ", ".join(map(repr, keys_for_command("VIEW_IN_BROWSER")))
-
-        full_rendered_message_keys = ", ".join(
-            map(repr, keys_for_command("FULL_RENDERED_MESSAGE"))
+        view_in_browser_keys = "[{}]".format(
+            ", ".join(map(str, keys_for_command("VIEW_IN_BROWSER")))
         )
-        full_raw_message_keys = ", ".join(
-            map(repr, keys_for_command("FULL_RAW_MESSAGE"))
+
+        full_rendered_message_keys = "[{}]".format(
+            ", ".join(map(str, keys_for_command("FULL_RENDERED_MESSAGE")))
+        )
+        full_raw_message_keys = "[{}]".format(
+            ", ".join(map(str, keys_for_command("FULL_RAW_MESSAGE")))
         )
         msg_info = [
             (
@@ -1535,22 +1537,22 @@ class MsgInfoView(PopUpView):
                     ("Date & Time", date_and_time),
                     ("Sender", msg["sender_full_name"]),
                     ("Sender's Email ID", msg["sender_email"]),
-                    (
-                        "View message in browser",
-                        f"Press {view_in_browser_keys} to view message in browser",
-                    ),
-                    (
-                        "Full rendered message",
-                        f"Press {full_rendered_message_keys} to view",
-                    ),
-                    (
-                        "Full raw message",
-                        f"Press {full_raw_message_keys} to view",
-                    ),
                 ],
-            ),
+            )
         ]
+
+        # actions for message info popup
+        viewing_actions = (
+            "Viewing Actions",
+            [
+                ("Open in web browser", view_in_browser_keys),
+                ("Full rendered message", full_rendered_message_keys),
+                ("Full raw message", full_raw_message_keys),
+            ],
+        )
+        msg_info.append(viewing_actions)
         # Only show the 'Edit History' label for edited messages.
+
         self.show_edit_history_label = (
             self.msg["id"] in controller.model.index["edited_messages"]
             and controller.model.initial_data["realm_allow_edit_history"]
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1535,31 +1535,38 @@ def __init__(
...
                 ],
             ),
         ]
+
+        # actions for message info popup
+        viewing_actions = (
+            "Viewing Actions",
+            [
+                (
+                    "View in browser",
```
neiljp, line 1547: Since we're adjusting this, what do you think about 'Open in web browser'?

zulipterminal/ui_tools/views.py
```diff
@@ -1610,13 +1619,13 @@ def __init__(
 
             # slice_index = Number of labels before topic links + 1 newline
             #               + 1 'Topic Links' category label.
-            slice_index = len(msg_info[0][1]) + 2
+            #               + 2 for Viewing Actions category label and its newline
+            slice_index = len(msg_info[0][1]) + len(msg_info[1][1]) + 2 + 2
```
neiljp, line 1623: Why 2+2 here, but 4 above?

zulipterminal/ui_tools/views.py
```diff
@@ -1610,13 +1619,13 @@ def __init__(
...
             slice_index += sum([len(w) + 2 for w in self.button_widgets])
             self.button_widgets.append(topic_links)
 
             widgets = widgets[:slice_index] + topic_link_widgets + widgets[slice_index:]
             popup_width = max(popup_width, topic_link_width)
-
```
neiljp, line 1619: This line removal looks unrelated.

zulipterminal/ui_tools/views.py
```diff
@@ -1535,31 +1535,38 @@ def __init__(
...
+        )
+        msg_info.append(viewing_actions)
         # Only show the 'Edit History' label for edited messages.
         self.show_edit_history_label = (
             self.msg["id"] in controller.model.index["edited_messages"]
             and controller.model.initial_data["realm_allow_edit_history"]
         )
         if self.show_edit_history_label:
             msg_info[0][1][0] = ("Date & Time (Original)", date_and_time)
-
```
neiljp, line 1560: Unrelated removal?

neiljp (commented):
@axrav I'm just jumping in with a review, since I saw you'd given this a go and had some visual feedback at least :)

I've left a few notes inline, but the main visual point I'd make is that the key style doesn't yet follow that I suggested in the second bullet of the issue. For examples, see the indications of search keys and those for the narrowing actions in the top-left of the UI.

However, this already looks *much* cleaner, with less repetition :+1:

Separate from the code, could you also please adjust the commit style to match the documented style.

zulipterminal/ui_tools/views.py
axrav: yeah, that makes sense, I have made the changes

zulipterminal/ui_tools/views.py
axrav: It was a blunder by my side, has been fixed

zulipterminal/ui_tools/views.py
axrav: regarding the unrelated removal, it has been fixed for both the cases

axrav:
@neiljp I have made the changes you requested, one of the workflows job failed, but the lint issues have already been fixed, can you have a look onto it.

zulipterminal/ui_tools/views.py
```diff
@@ -1535,31 +1537,37 @@ def __init__(
...
             ),
         ]
+
+        # actions for message info popup
+        viewing_actions = (
+            "Viewing Actions",
+            [
+                (
+                    "Open in web browser",
+                    view_in_browser_keys,
```
neiljp, line 1550: If you remove the commas at the end of some of these lines, `black` may be allow them all on one line.

neiljp:
@axrav Thanks for the update; the visual change looks better :+1:

As per the inline comment, we can likely combine lines; black treats trailing commas as an instruction to not combine lines in some cases, but since the code is now shorter, it may be able to do so.

The failing test is since each commit does not pass linting and tests individually, even though the overall PR does. That occurs if you fix a linting issue in a separate commit to the commit where the change was made, for example. You can check this locally using `tools/check-branch`.

So you'll want to squash the changes into the original commit. As it stands, these commits look like local commits, rather than those easier to read and review. Please read the guidelines in the README, in particular the Structuring Commits section. Note that not all changes are always squashed into a single commit, as per those notes. Please also update the commit text for the final commit into the preferred style.

axrav:
@neiljp if everything looks good, can you merge this up, so I ll be able to assign myself new issues.

neiljp:
@axrav Thanks for working on this - This does look much better :+1:

I wanted to correct the commit title/body first, hence the minor delay, but I'm merging now 🎉 

Please take a look at the merged commit text for how it differs from your pushed one - the less tidying at merging, the faster everything is :)

PR #1134, opened 2021-09-07
Add support for copying code snippets

kingjuno:
**What does this PR do?**  
This adds a feature to copy code snippets as mentioned in [#1123 ](https://github.com/zulip/zulip-terminal/issues/1123)

copy_code_snippet

Work in progress
**Tested?** 
- [ ] Manually
- [ ] Existing tests (adapted, if necessary)
- [ ] New tests added (for any new behavior)
- [ ] Passed linting & tests (each commit)

Issue #1123, opened 2021-08-20
Add support for copying code snippets


neiljp:
Often it's useful to extract code - or maybe any quoted text? - from a message.

The likely initial way to implement this would be to ultimately add an extra (optional) section to the message information popup. However, as with in-message links or topic links, this feature requires solid implementation of extraction of code blocks first into a format the popup can show.

Things to consider:
* each code section may have a language (eg. bash, python), which it would be good to indicate in an entry
* we don't want to list the full section in the message information popup, just enough to be unique
* in the short term, activating the entry for the code could copy it
* in the longer term we may want to extend the copy feature to other links too, and activation could instead trigger a feature like #985

**NOTE: I've marked this as a good first issue, but this may require a deeper dive into multiple areas!**

zulipbot:
Hello @kingjuno, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

avats-dev:
@zulipbot claim

zulipbot:
Welcome to Zulip, @avats-dev! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

kingjuno:
@avats-dev this issue is already somewhat half-implemented except for the test.

avats-dev:
@kingjuno yeah sure. Sorry, I thought it was unaddressed. 
@zulipbot abandon

zulipbot:
**ERROR:** You have not claimed this issue to work on yet.

[...]

PR #1353, opened 2023-03-24
buttons/messages/views: Add support for copying code snippets.

yuktasarode:
**What does this PR do?**  
This PR aims to add support for copying code snippets and is based off on the work done by @kingjuno and @neiljp in the PR #1134. This PR mainly fixes the tests and linting issues and bring the PR to a merge-able state. 

**Tested?** 
- [x] Manually

This may fix the core elements of #1123.

### DIFF
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -647,6 +647,59 @@ class MessageLinkButton(urwid.Button):
                 self.controller.exit_popup()
 
 
+class CodeSnippetButton(urwid.Button):
+    def __init__(
+        self,
+        *,
+        controller: Any,
+        caption: str,
+        display_code: List[Tuple[str, str]],
+        copy_code: str,
+        display_attr: Optional[str],
+    ) -> None:
+        self.controller = controller
+        self.model = self.controller.model
+        self.view = self.controller.view
+        self.caption = caption
+        self.display_code = display_code
+        self.copy_code = copy_code
+
+        super().__init__("")
+
+        self.update_widget(display_code, display_attr)
+        urwid.connect_signal(self, "click", self.copy_to_clipboard)
+
+    def update_widget(
+        self, display_code: List[Tuple[str, str]], display_attr: Optional[str] = None
+    ) -> None:
+        """
+        Overrides the existing button widget for custom styling.
+        """
+        # Set cursor position next to len(caption) to avoid the cursor.
+        icon = urwid.SelectableIcon(display_code, cursor_position=len(display_code) + 1)
+        self._w = urwid.AttrMap(icon, display_attr, focus_map="selected")
+
+    def copy_to_clipboard(self, *_: Any) -> None:
+        self.controller.copy_to_clipboard(self.copy_code, "code")
+
+    def get_code_from_snippet(
+        self, snippet_list: List[Tuple[str, str]]
+    ) -> Tuple[List[Tuple[str, str]], str]:
+        """
+        Returns two lines of code from the given snippet list
+        as to summarize the code and
+        returns code in text format to be copied.
+        """
+        snippet = snippet_list[:]
+        no_of_lines = 0
+        code = "".join(snip[1] for snip in snippet)
+        for index, snip in enumerate(snippet):
+            no_of_lines += 1 if "\n" in snip[1] else 0
+            if no_of_lines == 2:
+                return snippet[:index] + [("pygments:w", "...")], code
+        return snippet, code
+
+
 class EditModeButton(urwid.Button):
     def __init__(self, *, controller: Any, width: int) -> None:
         self.controller = controller
--- a/zulipterminal/ui_tools/messages.py
+++ b/zulipterminal/ui_tools/messages.py
@@ -57,6 +57,7 @@ class MessageBox(urwid.Pile):
         self.user_id: Optional[int] = None
         self.message_links: Dict[str, Tuple[str, int, bool]] = dict()
         self.topic_links: Dict[str, Tuple[str, int, bool]] = dict()
+        self.code_snippets: List[Tuple[str, List[Tuple[str, str]]]] = list()
         self.time_mentions: List[Tuple[str, str]] = list()
         self.last_message = last_message
         # if this is the first message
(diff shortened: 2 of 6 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1646,12 +1665,68 @@ def create_link_buttons(
...
+                #     display_code[-1][-1] = (
+                #         display_code[-1][-1][:-1]
+                #         if display_code[-1][-1][-1] == "\n"
+                #         else display_code
+                #     )
+                #     display_code[-1] = tuple(display_code[-1])
+            code = f"{str(index+1)}: {language}\n"
+            button = CodeSnippetButton([("pygments:w", code)] + display_code)
+            urwid.connect_signal(button, "click", self.copy_to_clipboard, copy_code)
+            display_attr = None if index % 2 else "popup_contrast"
```
neiljp, line 1693: It would be cleaner if we could have this kind of behavior, and that in a few other methods below, defined in the CodeSnippetButton itself instead, like we do with other buttons - I can't see a reason not to do this.

Note how the create_link_button function is simpler compared to this one.

The CodeSnippetButton does very little that's unique right now - it's very close to being a very simple Button.

zulipterminal/ui_tools/views.py
```diff
@@ -1646,12 +1665,68 @@ def create_link_buttons(
...
+                    code_snippet_width, len(max(display_code, key=len))
+                )
+                # if language == "":
+                #     display_code[-1] = list(display_code[-1])
+                #     display_code[-1][-1] = (
+                #         display_code[-1][-1][:-1]
+                #         if display_code[-1][-1][-1] == "\n"
+                #         else display_code
+                #     )
+                #     display_code[-1] = tuple(display_code[-1])
```
neiljp, line 1689: We should remove commented code from before; you can split this into a later commit, in case we want to refer to it for a later feature, or remove it and we can refer to the original PR instead.

zulipterminal/ui_tools/messages.py
```diff
@@ -551,8 +560,16 @@ def soup2markup(
...
                         css_style = code_element.attrs.get("class", ["w"])
                         markup.append((f"pygments:{css_style[0]}", code_text))
+                        code_snippet.append((f"pygments:{css_style[0]}", code_text))
                     else:
                         markup.append(("pygments:w", code_text))
+                        code_snippet.append(("pygments:w", code_text))
+                try:
+                    code_language = element.attrs["data-code-language"]
+                except KeyError:
+                    code_language = None
```
neiljp, line 570: We can simplify this, since attrs is essentially a dict, using .get.

zulipterminal/ui_tools/buttons.py
```diff
@@ -647,6 +647,40 @@ def handle_narrow_link(self) -> None:
                 self.controller.exit_popup()
 
 
+class CodeSnippetButton(urwid.Text):
```
neiljp, line 650: As mentioned elsewhere, this works, but it would be good to migrate to a button with the specific functionality we want.

neiljp (commented):
@yuktasarode Thanks for tidying this up further! :+1:

It's good to see this passing linting and tests :) It's also working well still, though I've not examined the functionality in great detail yet.

I've left some comments inline, but you may want to first rebase against `main`, since I recently merged a PR regarding how the 'Action' keys are shown and grouped, which affects this UI.

My note in the other PR (https://github.com/zulip/zulip-terminal/pull/1134#issuecomment-1472479750) mention a few other changes; I'm not sure if you're already looking at those aspects, but I've left some details below.

I'd suggest the next step to make this easier to work on and read would be to split this into multiple commits. You'll want to confirm how the code links together yourself, but my take would be that we're generating additional data via the message, which will need to go into a button, both of which are combined into the end result - so 3 commits might be a good starting point.

Moving the behavior into the CodeSnippetButton may be another good step, or at least inheriting via `urwid.Button` like the MessageLinkButton does in the first instance, since I believe it should achieve most of what the custom button does already?

My note in the other PR (https://github.com/zulip/zulip-terminal/pull/1134#issuecomment-1472479750) also notes regarding adding tests for each commit/change; if you wish to start on that, that's fine, but I expect it'll be a lot easier to keep track of when the commit size is smaller, and you can separate which code change/addition is related to which test change/addition.

zulipterminal/ui_tools/views.py
yuktasarode: I have deleted the commented code.

neiljp:
@yuktasarode Is this ready for review? If so, please change the labels as per my message in the GSoC 2023 topic in zulip-terminal (ie. remove needing update and add needs review).

yuktasarode:

yuktasarode:

zulipterminal/ui_tools/buttons.py
yuktasarode: I did trying inherting urwid.Button and Moving the behavior into the CodeSnippetButton but I could only move the method get_code_from_snippet() from views.py to buttons.py, since all other methods require the instance of controller to be taken into account. Kindly let me know anyother way to do this.

neiljp: Right now, all the basic Button functionality currently in `CodeSnippetButton` except for `get_code_from_snippet` is as per the default in `urwid.Button` (you can compare yourself, to `wimp.py` in urwid).

Look at the initializer for `MessageLinkButton` - that is how we can set up a controller via the button. We may also want to store the code content (snippet information) in there, for example.

Right now you're using the class name to call a method in a class, and explicitly passing the `self` value. Normally that'd be a bad idea and often cause an error, but you're not using `self` in `get_code_from_snippet` here! Instead, we should be creating `CodeSnippetButton`s with the relevant data, then using their properties as we need them.

I understand you've inherited this PR, but this (and the UI, using urwid) is using OOP in Python, so you may need to learn more about this if you've not worked on this before.

neiljp:
@yuktasarode Could you please rebase this after the recent changes we had in `main`, and further explore how you might simplify and initialize the snippet button like we do with the `MessageLinkButton`? (given my comment above)

You're welcome to take up another issue also, of course :+1:

neiljp (commented):
@yuktasarode You didn't note this for review, but I saw you had pushed. This looks much closer than before :tada:

PR #1365, opened 2023-03-30
views/keys/hotkeys: Add key to copy message content

### LINKED ISSUE #546: Message copy[/paste]
At least within my terminal, copying a message using terminal selection does not work across multiple lines, as it also selects text in one or both of the side bars.

While we could work around this temporarily using autohide, it would be useful to be able to copy selected message(s) independently of how the UI is set up.

This could be for quoting in another stream, or external to ZT.

A feature like this has also been discussed for the webapp, so there may be reference on chat.zulip.org to this, particularly with respect to copying multiple messages (which could be a followup)

This is sort of a UI 'export', so we may want to explore different targets/formats, to include at least some kind of internal message buffer/reference and the native copy/paste buffer.

Sushmey:
Copying a message using terminal selection does not work across multiple lines, as it also selects text in one or both of the side bars.

It would be useful to be able to copy selected message(s) independently of how the UI is set up.
This could be for quoting in another stream, or external to ZT.

Adding a key to copy message content which can be then pasted. Implemented using preexisting `copy_to_clipboard` function.

### Outstanding aspect(s)    

This seems to not work for quotes messages because of the nested list

Unsure as to how to navigate about specific cases like 

- [ ] Fully fixes #
- [x] Partially fixes issue #546 
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

[asciicast](https://asciinema.org/a/HkVCgXI6U6FCbWDf5OY27H4XM?t=10)

### DIFF
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -396,6 +396,12 @@ KEY_BINDINGS: 'OrderedDict[str, KeyBinding]' = OrderedDict([
         'help_text': 'Show/hide full raw message (from message information)',
         'key_category': 'msg_actions',
     }),
+    ('COPY_MESSAGE', {
+        'keys': ['C'],
+        'help_text':
+            'Copy message content to clipboard',
+        'key_category': 'msg_actions',
+    }),
 ])
 # fmt: on
 
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -1530,6 +1530,9 @@ class MsgInfoView(PopUpView):
         full_raw_message_keys = "[{}]".format(
             ", ".join(map(str, keys_for_command("FULL_RAW_MESSAGE")))
         )
+        copy_message_keys = "[{}]".format(
+            ", ".join(map(str, keys_for_command("COPY_MESSAGE")))
+        )
         msg_info = [
             (
                 "",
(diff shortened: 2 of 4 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1678,6 +1682,17 @@ def keypress(self, size: urwid_Size, key: str) -> str:
...
+            rendered_content, *_ = MessageBox.transform_content(
+                self.msg["content"], self.controller.model.server_url
+            )
+            content = []
+            for word in rendered_content[1]:
+                if isinstance(word, tuple):  # if msg content has markup
+                    content.append(word[1])
+                else:
+                    content.append(word)
+            self.controller.copy_to_clipboard("".join(content), "Message Content")
```
neiljp, line 1695: The principle here is just fine, but we do perform a few extra adjustments to content which may not follow the principles that you assume here.

That's likely what gave rise to the crash I reported in the stream.

zulipterminal/config/keys.py
```diff
@@ -396,6 +396,12 @@ class KeyBinding(TypedDict):
         'help_text': 'Show/hide full raw message (from message information)',
         'key_category': 'msg_actions',
     }),
+    ('COPY_MESSAGE', {
+        'keys': ['C'],
+        'help_text':
+            'Copy message content to clipboard',
+        'key_category': 'msg_actions',
```
neiljp, line 403: We've discussed this in the stream, but
- `c` should not need capitalization, with some adjustment, if we stick with that hotkey, though key choice right now is not important
  - though note that we already have `c` for copying stream email address
  - plus we discussed something similar for the copy non-markdown content version
- the help_text likely doesn't need to wrap? That's only been done for long lines
- we may want the COPY_MESSAGE to be more meaningful if we have two versions :)

neiljp (commented):
@Sushmey We've discussed this in the stream but I wanted to just read over what you'd done so far - though I understand the approach may change, things look fairly reasonable except for the extraction of the content :+1:

Issue #1358, opened 2023-03-26
Support multi-word matching when autocompleting topics


neiljp:
When autocompleting other items such as streams, we check the search term against different 'words' in the possible matches. For example, `doc` can match both `documentation` and `api documentation`, `cod` can match `code review` and `live coding`, and `term` can match both `terminal clients` and `zulip-terminal`.

For topics, we do not do so, so that `PR ` only matches topics starting with exactly those characters, not any later 'word'. You should be able to confirm this on chat.zulip.org using zulip terminal.

We should extend the current topic autocompletion logic to behave similarly to that for streams, to handle these kinds of cases - likely with the same ordering as with streams, though that may need confirming since there could be a *lot* of topics compared to streams, and we currently only show up to 10 matches. The web app logic may be useful for comparison.

The UI logic for this is in `boxes.py`, so that's a good place to start. The logic is likely common between the topic searches, but extra test cases to ensure this is working would be important. However, if the matching logic can be completely unified with topics then we may be able to remove some tests later.

AmeliaTaihui:
@zulipbot claim

zulipbot:
Welcome to Zulip, @AmeliaTaihui! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

ifarabi:
@zulipbot claim

zulipbot:
@ifarabi This issue cannot be claimed, as someone else is already working on it. Please see our [contributor guide](https://zulip.readthedocs.io/en/latest/overview/contributing.html#your-first-codebase-contribution) for advice on finding an issue to work on. Thanks!

zulipbot:
[...]

PR #1371, opened 2023-04-08
test_boxes.py/helper.py: support and test multi-word matching without any ordering algorithm

AmeliaTaihui:
This PR supports multiword match in topic search and created and changes a few test cases to support it.

### Outstanding aspect(s)    
 

 

- [ ] not sure if we should prioritize words if they appear earlier
- [ ] it does not have a decent ordering algorithm yet.

- [x] Discussed in **#zulip-terminal** in `Multi-word matching for topic autocomplete #T1358 #T1371`
- [ ] Fully fixes #
- [x] Partially fixes issue #1358
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -96,10 +96,18 @@ class TestWriteBox:
         [
             ("#**Stream 1>T", 0, True, "#**Stream 1>Topic 1**"),
             ("#**Stream 1>T", 1, True, "#**Stream 1>This is a topic**"),
-            ("#**Stream 1>T", 2, True, None),
-            ("#**Stream 1>T", -1, True, "#**Stream 1>This is a topic**"),
-            ("#**Stream 1>T", -2, True, "#**Stream 1>Topic 1**"),
-            ("#**Stream 1>T", -3, True, None),
+            # ("#**Stream 1>T", 2, True, None),
+            # changed test
+            ("#**Stream 1>T", 2, True, "#**Stream 1>Hello there!**"),
+            # ("#**Stream 1>T", -1, True, "#**Stream 1>This is a topic**"),
+            # changed test
+            ("#**Stream 1>T", -1, True, "#**Stream 1>Hello there!**"),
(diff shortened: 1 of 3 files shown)


mounilKshah (commented):
Thank you @AmeliaTaihui for working on this. I tried running this on my local machine and it worked well. Even the extra test case that you have added in `test_boxes` looks fine to me.

You can change the commit title and add a short message as per the [commit message guidelines](https://github.com/zulip/zulip-terminal#commit-message-style) in the README.

neiljp (commented):
@AmeliaTaihui Thanks for working on this - it works better than what we have so far :+1:

More specifically, this is an improvement in matching a single search word to one of multiple later target words. See below the line for extensions from this, namely multiple search words to other target words, and thinking of relative ordering.

I've left some comments inline, though haven't done a full review of the tests to think of cases that aren't currently covered. What you have looks good, though we'll likely want to add some other test cases to handle the more complex searches that are now possible.

I expanded your PR questions into the checkboxes, which I hope was accurate?

Regarding if we should likely prioritize search words mentioned earlier, I think we need to ensure that we continue to factor in some element of recent topics being prioritized, which we might want to discuss further in the stream.

---
Another factor, which I'm not sure if we should focus on addressing in this PR, is whether we should care about handling multiple search words better.

This already appears to work for streams in ZT - eg. `api d` in the stream compose box or after `#` in the message body. However, while `support custom` does find the correct topic in the topic box, `custom profile` does not - even with this commit. Similarly, `#T1358 #T1371` appears to give no results, while `#T1358` does so - both near the end of the topic name.

Other than multiple search words later in the topic name, we could also consider the relative ordering of the words in each case. The web app appears to care about ordering - a good example of this is that as with this PR, it allows one to search for `#T1371`. However, `#T1371 #T1358` gives no results in the web app, since the words are the opposite order in the topic name!

neiljp:
This was replaced by #1378 and ultimately #1385 with broadly the same functionality :+1:

PR #1374, opened 2023-04-08
Load and show custom profile fields in user information popup

### LINKED ISSUE #1338: Load, update and show Custom Profile Fields in user information popup
Our current user information popup (`i` on user list) shows a lot of information, but no [custom profile fields](https://docs.zulip.com/help/custom-profile-fields).

This work likely comprises at least 3 steps, each of which should be accompanied by test updates and new tests.

Firstly, the definition of the current fields can be obtained when we first connect to the server; if they are not already in the Model initial_data, then see the API docs for how these can be added by adjusting the data we initially request: https://docs.zulip.com/api/register-queue#response

We also should add an event handler for when these are updated, to ensure they stay up to date in the running application; see the Model for other event handlers, the api_types.py file, and the API docs for the event: https://docs.zulip.com/api/get-events#custom_profile_fields

Once the data is available and updating, we can update the Model method which processes user data into appropriate fields, which is called by the user-information popup to show the user information.

theViz343:
This PR adds custom profile fields to the user info popup, which previously only displayed basic user information. This will include information about the user which, currently, has to be accessed by using the webapp. This PR also adds update event handling for the custom profile data.

### Outstanding aspect(s)    

- [x] Update field values in current session using events.
- [x] Add/update tests.
- [x] Sort fields using 'order'
- [x] Simplify `_clean_custom_profile_data`

- [x] Discussed in **#zulip-terminal** in [`Support custom profile fields #T1338`](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Support.20custom.20profile.20fields.20.23T1338)
- [x] Fully fixes #1338
- [ ] Partially fixes issue 
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,13 +6,18 @@ import pytest
 from pytest_mock import MockerFixture
 from urwid import Widget
 
-from zulipterminal.api_types import Message, MessageType
+from zulipterminal.api_types import (
+    CustomFieldValue,
+    CustomProfileField,
+    Message,
+    MessageType,
+)
 from zulipterminal.config.keys import (
     ZT_TO_URWID_CMD_MAPPING,
     keys_for_command,
     primary_key_for_command,
 )
-from zulipterminal.helper import Index, TidiedUserInfo
+from zulipterminal.helper import CustomProfileData, Index, TidiedUserInfo
 from zulipterminal.helper import initial_index as helper_initial_index
 from zulipterminal.ui_tools.buttons import StreamButton, TopicButton, UserButton
 from zulipterminal.ui_tools.messages import MessageBox
@@ -624,6 +629,224 @@ def mentioned_messages_combination(request: Any) -> Tuple[Set[int], Set[int]]:
     Returns a combination of mentioned and wildcard_mentioned messages
     """
     return deepcopy(request.param)
+
+
+@pytest.fixture
+def custom_profile_fields_fixture() -> List[CustomProfileField]:
+    return [
+        {
+            # Short text: For one line responses, like "Job title".
+            # Responses are limited to 50 characters.
+            "id": 1,
+            "name": "Phone number",
+            "type": 1,
+            "hint": "",
+            "field_data": "",
+            "order": 1,
+        },
+        {
+            # Long text: For multiline responses, like "Biography".
+            "id": 2,
+            "name": "Biography",
+            "type": 2,
+            "hint": "What are you known for?",
+            "field_data": "",
+            "order": 2,
+        },
+        {
+            # Another example of Short text field.
+            "id": 3,
+            "name": "Favorite food",
+            "type": 1,
+            "hint": "Or drink, if you'd prefer",
+            "field_data": "",
+            "order": 3,
+        },
+        {
+            # List of options: Creates a dropdown with a list of options.
+            "id": 4,
+            "name": "Favorite editor",
+            "type": 3,
+            "hint": "",
+            "field_data": (
+                '{"0":{"text":"Vim","order":"1"},"1":{"text":"Emacs","order":"2"}}'
+            ),
+            "order": 4,
+        },
+        {
+            # Date picker: For dates, like "Birthday".
+            "id": 5,
+            "name": "Birthday",
+            "type": 4,
+            "hint": "",
+            "field_data": "",
+            "order": 5,
+        },
+        {
+            # Link: For links to websites.
+            "id": 6,
+            "name": "Favorite website",
+            "type": 5,
+            "hint": "Or your personal blog's URL",
+            "field_data": "",
+            "order": 6,
+        },
+        {
+            # Person picker: For selecting one or more users,
+            # like "Manager" or "Direct reports".
+            "id": 7,
+            "name": "Manager",
+            "type": 6,
+            "hint": "Only one person",
+            "field_data": "",
+            "order": 7,
+        },
+        {
+            # Another person picker: Use case with multiple users.
+            "id": 8,
+            "name": "Mentor",
+            "type": 6,
+            "hint": "",
+            "field_data": "",
+            "order": 8,
+        },
+        {
+            # External account: For linking to GitHub, Twitter, etc.
+            # GitHub example
+            "id": 9,
+            "name": "GitHub username",
+            "type": 7,
+            "hint": "",
+            "field_data": '{"subtype":"github"}',
+            "order": 9,
+        },
+        {
+            # Twitter example
+            "id": 10,
+            "name": "Twitter username",
+            "type": 7,
+            "hint": "",
+            "field_data": '{"subtype":"twitter"}',
+            "order": 10,
+        },
+        {
+            # Custom example
+            "id": 11,
+            "name": "Reddit username",
+            "type": 7,
+            "hint": "",
+            "field_data": '{"subtype":"custom", "url_pattern":"https://www.reddit.com/u/%(username)s"}',
+            "order": 11,
+        },
+        {
+            # Pronouns: What pronouns should people use to refer to the user?
+            "id": 12,
+            "name": "Pronouns",
+            "type": 8,
+            "hint": "What pronouns should people use to refer to you?",
+            "field_data": "",
+            "order": 12,
+        },
+    ]
+
+
+@pytest.fixture
+def custom_profile_data_fixture() -> Dict[str, CustomFieldValue]:
+    return {
+        "1": {"value": "6352813452", "rendered_value": "<p>6352813452</p>"},
+        "2": {
+            "value": "Simplicity\nThis is a multiline field",
+            "rendered_value": "<p>Simplicity\nThis is a multiline field</p>",
+        },
+        "3": {"value": "cola", "rendered_value": "<p>cola</p>"},
+        "4": {"value": "0"},
+        "5": {"value": "2023-04-22"},
+        "6": {"value": "https://www.google.com"},
+        "7": {"value": "[11]"},
+        "8": {"value": "[11, 13]"},
+        "9": {"value": "gitmaster"},
+        "10": {"value": "twittermaster"},
+        "11": {"value": "redditmaster"},
+        "12": {"value": "he/him"},
+    }
+
+
+@pytest.fixture
+def clean_custom_profile_data_fixture() -> List[CustomProfileData]:
+    return [
+        {
+            "label": "Phone number",
+            "value": "6352813452",
+            "type": 1,
+            "order": 1,
+        },
+        {
+            "label": "Biography",
+            "value": "Simplicity\nThis is a multiline field",
+            "type": 2,
+            "order": 2,
+        },
+        {
+            "label": "Favorite food",
+            "value": "cola",
+            "type": 1,
+            "order": 3,
+        },
+        {
+            "label": "Favorite editor",
+            "value": "Vim",
+            "type": 3,
+            "order": 4,
+        },
+        {
+            "label": "Birthday",
+            "value": "2023-04-22",
+            "type": 4,
+            "order": 5,
+        },
+        {
+            "label": "Favorite website",
+            "value": "https://www.google.com",
+            "type": 5,
+            "order": 6,
+        },
+        {
+            "label": "Manager",
+            "value": [11],
+            "type": 6,
+            "order": 7,
+        },
+        {
+            "label": "Mentor",
+            "value": [11, 13],
+            "type": 6,
+            "order": 8,
+        },
+        {
+            "label": "GitHub username",
+            "value": "https://github.com/gitmaster",
+            "type": 7,
+            "order": 9,
+        },
+        {
+            "label": "Twitter username",
+            "value": "https://twitter.com/twittermaster",
+            "type": 7,
+            "order": 10,
+        },
+        {
+            "label": "Reddit username",
+            "value": "https://www.reddit.com/u/redditmaster",
+            "type": 7,
+            "order": 11,
+        },
+        {
+            "label": "Pronouns",
+            "value": "he/him",
+            "type": 8,
+            "order": 12,
+        },
+    ]
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -125,7 +125,10 @@ def msg_box(
 
 
 @pytest.fixture
-def users_fixture(logged_on_user: Dict[str, Any]) -> List[Dict[str, Any]]:
+def users_fixture(
+    logged_on_user: Dict[str, Any],
+    custom_profile_data_fixture: Dict[str, CustomFieldValue],
+) -> List[Dict[str, Any]]:
     users = [logged_on_user]
     for i in range(1, 3):
         users.append(
@@ -154,11 +157,19 @@ def users_fixture(logged_on_user: Dict[str, Any]) -> List[Dict[str, Any]]:
                 "is_admin": False,
             }
         )
+    # Add custom profile data to user 12
+    for user in users:
+        if user["user_id"] == 12:
+            user["profile_data"] = custom_profile_data_fixture
+        else:
+            user["profile_data"] = {}
     return users
 
 
 @pytest.fixture
-def tidied_user_info_response() -> TidiedUserInfo:
+def tidied_user_info_response(
+    clean_custom_profile_data_fixture: List[CustomProfileData],
+) -> TidiedUserInfo:
     # FIXME: Refactor this to use a more generic user?
     return {
         "full_name": "Human 2",
@@ -170,6 +181,7 @@ def tidied_user_info_response() -> TidiedUserInfo:
         "bot_type": None,
         "bot_owner_name": "",
         "last_active": "",
+        "custom_profile_data": clean_custom_profile_data_fixture,
     }
 
 
@@ -855,6 +867,7 @@ def initial_data(
     users_fixture: List[Dict[str, Any]],
     streams_fixture: List[Dict[str, Any]],
     realm_emojis: Dict[str, Dict[str, Any]],
+    custom_profile_fields_fixture: List[Dict[str, Union[str, int]]],
 ) -> Dict[str, Any]:
     """
     Response from /register API request.
@@ -1029,6 +1042,7 @@ def initial_data(
         "zulip_version": MINIMUM_SUPPORTED_SERVER_VERSION[0],
         "zulip_feature_level": MINIMUM_SUPPORTED_SERVER_VERSION[1],
         "starred_messages": [1117554, 1117558, 1117574],
+        "custom_profile_fields": custom_profile_fields_fixture,
     }
 
 
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1570,6 +1570,19 @@ class TestModel:
         }
         assert user_group_names == ["Group 1", "Group 2", "Group 3", "Group 4"]
 
+    def test__clean_and_order_custom_profile_data(
+        self,
+        model,
+        custom_profile_fields_fixture,
+        custom_profile_data_fixture,
+        clean_custom_profile_data_fixture,
+    ):
+        model.initial_data["custom_profile_fields"] = custom_profile_fields_fixture
+        assert (
+            clean_custom_profile_data_fixture
+            == model._clean_and_order_custom_profile_data(custom_profile_data_fixture)
+        )
+
     @pytest.mark.parametrize(
         ["to_vary_in_each_user", "key", "expected_value"],
         [
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1,3 +1,4 @@
+import copy
 import json
 from collections import OrderedDict
 from copy import deepcopy
@@ -3614,7 +3615,7 @@ class TestModel:
             "delivery_email",
         ],
     )
-    def test__handle_realm_user_event(
+    def test__handle_realm_user_event__general(
         self, person, event_field, updated_field_if_different, model, initial_data
     ):
         # id 11 matches initial_data["realm_users"][1] in the initial_data fixture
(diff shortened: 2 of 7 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1165,6 +1165,10 @@ def _fetch_user_data(controller: Any, user_id: int) -> Dict[str, Any]:
             )
             display_data["Local time"] = local_time[11:]
 
+        if data["custom_profile_data"]:
+            for field in data["custom_profile_data"]:
+                display_data[field] = data["custom_profile_data"][field]
```
neiljp, line 1170: I know you mentioned that the last-online field is useful to have last, this place isn't the best to accomplish this.

This function is concerned with preparing the TidiedUserInfo for output, so we don't need to enforce the ordering in this function. A more general response would be to return the regular data and this data as a tuple. A simple way to group the custom profile data would then be to put it into a group like we do with the actions/links/reactions in the message info popup.

zulipterminal/model.py
```diff
@@ -979,6 +1017,10 @@ def get_user_info(self, user_id: int) -> Optional[TidiedUserInfo]:
         else:
             user_role = raw_user_role
 
+        custom_profile_data = api_user_data.get("profile_data", {})
+        cleaned_custom_profile_data = self._clean_custom_profile_data(
+            custom_profile_data
+        )
```
neiljp, line 1023: This change and the TidiedUserInfo change in an earlier commit can go together? The big new function above should go into an earlier commit, with tests.

I understand this works for getting things working :)

zulipterminal/model.py
```diff
@@ -960,6 +961,43 @@ def get_other_subscribers_in_stream(
                 if sub != self.user_id
             ]
 
+    def _clean_custom_profile_data(
+        self, custom_profile_data: Dict[str, Dict[str, str]]
+    ) -> Dict[str, str]:
```
neiljp, line 966: As per another comment, this model function would be concerned with cleaning the data, not necessarily making it into just an output form. For example, if I want a list of users, it would be simpler to just end up with a list of ids - and this function may become simpler :)

That may mean defining additional TypedDicts for use with TidiedUserInfo and this function would be helpful.

zulipterminal/model.py
```diff
@@ -960,6 +961,43 @@ def get_other_subscribers_in_stream(
...
+                field_options = json.loads(profile_fields[field_id]["field_data"])
+                field_subtype = field_options["subtype"]
+                if field_subtype == "github":
+                    field_value = "https://github.com/" + value
+                elif field_subtype == "twitter":
+                    field_value = "https://twitter.com/" + value
+                else:
+                    field_value = value
+            else:
+                field_value = value
```
neiljp, line 997: Are 3/6/7 the only ones needing special handling? This will be easier to observe with tests.

neiljp (commented):
@theViz343 Thanks for looking at this :+1:

This works fairly well as a v1 already :) However, given the variation in inputs then as you might expect, I'd like to see tests!

Generally the code could be a little cleaner and easier to read, though I suspect that when the functions are adjusted and tested, some of that may resolve itself.

Other than the updates to user data and server fields, this doesn't use `order` yet - I suspect czo uses that for some fields, and might affect how we show the output ourselves.

zulipterminal/model.py
theViz343: Yes, from manual testing, it seems so. I'm currently working on adding tests.

zulipterminal/model.py
```diff
@@ -960,6 +961,43 @@ def get_other_subscribers_in_stream(
...
+                        user_name = user_data.get("full_name", "(No name)")
+                        users.append(user_name)
+                field_value = ", ".join(users)
+            elif field_type == 7:
+                field_options = json.loads(profile_fields[field_id]["field_data"])
+                field_subtype = field_options["subtype"]
+                if field_subtype == "github":
+                    field_value = "https://github.com/" + value
+                elif field_subtype == "twitter":
+                    field_value = "https://twitter.com/" + value
```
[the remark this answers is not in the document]
theViz343: Yes, for the `external account` field, the web app has three subtypes by default: `github`, `twitter` and `custom`. The `github` and `twitter` subtypes require just the user name and it appends the appropriate prefix while displaying. The `custom` sub type requires a URL pattern while creating the field. This pattern then gets prefixed to the username a user enters. I've yet to handle the `custom` case.

zulipterminal/model.py
theViz343: Should I keep this function and just use it to clean the data, and maybe create another function where I process it to get display information? 
Like the example that you gave here - when this function cleans the `list of persons` field, it then produces a list of ids in it's place. Then, a second function would fetch the names of the users with the user ids in the list, and create a string (concatenate the names together with commas) for display. 
Did I understand you correctly?

neiljp: Yes, that's what I meant. There is a function which already does this - to some extent - in UserInfoView.

theViz343:

theViz343:

tests/conftest.py
```diff
@@ -619,6 +619,160 @@ def mentioned_messages_combination(request: Any) -> Tuple[Set[int], Set[int]]:
...
+            "field_data": "",
+            "order": 3,
+        },
+        {
+            "id": 4,
+            "name": "Favorite editor",
+            "type": 3,
+            "hint": "",
+            "field_data": '{"0":{"text":"Vim","order":"1"},\
+            "1":{"text":"Emacs","order":"2"}}',
```
neiljp, line 666: What is the source of these? Is the `\` character for a line break, or in the raw data?

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: These seem OK to include as prefixes, though the webapp treats them as special links - did you do a prefix to ensure the link eg. highlights in terminals for easy copy/paste?

neiljp (commented):
@theViz343 This is a partial review, since there's now a lot here!

The ordering of the new fields in the new category looks great and it's good to see the tests too :+1:

I think I like the extra UI section, though we can discuss the title - `Additional details`? Optional? Web/mobile avoid this issue.

Adjusting the balance between processing in the model and view method should simplify the code again.

Some inline feedback is code tidying, so not as much to be concerned with just some advice.

tests/conftest.py
theViz343: It's for a line break, to keep the linter from complaining.

zulipterminal/model.py
```diff
@@ -962,6 +964,39 @@ def get_other_subscribers_in_stream(
                 if sub != self.user_id
             ]
 
+    def _clean_custom_profile_data(
+        self, custom_profile_data: Dict[str, CustomFieldValue]
+    ) -> Dict[str, CustomProfileData]:
```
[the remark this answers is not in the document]
theViz343: This could be changed to return a list. This would also make it easier to keep it sorted according to order. Also, we could remove the field_id since we don't use it anymore. I've made the change

theViz343:

zulipterminal/model.py
```diff
@@ -1907,6 +1907,15 @@ def _handle_realm_user_event(self, event: Event) -> None:
                     # realm_users has 'email' attribute and not 'new_email'
                     if "new_email" in updated_details:
                         realm_user["email"] = updated_details["new_email"]
+                    elif "custom_profile_field" in updated_details:
```
neiljp, line 1910: This isn't accompanied by a change to the API, which was very confusing initially as mypy isn't complaining! I feel this is really a bug and lack of coverage in mypy, but the underlying issue appears to be since the TypedDict is defined in such a way that the keys are *definitely* present since it is 'total', so this branch isn't checked at all by mypy.

We have Message as total=False, but ideally we'd avoid that since the type-checking isn't as good. Would you be up for improving the typing of that?

neiljp (commented):
@theViz343 I've not checked as much as I'd like yet, but as per my comment in the stream this is looking great!

If anything, we could extract the updating into another PR, but the core data extraction and UI output (ie. 3 of the commits) seem mostly fine.

I'm mostly concerned by the event handling, partially due to not having the chance to manually check it yet, but also due to the type issue I mentioned.

If you do want to split, you could open another PR which builds upon this one for the updating part, but that's not strictly necessary right now - we can always do that later, as long as the updating is working :)

zulipterminal/ui_tools/views.py
```diff
@@ -1178,7 +1189,23 @@ def _fetch_user_data(controller: Any, user_id: int) -> Dict[str, Any]:
...
-        return display_data
+        custom_profile_data = data["custom_profile_data"]
+
+        if custom_profile_data:
+            for field in custom_profile_data:
+                # Get label, type, order and value of field data
+                if field["type"] == 6:  # Person picker
+                    user_names = []
+                    for user in field["value"]:
+                        user_data = controller.model._all_users_by_id.get(user)
```
neiljp, line 1200: This should use `user_name_from_id` to avoid accessing the implementation.

I'd be open to adjusting the error behavior in that method to make this cleaner; perhaps providing an optional value different would return that instead of raising the exception?
(it looks like there are some other calling sites we could improve in any case)

I suspect that would enable making this into a simple list comprehension, or even generator expression with the join.

zulipterminal/ui_tools/views.py
```diff
@@ -1178,7 +1189,23 @@ def _fetch_user_data(controller: Any, user_id: int) -> Dict[str, Any]:
             if data["last_active"]:
                 display_data["Last active"] = data["last_active"]
 
-        return display_data
+        custom_profile_data = data["custom_profile_data"]
```
neiljp, line 1192: At the risk of increased nesting we could nest all of this section in the non-bot area, since bots cannot have custom fields, but a comment should suffice since the data should simply not have the fields in the first place.

zulipterminal/ui_tools/views.py
```diff
@@ -1,7 +1,6 @@
 """
 UI views for larger elements such as Streams, Messages, Topics, Help, etc
 """
-
```
neiljp, line 4: Unrelated - and inconsistent?

tests/ui_tools/test_popups.py
```diff
@@ -343,14 +355,66 @@ def test__fetch_user_data(
...
+            self.controller, 1
+        )
+
+        assert custom_profile_data == expected_value
+
     def test__fetch_user_data_USER_NOT_FOUND(self, mocker: MockerFixture) -> None:
         mocker.patch.object(self.controller.model, "get_user_info", return_value=dict())
 
-        display_data = self.user_info_view._fetch_user_data(self.controller, 1)
+        display_data, custom_profile_data = self.user_info_view._fetch_user_data(
```
neiljp, line 415: assert on custom_profile_data? The other two tests account for each part individually, but this is the only test for invalid user.

tests/ui_tools/test_popups.py
```diff
@@ -271,6 +275,14 @@ def mock_external_classes(
             return_value="Tue Mar 13 10:55 AM",
         )
 
+        mocker.patch.object(self.controller.model, "_all_users_by_id", _all_users_by_id)
+
+        mocker.patch.object(
+            self.controller.model,
+            "initial_data",
+            initial_data,
+        )
```
neiljp, line 284: This seems unnecessary?

tests/ui_tools/test_popups.py
```diff
@@ -271,6 +275,14 @@ def mock_external_classes(
             return_value="Tue Mar 13 10:55 AM",
         )
 
+        mocker.patch.object(self.controller.model, "_all_users_by_id", _all_users_by_id)
```
neiljp, line 278: This line will change based on another comment, and being consistent is great.

So it's worth keeping this consistent like you have already, and the code above does patch the object, but more generally we can directly set values on the controller, since it's a Mock(). That's a lot less verbose than repeatedly patching.

tests/ui_tools/test_popups.py
```diff
@@ -343,14 +355,66 @@ def test__fetch_user_data(
...
+                ],
+                {"Biography": "Simplicity", "Mentor": "Human 1, Human 2"},
+                id="user_has_custom_profile_data",
+            ),
+        ],
+    )
+    def test__fetch_user_data_custom_profile_data(
+        self,
+        mocker: MockerFixture,
+        to_vary_in_each_user: Dict[str, Any],
```
neiljp, line 398: This is the wrong type. Unfortunately mypy/pytest doesn't check the parametrize(s) to the parameters.

neiljp (commented):
@theViz343 This review should just be for the last commit, mostly small points and feedback :+1:

zulipterminal/model.py
theViz343: Since the realm user update event does not have "new_email" as a compulsory field, changing it to a ```NotRequired``` field in ```RealmUserEventPerson``` fixes this issue. Also, similar to "new_email", other fields are also not guaranteed to be present, except for ```user_id```. This means changing every field (except ```user_id```) to be a ```NotRequired``` field would also make sense.

neiljp: Yes, this is definitely a reasonable approach. If the different updates cannot be in the same event (eg. role and timezone?) then it would be stricter to use a Union of classes for each option. Otherwise NotRequired on lots of these indicates that somewhere between all or none of them could be present.

neiljp (commented):
@theViz343 This seems much more complete than before :+1:

While I left a lot of comments, I hope most are likely very quick changes or feedback. If there are any that you feel could be better documented more generally, please do propose that in the stream :)

I left feedback on some commit messages, so you may wish to look via GitHub for those comments well before pushing again, since they're unfortunately associated with the top of the page rather than a particular line of code. In future it's likely better for me to place those comments in my general feedback here.

The only model-based followup to this would be handling updates to the fields, which would round off the set of changes quite nicely. However, I'm unsure if this requires careful handling around synchronization of fields and user data, to avoid looking up absent fields (if possible to remove), or indexing into updated lists of options for a given field. I remember a czo topic mentioning some of this, but it's unclear which of these situations were covered, and how much is required to be handled on the client side. We should file an issue for this if we don't get to it.

zulipterminal/api_types.py
```diff
@@ -248,6 +248,20 @@ class Subscription(TypedDict):
 # NOTE: Responses between versions & endpoints vary
 
 
+class CustomProfileField(TypedDict):
```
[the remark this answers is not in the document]
theViz343: The comment on the CustomProfileField typeddict is wrong. The example data in the API docs has `field_data` missing in the type 8 (pronouns) field, but the API does have it. I'll make the change.

tests/conftest.py
```diff
@@ -626,6 +631,175 @@ def mentioned_messages_combination(request: Any) -> Tuple[Set[int], Set[int]]:
...
+            "field_data": '{"subtype":"github"}',
+            "order": 8,
+        },
+        {
+            # Pronouns: What pronouns should people use to refer to the user?
+            "id": 9,
+            "name": "Pronouns",
+            "type": 8,
+            "hint": "What pronouns should people use to refer to you?",
+            "field_data": "",
```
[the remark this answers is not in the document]
theViz343: The comment on the CustomProfileField typeddict is wrong. The example data in the API docs has `field_data` missing in the type 8 (pronouns) field, but the API does have it. I'll make the change.

zulipterminal/model.py
```diff
@@ -975,6 +977,62 @@ def get_other_subscribers_in_stream(
...
+        }
+
+        cleaned_profile_data = []
+        for field_id in custom_profile_data:
+            # Get label, type, order and value of field data
+            field_data = profile_fields[field_id]
+            raw_value = custom_profile_data[field_id]["value"]
+
+            if field_data["type"] == 3:  # List of options
+                field_options = json.loads(profile_fields[field_id]["field_data"])
```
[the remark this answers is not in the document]
theViz343: Renamed `field_data` to just `field` to simplify.

zulipterminal/model.py
```diff
@@ -975,6 +977,62 @@ def get_other_subscribers_in_stream(
...
+                    field_value = "https://github.com/{username}".format(
+                        **username_dict
+                    )
+                elif field_subtype == "twitter":
+                    field_value = "https://twitter.com/{username}".format(
+                        **username_dict
+                    )
+                elif field_subtype == "custom":
+                    url_pattern = field_options["url_pattern"]
+                    field_value = url_pattern % username_dict
```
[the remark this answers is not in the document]
theViz343: For the user-picker, the server sends a RealmUser update event, which has the field and it's value set to None. I think this means that the field can be considered deleted and is not needed to be shown. This behavior is the same as in the webapp, and is added to this PR in a later commit (event handling one).Therefore, the case where the value is an empty list is not possible.

zulipterminal/model.py
```diff
@@ -1933,6 +1933,21 @@ def _handle_realm_user_event(self, event: Event) -> None:
...
+                            "value": profile_field_data["value"]
+                        }
+                        if "rendered_value" in profile_field_data:
+                            updated_profile_data[profile_field_id][
+                                "rendered_value"
+                            ] = profile_field_data["rendered_value"]
+                        if profile_field_data["value"] is None:
+                            realm_user["profile_data"].pop(profile_field_id)
+                        else:
+                            realm_user["profile_data"].update(updated_profile_data)
```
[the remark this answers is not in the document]
theViz343: > The conditional part for the internal dict may be clearer as a comprehension with a filter.
This was my initial intention, but strangely enough, mypy doesn't seem to understand that the `key` exists in the typeddict while doing something like -  
`updated_profile_data[profile_field_id] = 
{
key: profile_field_data[key]
for key in profile_field_data
if key != "id"
}`
Instead, it throws an error saying - `error: TypedDict key must be a string literal; expected one of ("value", "rendered_value", "id")  [literal-required]`

zulipterminal/ui_tools/views.py
```diff
@@ -1144,7 +1156,22 @@ def _fetch_user_data(controller: Any, user_id: int) -> Dict[str, Any]:
             if data["last_active"]:
                 display_data["Last active"] = data["last_active"]
 
-        return display_data
+        # This will be an empty dict in case of bot users
+        custom_profile_data = data["custom_profile_data"]
+
+        if custom_profile_data:
+            for field in custom_profile_data:
+                # Get label, type, order and value of field data
```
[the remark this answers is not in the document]
theViz343: I forgot to remove it after I made changes in the previous version. I'll remove it

zulipterminal/ui_tools/views.py
```diff
@@ -1144,7 +1156,22 @@ def _fetch_user_data(controller: Any, user_id: int) -> Dict[str, Any]:
...
+                if field["type"] == 6:  # Person picker
+                    user_names = [
+                        controller.model.user_name_from_id(user)
+                        for user in field["value"]
+                    ]
+                    field["value"] = ", ".join(user_names)
+
+                display_custom_profile_data[field["label"]] = str(field["value"])
```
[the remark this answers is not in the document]
theViz343: It's only purpose is to indicate to mypy that field["value"] is a string instead of being a `Union[str, List[int]]` since `display_custom_profile_data` is typed as being  a `Dict[str,str]` (part of return value of the function).
I could change the type to `Dict[str, Any]` which would allow `str` to be removed but that might make the typing a bit weaker.

zulipterminal/ui_tools/views.py
```diff
@@ -1099,16 +1108,19 @@ def __init__(self, controller: Any, user_id: int, title: str, command: str) -> N
         super().__init__(controller, widgets, command, popup_width, title)
 
     @staticmethod
-    def _fetch_user_data(controller: Any, user_id: int) -> Dict[str, Any]:
+    def _fetch_user_data(
+        controller: Any, user_id: int
+    ) -> Tuple[Dict[str, Any], Dict[str, str]]:
```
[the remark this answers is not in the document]
theViz343: display_data can be `Dict[str,str]` but there is an issue with typing display_custom_profile_data as `Dict[str,str]` as I replied in another comment.

tests/ui_tools/test_popups.py
```diff
@@ -9,7 +9,7 @@
 from zulipterminal.api_types import Message
```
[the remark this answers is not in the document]
theViz343: > Is there an indication to use a ui: area somewhere? It's not unreasonable, but we've not included one in the recent past

No, it just felt natural to include it, so I added it. Thinking about it now, it might be a bit confusing to have an area be the same as a filename. I'll remove it for now.

zulipterminal/model.py
```diff
@@ -1927,6 +1927,24 @@ def _handle_realm_user_event(self, event: Event) -> None:
...
+                        else:
+                            updated_data = {"value": profile_field_data["value"]}
+                            if "rendered_value" in profile_field_data:
+                                updated_data["rendered_value"] = profile_field_data[
+                                    "rendered_value"
+                                ]
+
+                            realm_user["profile_data"][profile_field_id].update(
+                                updated_data
+                            )
```
neiljp, line 1946: Does this code miss the situation where a user starts with no custom data?
For example, what happens if you write a test which updates (adds) any custom field to user 11?

Similarly, while it shouldn't ideally happen, what if the user has no custom fields, or a given field has no data, and the event data is None?
This should only happen if we receive duplicate events or similar, but I'd hope we try and recover gracefully.

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: I see you made the other suggested changes. The remaining change isn't as important, though may make it easier to read.

Re the typing issue, one issue could be that we're moving from the event type to the profile data type? What is now `updated_data` is not typed, after all.

However, I suspect it's a wider problem than that since key is a variable in that comprehension. One alternative would be to type the keys somehow as Literal[...], or simply use an explicit form like:
```python3
if condition:
    some_dict = {"value": ...}
else:
    some_dict = {"value": ..., "rendered_value": ...}

File: zulipterminal/model.py (hunk 19 above)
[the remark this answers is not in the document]
Comment (neiljp) [reply zulipterminal/model.py]: In that case, I'd suggest adding defensive code which checks or asserts on the length of the existing/updated dict in various locations. An empty dict is a little False-y to ignore.

**Comment by neiljp** [comment]
@theViz343 I was half way through reviewing this, and reloading the main PR page submitted this review :/ Looking at the rest now.

File: tests/model/test_model.py [hunk 26]
```diff
@@ -3634,6 +3634,149 @@ def test__handle_realm_user_event(
...
+            "Link",
+            "Person Picker",
+            "External Account",
+            "Pronouns",
+        ],
+    )
+    def test__handle_realm_user_event_custom_profile_data(
+        self, person, profile_field_id, model, initial_data
+    ):
+        # id 12 matches initial_data["realm_users"][2] in the initial_data fixture
```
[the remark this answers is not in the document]
neiljp: Normally the linter complains about capitalized names in functions, since that is reserved for at module-level, but we may have it turned off (we can change it to a case-by-case basis).

In other tests we've erred towards:
```python3
def test_...(a, b, c, ..., constant=5, another=6):
```
or simply defining them in the body of the function, without capitalization.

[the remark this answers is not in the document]
neiljp: The general changes are an improvement, this is really a nit-pick :)

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
neiljp: Ah, I see the issue here. Generally if I see an explicit type conversion in code like this then I'm expecting there's a runtime reason, like the conversion from an int for lookup in the string-int index of custom profile fields.

Using `str()` is totally valid, and so mypy will also accept that as a 'fix' in this case.

However, for this kind of situation it's useful to use either a `cast` (or similar) as an indication to mypy and the reader what is happening and add a comment, or fix the type issue ;) We don't need an actual runtime conversion, as it's a string already, and a cast avoids that and indicates "I know what I'm doing - see the comment" ;)

Fixing the typing fully is likely too much work here, since we'd have to map from literals to the type of the data.

Another approach which is used a lot instead of `cast` is to use an assert on the type. For example, here you're handling the only non-string case specially, so an `else` should be fine:
```python3
for field in custom_profile_data:
    if field["type"] == 6:  # Person picker
        ...
    else:
        assert isinstance(field["value"], str)
```
Or, since you're reassigning `field["value"]`, you can likely move the assert outside of the conditional.
A comment may still be useful though :)
This is (typically) a runtime check, but doesn't implicitly convert anything like `str` would.

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
neiljp: This seems resolved now?

zulipterminal/model.py
theViz343: All users start with an empty dict if there is no custom profile data, ie `initial_data["realm_users"]["profile_data"]` = `{}`. So, for the first case, updating the dict new information won't be an issue. There is no existing test for it and the user fixture doesn't have the empty dict by default, so I'll add them.

For the second case, I'll provide a default `None` argument for pop() so that it doesn't throw a keyerror. I'll also add a comment for it.

neiljp: I think we'd have caught this if we had a good type for users, but unfortunately we're not applying RealmUser here, and it does differ between different situations in any case (eg. `is_active` being present).

neiljp:
@theViz343 Happy with the changes I proposed in the branch I linked in the stream? If so, I'll push here and merge :+1:

theViz343:
Yeah, the commit message changes and the fixup commits look good :+1: . Just a minor thing - in the `fixup: update: simplify tests.` commit, we no longer use the `custom_profile_data_fixture` directly so we could remove it from the test parameters.

neiljp:
@theViz343 Thanks for your work on this - just merged! :tada:

PR #1378, opened 2023-04-13
tests: conftest/test_boxes: Improve test for special characters. 

AmeliaTaihui:
Added tests that support special characters in topic matching

- [x] Discussed in **#zulip-terminal** in `Multi-word matching for topic autocomplete #T1358 #T1371 #T1378`
- [ ] Fully fixes #
- [x] Partially fixes issue #1358
- [x] Builds upon previous unmerged work in PR #1371 
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -5,11 +5,13 @@ from pathlib import Path
 from typing import Callable, Dict, Generator, List, Optional, Tuple
 
 import pytest
+import requests
 from pytest import CaptureFixture
 from pytest_mock import MockerFixture
 
 from zulipterminal.api_types import ServerSettings
 from zulipterminal.cli.run import (
+    NotAZulipOrganizationError,
     _write_zuliprc,
     exit_with_error,
     get_login_label,
@@ -90,13 +92,30 @@ def test_get_server_settings(
     server_settings_minimal: ServerSettings,
     realm_url: str = "https://chat.zulip.org",
 ) -> None:
-    response = mocker.Mock(json=lambda: server_settings_minimal)
+    response = mocker.Mock(
+        status_code=requests.codes.OK, json=lambda: server_settings_minimal
+    )
     mocked_get = mocker.patch("requests.get", return_value=response)
 
     result = get_server_settings(realm_url)
 
     assert mocked_get.called_once_with(url=realm_url + "/api/v1/server_settings")
     assert result == server_settings_minimal
+
+
+def test_get_server_settings__not_a_zulip_organization(
+    mocker: MockerFixture, realm_url: str = "https://google.com"
+) -> None:
+    response = mocker.Mock(
+        status_code=requests.codes.bad_request  # FIXME: Test others?
+    )
+    mocked_get = mocker.patch("requests.get", return_value=response)
+
+    with pytest.raises(NotAZulipOrganizationError) as exc:
+        get_server_settings(realm_url)
+
+    assert mocked_get.called_once_with(url=realm_url + "/api/v1/server_settings")
+    assert str(exc.value) == realm_url
 
 
 @pytest.mark.parametrize("options", ["-h", "--help"])
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -220,15 +220,24 @@ def get_login_label(server_properties: ServerSettings) -> str:
         return "Email: "
 
 
+class NotAZulipOrganizationError(Exception):
+    pass
+
+
 def get_server_settings(realm_url: str) -> ServerSettings:
-    server_properties = requests.get(url=f"{realm_url}/api/v1/server_settings").json()
-    return server_properties
+    response = requests.get(url=f"{realm_url}/api/v1/server_settings")
+    if response.status_code != requests.codes.OK:
+        raise NotAZulipOrganizationError(realm_url)
+    return response.json()
 
 
 def get_api_key(realm_url: str) -> Optional[Tuple[str, str, str]]:
     from getpass import getpass
 
-    server_properties = get_server_settings(realm_url)
+    try:
+        server_properties = get_server_settings(realm_url)
+    except NotAZulipOrganizationError:
+        exit_with_error(f"No Zulip Organization found at {realm_url}.")
 
     # Assuming we connect to and get data from the server, use the realm_url it suggests
     # This avoids cases where there are redirects between http and https, for example


zulipterminal/helper.py
```diff
@@ -535,9 +535,18 @@ def match_emoji(emoji: str, text: str) -> bool:
 
 
 def match_topics(topic_names: List[str], search_text: str) -> List[str]:
-    return [
-        name for name in topic_names if name.lower().startswith(search_text.lower())
-    ]
+    res = []
```
neiljp, line 538: This is minor, but it's good for readability to use full words (or phrases) for variables. It can be figured out, but the less ambiguous, the quicker to read :)

Does this stand for Result? Results? Something else?

Since this represents the return value, `matches`? `matching_topics` even?

zulipterminal/helper.py
```diff
@@ -535,9 +535,18 @@ def match_emoji(emoji: str, text: str) -> bool:
...
+    delimiters = "-_/"
+    trans = str.maketrans(delimiters, len(delimiters) * " ")
+    for topic_name_raw in topic_names:
```
neiljp, line 541: Again minor re naming:
- raw? `full_topic_name`? `complete_topic_name`? `entire_topic_name`?
- topic_name_lst? Sure it's a list, and we could add an `i`, but it's not a list of topic names... or is the `lst` suggesting something else?

tests/ui_tools/test_boxes.py
```diff
@@ -96,22 +96,27 @@ def test_not_calling_typing_method_without_recipients(
...
             ("#**Stream 1>T", 0, True, "#**Stream 1>Topic 1**"),
             ("#**Stream 1>T", 1, True, "#**Stream 1>This is a topic**"),
-            ("#**Stream 1>T", 2, True, None),
-            ("#**Stream 1>T", -1, True, "#**Stream 1>This is a topic**"),
-            ("#**Stream 1>T", -2, True, "#**Stream 1>Topic 1**"),
-            ("#**Stream 1>T", -3, True, None),
+            ("#**Stream 1>T", 2, True, "#**Stream 1>Hello there!**"),
+            ("#**Stream 1>T", -1, True, "#**Stream 1>He-llo there!**"),
+            ("#**Stream 1>T", -2, True, "#**Stream 1>Hello there!**"),
+            ("#**Stream 1>T", -3, True, "#**Stream 1>This is a topic**"),
```
neiljp, line 102: Note that the previous test cases were both forward and backward tests (0 to 2, -1 to -3), which cycles around. This loses the testing of those cases, since the -1 and -2 cases are now presumably the equivalent of the 3 and 4 cases?

tests/ui_tools/test_boxes.py
```diff
@@ -96,22 +96,27 @@ def test_not_calling_typing_method_without_recipients(
...
+            ("#**Stream 1>He", 0, True, "#**Stream 1>Hello there!**"),
+            ("#**Stream 1>He", 1, True, "#**Stream 1>He-llo there!**"),
+            ("#**Stream 1>He-", 0, True, "#**Stream 1>He-llo there!**"),
+            ("#**Stream 1>llo", 0, True, "#**Stream 1>He-llo there!**"),
             ("#**Stream 1>", 0, True, "#**Stream 1>Topic 1**"),
             ("#**Stream 1>", 1, True, "#**Stream 1>This is a topic**"),
-            ("#**Stream 1>", -1, True, "#**Stream 1>Hello there!**"),
-            ("#**Stream 1>", -2, True, "#**Stream 1>This is a topic**"),
+            ("#**Stream 1>", -1, True, "#**Stream 1>He-llo there!**"),
+            ("#**Stream 1>", -2, True, "#**Stream 1>Hello there!**"),
```
neiljp, line 113: It's the same issue here - there are more matches.

neiljp (commented):
@AmeliaTaihui Thanks for the tidying up :+1: I left some queries in the stream too, since I'm not fully sure what the intention was with a second PR.

This looks good. I left a few points inline - overall since the commits are separate still - mainly regarding naming and possible extra test cases we could add here.

This is by no means necessary for this PR, but if you are interested, either before this PR or after, you might see what is done in the test whose body starts at line 920 - test_generic_autocomplete_streams. That was refactored to provide a dict of results for a specific search string, which avoids a lot of repetition you're seeing here.

tests/ui_tools/test_boxes.py
```diff
@@ -1306,12 +1311,22 @@ def test__stream_box_autocomplete_with_spaces(
...
+            ("llo", ["He-llo there!"]),
         ],
         ids=[
             "no_search_text",
             "single_word_search_text",
+            "dash_sign_other_words_test",
+            "dash_sign_splited_or_not_0_test",
+            "dash_sign_splited_or_not_1_test",
+            "dash_sign_in_search_test",
+            "dash_sign_space_test",
```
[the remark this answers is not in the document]
AmeliaTaihui: I'm not exactly familiar with pytest, is there any document on how to use "case" or "param"(as shown in test_boxes.py,` from pytest import param as case`) to tidy up the ids? I did some search and couldn't find any. Or do you mean that I can do that myself by naming them in a more distinguished way?

tests/conftest.py
```diff
@@ -587,7 +587,7 @@ def message_history(request: Any) -> List[Dict[str, Any]]:
 
 @pytest.fixture
 def topics() -> List[str]:
-    return ["Topic 1", "This is a topic", "Hello there!"]
+    return ["Topic 1", "This is a topic", "Hello there!", "He-llo there!"]
```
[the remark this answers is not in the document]
AmeliaTaihui: I think if we add `Hello th-ere`, then there will be one more match with `th`. If we delete `He-llo there` and add `Hello t/here`, then there will be one less match with `th`. But this way we would lose having splitting characters in the first word like in `He-llo there`. Do you think we should delete `He-llo there` ?

tests/ui_tools/test_boxes.py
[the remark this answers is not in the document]
neiljp: I imported as `case` since `param` is used everywhere in pytest :)

This allows a test case that would be one list item as a tuple `(value1, value2, value3)` mapping to one list item in the `ids`, eg. `"some_test_case"`, to be written instead as
```python3
case(value1, value2, value3, id="some_test_case"),
case(value1a, value2, value3, id="another_test"),
...
```

That maintains locality of the ids with each specific test case :)

PR #1385, opened 2023-04-21
autocomplete: Match topic search text against start of all words

### LINKED ISSUE #1358: Support multi-word matching when autocompleting topics

AmeliaTaihui:
Support single search word matching words in candidate topics in different position

- [x] Discussed in **#zulip-terminal** in `Multi-word matching for topic autocomplete #T1358 #T1371`
- [x] Fully fixes #1358 
- [ ] Partially fixes issue #
- [x] Builds upon previous unmerged work in PR #1371 #1378
- [ ] Requires merge of PR #

### How did you test this?

- [] Existing automated tests should already cover this (*only a refactor of tested code*)


neiljp:
@AmeliaTaihui Thanks for working on this - as per the message in the stream, I squashed the commits from 1378 that I had from your previous push, and updated the commit message slightly.

This feature will be useful to have merged :tada:

Issue #1348, opened 2023-03-22
Add markers before stream names (& DMs) in message headers


neiljp:
In the current UI, we don't use a prefix to indicate the type of stream (if relevant):
- ~in the narrow title (eg. `zulip-terminal: topic narrow`)~
- in the 'Current message recipients' area (eg. `zulip-terminal > some topic`)
- in the recipient header which is present in the message list in interleaved views (eg. scrolling through multiple topics in a stream narrow)

In contrast, see the stream list in the left panel, or the marker which we add when composing a message to different types of stream.

The characters used are looked up using `STREAM_ACCESS_TYPE`; reviewing how that is used elsewhere should guide how to add it in these cases.

Not all of the above cases needed to be fixed in one PR.

This was previously raised as part of discussion in **# zulip-terminal > Stream prefixes (messages)** and resulted in #719, from which this is split out from (I'm not aware of any outstanding PRs for that). That issue and discussion may provide some additional clarification on the intent here.

yuktasarode:
@zulipbot claim

zulipbot:
**ERROR:** Unexpected response from GitHub API.

supascooopa:
Hello, I am new to open source and I'd like to give this issue a try if it's still available.

supascooopa:
@zulipbot claim

zulipbot:
Welcome to Zulip, @supascooopa! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

zulipbot:
@supascooopa We noticed that you have not made any updates to this issue or linked PRs for 10 days. Please comment here if you are still actively working on it. Otherwise, we'd appreciate a quick `@zulipbot abandon` comment so that someone else can claim this issue and continue from where you left off.

If we don't hear back, you will be automatically unassigned in 4 days. Thanks!

[...]

PR #1386, opened 2023-04-24
Adding prefix icons to current message headers

supascooopa:
Adds prefix icons to currents message headers

### Outstanding aspect(s)    

There are other parts of the ui that also needs the prefix icons but the person who opened the issue said that each section could be done with multiple pull request so that is how I chose to do it.

- [ ] 

- [ ] Fully fixes #
- [x] Partially fixes issue #1348 
- [ ] Requires merge of PR #

### How did you test this?

- [ ] It has a commit summary following the [documented style](https://github.com/zulip/zulip-terminal#structuring-commits-
--speeding-up-reviews-merging--development) (title & body)

### Visual changes    

Screenshot 2023-04-24 at 16 18 38

Screenshot 2023-04-24 at 16 18 38

Screenshot 2023-04-24 at 16 20 30

### DIFF
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -11,6 +11,7 @@ from zulipterminal.config.keys import keys_for_command
 from zulipterminal.config.symbols import (
     QUOTED_TEXT_MARKER,
     STATUS_INACTIVE,
+    STREAM_MARKER_PUBLIC,
     STREAM_TOPIC_SEPARATOR,
     TIME_MENTION_MARKER,
 )
@@ -28,6 +29,7 @@ class TestMessageBox:
     def mock_external_classes(self, mocker, initial_index):
         self.model = mocker.MagicMock()
         self.model.index = initial_index
+        self.model.stream_access_type.return_value = "public"
 
     @pytest.mark.parametrize(
         "message_type, set_fields",
@@ -893,19 +895,24 @@ class TestMessageBox:
     @pytest.mark.parametrize(
         "msg_narrow, msg_type, assert_header_bar, assert_search_bar",
         [
-            ([], 0, f"PTEST {STREAM_TOPIC_SEPARATOR} ", "All messages"),
+            (
+                [],
+                0,
+                f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR} ",
+                "All messages",
+            ),
             ([], 1, "You and ", "All messages"),
             ([], 2, "You and ", "All messages"),
             (
                 [["stream", "PTEST"]],
                 0,
-                f"PTEST {STREAM_TOPIC_SEPARATOR} ",
+                f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR} ",
                 ("bar", [("s#bd6", "PTEST")]),
             ),
             (
                 [["stream", "PTEST"], ["topic", "b"]],
                 0,
-                f"PTEST {STREAM_TOPIC_SEPARATOR}",
+                f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR}",
                 ("bar", [("s#bd6", "PTEST"), ("s#bd6", ": topic narrow")]),
             ),
             ([["is", "private"]], 1, "You and ", "All direct messages"),
@@ -925,7 +932,7 @@ class TestMessageBox:
             (
                 [["is", "starred"]],
                 0,
-                f"PTEST {STREAM_TOPIC_SEPARATOR} ",
+                f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR} ",
                 "Starred messages",
             ),
             ([["is", "starred"]], 1, "You and ", "Starred messages"),
@@ -934,10 +941,15 @@ class TestMessageBox:
             (
                 [["search", "FOO"]],
                 0,
-                f"PTEST {STREAM_TOPIC_SEPARATOR} ",
+                f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR} ",
                 "All messages",
             ),
-            ([["is", "mentioned"]], 0, f"PTEST {STREAM_TOPIC_SEPARATOR} ", "Mentions"),
+            (
+                [["is", "mentioned"]],
+                0,
+                f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR} ",
+                "Mentions",
+            ),
             ([["is", "mentioned"]], 1, "You and ", "Mentions"),
             ([["is", "mentioned"]], 2, "You and ", "Mentions"),
             ([["is", "mentioned"], ["search", "FOO"]], 1, "You and ", "Mentions"),
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -9,6 +9,7 @@ from urwid import Columns, Divider, Padding, Text
 
 from zulipterminal.config.keys import keys_for_command
 from zulipterminal.config.symbols import (
+    DIRECT_MESSAGE_MARKER,
     QUOTED_TEXT_MARKER,
     STATUS_INACTIVE,
     STREAM_MARKER_PUBLIC,
@@ -901,8 +902,8 @@ class TestMessageBox:
                 f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR} ",
                 "All messages",
             ),
-            ([], 1, "You and ", "All messages"),
-            ([], 2, "You and ", "All messages"),
+            ([], 1, f" {DIRECT_MESSAGE_MARKER} You and ", "All messages"),
+            ([], 2, f" {DIRECT_MESSAGE_MARKER} You and ", "All messages"),
             (
                 [["stream", "PTEST"]],
                 0,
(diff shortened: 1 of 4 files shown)


neiljp:
@supascooopa Thanks for working on this, it looks like a good start on the feature! :+1:

I suspect it would be clearer visually if we style the icon as part of the name, so that it looks like part of the 'bar' and doesn't look too similar to the user presence icons - particular with web-public streams.

Keeping the commits separate is fine for now, though the changes may well be small enough to squash together.

However, we will need to fix the test failure - are you planning to look at that next?

Please do drop into the stream on chat.zulip.org and ask if you have questions :)

supascooopa:
I was not sure where to put the icon and was torn between it being inside the rectangle/bar or out. So I could do that switch no problem. 

I'll keep in mind the size of the commits for next time.

I'd like to resolve the test failure but I'm not sure what's causing it and I'd appreciate some help resolving it.

neiljp:
@supascooopa You may want to wait until you've made the changes before looking at the test, in case the fix is different.

Debugging the test failure may be easier if you're doing it locally? Did you use `make check`, `pytest` or similar? It's difficult to know what you're wanting help with without more context.

supascooopa:
Before I committed any changes I wanted to make sure that the following is what you @neiljp  meant, please have a look:
Screenshot 2023-04-25 at 11 20 20

I've very little experience with pytest, and don't understand why the test are failing. I've ran the pytest after I've added the icon into the rectangle but the problem is persisting.

supascooopa:
Hello, I have found a way to pass the tests. The problem was a invalid key being passed in STREAM_ACCESS_TYPE[stream_access_type]["icon"] by MagicMock object. I've looked through  TestMessageBox class to fix the failed tests but wasn't able to make much sense of it. Instead I was able to clear this issue by writing a try except block in stream_header method where the key error was coming from. Which cleared test error but others popped up. but I was quickly able to identify which test were failing and made adjustment to them and now the test all run without a hitch.

neiljp:
> Before I committed any changes I wanted to make sure that the following is what you @neiljp meant, please have a look: Screenshot 2023-04-25 at 11 20 20

This looks better :+1: We could also consider an additional space before this, what do you think?

supascooopa:
Apologize for the late reply, I had some issues launching ZT but I managed to solve it. Yes I think adding a space before the icon looks better. Before it was pressing up against the left most border of the rectangle which gave it a claustrophobic feeling. please have a look below and let me know what you think:
Before:
Screenshot 2023-04-28 at 18 23 49


After:
Screenshot 2023-04-28 at 18 17 03

zulipterminal/ui_tools/messages.py
```diff
@@ -152,9 +153,15 @@ def stream_header(self) -> Any:
         assert self.stream_id is not None
         color = self.model.stream_dict[self.stream_id]["color"]
         bar_color = f"s{color}"
+        stream_access_type = self.model.stream_access_type(self.stream_id)
+        try:
+            stream_icon = STREAM_ACCESS_TYPE[stream_access_type]["icon"]
+        except KeyError:
+            stream_icon = INVALID_MARKER
```
neiljp, line 160: Is this just for the tests?

The model method should only ever return a valid index into STREAM_ACCESS_TYPE, so we shouldn't need this try/except otherwise.

supascooopa: Yes, I've added the try/except in order to clear tests. Every time I ran the tests it gave a key error on the line with STREAM_ACCESS_TYPE. The error message I kept getting was :
`KeyError: <MagicMock name='mock.stream_access_type()' id='4817710544'>`
 I suspect that the test is missing a parameter to access the STREAM_ACCESS_TYPE, that's why the test results in a key error. I tried to look over the test to figure it out but wasn't able to. 
Do you have any ideas?

neiljp: You have the right idea as to the fundamental issue here - since the mock is very general, it doesn't have specific values and so doesn't correctly key into the dict. The simplest way to do this is likely to set `stream_access_type.return_value="public"` in a convenient location, at least to make the tests run and enable you to adjust the test cases.

It'd be useful to have tests with other icons, but that's a different change.

supascooopa: That seems to be working, but I have to fiddle with it more to see where exactly to put it so it works. I'm currently putting it in several test in TestMessageBox class.

neiljp: The simplest approach may be to insert it into the two fixtures which are used in all/most of the tests.

supascooopa: I have looked over conftest.py came to the conclusion that msg_box is the fixture to use for this job. But I'm having trouble getting it to work as I can't get it to return "public" when it's called. Do you have any suggestions?

neiljp: If you use something like I mentioned above it should work. I'd suggest you push another commit with the test changes, as that way people can see what you have and how you might fix it.

supascooopa:
I'm not sure why the linting test is failing now, any ideas @neiljp?

neiljp:
@supascooopa The failing part is 'Ensure isolated PR commits', which as per the name, implies that each commit should individually pass linting/tests. So if you have added a commit to fix a failing earlier commit, that earlier commit will still fail this part of the automated testing. The typical solution is to combine (squash) the commits that 'fix' a previous commit with that earlier commit, using an interactive rebase. You can then force push to the PR with the revised branch.

supascooopa:
@neiljp  Thank you for the suggestion. I'm working on it locally and it looks like I'll be left with 2 commits. Because the cause of the error is pretty early on the commit history. I'd have to squash 457761b to 95c6c3f and then 9776416 to 64b0018. If that's okay with you I'll do it this way. If you have any other idea's please let me know.

tests/ui_tools/test_messages.py
```diff
@@ -1725,6 +1737,7 @@ def test_reactions_view(
         expected_attributes,
     ):
         self.model.user_id = 1
+
```
neiljp, line 1740: Unrelated addition.

zulipterminal/ui_tools/messages.py
```diff
@@ -152,9 +152,12 @@ def stream_header(self) -> Any:
         assert self.stream_id is not None
         color = self.model.stream_dict[self.stream_id]["color"]
         bar_color = f"s{color}"
+        stream_access_type = self.model.stream_access_type(self.stream_id)
+        stream_icon = STREAM_ACCESS_TYPE[stream_access_type]["icon"]
         stream_title_markup = (
             "bar",
             [
+                (bar_color, f" {stream_icon} "),
                 (bar_color, f"{self.stream_name} {STREAM_TOPIC_SEPARATOR} "),
```
neiljp, line 161: It would be nice to combine these two lines, but it does put it over the line limit, so this is fine as it is :+1:

neiljp (approved):
@supascooopa The test change does look good, certainly in terms of adapting the tests and showing the public-stream case :+1:

I tried with the two resulting commits you mention, but it doesn't pass `tools/check-branch`. It also seems like the resulting second commit would effectively resolve the first approach you took to solve the problem (try/except). In that situation I'd recommend you consider squashing all the commits together, since we aim for multiple commits in PRs to relate to specific changes rather than exploratory work towards solving a problem (see https://github.com/zulip/zulip-terminal#structuring-commits---speeding-up-reviews-merging--development).

I've left a few other comments, but when squashed and with a good commit title/summary this would be a useful starting point to merge :+1:

tests/ui_tools/test_messages.py
supascooopa: Probably a mistyping. I'll delete it with a new commit before I squash all the commits together?

supascooopa:
I'm currently looking through the git log and reading up on the commit message styles. I'll come up with a coherent commit message that adheres to those styles and squash all the commits soon.

supascooopa:
My commit message would be something like the following, if @neiljp you think it's good I'll go ahead with it:

messages: Adding stream markers to message headers

This commit aims to improve the current UI, by adding prefix stream markers to message headers in messages view. This is achieved by the following changes made in stream_header method:
- Calling the stream access type from the model using the stream_id.
- Using that access type to call the kind of icon from the STREAM_ACCESS_TYPE dictionary. 
- Finally adding the icon to stream_title_markup.

tests and fixtures are updated accordingly.

tests/ui_tools/test_messages.py
neiljp: Yes, or you can use eg `git commit --amend -a` to update the previous commit code and text, whether normally at the tip of a branch, or if in the middle of an interactive rebase.

neiljp:
@supascooopa It's fine to go with something like that - note that you can update commit messages in your branch at any time using the `--amend` I noted above, or specific commits by changing the 'pick' in an interactive rebase to `r` (or `reword`). The necessary force-push to this feature branch (not `main`) after doing the squash would be just the same as if you adjusted the commit text repeatedly and pushed each time :)

In terms of the text specifically, you might try installing and setting up gitlint in your development environment (see the README) for some automated feedback, particularly if you intend to continue contributing.

Typically the body of a commit message does not need to repeat the logic of the code in text form, but specific notes on the reasoning for the change in behavior, or why a certain implementation was chosen.

Note that someone merging code may adjust the text/code slightly before merging in any case :)

zulipterminal/ui_tools/messages.py
```diff
@@ -176,7 +176,11 @@ def stream_header(self) -> Any:
     def private_header(self) -> Any:
         title_markup = (
             "header",
-            [("general_narrow", "You and "), ("general_narrow", self.recipients_names)],
+            [
+                ("general_narrow", " § "),
```
neiljp, line 180: It'd be good to add this new symbol in the centralized config/symbols.py.

supascooopa: Right, I'll get on that.

supascooopa:
@neiljp Can I help to resolve the conflicts? Would you like for me to squash the last two commits to 594808507a9d45a8d2e128e68046960aac6bfcc4.  Any feed back is welcome! :)

neiljp:
@supascooopa Are you asking for help resolving the conflicts?

The last two commits should be squashed together, since the second is a refinement of the first.

I'm not suggesting you do the following, it is an example. If you had made the addition to symbols.py in one commit, and then used it in the next commit, this wouldn't have been a refinement in the same way. The next would have depended on the earlier commit, but was a separate change. The documentation tries to describe this in the 'Structuring commits' section.

supascooopa:
@neiljp Yes I'd like some help resolving the conflicts. Should I rebase my branch to the latest commit on the main branch and resolve the conflicts that way?
 I will squash the last two commits together.
 Thank you for the suggestion, I try to follow the rules/documentation but keep forgetting some while following others...

neiljp:
@supascooopa Thanks for working on this :+1: I slightly adjusted the symbols file and commit text, and just merged this :tada:

Issue #357, opened 2019-05-15
Windows OS support


neiljp:
Currently ZT does not run in Windows environments, excepting eg. virtual machines running a supported OS. One of the reasons noted for the lack of support is that `urwid` appears to require the `fcntl` call in some cases urwid/urwid#152.

One possible solution involves a cross-platform wrapper such as `portalocker` (available on PyPI and at https://github.com/WoLpH/portalocker). This may need support in `urwid` rather than ZT, but may be the first step forward.

Note that there may be other issues, but the above is a definite blocker to Windows support.

For reference, also see #205 for testing on windows.

rht:
Or use WSL for now: https://github.com/zulip/zulip-terminal/issues/178#issuecomment-506905623. Since Dragon NaturallySpeaking (for accessibility purpose, and there doesn't seem to be other viable alternative for dictation?) runs on Windows, this is the only way to run both.

amanagr:
I use WSL and zt works well without any extra setup on it. I actually test any cross-platform changes on WSL for zt.

neiljp:
I've renamed this issue to narrow its applicability; right now ZT should run fine in a WSL environment.

I'm leaving this open to track the issue, but it's certainly not a priority given we have WSL support now.

However, anyone is welcome to contribute a PR to make this reality :+1:

It is possible that urwid could provide a workaround for the fcntl issue, but otherwise a PR would need to do so. We'd also want at least some documentation and CI support.

PR #1394, opened 2023-05-05
Edit message in a terminal editor subprocess

mek-yt:
Create a new command OPEN_IN_TERMINAL_EDITOR (default `ctrl o`) for editing the current message in terminal editor with python tempfile.

I am using helix as my terminal editor and when I write in terminal I found myself using it's shortcut. 

- [x] Discussed in **#zulip-terminal** in `Composing in terminal editor #T1394`
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

[Composing in terminal editor](https://asciinema.org/a/3CsC3bfnwZl2gZFHYoxSP22tM)

### DIFF
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -428,6 +428,11 @@ KEY_BINDINGS: Dict[str, KeyBinding] = {
         'help_text': 'Insert new line',
         'key_category': 'compose_box',
     },
+    'OPEN_EXTERNAL_EDITOR': {
+        'keys': ['ctrl o'],
+        'help_text': 'Open an external editor to edit the message content',
+        'key_category': 'compose_box',
+    },
     'FULL_RENDERED_MESSAGE': {
         'keys': ['f'],
         'help_text': 'Show/hide full rendered message (from message information)',
--- a/zulipterminal/ui_tools/boxes.py
+++ b/zulipterminal/ui_tools/boxes.py
@@ -3,9 +3,13 @@ UI boxes for entering text: WriteBox, MessageSearchBox, PanelSearchBox
 """
 
 import re
+import shlex
+import shutil
+import subprocess
 import unicodedata
 from collections import Counter
 from datetime import datetime, timedelta
+from tempfile import NamedTemporaryFile
 from time import sleep
 from typing import Any, Callable, Dict, List, NamedTuple, Optional, Tuple
 
@@ -839,6 +843,47 @@ class WriteBox(urwid.Pile):
         elif is_command_key("MARKDOWN_HELP", key):
             self.view.controller.show_markdown_help()
             return key
+        elif is_command_key("OPEN_EXTERNAL_EDITOR", key):
+            editor_command = self.view.controller.editor_command
+
+            # None would indicate for shlex.split to read sys.stdin for Python < 3.12
+            # It should never occur in practice
+            assert isinstance(editor_command, str)
+
+            if editor_command == "":
+                self.view.controller.report_error(
+                    "No external editor command specified; "
+                    "Set 'editor' in zuliprc file, or "
+                    "$ZULIP_EDITOR_COMMAND or $EDITOR environment variables."
+                )
+                return key
+
+            editor_command_line: List[str] = shlex.split(editor_command)
+            if not editor_command_line:
+                fullpath_program = None  # A command may be specified, but empty
+            else:
+                fullpath_program = shutil.which(editor_command_line[0])
+            if fullpath_program is None:
+                self.view.controller.report_error(
+                    "External editor command not found; "
+                    "Check your zuliprc file, $EDITOR or $ZULIP_EDITOR_COMMAND."
+                )
+                return key
+            editor_command_line[0] = fullpath_program
+
+            with NamedTemporaryFile(suffix=".md") as edit_tempfile:
+                with open(edit_tempfile.name, mode="w") as edit_writer:
+                    edit_writer.write(self.msg_write_box.edit_text)
+                self.view.controller.loop.screen.stop()
+
+                editor_command_line.append(edit_tempfile.name)
+                subprocess.call(editor_command_line)
+
+                with open(edit_tempfile.name, mode="r") as edit_reader:
+                    self.msg_write_box.edit_text = edit_reader.read().rstrip()
+            self.view.controller.loop.screen.start()
+            return key
+
         elif is_command_key("SAVE_AS_DRAFT", key):
             if self.msg_edit_state is None:
                 if self.compose_box_status == "open_with_private":
--- a/README.md
+++ b/README.md
@@ -251,6 +251,10 @@ color-depth=256
 ## This is highly dependent on a suitable terminal emulator, and support in the selected theme
 ## Terminal emulators without this feature may show an arbitrary solid background color
 transparency=disabled
+
+## Editor: set external editor command, to edit message content
+## If not set, this falls back to the $ZULIP_EDITOR_COMMAND then $EDITOR environment variables
+# editor: nano
 ```
 
 > **NOTE:** Most of these configuration settings may be specified on the
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -13,6 +13,7 @@
   - [When are messages marked as having been read?](#when-are-messages-marked-as-having-been-read)
   - [How do I access multiple servers?](#how-do-i-access-multiple-servers)
   - [What is autocomplete? Why is it useful?](#what-is-autocomplete-why-is-it-useful)
+  - [Can I compose messages in another editor?](#can-i-compose-messages-in-another-editor)
 - Something is not working!
   - [Colors appear mismatched, don't change with theme, or look strange](#colors-appear-mismatched-dont-change-with-theme-or-look-strange)
   - [Symbols look different to in the provided screenshots, or just look incorrect](#symbols-look-different-to-in-the-provided-screenshots-or-just-look-incorrect)
@@ -372,6 +373,55 @@ through autocomplete depend upon the context automatically.
 
 **NOTE:** If a direct message recipient's name contains comma(s) (`,`), they
 are currently treated as comma-separated recipients.
+
+## Can I compose messages in another editor?
+
+In the `main` branch of zulip-terminal, you can now use an external editor of
+your choice to compose your messages using the `Ctrl o` hotkey.
+
+The editor command is looked for in the `ZULIP_EDITOR_COMMAND` and `EDITOR`
+environment variables (in that order), or the `editor` entry of the zuliprc
+file (which overrides both of those).
+
+It should work directly for most terminal editors with only the program name,
+eg. `vim`, `nano`, `helix`, `kakoune`, `nvim`, etc.
+
+It can also be used with a desktop editor with some constraints: the program
+must not fork or detach from the running terminal and should open in a new
+window. For this reason it may be useful to use `ZULIP_EDITOR_COMMAND` (or the
+zuliprc setting) to avoid conflict with the more widely used `EDITOR`
+environment variable. Some examples include:
+
+- [lapce](https://github.com/lapce/lapce) with `lapce -n -w`
+- [sublime-text](https://www.sublimetext.com/) with `subl -n -w`
+- [marker](https://github.com/fabiocolacio/Marker) with `marker`
+- [vim](https://github.com/vim/vim) with `vim -g -f` or `gvim -f`
+- [vscode](https://github.com/microsoft/vscode) with `code -n -w`
+
+**NOTE:** Backslashing white space (`\ `) is needed when using an executable
+containing them, for example Sublime Text on macOS can be configured with
+`/Applications/Sublime\ Text.app/Contents/SharedSupport/bin/subl`.
+
+This feature works by sending the message content back and forth using
+temporary files:
+- A temporary file is created, filled with any existing message content;
+- The editor command is run, with the filename appended (standard practice for editors);
+- You edit your text in the editor of your choice;
+- When the external editor process ends (closing the window, or quitting the terminal
+editor), the compose box will be updated with the new message content from
+the temporary file.
+
+> NOTE 1: If using a terminal-based editor, it will temporarily take over the
+> entire terminal and replace the application. However, upon exiting that
+> editor, the application should return to the terminal.
+
+> NOTE 2: Whichever type of editor you use (terminal or desktop/graphical), the
+> application will temporarily pause while you are editing. The application
+> should resume as normal once you exit the external editor and the text you
+> entered appears in the compose area.
+
+**If you have success with other editors, these examples do not work, or have
+other problems, please let us know!**
 
 ## Colors appear mismatched, don't change with theme, or look strange
 
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -97,6 +97,7 @@
 |Cycle through autocomplete suggestions in reverse|<kbd>Ctrl</kbd> + <kbd>r</kbd>|
 |Exit message compose box|<kbd>Esc</kbd>|
 |Insert new line|<kbd>Enter</kbd>|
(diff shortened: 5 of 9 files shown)


zulipterminal/ui_tools/boxes.py
```diff
@@ -810,6 +812,17 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+        elif is_command_key("OPEN_IN_TERMINAL_EDITOR", key):
+            edit_tempfile = NamedTemporaryFile(suffix=".md")
+            with open(edit_tempfile.name, mode="w") as edit_writer:
+                edit_writer.write(self.msg_write_box.edit_text)
+            self.view.controller.loop.screen.stop()
+            subprocess.call([f"$EDITOR {edit_tempfile.name}"], shell=True)
+            with open(edit_tempfile.name, mode="r") as edit_reader:
+                self.msg_write_box.edit_text = edit_reader.read()
+            edit_tempfile.close()
+            self.view.controller.loop.screen.start()
```
neiljp, line 824: One potential concern I have here is whether there is anything that could block getting to this point and restarting the application loop.

neiljp (commented):
@neruyzo Thanks for thinking of this feature and contributing :+1:

Once I had set $EDITOR this worked for me with `vim`, though prior to that it simply failed with no error.

Was `ctrl o` the shortcut in your editor, or from something else? Selecting a new shortcut key is often challenging as it then is difficult to change later.

One issue I found is that there seems to be a blank line added after the text in the compose box, after resuming from the editor?

One followup we could discuss may be to have a header line automatically generated in the message to indicate the recipients/context, which would then be stripped out later. At an extreme we could even defer compose to an editor, though we'd lose features like validation and autocomplete that we currently have.

zulipterminal/ui_tools/boxes.py
neruyzo: I don't think this is really a problem, the only case it will lock is if the external editor freeze and in that case the user would already try to kill it unlocking zulip.

zulipterminal/ui_tools/boxes.py
```diff
@@ -810,6 +812,17 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
         elif is_command_key("MARKDOWN_HELP", key):
             self.view.controller.show_markdown_help()
             return key
+        elif is_command_key("OPEN_IN_TERMINAL_EDITOR", key):
+            edit_tempfile = NamedTemporaryFile(suffix=".md")
+            with open(edit_tempfile.name, mode="w") as edit_writer:
+                edit_writer.write(self.msg_write_box.edit_text)
+            self.view.controller.loop.screen.stop()
+            subprocess.call([f"$EDITOR {edit_tempfile.name}"], shell=True)
```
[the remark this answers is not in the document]
neruyzo: I now check with `shutil.which` the full path to the program and `report_error` when it return `None`.

neruyzo:
I update the pull request by adding an environment variable `$ZULIP_EDITOR_COMMAND` for custom command like `lapce -n -w TEMPFILE` which allow even desktop application , check and report if the program is not found.

I rename the command `OPEN_EXTERNAL_EDITOR`.

zulipterminal/config/keys.py
```diff
@@ -385,6 +385,11 @@ class KeyBinding(TypedDict):
         'help_text': 'Clear compose box',
         'key_category': 'msg_compose',
     },
+    'OPEN_EXTERNAL_EDITOR': {
+        'keys': ['ctrl o'],
+        'help_text': 'Open the message in external editor',
```
neiljp, line 390: I like the change to 'external editor' rather than 'terminal editor' here, particularly if we're working with external editors more generally. I did try gvim, but it didn't seem to work for me, though vim still does.

'Open' vs 'Edit' is tricky; maybe use 'Compose'?

A minor point is that this may make more sense to have further up in the file, so that in the help and hotkeys.md it appears above the 'navigation' compose keys (eg. "Jump to start of line")

zulipterminal/ui_tools/boxes.py
```diff
@@ -810,6 +814,40 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+            with open(edit_tempfile.name, mode="w") as edit_writer:
+                edit_writer.write(self.msg_write_box.edit_text)
+            self.view.controller.loop.screen.stop()
+            if "TEMPFILE" not in editor:
+                editor_splits.append(edit_tempfile.name)
+            else:
+                editor_splits = [
+                    part if part != "TEMPFILE" else edit_tempfile.name
+                    for part in editor_splits
+                ]
```
neiljp, line 844: Is "TEMPFILE" related to a special case for `lapce`? Does appending the filename not work?

zulipterminal/ui_tools/boxes.py
```diff
@@ -810,6 +814,40 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
             return key
+        elif is_command_key("OPEN_EXTERNAL_EDITOR", key):
+            editor = os.environ.get("ZULIP_EDITOR_COMMAND", os.environ.get("EDITOR"))
+            if editor is None:
+                self.view.controller.report_error(
+                    "Configure $EDITOR or $ZULIP_EDITOR_COMMAND shell environment."
+                )
+                return key
+            editor_splits = editor.split(" ")
+            if len(editor_splits) >= 1:
```
neiljp, line 825: What case are you trying to avoid here? Or equally, what should happen if `len(editor_splits)<1` ?

zulipterminal/ui_tools/boxes.py
neiljp: I agree this is OK for now, though if an external editor is opened (eg. gvim) then I can switch back to the application and it allows me to keep typing.

That may or may not be related to the issue with gvim not working more generally - it leaves any text I entered *while gvim was open* in the text box. Maybe since it is running in the background, the subprocess exits already?

neiljp (commented):
@neruyzo Thanks for improving this - I left a few more comments.

The other point that occurs for me is the blank line issue I mentioned in the previous review - do you see this too?

For example, for `This text` in vim, becomes
```
This text[]

```
in the compose box (`[]` being the cursor)

zulipterminal/config/keys.py
neruyzo: I think using the first letter of the command as the shortcut is easier to remember if 'Open' keyword is okay in this context.

zulipterminal/ui_tools/boxes.py
neruyzo: I remove the condition and keep only appending.
I was like 'maybe some people would like to do pipe command' but it will not work with shell off.

I test it with lapce / sublime text / marker and all don't need to position the file inside the command.

zulipterminal/ui_tools/boxes.py
neruyzo: Agree, we only need to catch the `None` and the non existing program, the `split` will never return an empty list.

neruyzo:
I updated it.

I missed before the trailing empty line that may be create by code editor, to fix that I now use `.rstrip()` to remove trailing empty line and space.

In addition to the comments, should I document $ZULIP_EDITOR_COMMAND somewhere, maybe in the README ?

I tested in most of my code editors I could use `TEMPFILE:LINE:CHARACTER` to position the cursor on opening.

Zulip terminal will try to position the cursor at the same place as before the message was edit. There is no problem when the new message is smaller than cursor position.

zulipterminal/ui_tools/boxes.py
neruyzo: Exactly, in that case, the gvim command will close directly. Maybe like for sublime text and lapce there is a `--watch` and `--new` flags to use in the command.

zulipterminal/config/keys.py
neiljp: I agree about 'o' for 'Open' being good, I'll have a think on the wording.

neiljp (commented):
@neruyzo For use with terminal editors this looks fine to me :+1:

I'll adjust the commit text if we merge this as-is soon, to match the style we use.

The extra line issue looks resolved, and cursor position isn't a problem :+1:

Re the description, I definitely agree that `Open` works in terms of `ctrl o`, and looking at adjacent entries, we use 'message' rather than 'composition' or similar, so it's at least consistent for now :)
(it's also better in the more prominent position :+1:)

I tried a little to debug the gvim issue, that it opens with no content. By the time that gvim opens with the PR right now, the file is empty/gone. If I add a sleep(20) after the open, ZT pauses and the content does enter the editor... and if (and only if) I edit and save the file within 20s does the content transfer back to ZT.

Other than these minor issues, which we can document, I do wonder if using a context manager with the NamedTemporaryFile may be cleaner/simpler - it already can be used to write to and read from a file, and is guaranteed to always clean up.

ZULIP_EDITOR_COMMAND works well enough, and we should document it, though a config file option may be easier in the long run. We can document in the README, or maybe just placing everything into a FAQ entry like 'Can I compose messages in another editor?' might work well enough. That can include how to set it up, as well as what values are known to work (or not).

Re the TEMPFILE:LINE:CHARACTER point, do you mean passing eg. `/tmp/tmp12345filename:5:1` as an option? That can certainly be an option, and I'd welcome an issue being filed with anything that we don't complete here.

zulipterminal/ui_tools/boxes.py
```diff
@@ -808,6 +812,32 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
             self.view.controller.show_markdown_help()
             return key
+        elif is_command_key("OPEN_EXTERNAL_EDITOR", key):
+            editor = os.environ.get("ZULIP_EDITOR_COMMAND", os.environ.get("EDITOR"))
+            if editor is None:
+                self.view.controller.report_error(
+                    "Configure $EDITOR or $ZULIP_EDITOR_COMMAND shell environment."
+                )
+                return key
+            editor_splits = editor.split(" ")
```
andersk, line 822: Splitting on all spaces seems problematic, especially on platforms with crazy paths like Windows (`"C:\Program Files (x86)\Notepad++\Notepad++.exe"`) and macOS (`"/Applications/Sublime Text.app"`).

On POSIX, the right behavior would be `editor_splits = shlex.split(editor)` ([documentation](https://docs.python.org/3/library/shlex.html#shlex.split)). I’m not sure if Windows users would find it intuitive that this requires them to quote `"C:\Windows\notepad.exe"`, though (python/cpython#44990). `shlex.split(editor, posix=False)` doesn’t really help.

```pycon
>>> # unintuitive?
>>> shlex.split(r'"C:\Windows\notepad.exe"')
['C:\\Windows\\notepad.exe']
>>> # too much splitting
>>> shlex.split(r'C:\Program Files (x86)\Notepad++\Notepad++.exe', posix=False)
['C:\\Program', 'Files', '(x86)\\Notepad++\\Notepad++.exe']
>>> # extra quotes left behind
>>> shlex.split(r'"C:\Program Files (x86)\Notepad++\Notepad++.exe"', posix=False)
['"C:\\Program Files (x86)\\Notepad++\\Notepad++.exe"']
```

So I’m not sure what to suggest.

Somehow Python on Windows manages to guess the intended parsing of `subprocess.check_call(r'C:\Program Files (x86)\notepad++\notepad++.exe test.txt')`, but I don’t understand what rules it uses to do that, and I can’t imagine how they could be secure…

andersk: Maybe the [`oslex` library](https://pypi.org/project/oslex/) does the right thing? You’d think there’d be a library with more than one GitHub star for this.

neruyzo: I took a look at oslex, even with one star the code looks good. Should we add this dependency or copy the 50 like lines of code in some utils (license is MIT).

neruyzo: But even if we include the os detection in zulip we would still need [mslex](https://github.com/smoofra/mslex) dependency.

neiljp: Based on the examples above, it appears that `shlex.split` fails to behave as one might expect for *any* path with an unescaped space in eg. `Sublime Text.app some_arg` gives two arguments to `Sublime`. So in any case we'd need to note that users must be sure to escape spaces in file names, in these environment variables, if we used some form of `shlex.split`. eg. `export EDITOR=Sublime\ Text.app some_arg`
- Is that already necessary if one wants to use Sublime as the default $EDITOR on Mac?
- Surely other tools which use $EDITOR have similar parsing challenges?

The `oslex` and `mslex` docs lead to some brain-melting reading on windows path parsing! A potential simplification is that we don't support Windows natively (#357), only via WSL, so we're always expecting unix-like Python behavior (except in a few cases with temporary files). By default, executables in the windows path can be run from WSL, and I expect would also be executable via somewhere in eg. `/mnt/c/`, which would suggest we could rely on unix-like quoting rules in those cases. That behavior suggests `C:\...`-style paths inside WSL are not usable? There are some nested commands which could be more complex, maybe eg. `CMD.exe`, but indications are that MS aims to integrate win32/linux binaries into whatever environment you're in, so we may be able to just avoid the windows parsing pain when in WSL.

One wrinkle is that environment variables can be shared from windows to WSL, and don't always do automatic path translation, but we can document that.

So:
- `shlex.split` seems better than basic `split`, but still needs special care
- It would be good to confirm, but I don't believe we need to handle windows-style paths via `mslex`, since our supported platforms couldn't access them in any case
- Can we just document expected parsing challenges from shlex and WSL? Would that be insecure?
- Can we find inspiration from other applications using `$EDITOR`? Are they simpler, or also potentially insecure? Are they less cross-platform?

Other than trusting users follow the documentation, we could explicitly expect them to enter different parts of the command, ie. require variables for the command and (separate) arguments. That could be messy with multiple environment variables, particularly if trying to reuse `$EDITOR`, though simpler with configuration options.

neruyzo:
Update using shlex instead of oslex

neiljp:
@mek-yt Unfortunately this great PR got a bit buried with other work over the Summer, but seems essentially good to go?

Re-reading this:
- We don't support running on windows right now, but we could also explicitly block this on windows if path handling is incomplete (or of concern)
- Alternatively, or in addition, we could sit this behind a config-file options, so that users know that one (or a specific) environment variable is in use
- Following from the previous point, if possible then a possible followup - at least for editors opening in a different location - would be to display a popup before pausing, which could indicate to the user why the screen looks the way that it does? That could show the external editor name (path?), full command used (without temp-file?) and the environment variable it was taken from.

@andersk Do you have any outstanding concerns regarding this in its current state?

mek-yt:
I updated the code with setting from ZULIPRC in `editor` key, it should be ZULIPRC > `$ZULIP_EDITOR_COMMAND` > `$EDITOR` read order and print the source at launching. There are linting errors, I will check later if it is because of me.

mek-yt:
@neiljp everything should be done and clean. I updated the message on opening and add

```text
external editor command 'hx' specified in zuliprc file.
```

or with new config source `environment`

```text
external editor command 'hx' specified from environment.
``` 

if there is no config in zuliprc or environment the message is

```text
external editor command '' specified from environment.
```

mek-yt:
I fix the conflicts.
Don't hesitate to tell me if there are some improvements need to merge this feature.

neiljp:
@mek-yt Thanks for all your work on this, it's really appreciated!

I'm working on resolving some conflicts and changes necessary to integrate with other recent commits, and plan to merge soon!

neiljp:
@mek-yt Current changes that I will shortly push back here before merging include:
- Split of FAQ out into a second commit, with rewording & extra notes
- Commit text adjustments
- Command description, README, and error text adjustments
- Use some more descriptive variables, and some defensive checks to avoid potential errors
- Updates of informative text to better represent the code

Generally my preference would be for a more spread-out sequence of commits, eg. we don't have environment variables as configuration, so that could be a clean separate addition.

In an ideal world I'd also request more tests, but I'd prefer we get this merged and move on with this useful feature! :tada:

Issue #1002, opened 2021-04-21
Make standard scrollable popups shortcut keys more obvious


neiljp:
Currently we indicate in the help menu that `?` 'show/hide the help menu', and indicate that 'up/down scrolls' in the title - and similarly for many other standard popups.

This has a few issues:
* the popup title is slightly mixed with the scrolling hint
* except for the help menu, there is no hint how to exit a popup

Given that the popup title is now a bar, one solution to the first two is moving/adding these hints into the top corners of the popup.
For other ideas, please discuss in the stream first.

zulipbot:
Hello @zulip/server-hotkeys members, this issue was labeled with the "area: hotkeys" label, so you may want to check it out!

PR #1404, opened 2023-05-24
Displays user information for sender of message on u hotkey

### LINKED ISSUE #1397: Show user popup for sender of message on `u` hotkey
This would match the behavior in the web app.

Look for the way in which this popup is opened from the user list, and apply the same approach but using the sender of message (only if one is selected).

Note that this is not `U`, which will be a key we use for marking unread, assuming we keep the same shortcut as in the web app.

israelsgalaxy:
This PR adds functionality which displays user information for the sender of a message when the 'u' hotkey is pressed on an active message box. This would match the behavior in the web app.

- [x] Fully fixes #1397
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    
Updated UI

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -314,7 +314,9 @@ class Controller:
 
     def show_user_info(self, user_id: int) -> None:
         self.show_pop_up(
-            UserInfoView(self, user_id, "User Information (up/down scrolls)"),
+            UserInfoView(
+                self, user_id, "User Information (up/down scrolls)", "USER_INFO"
+            ),
             "area:user",
         )
 
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -320,6 +320,17 @@ class Controller:
             "area:user",
         )
 
+    def show_msg_sender_info(self, user_id: int) -> None:
+        self.show_pop_up(
+            UserInfoView(
+                self,
+                user_id,
+                "Message Sender Information (up/down scrolls)",
+                "MSG_SENDER_INFO",
+            ),
+            "area:user",
+        )
+
     def show_full_rendered_message(
         self,
         message: Message,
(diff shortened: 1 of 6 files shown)


zulipterminal/core.py
```diff
@@ -312,9 +312,9 @@ def show_about(self) -> None:
             "area:help",
         )
 
-    def show_user_info(self, user_id: int) -> None:
+    def show_user_info(self, user_id: int, cmd: str = "USER_INFO") -> None:
```
neiljp, line 315: This doesn't need to be a default value, and would be clearer if it were not.

There's also no need to abbreviate the text to 'cmd' - presumably this is command?

This is a refactor commit - it doesn't change behavior - so it's good to note it as such as an additional commit title area.

neiljp (commented):
@israelsgalaxy Thanks for looking at another issues - this is functional :+1:

Exploring using this, my only thought is that it may be useful to change the title of the popup to represent that this is the information of the message sender, not simply that of a selected user as before.

One of the things which may be contributing to my feeling about this is that now we have a popup coming from two different places (user list, message list) the key to exit this popup is no longer unique. So from a user perspective, it's not clear which key to use, and it isn't shown in the popup even though it differs. This suggests working on #1002 would be useful.

israelsgalaxy:
Thank you for the review @neiljp . I have changed the title of the popup to give users more clarity. I have also removed the default value in the show_user_info method.

neiljp:
@israelsgalaxy Thanks for the adjustment - I reordered the commits and slightly adjusted the commit text for accuracy and have pushed back here, and will merge now CI is complete :tada:

Other than #1002, as mentioned above, a more directly connected follow-up that would relate to this would be to explore turning the user information details in the message information into a single button which would also open on `u` from there, much like links. A simpler first approach would be to show it (and allow it to work) in the 'Viewing Actions' list there. It would likely be useful to turn those into buttons that work as part of the overall list of scrollable items which can be activated. So the less simple path of turning the user information into a button would fit with that direction, particularly since it already has a place in the popup.

PR #1405, opened 2023-05-25
Refactor: Zulip Feature Level simplified

### LINKED ISSUE #1364: Simplify Zulip Feature Level (ZFL) handling (treat absent as zero)
Historically handling features based on ZFL was introduced slowly, using `None` for when it was absent. However, for simple comparison purposes it is simpler if we move towards a pure integer representation, as recommended elsewhere.

This should be a fairly straightforward refactor, and enable simplification of quite a lot of conditionals.

Check the Model for where this is first set.

LucasC-B:

This issue was about changing the zulip feature level absence from None to zero.

Fixes #1364.

- [x] Fully fixes #1364
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/zulipterminal/api_types.py
+++ b/zulipterminal/api_types.py
@@ -568,8 +568,8 @@ class ServerSettings(TypedDict):
     # Added in Zulip 2.1.0
     external_authentication_methods: List[ExternalAuthenticationMethod]
 
-    # Refactor ZFL to default to zero DONE
-    zulip_feature_level: int  # New in Zulip 3.0, ZFL 1
+    # TODO Refactor ZFL to default to zero
+    zulip_feature_level: NotRequired[int]  # New in Zulip 3.0, ZFL 1
     zulip_version: str
     zulip_merge_base: NotRequired[str]  # New in Zulip 5.0, ZFL 88
 
--- a/zulipterminal/api_types.py
+++ b/zulipterminal/api_types.py
@@ -568,8 +568,8 @@ class ServerSettings(TypedDict):
     # Added in Zulip 2.1.0
     external_authentication_methods: List[ExternalAuthenticationMethod]
 
-    # TODO Refactor ZFL to default to zero
-    zulip_feature_level: NotRequired[int]  # New in Zulip 3.0, ZFL 1
+    # Refactor ZFL to default to zero DONE
+    zulip_feature_level: int  # New in Zulip 3.0, ZFL 1
     zulip_version: str
     zulip_merge_base: NotRequired[str]  # New in Zulip 5.0, ZFL 88
 
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -183,7 +183,7 @@ class Model:
         # NOTE: The date_created field of stream has been added in feature
         # level 30, server version 4. For consistency we add this field
         # on server iterations even before this with value of 0.
-        if self.server_feature_level == 0 or self.server_feature_level < 30:
+        if self.server_feature_level < 30:
             for stream in self.stream_dict.values():
                 stream["date_created"] = None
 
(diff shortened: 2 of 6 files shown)


zulipterminal/model.py
```diff
@@ -1686,7 +1685,7 @@ def _handle_update_message_flags_event(self, event: Event) -> None:
         Handle change to message flags (eg. starred, read)
         """
         assert event["type"] == "update_message_flags"
-        if self.server_feature_level is None or self.server_feature_level < 32:
+        if self.server_feature_level == 0 or self.server_feature_level < 32:
```
neiljp, line 1688: One of the benefits of converting to use zero instead of None is that we don't need multiple conditions in positions like this, since 0 < 32 too.

zulipterminal/model.py
```diff
@@ -583,7 +582,7 @@ def update_stream_message(
         if content is not None:
             request["content"] = content
 
-        if self.server_feature_level is not None and self.server_feature_level >= 9:
+        if self.server_feature_level != 0 and self.server_feature_level >= 9:
```
neiljp, line 585: This can be simplified by writing a single conditional expression
(https://docs.python.org/3/reference/expressions.html#comparisons)

zulipterminal/model.py
```diff
@@ -183,8 +182,8 @@ def __init__(self, controller: Any) -> None:
 
         # NOTE: The date_created field of stream has been added in feature
         # level 30, server version 4. For consistency we add this field
-        # on server iterations even before this with value of None.
-        if self.server_feature_level is None or self.server_feature_level < 30:
+        # on server iterations even before this with value of 0.
+        if self.server_feature_level == 0 or self.server_feature_level < 30:
```
neiljp, line 186: This original comment is not great, but this is a good time to adjust to say for earlier versions (in general, not just None or zero).

zulipterminal/model.py
```diff
@@ -164,12 +164,11 @@ def __init__(self, controller: Any) -> None:
         # Register to the queue before initializing further so that we don't
         # lose any updates while messages are being fetched.
         self._fetch_initial_data()
-
```
neiljp, line 167: This is an unrelated change. It's worth checking on GitHub before/after you submit/update a PR, or locally on your computer using eg. `tig` or other tools, for unrelated changes in a given commit/PR.

zulipterminal/api_types.py
```diff
@@ -568,8 +568,8 @@ class ServerSettings(TypedDict):
     # Added in Zulip 2.1.0
     external_authentication_methods: List[ExternalAuthenticationMethod]
 
-    # TODO Refactor ZFL to default to zero
-    zulip_feature_level: NotRequired[int]  # New in Zulip 3.0, ZFL 1
+    # Refactor ZFL to default to zero DONE
+    zulip_feature_level: int  # New in Zulip 3.0, ZFL 1
```
neiljp, line 572: For TODO/FIXME comments, these can be removed when complete.

Note that `NotRequired` is related to the typing, namely that it may not be present in the object (dict), which remains true until 3.0, so should remain.

For reference, we don't yet have a type for initial_data, which is where the model is obtaining the feature level from at run-time; this location is for unauthenticated pre-fetch of the server information, some of which we use when generating a user zuliprc for them.

neiljp (commented):
@LucasC-B Thanks for looking at this :+1:

I've left some initial feedback, but have not yet confirmed if you have updated all the necessary locations.

zulipterminal/api_types.py
neiljp: To clarify, as part of this change it is fine to remove this comment entirely. That is, this PR should end up changing our default to zero.

neiljp (commented):
@LucasC-B Thanks for the update :+1:

Apologies for the delay. I left further comments, not all requiring changes. The biggest change left is likely that not included yet, related to the constants in version.py, which affect tests and from there suggest that other code changes may also be necessary.

Issue #1337, opened 2023-03-19
Add type hints to `test_messages.py`


neiljp:
Many of our tests have type hints, primarily on fixtures and the test function parameters, but we have some remaining exclusions, as noted in #1229.

It is worth noting that consistency between elements are *not all* checked by mypy or pytest right now, so attention to detail will be important.

This *could* be an ideal first issue for you, but likely **only if you have a background in Python Type Hints and pytest**.

That said, the parameters to these test functions generally come in one of 4 categories:
- parameters with default values: that makes determining the type to use more straightforward
- parameters from a parametrize decorator: check the data in the list of values above the function name
- our fixtures: these will be in the test file, or in `conftest.py`; look for the return type
- other fixtures: this is something like `mocker`, which is a `MockerFixture` from `pytest_mock`

While working, you'll want to disable excluding the file from being checked by mypy by removing that line from `pyproject.toml` and adding to a separate commit before anything else. Check the commit log for the style used for other typing of test files (mainly by prah23).

It's likely worth working on one test function at a time in a separate commit, which we may squash together before merging. Once the file fully passes mypy cleanly, we can then move the first commit to the end, so that mypy doesn't complain before the file is fixed. That commit will then ensure that changes to that file will require type hints in future.

> This is one element of #1229: please focus on this **one** file of that issue first, and if things go well, then you would be well situated to cover the remaining files later as followup PR(s). If you wish to work on a different file then please discuss in **#zulip-terminal** first.

zulipbot:
Hello @zulip/server-refactoring members, this issue was labeled with the "area: refactoring" label, so you may want to check it out!

abdulmoiz37:
@zulipbot claim

zulipbot:
Welcome to Zulip, @abdulmoiz37! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

[...]

PR #1407, opened 2023-05-31
Add type hints to test_message.py

SBNetto01:
This PR adds the adequate type hints in each function from test_message.py, solving issue #1337 

- [x] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1,11 +1,15 @@
 from collections import OrderedDict, defaultdict
 from datetime import date
+from typing import List, Tuple, Any, Optional, Dict, Callable
+from pytest_mock import MockerFixture
+from zulipterminal.api_types import MessageType, Message
+from zulipterminal.urwid_types import urwid_Size
 
 import pytest
 import pytz
 from bs4 import BeautifulSoup
 from pytest import param as case
-from urwid import Columns, Divider, Padding, Text
+from urwid import Columns, Divider, Padding, Text, Widget
 
 from zulipterminal.config.keys import keys_for_command
 from zulipterminal.config.symbols import (
@@ -25,7 +29,7 @@ SERVER_URL = "https://chat.zulip.zulip"
 
 class TestMessageBox:
     @pytest.fixture(autouse=True)
-    def mock_external_classes(self, mocker, initial_index):
+    def mock_external_classes(self, mocker: MockerFixture, initial_index: int) -> None:
         self.model = mocker.MagicMock()
         self.model.index = initial_index
 
@@ -36,7 +40,12 @@ class TestMessageBox:
             ("private", [("email", ""), ("user_id", None)]),
         ],
     )
-    def test_init(self, mocker, message_type, set_fields):
+    def test_init(
+        self, 
+        mocker: MockerFixture, 
+        message_type: str, 
+        set_fields: List[Tuple[str, Any]]
+        ) -> None:
         mocker.patch.object(MessageBox, "main_view")
         message = dict(
             display_recipient=[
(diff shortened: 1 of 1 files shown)


neiljp:
@SBNetto01 Thanks for looking into this :+1:

Did you see the other PR for this? I think there is only one other?

Were you wanting feedback at this point, or unsure how to proceed given the failing linting/tests?

Do you understand why the linting and tests are failing? I'd be happy to discuss briefly here, or chat in more depth on chat.zulip.org if you need help with this.

SBNetto01:
Thank you for your reply and sorry about the delay! I was waiting on some feedback, since the failed linting/tests are a bit weird. I wasn't able to use the --fix recommended on a few of the documentations, and all of them are related to "empty spaces" if I'm not mistaken, but it's not related to my modifications whatsoever. Would love to know how to proceed with this.

tests/ui_tools/test_messages.py
```diff
@@ -36,7 +40,12 @@ def mock_external_classes(self, mocker, initial_index):
             ("private", [("email", ""), ("user_id", None)]),
         ],
     )
-    def test_init(self, mocker, message_type, set_fields):
+    def test_init(
+        self, 
+        mocker: MockerFixture, 
+        message_type: str,
```
neiljp, line 46: For `message_type`, we likely want the new `MessageType` in `api_types.py`; that's also a string, but a more constrained type.

tests/ui_tools/test_messages.py
```diff
@@ -945,13 +975,13 @@ def test_main_view_generates_PM_header(
...
-        mocker,
-        messages_successful_response,
-        msg_type,
-        msg_narrow,
-        assert_header_bar,
-        assert_search_bar,
-    ):
+        mocker: MockerFixture,
+        messages_successful_response: Dict[str, Any],
+        msg_type: MessageType,
```
neiljp, line 980: Ah, I see you used `MessageType` here - but note that the variable type is different here, which I can tell from reading the parametrize (as per other comments).

tests/ui_tools/test_messages.py
```diff
@@ -945,13 +975,13 @@ def test_main_view_generates_PM_header(
...
+        msg_narrow: list,
```
neiljp, line 981: I think mypy will complain if you use a list with no typing, even if that typing is `Any` (which we may need here) - and with our current setup we likely need `List` too.

tests/ui_tools/test_messages.py
```diff
@@ -1827,8 +1873,12 @@ def test_reactions_view(
         ],
     )
     def test_footlinks_view(
-        self, message_links, expected_text, expected_attrib, expected_footlinks_width
-    ):
+        self, 
+        message_links: List[str], 
+        expected_text: Optional[str], 
+        expected_attrib: Optional[Dict[str, Any]],
+        expected_footlinks_width: Optional[int]
```
neiljp, line 1880: These need you to understand parametrize.

neiljp (commented):
@SBNetto01 I've unfortunately only just got back to this, but here are some details if you're still interested!

I left some details inline - generally you seem to have a good grasp of where the typing goes in the code and what some of the types are, but others appear to not match at all - did you use a tool for this? As I mention inline, if you can understand how pytest uses parametrize better then I suspect you'll be able to determine a lot of the types more fully.

---

One challenge that I forgot from the previous change where we typed other tests was moving from a simple Dict to a TypedDict (like Message). The main challenge with those is that historically we used an approach of base data that is updated in various different ways in each test case, but mypy doesn't like using certain general dict operations where literal keys are not involved.

So, you may wish to keep some types looser (eg. Dict[str, Any] for what may become Message) in a first pass if that helps you get further, and then we can refine from there.

For reference, the general solution for the latter case generally involves updating the tests:
a) to refer to specific literal strings as keys into the dict (where possible), or
b) using factory functions/fixtures where the desired values are passed in (constructed as required, rather than modified later)

If you're on chat.zulip.org, do start a topic in #zulip-terminal if you're interested :)

PR #1387, opened 2023-04-24
Handle (un)subscribe events

theViz343:
### What does this PR do, and why?
This PR adds event handling for (un)subscribe events. Without event handling, ZT crashes
if the user narrows onto a message in a stream which is unsubscribed from. This happens mostly 
when a user is logged in from multiple clients, but can also occur if an admin unsubscribes the user.

### Outstanding aspect(s)    

- [ ] Updated subscribed streams frequently have the wrong/default color. (This seems to be an issue of
ZT not loading the colors.)
- [x] ZT crashes when the currently narrowed message is unsubscribed from.
- [ ] Should we notify user of the new (un)subscription? If yes, then where?(As a popup or as a message in the footer)

### External discussion & connections

- [x] Discussed in **#zulip-terminal** in `Adding streams throws key error`
- [x] Fully fixes #816
- [ ] Partially fixes issue #
- [x] Builds upon previous unmerged work in PR #1408 
- [ ] Is a follow-up to work in PR #
- [ ] Requires merge of PR #
- [ ] Merge will enable work on #

### How did you test this?

- [x] Manually - Behavioral changes
- [x] Manually - Visual changes
- [x] Adapting existing automated tests
- [x] Adding automated tests for new behavior (or missing tests)
- [ ] Existing automated tests should already cover this (*only a refactor of tested code*)

### Self-review checklist for each commit
- [x] It is a [minimal coherent idea](https://github.com/zulip/zulip-terminal#structuring-commits---speeding-up-reviews-merging--development)
- [x] It has a commit summary following the [documented style](https://github.com/zulip/zulip-terminal#structuring-commits---speeding-up-reviews-merging--development) (title & body)
- [x] It has a commit summary describing the  motivation and reasoning for the change
- [x] It individually passes linting and tests
- [x] It contains test additions for any new behavior
- [x] It flows clearly from a previous branch commit, and/or prepares for the next commit

### Visual changes    

[Preview](https://asciinema.org/a/qSfrl15TUAozL0kmLAb6kmfim)

PR #1408, opened 2023-06-08
Add support for all streams

theViz343:
This PR adds support for unsubscribed streams and never_subscribed streams in ZT. It creates new data structures and accessor methods to access them. This PR is needed for further work on reducing stream related issues and adding support for stream (un)subscription.

- [x] Discussed in **#zulip-terminal** in `Adding streams throws key error`
- [ ] Fully fixes #
- [x] Partially fixes issue #816 
- [ ] Requires merge of PR #
- [x] Merge will enable work on #1387 

### How did you test this?

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1973,6 +1973,89 @@ class TestModel:
         assert model.is_stream_web_public(stream_id) == expected_value
         model._get_stream_from_id.assert_called_once()
 
+    @pytest.mark.parametrize(
+        "stream_id, to_vary, expected_value",
+        [
+            case(
+                1000,
+                {
+                    "message_retention_days": None,
+                },
+                None,
+            ),
+            case(
+                3,
+                {
+                    "message_retention_days": 33,
+                },
+                33,
+            ),
+            case(
+                5,
+                {
+                    "message_retention_days": 35,
+                },
+                35,
+            ),
+        ],
+    )
+    def test_get_stream_message_retention_days(
+        self,
+        model,
+        mocker,
+        get_stream_from_id_fixture,
+        stream_id,
+        to_vary,
+        expected_value,
+    ):
+        get_stream_from_id_fixture.update(to_vary)
+        mocker.patch(
+            MODEL + "._get_stream_from_id", return_value=get_stream_from_id_fixture
+        )
+        assert model.get_stream_message_retention_days(stream_id) == expected_value
+        model._get_stream_from_id.assert_called_once()
+
+    @pytest.mark.parametrize(
+        "stream_id, to_vary",
+        [
+            case(
+                1000,
+                {
+                    "message_retention_days": None,
+                },
+            ),
+            case(
+                3,
+                {
+                    "message_retention_days": 33,
+                },
+            ),
+            case(
+                5,
+                {
+                    "message_retention_days": 35,
+                },
+            ),
+        ],
+    )
+    def test_set_stream_message_retention_days(
+        self,
+        model,
+        mocker,
+        get_stream_from_id_fixture,
+        stream_id,
+        to_vary,
+        value_to_be_changed=30,
+    ):
+        get_stream_from_id_fixture.update(to_vary)
+        mocker.patch(
+            MODEL + "._get_stream_from_id", return_value=get_stream_from_id_fixture
+        )
+
+        model.set_stream_message_retention_days(stream_id, value_to_be_changed)
+        model._get_stream_from_id.assert_called_once()
+        assert model.get_stream_message_retention_days(stream_id) == 30
+
     @pytest.mark.parametrize("muted", powerset([99, 1000]))
     @pytest.mark.parametrize("visual_notification_enabled", powerset([99, 1000]))
     def test__subscribe_to_streams(
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -2248,6 +2248,37 @@ class TestModel:
         assert model.get_stream_weekly_traffic(stream_id) == expected_value
         model._get_stream_from_id.assert_called_once()
 
+    @pytest.mark.parametrize(
+        "stream_id, expected_value",
+        [
+            case(
+                1000,
+                "General Stream",
+            ),
+            case(
+                3,
+                "A description of stream 3",
+            ),
+            case(
+                5,
+                "A description of stream 5",
+            ),
+        ],
+    )
+    def test_get_stream_rendered_description(
+        self,
+        model,
+        mocker,
+        get_stream_from_id_fixture,
+        stream_id,
+        expected_value,
+    ):
+        mocker.patch(
+            MODEL + "._get_stream_from_id", return_value=get_stream_from_id_fixture
+        )
+        assert model.get_stream_rendered_description(stream_id) == expected_value
+        model._get_stream_from_id.assert_called_once()
+
     @pytest.mark.parametrize("muted", powerset([99, 1000]))
     @pytest.mark.parametrize("visual_notification_enabled", powerset([99, 1000]))
     def test__subscribe_to_streams(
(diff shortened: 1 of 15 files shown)


zulipterminal/model.py
```diff
@@ -1243,6 +1243,13 @@ def get_subscription_color(self, stream_id: int) -> Optional[str]:
             return self._unsubscribed_streams[stream_id]["color"]
         return None
 
+    def get_subscription_email(self, stream_id: int) -> Optional[str]:
+        if stream_id in self.stream_dict:
+            return self.stream_dict[stream_id]["email_address"]
+        elif stream_id in self._unsubscribed_streams:
+            return self._unsubscribed_streams[stream_id]["email_address"]
+        return None
```
prah23, line 1251: Since this method is only being used in the `StreamInfoView` and you haven't added the feature of viewing stream info for a never subscribed stream, can the base condition of returning `None` be avoided? The return value of this method is currently attached to a copy command, and having to copy a `None` to the clipboard doesn't seem appropriate.

prah23 (changes requested):
Thanks for the PR @theViz343!
This is a neat refactor; I've just left some minor points of feedback.

zulipterminal/model.py
theViz343: I've made it so that the method raises a RuntimeError if the stream is not a subscription (or does not exist).

tests/model/test_model.py
```diff
@@ -1779,6 +1779,32 @@ def test_get_all_stream_ids(
...
+        ],
+    )
+    def test_get_stream_name(
+        self, model, mocker, get_stream_from_id_fixture, stream_id, expected_value
+    ):
+        mocker.patch(
+            MODEL + "._get_stream_from_id", return_value=get_stream_from_id_fixture
+        )
+        assert model.get_stream_name(stream_id) == expected_value
+        model._get_stream_from_id.assert_called_once()
```
Subhasish-Behera, line 1813: can `assert_called_once_with` be used here?

Subhasish-Behera:
@theViz343  Thanks for working on this! This seems to work well and was pretty easy to understand because of the commit structure and commit messages.
Apart from that, I have some minor code suggestions in the comments.

zulipterminal/model.py
```diff
@@ -1215,6 +1215,9 @@ def set_stream_message_retention_days(
     ) -> None:
         self._get_stream_from_id(stream_id)["message_retention_days"] = value
 
+    def is_stream_invite_only(self, stream_id: int) -> bool:
+        return self._get_stream_from_id(stream_id)["invite_only"]
```
neiljp, line 1222: Similarly to is_web_public, we already have a method for this, which you've updated to use this accessor.

There is no use of this data outside of the model in any case, other than through stream_access_type(), so there is no motivation for introducing this accessor.

zulipterminal/ui_tools/views.py
```diff
@@ -1292,8 +1292,10 @@ def __init__(self, controller: Any, stream_id: int) -> None:
...
         ]
 
-        if "stream_post_policy" in stream:
-            stream_policy = STREAM_POST_POLICY[stream["stream_post_policy"]]
+        if controller.model.get_stream_post_policy(self.stream_id):
+            stream_policy = STREAM_POST_POLICY[
+                controller.model.get_stream_post_policy(self.stream_id)
+            ]
         else:
             if stream.get("is_announcement_only"):
```
neiljp: This existing UI code incorporates implicit checking of the server version, which would be better wrapped into the model under `get_stream_post_policy`, rather than the two functions introduced in this commit and the next.

zulipterminal/model.py
```diff
@@ -174,7 +174,7 @@ def __init__(self, controller: Any) -> None:
 
         self.users = self.get_all_users()
 
-        self.stream_dict: Dict[int, Any] = {}
+        self.stream_dict: Dict[int, Subscription] = {}
```
neiljp: This change, and possibly others in this commit seem much more related to those earlier in the PR than here at the end?

zulipterminal/ui_tools/views.py
```diff
@@ -1268,7 +1268,7 @@ class StreamInfoView(PopUpView):
     def __init__(self, controller: Any, stream_id: int) -> None:
         self.stream_id = stream_id
         self.controller = controller
-        stream = controller.model.stream_dict[stream_id]
+        stream = controller.model._subscribed_streams[stream_id]
```
neiljp, line 1271: This suggests we haven't covered accessing the stream or subscription properties sufficiently?

neiljp (commented):
@theViz343 This has grown quite a lot since I last looked! I started a review of this before Hari, so apologies if this overlaps a little. The end result is great, though I think we can shrink it a little.

The PR seems to be split into two parts:
- adding the un/never-subscribed data, improving the data types, and generalizing the current code to avoid eg. the KeyErrors
- adding the encapsulation via the accessors, followed by the rename

Both of these parts have value, but I'd suggest grouping the first set together at the start so that we can consider them together, and so potentially merge the earlier commits in this PR first.

There are a lot of commits, particularly for the encapsulating functions, and isolating adding different methods and tests per commit is great. However, some of these functions are not used outside the model, and don't appear to be preparing for anything, or the effect is replicated in other functions already. You also have some 'setters', which we likely don't need for users of the model.

With a smaller set of new methods I think the changes will be clearer, but it may be worth discussing whether adding so many functions is the way to go, or rather using more general stream_properties and subscription_properties functions?

zulipterminal/api_types.py
```diff
@@ -197,37 +197,26 @@ class Message(TypedDict, total=False):
...
 
 
-class Subscription(TypedDict):
+class Stream(TypedDict):
     stream_id: int
     name: str
     description: str
     rendered_description: str
-    date_created: int  # NOTE: new in Zulip 4.0 / ZFL 30
+    date_created: Optional[int]  # NOTE: new in Zulip 4.0 / ZFL 30
```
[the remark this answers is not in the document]
theViz343: Yes, the part where we add the `date_created` key and set it to None fails the typecheck. Since we get the data from the server initially, mypy doesn't catch that `date_created` is not present. We could account for this by using `NotRequired[Optional[int]]` instead, but that would purely be for documentation since it passes the typecheck with just `Optional[int]`. I'll also add a comment.

zulipterminal/model.py
```diff
@@ -1176,6 +1176,22 @@ def _register_non_subscribed_streams(
...
+        if stream_id in self.stream_dict:
+            return self.stream_dict[stream_id]
+        elif stream_id in self._unsubscribed_streams:
+            return self._unsubscribed_streams[stream_id]
+        elif stream_id in self._never_subscribed_streams:
+            return self._never_subscribed_streams[stream_id]
+        else:
+            raise RuntimeError(f"Stream with id {stream_id} does not exist!")
+
+    def get_all_stream_ids(self) -> List[int]:
```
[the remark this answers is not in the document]
theViz343: Yeah, you're right, both are helper functions and are used internally only. I'll change the commit description to indicate that, and also maybe change the name for the `get_all_stream_ids` function to `_get_all_stream_ids`.

zulipterminal/model.py
```diff
@@ -1156,6 +1164,18 @@ def user_name_from_id(self, user_id: int) -> str:
 
         return self.user_dict[user_email]["full_name"]
 
+    def _register_non_subscribed_streams(
+        self,
+        unsubscribed_streams: List[Subscription],
+        never_subscribed_streams: List[Stream],
+    ) -> None:
+        self._unsubscribed_streams = {
+            stream["stream_id"]: stream for stream in unsubscribed_streams
```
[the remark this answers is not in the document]
neiljp: Resolved in #1419 :+1:

zulipterminal/model.py
```diff
@@ -174,11 +175,18 @@ def __init__(self, controller: Any) -> None:
         self.users = self.get_all_users()
 
         self.stream_dict: Dict[int, Any] = {}
```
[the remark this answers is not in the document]
neiljp: Resolved in #1419 :+1:

Issue #1332, opened 2023-03-18
TRACKING: Improve algorithm for 'next unread topic' (`n`)


neiljp:
For a small number of unreads, this feature works reasonably well, and #1245 migrated older work to skip muted topics and streams.

However, as discussed in the related [topic](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/.E2.9C.94.20Skip.20muted.20topics.20and.20streams.20on.20'n'.20.23T442.20.23T1239.20.23T1245) and elsewhere, the current algorithm for next-unread-topic, via the <kbd>n</kbd> hotkey, is lacking in various ways.

These include:
- [x] there is no context based on the current narrow/message, but rather separate state stored in the model; this leads to various behavior, including typically jumping out of the current stream, even if it still has unread topics: one must fully read that stream via this method first [**fixed in #1356**]
- [x] the ordering of moving between streams is based on *stream id*, not *stream name*; this causes the order of streams being fully read to jump around and not match the order in the left panel (if visible) [**fixed in #1412**]
- [ ] #1582

These elements are obvious deviations from the behavior of the web app, but others may be present.

PR #1412, opened 2023-06-23
Use stream name instead of stream id to order unread topics in the next_unread_topic_from_message_id algorithm.

theViz343:

This PR sorts the list of unread topics using stream name instead of stream id. This makes the order of moving between streams similar to that in the left panel.

- [x] Discussed in **#zulip-terminal** in [`Improve algorithm for 'next unread topic' (n)`](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Improve.20algorithm.20for.20'next.20unread.20topic'.20.28n.29.20.23T1333.20.23T1356)
- [ ] Fully fixes #
- [x] Partially fixes issue #1332 
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

[asciicast](https://asciinema.org/a/ML0Qj2WTSh5krcyCILzyzqdnv)

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -256,7 +256,7 @@ def test_powerset(
         ),
         case(
             {(999, "topic3"): 1, (1000, "topic2"): 1, (1, "topic1"): 1},
-            [(1, "topic1"), (999, "topic3"), (1000, "topic2")],
+            [(1000, "topic2"), (1, "topic1"), (999, "topic3")],
             id="multiple_unread_topics",
         ),
         case(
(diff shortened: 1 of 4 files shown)


neiljp:
@theViz343 Gave this a try - left more feedback in the stream :)

neiljp:
@theViz343 I updated the labels since it doesn't appear to be in a state requiring feedback - the tests are failing?

neiljp (commented):
@theViz343 Thanks for the update - this looks to work better now :+1:

After following up on my feedback, I think this should be easily mergeable.

Other than the inline comments, my only other thought was to let you consider how the commit structure may be clearer if done slightly differently. For example: a refactor to extract the current sorting behavior into a function (with simple tests), followed by then updating that function and tests for the new behavior. That's not critical here, but it helps with isolating what each change does, for someone reviewing a commit now or later. In this case, a reader would know that the first refactor commit should just be moving code around and be able to focus on confirming it only does that, then easily see a side by side comparison of the change in sort behavior in the function in the second commit - as well as noting that the behavior has changed or been extended, since the tests would change.

neiljp:
@theViz343 Thanks - this was good to go! :tada:

I only made a few very minor changes to the commit text and added an assert for the extra stream_dict data you added.

PR #1414, opened 2023-07-04
upload: Enable File Upload for users.

### LINKED ISSUE #529: Uploading/Attaching Files
How can we attach files with messages? I believe this is not possible right now. Is this possible in future or being worked upon?

Subhasish-Behera:

- [x] Discussed in **#zulip-terminal** in `File Upload #T529 #T1414`
- [x] Fully fixes #529 
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1,5 +1,7 @@
 import copy
 import json
+import os
+import tempfile
 from collections import OrderedDict
 from copy import deepcopy
 from typing import Any, List, Optional, Tuple
@@ -809,6 +811,49 @@ class TestModel:
             self.notify_if_message_sent_outside_narrow.assert_called_once_with(
                 req, self.controller
             )
+
+    @pytest.mark.parametrize(
+        "file_name, upload_result, expected_result",
+        [
+            case(
+                "existing_file.txt",
+                {"result": "success", "uri": "http://example.com/success_uri"},
+                "http://example.com/success_uri",
+                id="exisiting_file_with_successful_response",
+            ),
+            case(
+                "existing_file.txt",
+                {"result": "failure", "error_message": "Upload failed"},
+                None,
+                id="exisiting_file_with_unsuccessful_response",
+            ),
+            case(
+                "non_existing_file.txt",
+                None,
+                None,
+                id="non_exisiting_file_with_no_response",
+            ),
+        ],
+    )
+    def test_get_file_upload_uri(
+        self, mocker, model, file_name, upload_result, expected_result
+    ):
+        self.client.upload_file = mocker.Mock(return_value=upload_result)
+        with tempfile.TemporaryDirectory() as temp_dir:
+            if upload_result is not None:
+                temp_file_path = os.path.join(temp_dir, file_name)
+                with open(temp_file_path, "w") as temp_file:
+                    temp_file.write("Test content")
+            else:
+                temp_file_path = f"random_path/{file_name}"
+
+            result = model.get_file_upload_uri(temp_file_path)
+
+        if upload_result is not None:
+            self.client.upload_file.assert_called_once()
+        else:
+            self.client.upload_file.assert_not_called()
+        assert result == expected_result
 
     @pytest.mark.parametrize(
         "response, return_value",
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -1,15 +1,19 @@
 from collections import OrderedDict
+from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
+from unittest.mock import MagicMock
 
 import pytest
 from pytest import param as case
 from pytest_mock import MockerFixture
 from urwid import Columns, Pile, Text, Widget
 
+from tests.ui_tools.test_boxes import TestWriteBox
 from zulipterminal.api_types import Message
 from zulipterminal.config.keys import is_command_key, keys_for_command
 from zulipterminal.config.ui_mappings import EDIT_MODE_CAPTIONS
 from zulipterminal.helper import CustomProfileData, TidiedUserInfo
+from zulipterminal.ui_tools.boxes import WriteBox
 from zulipterminal.ui_tools.messages import MessageBox
 from zulipterminal.ui_tools.views import (
     AboutView,
@@ -17,6 +21,7 @@ from zulipterminal.ui_tools.views import (
     EditHistoryView,
     EditModeView,
     EmojiPickerView,
+    FileUploadView,
     FullRawMsgView,
     FullRenderedMsgView,
     HelpView,
@@ -870,6 +875,90 @@ class TestMarkdownHelpView:
         self.markdown_help_view.keypress(size, key)
 
         assert self.controller.exit_popup.called
+
+
+class TestFileUploadView:
+    @pytest.fixture(scope="class")
+    def write_box(self) -> Any:
+        return TestWriteBox().write_box
+
+    @pytest.fixture(autouse=True)
+    def mock_external_classes(self, mocker: MockerFixture, write_box: WriteBox) -> None:
+        self.controller = mocker.Mock()
+        mocker.patch.object(
+            self.controller, "maximum_popup_dimensions", return_value=(64, 64)
+        )
+        mocker.patch(LISTWALKER, return_value=[])
+        self.file_upload_view = FileUploadView(
+            self.controller, write_box, "Upload File"
+        )
+
+    def test_keypress_any_key(
+        self, widget_size: Callable[[Widget], urwid_Size]
+    ) -> None:
+        key = "a"
+        size = widget_size(self.file_upload_view)
+        self.file_upload_view.keypress(size, key)
+        assert not self.controller.exit_popup.called
+
+    @pytest.mark.parametrize("key", {*keys_for_command("GO_BACK")})
+    def test_keypress_exit_popup(
+        self, key: str, widget_size: Callable[[Widget], urwid_Size]
+    ) -> None:
+        size = widget_size(self.file_upload_view)
+        self.file_upload_view.keypress(size, key)
+        assert self.controller.exit_popup.called
+
+    @pytest.mark.parametrize(
+        "file_location, expected_uri, expected_error_message",
+        [
+            case(
+                "example.txt",
+                "http://example.txt/uploaded_file",
+                None,
+                id="txt_file_with_successful_uri",
+            ),
+            case(
+                "example.pdf",
+                "http://example.pdf/uploaded_file",
+                None,
+                id="pdf_file_with_successful_uri",
+            ),
+            case(
+                "invalid.txt",
+                "",
+                ["ERROR: Unable to get the URI"],
+                id="invalid_txt_file_with_error",
+            ),
+            case(
+                "invalid.pdf",
+                "",
+                ["ERROR: Unable to get the URI"],
+                id="invalid_pdf_file_with_error",
+            ),
+        ],
+    )
+    def test__handle_file_upload(
+        self,
+        file_location: str,
+        expected_uri: str,
+        expected_error_message: Optional[str],
+    ) -> None:
+        self.file_upload_view.write_box = MagicMock()
+        self.controller.model.get_file_upload_uri.return_value = expected_uri
+
+        self.file_upload_view._handle_file_upload(file_location)
+
+        self.controller.model.get_file_upload_uri.assert_called_once_with(file_location)
+        if not expected_error_message:
+            file_name = Path(file_location).name
+            self.file_upload_view.write_box.append_uri_and_filename.assert_called_once_with(
+                file_name, self.file_upload_view.uri
+            )
+        else:
+            self.controller.append_uri_and_filename.assert_not_called()
+            self.controller.report_error.assert_called_with(expected_error_message)
+        self.controller.exit_popup.assert_called()
 
 
 class TestHelpView:
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -29,12 +29,14 @@ from zulipterminal.helper import asynch, suppress_output
 from zulipterminal.model import Model
 from zulipterminal.platform_code import PLATFORM
 from zulipterminal.ui import Screen, View
+from zulipterminal.ui_tools.boxes import WriteBox
 from zulipterminal.ui_tools.utils import create_msg_box_list
 from zulipterminal.ui_tools.views import (
     AboutView,
     EditHistoryView,
     EditModeView,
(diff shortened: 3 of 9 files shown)


zulipterminal/core.py
```diff
@@ -274,6 +291,11 @@ def show_msg_info(
         )
         self.show_pop_up(msg_info_view, "area:msg")
 
+    def show_file_upload_popup(self) -> None:
+        file_upload_view = FileUploadView(self, "Upload File(Enter the location)")
+        self.show_pop_up(file_upload_view, "area:msg")
+
+        return ""  # Supposed to return the Uri
```
mounilKshah, line 298: There doesn't seem a need for the return statement at the bottom.

zulipterminal/ui_tools/views.py
```diff
@@ -1192,6 +1192,11 @@ def __init__(
             # urwid.Pile(msg_box.footer),
         )
 
+    def _handle_file_upload(self, file_location):
+        file_location1 = file_location
+        self.uri = self.model.get_file_upld_uri(file_location1)
+        self.controller.set_uri(self.uri)  # Update the uri in the controller
```
mounilKshah, line 1198: This function can be called when `keypress()` detects `ENTER` and the popup can be closed upon encountering this event.

zulipterminal/ui_tools/views.py
```diff
@@ -1168,7 +1168,32 @@ def __init__(self, controller: Any, title: str) -> None:
...
+            columns,
+            "FILE_UPLOAD",
+            max_cols,
+            title,
+            # urwid.Pile(msg_box.header),
+            # urwid.Pile(msg_box.footer),
+        )
 
+    def keypress(self, size: urwid_Size, key: str) -> str:
+        return super().keypress(size, key)
```
mounilKshah, line 1201: Maybe execute a function/some logic when the user hits `ENTER`

zulipterminal/core.py
```diff
@@ -109,6 +118,15 @@ def __init__(
...
         signal.signal(signal.SIGINT, self.exit_handler)
 
+
+    @property
+    def uri(self) -> str:
+        return self._uri
+
+    def set_uri(self, value: str) -> None:
+        self._uri = value
+        self.uri_updated_event.set()
```
mounilKshah, line 128: Instead of simply setting the URI, is it possible for you to add the link generated for the uploaded file and append it to the `write_box`? Maybe that can eliminate the need to wait for the upload API to provide response.

mounilKshah (commented):
Thank you for working on this @Subhasish-Behera
I ran the code from the PR on my local machine. I was able to view the link generated when a file is uploaded by entering the path in the popup.

Further steps can be manipulating the `write_box` such that the generated link gets attached in the message editor.

zulipterminal/ui_tools/views.py
Subhasish-Behera: @mounilKshah  The `FileUploadView` is derived from `PopUpView`. `PopUpView` is supposed to get closed on pressing the same key which opened. So I think it is natural that the user wants to close as soon as he has uploaded the file.
As Neil is suggesting considering using something different [(relevant chat)](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/File.20Upload.20.23T529.20.23T1414/near/1607421) than `PopUpView`, we should first discuss a bit about the Base classes to inherit from and then come back to chose appropriate keys and their actions.

zulipterminal/ui_tools/boxes.py
```diff
@@ -710,6 +710,14 @@ def autocomplete_emojis(
 
         return emoji_typeahead, emojis
 
+    def insert_uri(self, file_name: str, uri: str) -> None:
+        edit_widget = self.contents[self.FOCUS_CONTAINER_MESSAGE][
+            self.FOCUS_MESSAGE_BOX_BODY
+        ]
+        edit_widget.edit_text += f"[{file_name}](/{str(uri)})"
```
mounilKshah, line 717: Upon sending the message  and opening the file, the URL wasn't able to fetch the file. This may be because of the extra `/` before the `uri`, because the URI returned from the server already has a `/` at the start.

Also, the function already has the type of `uri` as `str`, so there might be no need to typecast it to string.

zulipterminal/model.py
```diff
@@ -557,8 +557,8 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
             notify_if_message_sent_outside_narrow(composition, self.controller)
         return message_was_sent
 
-    def get_file_upld_uri(self, file_loctn: str) -> str:
-        with open(file_loctn, "rb") as fp:
+    def get_file_upld_uri(self, file_loc: str) -> str:
+        with open(file_loc, "rb") as fp:
```
mounilKshah, line 561: I think this can be added to the second commit, where the function has been declared?

zulipterminal/ui_tools/views.py
```diff
@@ -1169,6 +1169,39 @@ def __init__(self, controller: Any, title: str) -> None:
...
+            controller,
+            columns,
+            "FILE_UPLOAD",
+            max_cols,
+            title,
+        )
+
+    def _handle_file_upload(self, file_location: str) -> None:
+        self.uri = self.model.get_file_upld_uri(file_location)
+        file_name = file_location.split("/")[-1]
```
mounilKshah, line 1196: This works for now, but I think in case of Windows, the file path contains [backward slashes](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats), and this approach may not work there.

mounilKshah (changes requested):
This version of the PR is much closer to the bare bones approach which I had mentioned on the stream earlier. The commit structure looks fine to me. I ran this on my local machine, and it worked fine when the extra `/` was not present in the appended URL.

Also, I think it was a good idea to change the position of the cursor after the URL is appended to the compose box content.

zulipterminal/core.py
```diff
@@ -274,6 +276,12 @@ def show_msg_info(
         )
         self.show_pop_up(msg_info_view, "area:msg")
 
+    def show_file_upload_popup(self, write_box: WriteBox) -> None:
+        file_upload_view = FileUploadView(
+            self, write_box, "Upload File(Enter the location)"
```
neiljp, line 281: There's no space before the parenthesis here, but let's keep the title simple and drop those and the text inside. The action should be clear from the internals of the popup?

zulipterminal/ui_tools/views.py
```diff
@@ -1169,6 +1169,39 @@ def __init__(self, controller: Any, title: str) -> None:
...
+        self.write_box.insert_uri(file_name, self.uri)
+
+    def keypress(self, size: urwid_Size, key: str) -> str:
+        if is_command_key("FILE_UPLOAD", key):
```
neiljp, line 1200: You're overriding the behavior of the keypress here, so you should be able to change the key we use here quite easily?

neiljp (commented):
@Subhasish-Behera As per Mounil's review, this works with a very minor change to the file prefix :+1:

Other than Mounil's and my inline comments, and ensuring the link is generated correctly, the next steps would be to add some tests for commits, in particular those handling the logic. That should help with thinking about the path entered and what the server expects and provides back to the application.

zulipterminal/ui_tools/boxes.py
```diff
@@ -710,6 +710,14 @@ def autocomplete_emojis(
 
         return emoji_typeahead, emojis
 
+    def insert_uri(self, file_name: str, uri: str) -> None:
```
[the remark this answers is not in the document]
Subhasish-Behera: Makes sense.

zulipterminal/ui_tools/views.py
Subhasish-Behera: Here to get the `file_name` we are using the file location and NOT the `url`. So not sure how `urllib` can be used directly to get the file name.

mounilKshah (commented):
Ran this code on my local machine and manually did a quick test. It works as per expectation 👍 

I think it is good to start with writing tests for this PR.

neiljp:
I left feedback in the stream, so removing needs-review flag, unless there are specific concerns?

tests/model/test_model.py
```diff
@@ -810,6 +812,49 @@ def test_send_stream_message(
...
+                None,
+                id="non_exisiting_file_with_no_response",
+            ),
+        ],
+    )
+    def test_get_file_upload_uri(
+        self, mocker, model, file_name, upload_result, expected_result
+    ):
+        self.client.upload_file = mocker.Mock(return_value=upload_result)
+        with tempfile.TemporaryDirectory() as temp_dir:
```
mounilKshah, line 842: Good thinking to use this.

mounilKshah (commented):
Thank you for implementing all the feedback and writing the tests. I tested the code on my local machine and it worked fine. I even tried uploading multiple files in the same message and that too worked without a glitch.
The tests seem fine to me. I think in the next iteration, apart from the minor suggestion, the commit structure can have the code block commits squashed together with their respective tests.

tests/model/test_model.py
```diff
@@ -810,6 +812,49 @@ def test_send_stream_message(
                 req, self.controller
             )
 
+    @pytest.mark.parametrize(
+        "file_name, upload_result, expected_result",
+        [
+            case(
+                "existing_file.txt",
```
neiljp, line 819: I don't see that we gain from using different `file_name` strings, when this variable is passed in as it is each time. Instead I'd suggest using a fixed string - if we continue to set a filename to create - and then using an alternative parameter here which may have a different path. That would allow using the same branch of the conditional each time, and using this new parameter to select the file that is generated or not.

neiljp (commented):
@Subhasish-Behera Thanks for the updates and the tests :+1:

This does work well, so I've focused more on the details here.

I agree with Mounil that squashing the appropriate test commits with the matching code would make this clearer, and likely also the two commits which build slightly on each other to wrap the upload.

Once your commits are squashed, I'd recommend checking their titles and any summary text, since many have not been updated while you've been working on the PR.

I'm looking forward to having this merged! :)

zulipterminal/ui_tools/boxes.py
```diff
@@ -710,6 +710,14 @@ def autocomplete_emojis(
 
         return emoji_typeahead, emojis
 
+    def append_uri_and_filename(self, file_name: str, uri: str) -> None:
```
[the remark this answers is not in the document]
Subhasish-Behera: will change the order of the arguments.

tests/ui_tools/test_popups.py
```diff
@@ -872,6 +877,90 @@ def test_keypress_exit_popup(
...
+        self.file_upload_view._handle_file_upload(file_location)
+
+        self.controller.model.get_file_upload_uri.assert_called_once_with(file_location)
+        if not expected_error_message:
+            file_name = Path(file_location).name
+            self.file_upload_view.write_box.append_uri_and_filename.assert_called_once_with(
+                file_name, self.file_upload_view.uri
+            )
+        else:
+            self.controller.append_uri_and_filename.assert_not_called()
```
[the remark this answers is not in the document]
Subhasish-Behera: Got it. will remove it.

tests/ui_tools/test_popups.py
```diff
@@ -872,6 +877,90 @@ def test_keypress_exit_popup(
...
+        expected_error_message: Optional[str],
+    ) -> None:
+        self.file_upload_view.write_box = MagicMock()
+        self.controller.model.get_file_upload_uri.return_value = expected_uri
+
+        self.file_upload_view._handle_file_upload(file_location)
+
+        self.controller.model.get_file_upload_uri.assert_called_once_with(file_location)
+        if not expected_error_message:
+            file_name = Path(file_location).name
```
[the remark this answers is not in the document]
Subhasish-Behera: Cool. I think you are suggesting this
```
    def test__handle_file_upload(
        self,
        file_location: str,
        file_name: str,
        expected_uri: str,
        expected_error_message: Optional[str],
    ) -> None:
```
and directly use the `file_name`  parameter.

tests/model/test_model.py
Subhasish-Behera: I probably thought along the lines of testing the code for different file types like `pdf` and `txt`.

Subhasish-Behera: I think you are suggesting that `file_name` doesn't help in testing that.I will change as you said.
 But should ZT-Client's code test if different file formats work?

tests/ui_tools/test_popups.py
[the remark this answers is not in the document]
neiljp: Right, the principle is that ideally a test is as simple as possible - less logic and calculation - and the test cases handle the input and corresponding expected output (`expected_*` parameters).

tests/ui_tools/test_popups.py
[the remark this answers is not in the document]
neiljp: If you do wish to assert on the mocked details, then my point here was more that the object you're asserting on does not exist - and is different to that a few lines above.

tests/model/test_model.py
neiljp: I agree that this illustrates that various file names should be acceptable to the function. However, how many would we include to show that? Also, file extensions do not necessarily correspond to file types - and I suspect the client/server accepts arbitrary files anyhow? So, if we want to demonstrate that, we could include a fixed string in the test - or as a default value to a test parameter - which looks fairly random.

In case it needs clarifying, using a mock for `client.upload` means that only ZT code sees this file in any case - it doesn't touch the actual `Zulip.Client`, and of course not any server.

I mentioned using a path parameter combined with a fixed file_name, since you're not testing eg. `./file_name`, `path_to/file_name` or similar vs just `file_name`. The `file_name` could be a constant, with the path varying in the test cases.

zulipterminal/model.py
```diff
@@ -560,10 +561,16 @@ def send_stream_message(self, stream: str, topic: str, content: str) -> bool:
             notify_if_message_sent_outside_narrow(composition, self.controller)
         return message_was_sent
 
-    def get_file_upload_uri(self, file_location: str) -> str:
-        with open(file_location, "rb") as fp:
-            result = self.client.upload_file(fp)
-        return result["uri"]
+    def get_file_upload_uri(self, file_location: str) -> Optional[str]:
+        if os.path.exists(file_location):
+            with open(file_location, "rb") as fp:
```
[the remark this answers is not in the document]
neiljp: Rereading this, it would definitely be better to use the exception approach here, and focus error handling for upload as occurring from the model, compared to error handling with accessing the file from the calling function.

It's possible there could be errors related to processing the upload related to the file handling, but that would be part of the upload process.

PR #1416, opened 2023-07-07
[WIP] Add DM panel to the left side view.

### LINKED ISSUE #1197: Add Private Messages side panel
This feature would provide the equivalent functionality to the web-app where clicking on the private messages button opens a drop-down.

However, in our case it seems cleaner (certainly as a first step) to integrate this as an alternative side panel to the streams/topics area. In this situation we likely need to add a key to switch back to the streams list, so may likely take over `S` for this purpose.

This issue comprises multiple steps, which could be resolved in one PR but may be more appropriate to merge over multiple PRs or cherry-picking commits from the one PR.

Possible steps:
- Add `Stream messages` total in top left; the web app does not have this, but this provides a great location to position a `S` (or similar) key when the PM side panel feature is fully added. This would need to handle the same intricacies as other totals in terms of muted elements.
- Generate a similar widget to the stream panel with a list of recent conversations with unread totals; buttons in that panel would load conversations a little like from the user list but without opening compose and would also need to cope with group PMs (see #220/#315)
- Switch to that widget when pressing `P` (shift p), NOT narrowing to PMs and focusing on a message; at the same time enable `S` (shift s) for showing the stream list

There are various nuances to consider, such as preserving state in lists between toggling, whether we consider stream+topic as a state (ie. S returns to topic list if previously in it)

Possible followup would be to enable the `p` key for what I believe we currently do for `P` (needs checking), ie. go to next unread PM. That would involve focusing on the message list, rather than switching the left panel. We could also consider using `s` for 'next unread stream message' and enable a 'streams narrow' (no PMs). This is similar but subtly different to the existing `n` key (for next unread topic), a feature which needs a little fine tuning eg. #442. Except for perhaps the first of these, these follow-ups would definitely benefit from further discussion.

I most recently proposed this feature in #**zulip-terminal > Streams/PMs view toggle**, where discussion may continue.

theViz343:
This PR is currently preparation for the addition of a collapsible DM panel in the left side UI view. 

### Outstanding aspect(s)    

- [ ] 

- [x] Discussed in **#zulip-terminal** in `Private messages sidepanel`
- [x] Fully fixes #1197 
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

[asciicast](https://asciinema.org/a/CmHP5b8aFhA6XRLEp0foNJYA9)

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -46,11 +46,11 @@ from zulipterminal.server_url import near_message_url
 from zulipterminal.ui_tools.boxes import PanelSearchBox
 from zulipterminal.ui_tools.buttons import (
     DMButton,
+    DMPanelButton,
     EmojiButton,
     HomeButton,
     MentionedButton,
     MessageLinkButton,
-    PMButton,
     StarredButton,
     StreamButton,
     StreamPanelButton,
@@ -310,7 +310,9 @@ class DMPanel(urwid.Pile):
     def __init__(self, submenu_view: List[Any], view: Any) -> None:
         self.view = view
         count = self.view.model.unread_counts.get("all_pms", 0)
-        self.view.pm_button = PMButton(controller=self.view.controller, count=count)
+        self.view.pm_button = DMPanelButton(
+            controller=self.view.controller, count=count
+        )
 
         self._contents = [
             ("pack", self.view.pm_button),
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -305,6 +305,31 @@ class MessageView(urwid.ListBox):
         self.model.mark_message_ids_as_read(read_msg_ids)
 
 
+class DMPanel(urwid.Pile):
+    def __init__(self, submenu_view: List[Any], view: Any) -> None:
+        self.view = view
+        count = self.view.model.unread_counts.get("all_pms", 0)
+        self.view.pm_button = PMButton(controller=self.view.controller, count=count)
+
+        self._contents = [
+            ("pack", self.view.pm_button),
+            ("pack", urwid.Divider(div_char=SECTION_DIVIDER_LINE)),
+            submenu_view,
+        ]
+
+        super().__init__(self.contents)
+
+
+class DMView(urwid.Frame):
+    def __init__(self, dm_btn_list: List[Any], view: Any) -> None:
+        self.view = view
+        self.log = urwid.SimpleFocusListWalker(dm_btn_list)
+        self.dm_btn_list = dm_btn_list
+        self.focus_index_before_search = 0
+        list_box = urwid.ListBox(self.log)
+        super().__init__(list_box)
+
+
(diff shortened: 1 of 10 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -882,11 +888,9 @@ def topics_view(self, stream_button: Any) -> Any:
         self.view.topic_w = TopicsView(topics_btn_list, self.view, stream_button)
         w = urwid.LineBox(
             self.view.topic_w,
-            title="Topics",
-            title_attr="column_title",
-            tlcorner=COLUMN_TITLE_BAR_LINE,
-            tline=COLUMN_TITLE_BAR_LINE,
-            trcorner=COLUMN_TITLE_BAR_LINE,
```
neiljp, line 870: This looks independent of this particular commit, and maybe belongs in the previous commit?

Also, this change essentially sets all the borders to be empty?

neiljp (commented):
@theViz343 I'd tested this previously for UX feedback, but not looked at the code in depth, until now :)

The DM Panel logic isn't quite complete functionally from what I remember, so I've focused on seeing if we can get some early commits from this PR which prepare for that work merged sooner, which is the first 3 commits.

Running the PR at that point appears to work broadly fine, with the new separate count and the stream and topic lists working as expected.

As per a comment, I'd drop the shortcut key text until the later DM-related commits, otherwise if we do a partial merge it's confusing - S is still set up to do a contextual change of narrow at that point.

Related to that, functionally there is a nit that we shouldn't be narrowing to all messages from the stream messages button - if anything, we should be narrowing to all non-DMs, but that's not implemented. At this point, you could make the button non-selectable instead. That may be relevant for the later commits, since we've not discussed the distinction between using the hotkeys (P,S) to switch panels vs ones to go to those narrows themselves - like P currently does.

In general the commits I looked at are good, but as per some comments, they might benefit from splitting/reordering. I did manage to trigger an exception when running at the first commit, since the button didn't yet exist.

theViz343:
The recent update to this PR addresses most of the earlier feedback.

The commits are now reordered so that the unread count functionality for the Stream Panel Button is added after the button's introduction.

I've removed the `S` hotkey from the early commits and only introduced it for the Stream Panel Button in the last commit which adds panel folding/unfolding. This might also benefit from some discussion on what to do for the current function of the `S` key (narrow to stream of the current message).

I've also addressed the focus position [issue](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Private.20messages.20sidepanel.20.23T1416/near/1642055). Now, the earlier focused position should be focused on when folding/unfolding panels. 

I've also added the user presence status to the DMButtons. A thing to note is that a DMButton is quite close in UI and functionality, to the UserButton class, with some differences such as the UserInfoView on pressing `i` and narrow to user vs opening compose box for user. This may warrant a parent class (still a child class of TopButton ) from which these two classes inherit. This may simplify the code a bit.

There are still a few tests to be added to cover the new code, but any initial testing is welcome :)

PR #1424, opened 2023-08-19
Add support for old format for narrow links

### LINKED ISSUE #1422: Support old format for narrow links (subject in addition to topic)
As documented in buttons.py (currently `_parse_narrow_link`), we support links formatted in the form:
```
... narrow/stream/[{stream-id}-]{stream-name}/topic/{encoded.20.topic.20name}[/near/{message_id}]
```

As I noted in [**#api documentation > Narrow URL formats**](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/Narrow.20URL.20formats) this doesn't handle an old format where `subject` was used in place of `topic`, which makes the URL in the quoted message fail to be identified and link properly in ZT.

Servers running newer versions of Zulip will not generate links of this format, which limits the impact of this issue, but equally anyone who has upgraded an older server will potentially still have messages with links of this form.

mounilKshah:
This PR contains commit which extends support to the old format for narrow links 
which existed before server version 2.1.0

[#zulip-terminal > Support old narrow links format #T1422 #T1424](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Support.20old.20narrow.20links.20format.20.23T1422.20.23T1424)
[#api-documentation > Narrow URL formats](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/Narrow.20URL.20formats)

- [ ] Discussed in **#zulip-terminal** in `Support old narrow links format #T1422 #T1424`
- [x] Fully fixes #1422 
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/ui_tools/test_buttons.py
+++ b/tests/ui_tools/test_buttons.py
@@ -687,30 +687,30 @@ class TestMessageLinkButton:
                 id="subject_narrow_link",
             ),
             case(
-                "/#narrow/stream/1-Stream-1/near/1",
+                "/#narrow/stream/1-Stream-1/near/987",
                 ParsedNarrowLink(
                     narrow="stream:near",
-                    message_id=1,
+                    message_id=987,
                     stream=DecodedStream(stream_id=1, stream_name=None),
                 ),
                 id="stream_near_narrow_link",
             ),
             case(
-                "/#narrow/stream/1-Stream-1/topic/foo/near/1",
+                "/#narrow/stream/1-Stream-1/topic/foo/near/789",
                 ParsedNarrowLink(
                     narrow="stream:topic:near",
                     topic_name="foo",
-                    message_id=1,
+                    message_id=789,
                     stream=DecodedStream(stream_id=1, stream_name=None),
                 ),
                 id="topic_near_narrow_link",
             ),
             case(
-                "/#narrow/stream/1-Stream-1/subject/foo/near/1",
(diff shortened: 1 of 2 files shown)


neiljp (commented):
@mounilKshah Thanks for taking this up :+1:

This seems to work just fine, and the tests back this up.

I was going to comment that the test cases you added were a little confusing in places since the message and stream ids are identical, but I note we do that in the other situations too. On that basis, it may be useful to do some cleanup on this test, perhaps as prep commits if you're comfortable with that, namely refactor commits for:
- inlining ids
- adjusting the expected message id to be different to the stream id
- moving the SERVER_URL to be used in the test body instead?

Re the commit text, note that notes re Fixes and the tests normally go on their own lines. I think ZT or Zulip document the former; the latter is likely assumed by Zulip in general, but can be useful to confirm and distinguish between new behavior coverage, adjustments/refactors, test cases vs tests, etc.

For followup:
- We've discussed in the stream that we could document test styles/guidelines. If you'd like to draw together some elements into a document, that'd be useful to direct users towards. Another approach which I briefly explored was enabling pytest checks in ruff.
- We may be able to either document or also add checks for Fixes and tests lines to the gitlint rules we use somehow, which would lower the need to give feedback on this as frequently.

neiljp:
@mounilKshah Thanks for the refactor :+1: Adjusted very slightly and added in a small followup commit for message vs stream id distinction I referred to.

Issue #1341, opened 2023-03-21
Add popup to confirm exiting the application


neiljp:
I raised a concern over this some time ago in [#zulip-terminal>Confirm before exiting? #T1341](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Confirm.20before.20exiting.3F.20.23T1341); see there for motivation.

The design for this could be as follows:
- reuse our existing confirmation popup to trigger on the exit keypress, instead of it directly exiting the application
- enable this by default, but add a configuration file option to specify a preference
- perhaps also a command-line option, to both enable and disable it (eg. to support default=confirm, config=no-confirm, command-line=confirm)

The first step of these would be a good first issue, and if that goes well the later points could be included in the first PR or follow-up(s).

The exit keypress (`QUIT`) is listed in the help but doesn't connect to a specific event like most other symbolic keys (eg. `HELP`), since it is actually us setting the `Ctrl c` (`SIGINT`) signal handler in `core.py` which controls what happens here. Intercepting that flow of functions would be the first thing to try.

We may later want to explore a different approach where we block that signal completely in the same way that we block Ctrl+z/s, and then use regular keypress handling instead, though if the urwid main loop blocks then we may lose the ability to shut down as cleanly.

AnonymousCodes911:
I would like to work on this, can you highlight the file(s) to target while working on this issue?

zulipbot:
Welcome to Zulip, @AnonymousCodes911! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

AnonymousCodes911:
@neiljp can you please help me locate the right file(s) to work on.
Thankyou!

AnonymousCodes911:
@neiljp 
`  def exit_window_confirmation_popup(self):
        self.show_pop_up(
[...]

PR #1427, opened 2023-08-29
Added Exit Confirmation Popup

rsashank:
Displays a confirmation popup when the (^C) interrupt signal is encountered, seeking confirmation prior to exiting the terminal client.

### Outstanding aspect(s)    

- [ ] 

- [ ] Fully fixes #
- [x] Partially fixes issue #1341
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

https://github.com/zulip/zulip-terminal/assets/110327345/f95a3f00-88ec-40cb-91ae-5ec8890a5985

### DIFF
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -619,10 +619,18 @@ class Controller:
     def deregister_client(self) -> None:
         queue_id = self.model.queue_id
         self.client.deregister(queue_id, 1.0)
+        sys.exit(0)
 
     def exit_handler(self, signum: int, frame: Any) -> None:
-        self.deregister_client()
-        sys.exit(0)
+        question = urwid.Text(
+            ("bold", " Please confirm that you wish to exit Zulip-Terminal "),
+            "center",
+        )
+        popup_view = PopUpConfirmationView(
+            self, question, self.deregister_client, location="center"
+        )
+        self.loop.widget = popup_view
+        self.loop.run()
 
     def _raise_exception(self, *args: Any, **kwargs: Any) -> Literal[True]:
         if self._exception_info is not None:


zulipterminal/core.py
```diff
@@ -617,11 +617,18 @@ def narrow_to_all_mentions(self) -> None:
         self._narrow_to(anchor=None, mentioned=True)
 
     def deregister_client(self) -> None:
-        queue_id = self.model.queue_id
-        self.client.deregister(queue_id, 1.0)
+        raise urwid.ExitMainLoop
```
neiljp, line 620: This does seem cleaner than the sys.exit we currently use.

However, without that I've experienced strange states where the client doesn't exit properly.

zulipterminal/core.py
```diff
@@ -617,11 +617,18 @@ def narrow_to_all_mentions(self) -> None:
...
+        question = urwid.Text(
+            ("bold", "Do you want to deregister the client?"),
+            "center",
+        )
+        popup_view = PopUpConfirmationView(
+            self, question, self.deregister_client, location="center"
+        )
+        self.loop.widget = popup_view
+        self.loop.run()
         sys.exit(0)
```
neiljp, line 632: I was surprised that we need the `loop.run()` here, but perhaps the signal handling interrupts the loop in some way. It seems to behave strangely without it?

neiljp (commented):
@rsashank Thanks for posting the code :+1:

This does appear to work quite well, though note that this has a regression in that it doesn't let the server know that we no longer need the queue. You can see this if you run in debug mode (-d) and look at the generated API log file.

There seem to be some nuances in using the signal handler with urwid code in the event loop. I want to like using `ExitMainLoop` alone for exiting, but it doesn't appear to reliably clear up properly in the absence of the existing `sys.exit`.

zulipterminal/core.py
rsashank: Thank you for your comment and for reviewing the code. I appreciate your feedback.
The absence of `loop.run()` results in the popup not appearing, it appears that you are correct, the signal handling is interrupting the loop in some way.

zulipterminal/core.py
rsashank: Thank you for your comment and for reviewing the code. I appreciate your feedback.
I've reverted to the existing code, the behavior seems apt now.

rsashank:
I've reverted to the original code for the `deregister_client` method. I believe it now results exits the client cleanly.

neiljp:
@rsashank Thanks for working on this and the update :+1:

Did you test this? The version you pushed had an explicit call to deregister_client from the signal handler before switching to the confirmation popup, so always exits before it has a chance to enter the popup.

I just pushed a version here which removes that line, as well as the existing sys.exit in the exit_handler, which should now be unnecessary. This version also has a user-friendly exit message - I would expect most users will not know about deregistering vs simply exiting :)

With an updated commit message, I plan to merge this shortly.

I was able to trigger a crash in an edge case of overusing ^C, which I suspect is related to the UI in the signal handler, but this works for now. We'll want to ensure we can configure this behavior next, as per the issue, and possibly discuss whether the default should be yes or no.

rsashank:
@neiljp Thanks for reviewing the code. 

I was testing calling the method in the exit_handler and inadvertently force-pushed the test code. I did test the code prior to that. I apologize for the novice oversight.

 I'll look into the edge case of overusing ^C, and will be looking forward to working on configuring this behavior next.

Thanks for the guidance, this is my first contribution to open source. Looking forward to the merge. :)

neiljp:
@rsashank Thanks for the contribution - I know it feels good to have code merged :)

As per my post in the stream, I suspect this will now be important to be configurable, initially in the config file, whether for existing users or developers who know what they are doing. Hopefully comments in the stream will help guide the importance of that.

You're welcome to dig into the ^C issue, but it seemed quite challenging to trigger, so I wouldn't think of it as a priority in comparison.

PR #1429, opened 2023-09-09
Upgrade spellcheckers

neiljp:

Other than keeping `codespell` and `typos` up to date with any extra spelling suggestions, the new version of `typos` allows configuration in `pyproject.toml`. Previous typos-specific workarounds were possible if adding a dedicated `typos.toml`, which I thought we'd previously added but avoided, which was the original motivation for looking at the latest typos version.

The pyproject.toml configuration is actually required for the upgrade due to `O_WRONLY` now appearing as a false positive. However, this also allows checking of the FAQ by typos, which was previously excluded from both tools due to the presence of `iterm`, by extending the allowed words list.

Ideally this would be simpler if tools added inline ignores (crate-ci/typos#316, codespell-project/codespell#1212), particularly if the format was common to multiple tools.

- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,10 @@ show_missing = true
 # See crate-ci/typos#744 and crate-ci/typos#773
 "O_WRONLY" = "O_WRONLY"
 
+[tool.typos.default.extend-words]
+# Allow iterm terminal emulator (ideally only per-file)
+"iterm" = "iterm"
+
 [tool.isort]
 py_version = 37
 # Use black for the default settings (with adjustments listed below)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,10 @@ precision = 1
 skip_covered = true
 show_missing = true
 
+[tool.typos.default.extend-identifiers]
+# See crate-ci/typos#744 and crate-ci/typos#773
+"O_WRONLY" = "O_WRONLY"
+
 [tool.isort]
 py_version = 37
 # Use black for the default settings (with adjustments listed below)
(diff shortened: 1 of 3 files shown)


epage:
> Ideally this would be simpler if tools added inline ignores ([crate-ci/typos#316](https://github.com/crate-ci/typos/issues/316), [codespell-project/codespell#1212](https://github.com/codespell-project/codespell/issues/1212)), particularly if the format was common to multiple tools.

Would love to have a common format for this.  In the mean time, I believe both tools have support for regex ignores and you can create your own directives with that.

neiljp:
@epage I was thinking of the regexes originally as primarily only for individual word cases.

However, I just saw https://github.com/codespell-project/codespell/issues/1212#issuecomment-1721152455, and I'm experimenting with relaxing the file exclusion using something like the following to maintain some potential compatibility with cspell (or if adopted more generally):
```python
  ....  # spellchecker:disable-line
```

That seems to work with codespell using a form of the regex in that comment, but using this in pyproject.toml doesn't appear to work for typos:
```
[tool.typos.default]
extend-ignore-re = [".*# spellchecker:disable-line$"]
```
Should those two forms be comparable? Is this related to a nuance with regexes? (in rust? in typos?)

epage:
To use anchors like that, you need to enable multi-line mode; try `extend-ignore-re = ["(?m)^.*# spellchecker:disable-line$"]`.  I tested this in crate-ci/typos#835

PR #1430, opened 2023-09-17
Add support for all streams minimal

### LINKED ISSUE #816: Adding new stream throws an Keyerror

theViz343:
This PR is very similar to #1408 and #1419 and combines elements of both. This PR keeps the initial commits from #1419 but replaces the general accessor methods with only an attribute-specific necessary method i.e "name".

### Outstanding aspect(s)    

- [x] Could add more attribute accessor methods (like "color")

- [x] Discussed in **#zulip-terminal** in [`Adding streams throws key error`](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Adding.20streams.20throws.20key.20error)
- [x] Fully fixes #816 
- [ ] Partially fixes issue #
- [x] Builds upon previous unmerged work in PR #1419 and PR #1408 
- [ ] Requires merge of PR #
- [x] Merge will enable work on #1387 

### How did you test this?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -416,6 +416,20 @@ def never_subscribed_streams_fixture() -> List[Stream]:
 
 
 @pytest.fixture
+def all_stream_ids(
+    streams_fixture: List[Subscription],
+    unsubscribed_streams_fixture: List[Subscription],
+    never_subscribed_streams_fixture: List[Stream],
+) -> List[int]:
+    return [
+        stream["stream_id"]
+        for stream in streams_fixture
+        + unsubscribed_streams_fixture
+        + never_subscribed_streams_fixture
+    ]
+
+
+@pytest.fixture
 def realm_emojis() -> Dict[str, Dict[str, Any]]:
     # Omitting source_url, author_id (server version 3.0),
     # author (server version < 3.0) since they are not used.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -233,7 +233,7 @@ def logged_on_user() -> Dict[str, Any]:
 
 
 @pytest.fixture
-def general_stream() -> Dict[str, Any]:
+def general_stream() -> Subscription:
     return {
         "name": "Some general stream",
         "date_created": 1472091253,
@@ -245,7 +245,6 @@ def general_stream() -> Dict[str, Any]:
         "audible_notifications": False,
         "description": "General Stream",
         "rendered_description": "General Stream",
-        "is_old_stream": True,
         "desktop_notifications": False,
         "stream_weekly_traffic": 0,
         "push_notifications": False,
@@ -253,13 +252,18 @@ def general_stream() -> Dict[str, Any]:
         "message_retention_days": 10,
         "subscribers": [1001, 11, 12],
         "history_public_to_subscribers": True,
+        "is_announcement_only": False,
+        "first_message_id": 1,
+        "email_notifications": False,
+        "wildcard_mentions_notify": False,
+        "is_web_public": False,
     }
 
 
 # This is a private stream;
 # only description/stream_id/invite_only/name/color vary from above
 @pytest.fixture
-def secret_stream() -> Dict[str, Any]:
+def secret_stream() -> Subscription:
     return {
         "description": "Some private stream",
         "stream_id": 99,
@@ -272,19 +276,23 @@ def secret_stream() -> Dict[str, Any]:
         "color": "#ccc",  # Color in '#xxx' format
         "is_muted": False,
         "audible_notifications": False,
-        "is_old_stream": True,
         "desktop_notifications": False,
         "stream_weekly_traffic": 0,
         "message_retention_days": -1,
         "push_notifications": False,
         "subscribers": [1001, 11],
         "history_public_to_subscribers": False,
+        "is_announcement_only": False,
+        "first_message_id": 1,
+        "email_notifications": False,
+        "wildcard_mentions_notify": False,
+        "is_web_public": False,
     }
 
 
 # Like public stream but with is_web_public=True
 @pytest.fixture
-def web_public_stream() -> Dict[str, Any]:
+def web_public_stream() -> Subscription:
     return {
         "description": "Some web public stream",
         "stream_id": 999,
@@ -297,7 +305,6 @@ def web_public_stream() -> Dict[str, Any]:
         "color": "#ddd",  # Color in '#xxx' format
         "is_muted": False,
         "audible_notifications": False,
-        "is_old_stream": True,
         "desktop_notifications": False,
         "stream_weekly_traffic": 0,
         "message_retention_days": -1,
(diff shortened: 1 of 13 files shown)


tests/ui/test_ui_tools.py
```diff
@@ -891,6 +891,7 @@ def test_keypress_NEXT_UNREAD_TOPIC_stream(
 
         mid_col_view.model.stream_dict = {1: {"name": "stream"}}
         mid_col_view.model.next_unread_topic_from_message_id.return_value = (1, "topic")
+        mid_col_view.model.get_stream_name_from_id.return_value = "stream"
```
neiljp, line 894: It's important to adjust the tests, not just to pass, but to represent what is now necessary. 

In this commit you're using the function instead of the stream_dict for the name - so is it now necessary to set the stream_dict in the test?

Many of the test changes need to be checked for this.

zulipterminal/model.py
```diff
@@ -1297,6 +1297,10 @@ def _get_all_stream_ids(self) -> List[int]:
         id_list.extend(stream_id for stream_id in self._never_subscribed_streams)
         return id_list
 
+    def get_stream_name_from_id(self, stream_id: int) -> str:
```
neiljp, line 1300: Based on the existing `stream_id_from_name`, let's drop the `get_`.

zulipterminal/api_types.py
```diff
@@ -244,9 +250,11 @@ class Subscription(Stream):
 
     is_muted: bool
 
-    role: int  # NOTE: new in Zulip 4.0 / ZFL 31
     color: str
 
+    # Deprecated fields
+    # role: int  # Removed in Zulip 6.0 (feature level 133)
```
neiljp, line 256: I think I noted somewhere that it's worth including here that it wasn't actively used by us (or the web app even?), and introduced at some other ZFL?

tests/core/test_core.py
```diff
@@ -124,18 +125,22 @@ def test_narrow_to_stream(
...
-            stream_id: {
+            stream_id: general_stream,
+        }
+        controller.model.stream_dict[stream_id].update(
+            {
                 "color": "#ffffff",
                 "name": stream_name,
             }
-        }
+        )
```
neiljp, line 143: It's good that this commit focuses on the stream/subscription migration, but
- it doesn't seem that the color is important to the test(s)?
- that could make this transition smaller/simpler
- can we replace the stream_dict code by setting the return_value in the model, in the later commits? (if we don't already)

neiljp (commented):
@theViz343 It's good to see another go at this one, though there are some comments from before to keep up with, including:
- https://github.com/zulip/zulip-terminal/pull/1419/commits/29fe057222efcfddacbcd6cb3b4fab0b40c33bd1#r1280046556
- https://github.com/zulip/zulip-terminal/pull/1419/commits/1ced8ec95217618b3e8f4e658728dfc4cf97d8f0#r1280070702

I'm happy with the core of this, but other than the points above was initially interested to see if the early new data in the Model you added could be confirmed in test_init using new fixtures. To see that you added fixtures which had similar data to test later code confirmed that would be useful, to test the data flow through the Model from the initial_data into the accessors.

zulipterminal/model.py
```diff
@@ -1281,6 +1281,22 @@ def _register_non_subscribed_streams(
             stream["stream_id"]: stream for stream in never_subscribed_streams
         }
 
+    def _get_stream_from_id(self, stream_id: int) -> Union[Subscription, Stream]:
```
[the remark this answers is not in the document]
theViz343: If we do that, should I cast the `subscribed` and `unsubscribed` streams returned by this method to `Stream`? That might make the usage of this method clearer too, ie. to only access stream properties (and not subscription properties) from the returned stream.

tests/core/test_core.py
theViz343: For the third point, which function's return_value are you referring to?

tests/conftest.py
```diff
@@ -243,21 +244,25 @@ def general_stream() -> Dict[str, Any]:
         "audible_notifications": False,
         "description": "General Stream",
         "rendered_description": "General Stream",
-        "is_old_stream": True,
```
theViz343, line 246: I could not find any documentation in the api docs or CZO discussion for this field. What was the field's original purpose, if you know? I can then add it's reason for removal in the commit text with more context.

neiljp: It's in the API changelog (https://zulip.com/api/changelog), and I found it on czo to be zulip/zulip#15409.

tests/core/test_core.py
neiljp: I meant that where you're moving to use the functions that access the name and color, instead of using stream_dict, can we now mock those functions and set the return value directly?

Essentially, we want to ensure we move away from the stream_dict fixture (or equivalent manual dicts), in addition to directly accessing the model attribute.

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: A cast shouldn't be necessary once you have the two different functions - you don't have that in the latest version of the code you just pushed?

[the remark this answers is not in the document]
theViz343: I've casted the stream_dict and the unsubscribed_streams to a `Stream` in the `_get_stream_from_id` in the latest commit, so that methods which use the `_get_stream_from_id` always access only the stream properties and not the subscription properties (which would then warrant the use of `_get_subscription_from_id`).

tests/core/test_core.py
```diff
@@ -124,18 +125,17 @@ def test_narrow_to_stream(
...
         controller.model.index = index_stream
         controller.view.message_view = mocker.patch("urwid.ListBox")
         controller.model.stream_dict = {
-            stream_id: {
-                "color": "#ffffff",
-                "name": stream_name,
-            }
+            stream_id: general_stream,
         }
+        controller.model.stream_dict[stream_id]["name"] = stream_name
```
neiljp, line 138: If it's necessary to keep this line for the test, can the stream_dict be removed instead later, when we switch to the new function?

neiljp (commented):
@theViz343 Thanks for the update again :+1:

Overall this is good, and the commits are generally clean and well tested.

I think most of my non-minor feedback has come about since this refactoring is complex - tidying types, correcting fixtures, fixing inconsistencies, adding data, using new accessors, removing old fixtures, integrating with existing code, updating old accessors - and so there are a lot of potential rough edges, some of which only surface later.

The stream_dict removals in the tests could be left for another time, certainly the complex cases.

I think clarifying the stream/subscription validity functions are likely the standout issue to address here, and which we need to be a little careful about since they're already in use, and we don't interact with all streams in every case.

zulipterminal/model.py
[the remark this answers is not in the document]
neiljp: Ah right, but I don't remember seeing an explicit `cast()` to work around mypy - I assume this works in the type system since a Stream is a subset of a Subscription.

PR #1432, opened 2023-10-01
Configuring the Exit Confirmation Popup

### LINKED ISSUE #1341: Add popup to confirm exiting the application

rsashank:
Configures the behavior of the exit confirmation popup, giving users the option to choose between exiting instantly or displaying a confirmation popup before exiting. This configuration can be managed through both the configuration file and command-line options.

### Outstanding aspect(s)    

- [ ] 

- [x] Fully fixes #1341
- [ ] Partially fixes issue #
- [x] Is a follow-up to work in PR #1427 
- [ ] Requires merge of PR #

### How did you test this?


zulipterminal/core.py
```diff
@@ -622,15 +624,18 @@ def deregister_client(self) -> None:
...
-        question = urwid.Text(
-            ("bold", " Please confirm that you wish to exit Zulip-Terminal "),
-            "center",
-        )
-        popup_view = PopUpConfirmationView(
-            self, question, self.deregister_client, location="center"
-        )
-        self.loop.widget = popup_view
-        self.loop.run()
+        if self.exit_confirmation:
```
neiljp, line 627: This conditional appears to work absolutely fine :+1:

However, we could avoid needing to do conditional work and change this method, if we treat this method (and maybe rename it) as a 'prompting_exit_handler', and attach a minimal method as an 'no_prompt_exit_handler' if we do not require confirmation. That moves the conditional into the initializer, rather than having it down here.

neiljp (commented):
@rsashank Thanks for the update, and apologies for my delay this time!

This looks to be working well, though as per the inline comment it may be cleaner to have two signal handlers which are attached depending on the configuration.

Other than the code, if you could look at our documentation and add anything relevant regarding the new configuration option that would be great as a second commit - we'll likely need at least an overview section in the README configuration example.

Re commit style, I'd suggest installing and setting up gitlint, though you have a good idea already :)

neiljp:
@rsashank Just to note that I've tagged this with 'awaiting update'; you can use @zulipbot to remove the tag - and add the similar 'review' tag or leave a message if it's ready again :+1:

neiljp:
@rsashank I just saw this error - I was looking into this recently, it's not related to you :)

zulipterminal/core.py
rsashank: Thank you for your review, that makes sense.  I have incorporated the new approach and force-pushed the changes as suggested.

rsashank:
> @rsashank I just saw this error - I was looking into this recently, it's not related to you :)


I appreciate the reassurance :P.  Anyways, I've made updates to the README. Do you have any additional recommendations for changes to the documentation?

neiljp:
@rsashank Thanks 👍 I made a small change to the text and squashed the commits into a new summary since the docs update was very small (and related).

Reviewing some other recent changes, I subsequently realized this didn't add the configuration option to the About popup yet. That would tidy up a loose end :)

neiljp:
> Got it, I'll be working on that. Should I make a separate PR for that?

Certainly - this PR is merged :)

Issue #1420, opened 2023-07-29
Add symbols for all supported narrows


neiljp:
This was originally noted as part of #1348, but deserves splitting out.

Very early in the history of the project, there were markers/icons/emoji for narrows, eg. for all messages there was a 'house' symbol, since at the time this was the 'home' in the web app.

These symbols were subsequently removed, since it is difficult to find unicode codepoints which are consistently viewable across terminals, such that running with a default install doesn't elicit "this doesn't look right" from a wide range of users.

This issue is concerned with considering the reintroduction of symbols, but for all the narrows which we currently support.

Originally symbols were only shown as prefixes in the top-left part of the application, and we'd likely want to include them there for consistency, which may lead to us increasing the width of the left panel slightly.

It would also be useful to show these symbols as prefixes to each narrow in the line below the "Messages" title line.

As a first step this will need identification of suitable symbols for each narrow; these include:
- `All messages` (via <kbd>a</kbd>)
- `All direct messages` (via <kbd>P</kbd>)
- `All stream messages` (new narrow to be added to the left side, likely via <kbd>S</kbd>)
- `Direct message conversation` (via <kbd>z</kbd>) [*]
- `Group direct message conversation` (via <kbd>z</kbd>) [*]
- `<stream-name>` (ie. stream narrow) (via <kbd>z</kbd>) [*]
- `<stream-name> > <topic-name>` (ie. topic narrow) (via <kbd>z</kbd>) [*]
- `(All) mentions` (via <kbd>#</kbd>)
- `Starred messages` (via <kbd>f</kbd>)

Note that in line with the original challenge - and why old symbols were removed - any new proposed symbols will need checking for compatibility on 'common' terminals/fonts.

Those marked with [*] have an existing prefix, and can be an extension of the work on #1348 / #1386. However, we'll likely want to introduce a set of symbols at once in order to reduce confusion, so **please discuss this in the stream before implementing - see `#zulip-terminal > Narrow symbols #T1420`**.

Connections to other issues:
- #1343 - For topic narrows, prefixing `<stream-name>: topic narrow` is not likely useful; similarly for direct message 'titles' like `Direct message conversation`, which #1343 would enable dropping.
- #1197 - This work will involve adding the 'All stream messages' narrow.
- #938 - If this was supported, we'd have fewer constraints on symbol choice, though we'd likely still aim for a commonly-viewable theme in addition to an 'emoji' and 'super basic' one.

neiljp:
[...]

PR #1441, opened 2023-11-10
messages: Adding prefix symbols to search bar.

supascooopa:
As previously done to other parts of the UI, this commit adds prefix symbols to the narrow bar before the search input area.

This is intended to make the UI more similar to the web app UI and unify the overall look.

This is applied to both streams, topics and main narrows.

Tests adjusted accordingly.

This PR adds prefix symbols to the narrow titles before the search bar. This is done in order to approximate the current UI to the web app and to unify the overall look of the UI.

- [x] Discussed in **#zulip-terminal** in `Narrow Symbols #T1420`
- [ ] Fully fixes #
- [x] Partially fixes issue #1420
- [ ] Requires merge of PR #

### How did you test this?

- [] Adding automated tests for new behavior (or missing tests)

### Visual changes    

Before:
image

After:
image


zulipterminal/ui_tools/messages.py
```diff
@@ -211,29 +214,33 @@ def top_search_bar(self) -> Any:
...
+            text_to_fill = f" {ALL_MESSAGES_MARKER} " + "All messages"
         elif len(curr_narrow) == 1 and curr_narrow[0][1] == "private":
-            text_to_fill = "All direct messages"
+            text_to_fill = f" {DIRECT_MESSAGE_MARKER} " + "All direct messages"
         elif len(curr_narrow) == 1 and curr_narrow[0][1] == "starred":
-            text_to_fill = "Starred messages"
+            text_to_fill = f" {STARRED_MESSAGES_MARKER} " + "Starred messages"
         elif len(curr_narrow) == 1 and curr_narrow[0][1] == "mentioned":
-            text_to_fill = "Mentions"
+            text_to_fill = f" {MENTIONED_MESSAGES_MARKER} " + "Mentions"
```
neiljp, line 223: These can all be single f-strings - no need to add them.

zulipterminal/ui_tools/messages.py
```diff
@@ -211,29 +214,33 @@ def top_search_bar(self) -> Any:
...
+                        (bar_color, f" {stream_icon} "),
                         (bar_color, self.stream_name),
                         (bar_color, ": topic narrow"),
                     ],
                 )
             else:
                 text_to_fill = (
                     "bar",  # type: ignore[assignment]
-                    [(bar_color, self.stream_name)],
+                    [(bar_color, f" {stream_icon} "), (bar_color, self.stream_name)],
```
neiljp, line 243: Similarly here, when the same style/attribute is applied, only one tuple is necessary, which means the strings can be combined.

You could also simplify the topic case above this in a similar way.

zulipterminal/ui_tools/messages.py
```diff
@@ -211,29 +214,33 @@ def top_search_bar(self) -> Any:
...
                 )
         elif len(curr_narrow) == 1 and len(curr_narrow[0][1].split(",")) > 1:
             text_to_fill = "Group direct message conversation"
```
neiljp, line 246: I noticed this wasn't updated during testing? I know we discussed different symbols, but for now I think we should have a symbol for all the narrows we support - otherwise it looks inconsistent.

neiljp (commented):
@supascooopa Thanks for the followup :+1:

This looks to work well and cover most narrows, except the DM subsets I refer to inline - did you intend to not cover these?

The leading space looks good, but I think it would be useful to borrow a space from that currently used to separate the narrow text from the `Search [/]:` (currently 2?). We can color it per the narrow to make it look a little more balanced, less abrupt after the end of the narrow name, and more similar to the spacing after the stream-topic separator. That will involve touching a different part of the code, so I'd suggest that be a separate commit after (or before) the current one.

supascooopa:
@neiljp Thank you for the feedback! I will look into all the suggestions you've made. 
I'm not sure what symbol we had decided on to be used for the DMs. U+2318? I remember discussing using multiple symbols and deciding on just using one in order to simplify the process.

zulipterminal/ui_tools/messages.py
supascooopa: I'm not %100 sure what you mean. Would you please elaborate more? If you mean using "#" sign for the search bar part of the testing, I could add a different symbol to represent all the narrows. Do you have any suggestions?

neiljp: Is this what you meant by the testing update you mentioned in the stream? This comment was regarding how the 'Group direct message conversation' has no prefix, as well as one other DM-related one also.

neiljp:
@supascooopa Thanks for this - it's looking good! 🎉 

As per my note in the stream, I squashed a couple of commits and adjusted the messages to fit our style and expand upon the intention. Note how I using 'Show' instead of 'Showing'.

Issue #1439, opened 2023-11-07
Enable remapping hotkeys


laomuon:
*** Description
Correct me if I'm wrong, but it seems like there's no way to remapping the hotkeys for the end user. If this feature has not been implemented yet, it would be nice to add, especially with the ones when composing messages, as I find them sometimes a little bit awkward to use

*** Proposed solution
Give the user the ability to define a dict of key mapping in their config file

neiljp:
@laomuon Thanks for the feedback - this is currently tracked as #680.

zulipbot:
Hello @zulip/server-hotkeys members, this issue was labeled with the "area: hotkeys" label, so you may want to check it out!

laomuon:
If there is no one working on this, can I create a PR for it?

neiljp:
@laomuon Nobody is working on this right now, so any contribution towards this would be welcome! :+1:

The latest summary for the approach towards this would appear to be https://github.com/zulip/zulip-terminal/issues/680#issuecomment-1655982277. If you need to discuss this further, I'd suggest discussing this in chat.zulip.org in #zulip-terminal in a new topic.

jamesdchu:
I had started to briefly work on this issue but forgot to comment/claim. Is it alright if I take it? @laomuon

laomuon:
@jamesdchu yes feel free to do it

jamesdchu:
Thank you, will do! 
@zulipbot claim

zulipbot:
Hello @jamesdchu!

Thanks for your interest in Zulip! You have attempted to claim an issue without the label "help wanted". You can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/zulip-terminal/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) label.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

PR #1446, opened 2023-12-05
Added custom override of hotkeys based on zuliprc file 

jamesdchu:
This PR edits the keys.py file to enable custom hotkey mapping. This means that users can change the zuliprc file to get their desired hotkey mappings. 

- [ ] Fully fixes #
- [x ] Partially fixes issue #1439
- [ ] Requires merge of PR #

### How did you test this?

- [x ] Adding automated tests for new behavior (or missing tests)
- [ x] Existing automated tests should already cover this (*only a refactor of tested code*)


neiljp:
@jamesdchu This looks like you're making progress towards this - are you looking for feedback yet?

zulipterminal/config/keys.py
```diff
@@ -471,6 +471,49 @@ def commands_for_random_tips() -> List[KeyBinding]:
...
+        for cmd, binding in existing_keybindings.items()
+        for key in binding["keys"]
+    }
+    requested_changes = {}
+    conflicts = {}
+
+    # Collect requested changes and detect conflicts
+    for command, new_key in custom_keybindings.items():
+        if command not in existing_keybindings:
+            raise InvalidCommand(f"Invalid command {command} in custom keybindings")
```
neiljp, line 488: Errors from this method should be handled and cleanly expressed to the user in run.py, not just left to escape unhandled.

neiljp (commented):
@jamesdchu Thanks for working on this :+1: I left the question re if you were asking for feedback in December, but only recently noticed you'd changed it from a draft to a regular PR before my comment, so here's some feedback in case you had intended that to be the case already :)

You currently have this as a lot of commits, so I've not left comments on individual commits, but rather the total set of changes. See the contributing notes in the README regarding how Zulip in general, and this repo in particular, structure commits and why. This is one reason why the PR is failing the 'isolated commits' phase of the automatic tests (CI).

I gave this a manual test and after adjusting a line in run.py I could get it to run OK, as per one comment inline. That comment discusses adding tests for that run.py part, which suggests one way to separate into two parts (commits) - implementing the functionality, and then integrating it into run.py, each commit with tests (and passing linting). The documentation part could also be separate, particularly since a FAQ entry could be useful too. Beyond that, you could also break the functionality part into stages, with tests added/adjusted as the functionality improves, but we can discuss that later if you're interested :)

Issue #1229, opened 2022-05-19
TRACKING: Add remaining type annotations to tests


neiljp:
This was started in Summer 2021 and is now mostly complete (exclusions are in `pyproject.toml`) 🎉 

Remaining exclusions are now in `pyproject.toml`:
- [ ] `model/test_model.py`
- [ ] `ui/test_ui_tools.py`
- [ ] `ui_tools/test_messages.py`

If taking up elements of this, you will need to examine test files that are annotated for comparison and add them to one file at a time, enabling checking by removing the related exclusion.

**Updated to translate into a tracking issue and removal personal tagging**

zulipbot:
Hello @zulip/server-refactoring members, this issue was labeled with the "area: refactoring" label, so you may want to check it out!

plugyawn:
@neiljp Could you explain this a little further? I'd like to take this up.

DaveOkpare:
@zulipbot claim

zulipbot:
Hello @DaveOkpare!

Thanks for your interest in Zulip! You have attempted to claim an issue without the label "help wanted". You can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/zulip-terminal/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) label.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

neiljp:
@DaveOkpare Apologies for the slow response, but either file should be good - though both are quite large! If you wish to take this up, you might wish to follow #1337, which is specific to another file, but could provide useful directions.

PR #1450, opened 2023-12-14
Zt emailapi

rsashank:
Added helper function to send request and get stream email_address, and called the helper function when user is viewing Stream Information.

- [x] Discussed in **#zulip-terminal** in `topic` https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Stream.20Information.20ZT.20Crash
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?


zulipterminal/ui_tools/views.py
```diff
@@ -1380,7 +1380,7 @@ def __init__(self, controller: Any, stream_id: int) -> None:
 
         # FIXME: This field was removed from the subscription data in Zulip 7.5 / ZFL226
         #   We should use the new /streams/{stream_id}/email_address endpoint instead
-        self._stream_email = stream.get("email_address", None)
+        self._stream_email = controller.model.stream_email_address(stream_id)
         if self._stream_email is None:
             stream_copy_text = "< Stream email is unavailable >"
         else:
```
neiljp, line 1386: With the change here, `self._stream_email` becomes a `str`, so the first branch of this conditional cannot occur. However, we should still handle this - if the call to the server to fetch the email address fails, it will now set the variable to the empty string. Similarly, the new code to limit copying the email address will not trigger properly.

The other point here is that this now only handles newer server versions. ZT is very broad in supported server versions, so we should aim to retain handling for the older version.

Note that the existing test I added will also need to be updated to handle this.

neiljp (commented):
@rsashank Thanks for the updates here :+1:

I tested this and it appears to manually work against the community server, but as noted inline this will not work as expected for older servers, or likely where the incoming email is not configured on newer servers.

To isolate the API-specific logic, one option would be to have `stream_email_address` handle the logic across server versions. That could then be tested separately, and the existing UI tests could remain very similar.

zulipterminal/ui_tools/views.py
rsashank: Updated the PR.

Initially, set the value by getting it from `self._stream_email = stream.get("email_address", None)`. 
If the value is `None`, then call the `stream_email_address` function from model to fetch it from the endpoint. And then  if the value is either an empty string `""` or `None`, assigned stream_copy_text to `"< Stream email is unavailable >"`.

Is this approach reasonable to retain handling for older versions?

I've also updated the test.

rsashank:

zulipterminal/model.py
```diff
@@ -868,6 +868,19 @@ def topics_in_stream(self, stream_id: int) -> List[str]:
 
         return list(self.index["topics"][stream_id])
 
+    def stream_email_address(self, stream_id: int) -> str:
+        """
+        Return a string containing the email address of a stream.
+        """
```
neiljp, line 874: Let's name this more explicitly with a `_fetch` prefix, to indicate this makes an API call to the server. The leading underscore will mean that we don't expect it to be called/used outside of the model.

My suggestion would be to add a separate model function which wraps the logic for determining if the data is present in the stream/subscription, and fetches it otherwise using the wrapped `.call_endpoint` method. You might consider the movement of the existing code in the popup into this new method to be a preliminary refactor; then once you have added the `_fetch` commit, you can update this other method to call it.

The docstring doesn't add much, compared to the method name and signature (ie. parameter & return value)?

zulipterminal/ui_tools/views.py
neiljp: The logic looks good, though as per the inline comment in my next/upcoming review, I'd suggest we refactor the popup logic here into a (tested) model method first, then expand upon it once we have the (tested) stream fetch code. That should also enable the view/popup test & code to be simplified, of course :)

neiljp (commented):
@rsashank This looks to have a good flow, but I think we should take the opportunity to move the logic that varies with server version in the model.

It would also be good to see independent tests for the model methods. 

When you update the suggested additional model method to call the `_fetch`, that would be a good time to include what is currently in the first commit - and using `NotRequired` would be a good change at that point too.

zulipterminal/model.py
rsashank: Thanks for the review! I've updated the PR accordingly. 

I'm still unsure about the docstrings though.

rsashank:

tests/ui_tools/test_popups.py
```diff
@@ -1398,9 +1398,11 @@ def test_stream_info_content__email_copy_text(
...
         stream_email_present: bool,
         expected_copy_text: str,
+        mocker: MockerFixture,
     ) -> None:
         if not stream_email_present:
             del general_stream["email_address"]
+            self.controller.model.stream_copy_text.return_value = ""
 
         model = self.controller.model
         stream_id = general_stream["stream_id"]
```
neiljp, line 1408: Can this be simplified further? The code doesn't access as much of the model data directly now.

neiljp (commented):
@rsashank This looks close to ready - most of my feedback should be clear to address, but let me know if not :)

neiljp:
>When you update the suggested additional model method to call the _fetch, that would be a good time to include what is currently in the first commit - and using NotRequired would be a good change at that point too.

Did you understand what I meant here? I missed following up on this.

Reading this again, I think my original idea was to add the model methods in two stages: firstly to wrap the direct data access (pure refactor) and secondly update that method and add the new one (so only in the model) to try and fetch the data instead - combined with the `NotRequired` change.

This isn't critical, so focus on the other points first, but consider how the code changes would read differently in two commits as above.

zulipterminal/model.py
```diff
@@ -891,6 +891,34 @@ def topics_in_stream(self, stream_id: int) -> List[str]:
...
+        Endpoint added in Zulip 7.5 (ZFL 226)
+        """
+        url = f"/streams/{stream_id}/email_address"
+
+        response = self.client.call_endpoint(url, method="GET")
+
+        if response.get("result") == "success":
+            email_address = response.get("email", "")
+            return str(email_address)
+        return "invalid"
```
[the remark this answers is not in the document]
rsashank: Updated :)

zulipterminal/model.py
```diff
@@ -891,6 +891,34 @@ def topics_in_stream(self, stream_id: int) -> List[str]:
...
+
+    def stream_copy_text(self, stream_id: int) -> str:
```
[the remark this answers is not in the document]
rsashank: Updated :)

zulipterminal/model.py
```diff
@@ -891,6 +891,34 @@ def topics_in_stream(self, stream_id: int) -> List[str]:
...
+        """
+        Attempts to retrieve stream email from stream_dict, fetching
+        from method if unavailable. (Retains handling across server versions).
+        """
+        stream = self.stream_dict[stream_id]
```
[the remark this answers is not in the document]
rsashank: Got it!

tests/ui_tools/test_popups.py
rsashank: I might be wrong, but I don't think we can simplify this. The `StreamInfoView` class calls get_stream_email_address in init, and unless we specify the endpoint return value as `None`, I don't think we can effectively test the possibility of no email being present.

        stream_info_view = StreamInfoView(self.controller, stream_id)

tests/model/test_model.py
```diff
@@ -561,6 +561,62 @@ def test_topics_in_stream(self, mocker, model, topics_index, fetched, stream_id=
...
+        self.client.call_endpoint = mocker.Mock(return_value=response)
+
+        result = model._fetch_stream_email_address(stream_id)
+
+        self.client.call_endpoint.assert_called_once_with(
+            f"/streams/{stream_id}/email_address", method="GET"
+        )
+        assert result == expected_return_value
+
+    @pytest.mark.parametrize(
```
Niloth-p, line 589: Would using the stream_dict fixture help simplify the parametrize for this, or complicate it?

Niloth-p (commented):
The function names `_fetch_stream_email_address` and `get_stream_email_address` look pretty similar.
Should we consider making them more clear on what they do and how they differ?
Maybe something like `_fetch_stream_email_from_endpoint`?
Not sure if we're already following some convention wrt words like 'fetch' and 'get'.

Lgtm otherwise.
Good PR, Sashank!

zulipterminal/model.py
```diff
@@ -891,6 +891,16 @@ def topics_in_stream(self, stream_id: int) -> List[str]:
 
         return list(self.index["topics"][stream_id])
 
+    def get_stream_email_address(self, stream_id: int) -> Union[None, str]:
```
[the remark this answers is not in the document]
rsashank: Oops, yes.

tests/model/test_model.py
```diff
@@ -561,6 +561,62 @@ def test_topics_in_stream(self, mocker, model, topics_index, fetched, stream_id=
...
+        self,
+        mocker,
+        model,
+        email_address_stream_dict,
+        fetched,
+        response,
+        expected_return_value,
+        stream_id=1,
+    ):
+        model.stream_dict[stream_id] = email_address_stream_dict
```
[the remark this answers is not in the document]
rsashank: The `RuntimeError` is not specific to this method, so it might not be necessary to include it here.

tests/model/test_model.py
rsashank: Since this test only concerns the presence of the stream email_address, using the Fixture might complicate things (and be a bit unnecessary). However, I'll wait for @neiljp's opinion on this.

rsashank:
Updated this PR! Thanks for the review @Niloth-p :)

zulipterminal/model.py
```diff
@@ -906,6 +906,34 @@ def topics_in_stream(self, stream_id: int) -> List[str]:
 
         return list(self.index["topics"][stream_id])
 
+    def _fetch_stream_email_from_endpoint(self, stream_id: int) -> Optional[str]:
+        """
+        Fetch stream's email address with specified stream_id and
+        returns stream_email_address as a String.
```
Niloth-p, line 912: Continuing from the other docstrings comment,
By looking only at the function definition line,
we know that the function fetches stream email, takes a stream_id as input, and returns a string optionally.
In that case, we wouldn't necessarily have to repeat it in the docstrings. If there are some latent aspects that could be overlooked, they could be mentioned instead on their own, like the server version, ZFL, etc you've mentioned.

I know it's commonly known that having docstrings is good, but for ZT, I think we're following the convention that we add docstrings only where necessary, or useful, which it might not be so for every function.

This is my interpretation.

rsashank: Most methods in `model.py` have docstrings, so I added one. But now that you mention it, I guess it can be removed. What do you think, @neiljp?

mounilKshah (commented):
Thanks for working on this @rsashank 
I tested this PR on my local machine and it worked fine! As for the code, the feedback provided so far seems to have been inculcated to the PR. Tests too worked fine, and look good to me. 
As for the docstrings, I think it is okay to keep it the way it is as we already have docstrings for functions which are similarly understandable through their name alone but still have a docstring.

neiljp (commented):
@rsashank Thanks for the updates :+1:

Most of these changes should be small, and give more experience with structuring multi-commit changes.

Note that type annotations for test_model.py are not checked right now (see pyproject.toml and #1229), though it's useful to have some starting point.

We could likely do with some passes on exploring consistency of docstrings, function prefixes (eg. get/fetch), and other similar aspects, and is certainly a reasonable thing to examine if anyone notices something like this, as you read through the code, or are working on a given area.

tests/model/test_model.py
```diff
@@ -563,6 +564,73 @@ def test_topics_in_stream(self, mocker, model, topics_index, fetched, stream_id=
...
+            ({}, "username@example.com", True, "username@example.com"),
+            ({}, None, True, None),
+        ],
+        ids=[
+            "email_present_in_dict",
+            "email_fetched_from_endpoint",
+            "email_not_present_and_not_fetched",
+        ],
+    )
+    def test_get_stream_email_address(
```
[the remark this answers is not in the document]
rsashank: Makes sense, thanks for explaining it out! :)

rsashank:
Thanks for the detailed review, @neiljp! 
I've updated this PR and added a new commit to change `fetched` to `expected_fetched` for another test written previously. Thanks for pointing out that type annotations aren't checked for test_model.py. I'll look into this at a later point.

zulipterminal/model.py
```diff
@@ -906,14 +906,32 @@ def topics_in_stream(self, stream_id: int) -> List[str]:
...
+
+        if response.get("result") == "success":
+            email_address = response.get("email", "")
+            return str(email_address)
+        return None
+
+    def get_stream_email_address(self, stream_id: int) -> Optional[str]:
+        """
+        Attempts to retrieve stream email from stream_dict, fetching
+        from method if unavailable. (Retains handling across server versions).
```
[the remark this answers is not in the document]
neiljp: I simplified the docstring further when I just pushed back here.

tests/model/test_model.py
neiljp: I think this would likely complicate it, since we wish to have the option to completely remove the email address from the dict, based on the parametrize.

This may become more complex when typing the tests, as the small dict fragment we're using here is not anything like a full Stream or Subscription object.

tests/model/test_model.py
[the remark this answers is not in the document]
neiljp: A test for this would be ideal, and I'd recommend looking at how you'd write it, but for this I'd prefer to get this merged and move on:
- this is an internal failure mode, not for control flow; if we get an error, it's the other function that is failing and should have a test
- tidying of the stream/subscription dict data may mean that these kinds of errors change, and so this would lead to a cascading need to add/remove/adjust these tests

tests/ui_tools/test_popups.py
neiljp: Coming back to this, I was referring to the model data, ie. whether we need to have the `del` statement there - it should not depend on the email_address in the stream.

We could rewrite this test to ensure this, but the rest of the function directly accesses the stream_dict for now, so this can likely wait until it gets tidied-away as per another comment.

neiljp:
@rsashank Thanks for your work on this, and everyone's reviews too :+1:

I made some minor changes and pushed back to merge. I dropped the last commit since while it was small and would not be a problem, this would likely make sense with other similar changes.

I'll be merging shortly, and plan to post the range-diff to the GSoC stream to discuss :)

Issue #1444, opened 2023-11-29
Add guest user suffix, depending on organization setting


neiljp:
From feature level 216 (Zulip 8.0), `realm_enable_guest_user_indicator` is present in the response from `register` (and in events), which should control whether we add a `(guest)` suffix to users with that role.

The web app UI needs checking for which locations this appears, which likely needs a development environment in order to explore behavior for when this setting is enabled or otherwise.

We should also confirm that
- the suffix wasn't present unconditionally prior to this
- this behaves correctly in a backwards-compatible way

However, this will also need tests, which may need to be feature level (ZFL) dependent.

AnkurPrabhu:
@neiljp can i pick this up ?

neiljp:
@AnkurPrabhu Sure! Apologies, I didn't see your message earlier!

neiljp:
This is implemented in mobile via zulip/zulip-mobile#5804, and based on the release announcement on chat.zulip.org, hides the guest suffix only when on 8.0+ and the organization has the setting to do so. There are also screenshots that suggest places that will need updating in zulip/zulip-mobile#5809.

AnkurPrabhu:
@zulipbot claim

zulipbot:
Welcome to Zulip, @AnkurPrabhu! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

AnkurPrabhu:
Hey @neiljp just wanted to say that i am still working on this issue some of the changes are done will raise a pr and share it if you approve it will begin writing test for it

zulipbot:
@AnkurPrabhu We noticed that you have not made any updates to this issue or linked PRs for 10 days. Please comment here if you are still actively working on it. Otherwise, we'd appreciate a quick `@zulipbot abandon` comment so that someone else can claim this issue and continue from where you left off.

[...]

PR #1460, opened 2024-01-12
ui: Add guest user suffix

AnkurPrabhu:
This commit is for adding guest user suffix in the UI if the user is a guest user and if realm_enable_guest_user_indicator setting is True. 

This PR Implements #1444  
The approach here is to update the function user_name_from_id to return the user's name with the suffix (guest) if the condition satisfies and calling the function in places where we display user's name.
The approach for Right column view is different from others we just pass the required arguments (realm_enable_guest_user_indicator  and is_guest) to UserButton and add suffix to the users name depending upon the arguments

### Outstanding aspect(s)    

- [ ] wanted to know if my approach is right 
- [ ] other scenarios where the display change needs to be done
- [ ] will fix the test cases once approach is approved 

- [x] Discussed in **#zulip-terminal** in `Add guest user suffix #T1444 #T1460`
- [ ] Fully fixes #
- [x] Partially fixes issue #1444
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

Screenshot 2024-01-13 at 2 15 00 AM
Screenshot 2024-01-13 at 2 15 07 AM
Screenshot 2024-01-13 at 2 15 24 AM
Screenshot 2024-01-13 at 2 15 39 AM
Screenshot 2024-01-13 at 2 20 13 AM

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -2094,3 +2094,12 @@ class Model:
                         self.controller.raise_exception_in_main_thread(
                             sys.exc_info(), critical=False
                         )
+
+    def is_guest_user(self, user_id: int) -> bool:
+        if self.initial_data.get("realm_enable_guest_user_indicator", None):
+            user = self.get_user_info(user_id)
+            if not user:
+                raise RuntimeError("Invalid user ID.")
+            role = user["role"]
+            return role == 600
+        return False
--- a/zulipterminal/themes/zt_dark.py
+++ b/zulipterminal/themes/zt_dark.py
@@ -68,6 +68,7 @@ STYLES = {
     'task:error'      : (Color.WHITE,               Color.DARK_RED),
     'task:warning'    : (Color.WHITE,               Color.BROWN),
     'ui_code'         : (Color.BLACK,               Color.WHITE),
+    'guest_suffix'    : (Color.WHITE,               Color.BLACK),
 }
 
 META = {
--- a/zulipterminal/ui_tools/buttons.py
+++ b/zulipterminal/ui_tools/buttons.py
@@ -269,6 +269,7 @@ class UserButton(TopButton):
         color: Optional[str] = None,
         count: int,
         is_current_user: bool = False,
+        is_guest: bool = False,
     ) -> None:
         # Properties accessed externally
         self.email = user["email"]
@@ -290,6 +291,11 @@ class UserButton(TopButton):
         if is_current_user:
             self.suffix_style = "current_user"
             self.suffix_text = "(you)"
+            self.update_widget()
+
+        if is_guest:
+            self.suffix_style = "current_user"
+            self.suffix_text = "(guest)"
             self.update_widget()
 
     def _narrow_with_compose(self) -> None:
--- a/zulipterminal/ui_tools/messages.py
+++ b/zulipterminal/ui_tools/messages.py
@@ -91,16 +91,24 @@ class MessageBox(urwid.Pile):
             if self._is_private_message_to_self():
                 recipient = self.message["display_recipient"][0]
                 self.recipients_names = recipient["full_name"]
+                self.recipients_names += (
+                    " (guest)" if self.model.is_guest_user(recipient["id"]) else ""
+                )
                 self.recipient_emails = [self.model.user_email]
                 self.recipient_ids = [self.model.user_id]
             else:
                 self.recipients_names = ", ".join(
                     [
-                        recipient["full_name"]
+                        (
+                            recipient["full_name"] + " (guest)"
+                            if self.model.is_guest_user(recipient["id"])
+                            else ""
+                        )
                         for recipient in self.message["display_recipient"]
                         if recipient["email"] != self.model.user_email
                     ]
                 )
+
                 self.recipient_emails = [
                     recipient["email"]
                     for recipient in self.message["display_recipient"]
@@ -683,11 +691,15 @@ class MessageBox(urwid.Pile):
         any_differences = any(different.values())
 
         if any_differences:  # Construct content_header, if needed
-            text_keys = ("author", "star", "time", "status")
+            text_keys = ("author", "suffix", "star", "time", "status")
             text: Dict[str, urwid_MarkupTuple] = {key: (None, " ") for key in text_keys}
 
             if any(different[key] for key in ("recipients", "author", "24h")):
                 text["author"] = ("msg_sender", message["this"]["author"])
+                if "sender_id" in self.message and self.model.is_guest_user(
+                    self.message["sender_id"]
+                ):
+                    text["suffix"] = ("guest_suffix", "(guest)")
 
                 # TODO: Refactor to use user ids for look up instead of emails.
                 email = self.message.get("sender_email", "")
@@ -713,7 +725,8 @@ class MessageBox(urwid.Pile):
             content_header = urwid.Columns(
                 [
                     ("pack", urwid.Text(text["status"])),
-                    ("weight", 10, urwid.Text(text["author"])),
+                    ("pack", urwid.Text(text["author"])),
+                    ("weight", 10, urwid.Text(text["suffix"])),
                     (26, urwid.Text(text["time"], align="right")),
                     (1, urwid.Text(text["star"], align="right")),
                 ],
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -733,6 +733,7 @@ class RightColumnView(urwid.Frame):
                 user["user_id"], 0
             )
             is_current_user = user["user_id"] == self.view.model.user_id
+            is_guest = self.view.model.is_guest_user(user["user_id"])
             users_btn_list.append(
                 UserButton(
                     user=user,
@@ -742,6 +743,7 @@ class RightColumnView(urwid.Frame):
                     color=f"user_{status}",
                     count=unread_count,
                     is_current_user=is_current_user,
+                    is_guest=is_guest,
                 )
             )
         user_w = UsersView(self.view.controller, users_btn_list)
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -1136,7 +1136,11 @@ class UserInfoView(PopUpView):
         user_details = [
             (key, value) for key, value in display_data.items() if key != "Name"
         ]
-        user_view_content = [(display_data["Name"], user_details)]
+        display_full_name = controller.model.user_name_from_id(user_id)
+        if controller.model.is_guest_user(user_id):
+            display_full_name += " (guest)"
+
+        user_view_content = [(display_full_name, user_details)]
 
         if display_custom_profile_data:
             user_view_content.extend(
@@ -1525,7 +1529,14 @@ class StreamMembersView(PopUpView):
         model = controller.model
 
         user_ids = model.get_other_subscribers_in_stream(stream_id=stream_id)
-        user_names = [model.user_name_from_id(id) for id in user_ids]
+        user_names = [
+            (
+                model.user_name_from_id(id) + " (guest)"
+                if model.is_guest_user(id)
+                else model.user_name_from_id(id)
+            )
+            for id in user_ids
+        ]
         sorted_user_names = sorted(user_names)
         sorted_user_names.insert(0, model.user_full_name)
         title = "Stream Members (up/down scrolls)"
@@ -1573,12 +1584,17 @@ class MsgInfoView(PopUpView):
         full_raw_message_keys = "[{}]".format(
             ", ".join(map(str, keys_for_command("FULL_RAW_MESSAGE")))
         )
+
+        sender_name_with_guest_suffix = msg["sender_full_name"]
+        if controller.model.is_guest_user(msg["sender_id"]):
+            sender_name_with_guest_suffix += " (guest)"
+
         msg_info = [
             (
                 "",
                 [
                     ("Date & Time", date_and_time),
-                    ("Sender", msg["sender_full_name"]),
+                    ("Sender", sender_name_with_guest_suffix),
                     ("Sender's Email ID", msg["sender_email"]),
                 ],
             )
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -234,6 +234,7 @@ class TestController:
         }
 
         emails = [user_email]
+        mocker.patch.object(controller.model, "is_guest_user", return_value=False)
 
         controller.narrow_to_user(recipient_emails=emails)
 
@@ -269,6 +270,7 @@ class TestController:
         }
         controller.model.muted_streams = set()
         mocker.patch(MODEL + ".is_muted_topic", return_value=False)
+        mocker.patch.object(controller.model, "is_guest_user", return_value=False)
 
         controller.narrow_to_all_messages(contextual_message_id=anchor)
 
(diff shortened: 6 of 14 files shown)


neiljp (commented):
@AnkurPrabhu Thanks for working on this :) :+1:

The UI you have in your initial screenshots look like a good starting point 👍 The main concern would be to have consistent styling for the guest suffix, and certainly a space from the name. Specifically, we need it to be distinct from the user name, so it is clear that it is not part of the user name! A new style may be warranted for this purpose, since keeping it distinct will change with theme (see the themes main file and individual themes).

I've noted inline some general points and also regarding keeping the model and UI logic separate from each other.

While it seems like this is a relatively small change, particularly when including tests you will likely find it clearer to split this into smaller commits and address the changes to the model and each UI area separately. That makes it easier to merge good parts earlier, if the commits are relatively independent and can be reordered, or set up with simpler commits first. The user list changes look like they may be more complex, for example.

We may need to decide in a followup or in this PR what to do if text wraps. I think the server had a similar issue with the user list at one point - eg. do we prioritize the user name or the guest marker, or both? This also applies if the active user is a guest!

zulipterminal/ui_tools/buttons.py
```diff
@@ -287,6 +289,12 @@ def __init__(
             show_function=self._narrow_with_compose,
             count=count,
         )
+
+        if realm_enable_guest_user_indicator and is_guest:
+            self.suffix_style = "current_user"
+            self.suffix_text = "(guest)"
+            self.update_widget()
```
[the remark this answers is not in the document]
AnkurPrabhu: the web app does not contain '(you)' for the current user but on hover it overrides the guest suffix take a look at the screenshot attached, Ill put this down the code so that '(you)' gets overridden if you want me to add guest for the active user as well.  
Screenshot 2024-01-16 at 7 42 13 PM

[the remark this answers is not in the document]
neiljp: Yeah, this will be part of the challenge with how to render this in the user list. I'm not sure there's an obvious solution given our platform constraints, but isolating these changes to their own commit will make this easier to examine separately.

zulipterminal/ui_tools/views.py
```diff
@@ -729,7 +730,12 @@ def users_view(self, users: Any = None) -> Any:
             unread_count = self.view.model.unread_counts["unread_pms"].get(
                 user["user_id"], 0
             )
+            realm_enable_guest_user_indicator = self.model.initial_data[
+                "realm_enable_guest_user_indicator"
+            ]
```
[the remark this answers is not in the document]
AnkurPrabhu: Hey for this ill use 
realm_enable_guest_user_indicator = self.model.initial_data.get(
                "realm_enable_guest_user_indicator", None
            )
so that it returns None when not present, also are you suggesting not to use this here ?

zulipterminal/ui_tools/messages.py
```diff
@@ -713,7 +718,8 @@ def main_view(self) -> List[Any]:
             content_header = urwid.Columns(
                 [
                     ("pack", urwid.Text(text["status"])),
-                    ("weight", 10, urwid.Text(text["author"])),
+                    ("pack", urwid.Text(text["author"])),
+                    ("weight", 10, urwid.Text(text["suffix"])),
```
AnkurPrabhu, line 722: added like a new widget

AnkurPrabhu:
Hey @neiljp i have tried a new approach in this one pls let me know your thoughts also as you told to split it in different prs this pr will be related to adding the guest suffix in the messages box (changes related to header and recipients).
For displaying the guest suffix in the right column that will be a separate pr but the changes in the button have been added here as I wanted to know your thoughts.

zulipterminal/model.py
```diff
@@ -1262,12 +1265,17 @@ def user_name_from_id(self, user_id: int) -> str:
...
-
-        return self.user_dict[user_email]["full_name"]
+        role = user["role"]
+        name = user["full_name"]
+        is_guest = role == 600
+        guest_user_indicator = self.initial_data["realm_enable_guest_user_indicator"]
+        if guest_user_indicator and is_guest:
+            return name + "(guest)"
+        else:
+            return name
```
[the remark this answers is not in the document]
neiljp: I mainly meant the name and the suffix as the two parts.

In terms of 'where the suffix goes', the styling and placement of the suffix is not something that the model should be handling - that is an issue for the UI. The model should only be concerned with providing an answer to "should I show a guest suffix for this user".

zulipterminal/ui_tools/views.py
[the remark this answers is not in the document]
neiljp: Right, that will avoid the exception.

As per another comment, in (new) code we are aiming to have server changes (different versions) isolated in the model, so that UI code isn't dependent on that.

zulipterminal/themes/zt_dark.py
```diff
@@ -68,6 +68,7 @@
     'task:error'      : (Color.WHITE,               Color.DARK_RED),
     'task:warning'    : (Color.WHITE,               Color.BROWN),
     'ui_code'         : (Color.BLACK,               Color.WHITE),
+    'guest_suffix'    : (Color.WHITE,               Color.BLACK),
```
neiljp, line 71: This is one place where a style can be added to a theme, but note there are others - it will need adding in each place.

For example, look for one of the other style names throughout the code.

See previous style-adding commits in the git log for examples - we tend to add a style in all places in a single commit.

zulipterminal/ui_tools/utils.py
```diff
@@ -30,6 +31,10 @@ def create_msg_box_list(
     last_msg = last_message
     muted_msgs = 0  # No of messages that are muted.
     for msg in message_list:
+        if model.initial_data.get("realm_enable_guest_user_indicator", None):
+            user_info = model.get_user_info(msg["sender_id"])
+            role = user_info["role"]
+            msg["sender_role"] = ROLE_BY_ID[role]["name"]
```
neiljp, line 37: I understand what you're trying to achieve here, ie. caching the name of the role in messages.

This does work, but the messages data structure already potentially has fields dedicated to various roles, as well as the main role field in late versions.

neiljp:
@AnkurPrabhu Thanks for following up. Unfortunate I didn't fetch the previous version, so it's difficult for me to identify precisely what you've changed in this version. I've left some notes inline, as well as responses to earlier comments.

Please note:
- I was mainly referring to multiple commits; these can be with the same PR if they work together
- some suffix work can certainly be left to another PR, but the foundational work likely wants to be in the first PR
- I know you left the button changes in, but you could still keep these in the PR, but in a different commit :)

For the next step, I'd suggest splitting your work here into multiple commits in this PR, and thinking how you might write tests for each new code. Firstly I'd advise adding a model method like I mentioned in a response to an earlier comment, and then adapting the rest of your code to use that method.

neiljp (commented):
@AnkurPrabhu I left my main comments in another comment, which refers to these comments too :)

zulipterminal/model.py
```diff
@@ -143,6 +143,9 @@ def __init__(self, controller: Any) -> None:
             # zulip_version and zulip_feature_level are always returned in
             # POST /register from Feature level 3.
             "zulip_version",
+            # realm_enable_guest_user_indicator is present  in
+            # POST /register from Feature level 216.
+            "realm_enable_guest_user_indicator",
```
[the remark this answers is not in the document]
AnkurPrabhu: yes its not necessary ill remove it

zulipterminal/themes/zt_dark.py
AnkurPrabhu: yes i saw there are a lot of styles i just wanted to show this to know if im on the right track or not, will add it in all the files

zulipterminal/ui_tools/utils.py
AnkurPrabhu: so ill make a separate method which will make the two parts of the name the name and suffix 
by default the suffix will be "" (blank)  so that way we do not have to check any setting from model and wherever we are displaying names we will call this method ? does this work ?

neiljp: Yes, that may be useful, though it may depend on how the UI works in different places, eg. our custom buttons do not accept the same input as a simple text box. If you apply it to one area first, then you can just have the logic at that position.

The logic may be quite small, once you have the new method in the model, to indicate if a guest suffix should be shown for a given user id.

AnkurPrabhu:
@neiljp this i have updated the pr with the latest changes:
1) have added style to all the current themes
2) currently showing guest user suffix at
- message header
- message recipients
- right column (button)
- message info view 
- message stream members view
- user info view 
let me know if I am missing the suffix at any place 

all these felt like similar changes only so once you approve my current approach will start writing tests for the same .
also i was thinking of making a constant in the model like guest_suffix= "( guest)" so that we can use this everywhere instead if writing it as a string everywhere... @neiljp  thoughts ?

AnkurPrabhu:
@neiljp i have fixed all the existing tests can you tell me how to fix this gitlint one ?

neiljp:
@AnkurPrabhu The test error appears to be due to your branch not including the setup code for the gitlint in CI. This should resolve if you fetch the latest `main` branch and rebase this branch against the updated `main` which has the appropriate commits to install gitlint in CI.

However, I suspect you will have gitlint errors locally too. See the docs, but eg try something like `gitlint --commits HEAD~3..` (or with `main..`, once you're up to date with main).

AnkurPrabhu:
> @neiljp this i have updated the pr with the latest changes:
> 
> 1. have added style to all the current themes
> 2. currently showing guest user suffix at
> 
> * message header
> * message recipients
> * right column (button)
> * message info view
> * message stream members view
> * user info view
>   let me know if I am missing the suffix at any place
> 
> all these felt like similar changes only so once you approve my current approach will start writing tests for the same . also i was thinking of making a constant in the model like guest_suffix= "( guest)" so that we can use this everywhere instead if writing it as a string everywhere... @neiljp thoughts ?

@neiljp you missed this i guess if the changes are good ill write tests and we can close this

neiljp:
@AnkurPrabhu Those locations look good; if we miss one then it can be added later.

I agree that extracting a common constant would be good, but note again that this is not something that would belong in the model, for the same reasons as before.

zulipterminal/model.py
```diff
@@ -2071,3 +2071,17 @@ def poll_for_events(self) -> None:
                         self.controller.raise_exception_in_main_thread(
                             sys.exc_info(), critical=False
                         )
+
+    def show_guest_suffix(self, user_id: int) -> bool:
+        """
+        Returns True if the realm_enable_guest_user_indicator
+        setting is enabled and the user is a guest
+        """
```
neiljp, line 2079: A docstring is for users of the function, not a comment as to how it works.

If you feel there is a need for a description, beyond the name of the function, then please adjust the docstring - but it should not describe the implementation.

zulipterminal/model.py
```diff
@@ -2071,3 +2071,17 @@ def poll_for_events(self) -> None:
...
+        user = self.get_user_info(user_id)
+        assert user is not None
```
neiljp, line 2081: Based on existing model code, let's raise a RuntimeError here.

We should move towards better exceptions for these cases in future though.

zulipterminal/model.py
```diff
@@ -2071,3 +2071,17 @@ def poll_for_events(self) -> None:
                         self.controller.raise_exception_in_main_thread(
                             sys.exc_info(), critical=False
                         )
+
```
neiljp, line 2074: You've split commits more here, but note the commit text - you're updating model.py, so 'ui' doesn't apply to the 'area' for the title.

neiljp (commented):
@AnkurPrabhu I just left some feedback in addition to answering your questions.

The new model method is in the right direction, and the split commits are an improvement :+1:

However, please do split the UI changes into separate ones for each UI part that you update.

zulipterminal/model.py
```diff
@@ -2071,3 +2071,17 @@ def poll_for_events(self) -> None:
...
+        setting is enabled and the user is a guest
+        """
+        user = self.get_user_info(user_id)
+        assert user is not None
+        role = user["role"]
+
+        return (
+            self.initial_data.get("realm_enable_guest_user_indicator", None)
+            and ROLE_BY_ID[role]["name"] == ROLE_BY_ID[600]["name"]
+        )
```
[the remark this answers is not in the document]
AnkurPrabhu: I am using ( and ) condition here so if the first one returns false it will never calculate the other half of the condition right ?  
yes it looks complicated ill make it more readable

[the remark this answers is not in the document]
neiljp: The expression won't calculate the later part if the former is false, but `role` itself will still be calculated in the part above this, independently of that.

zulipterminal/themes/zt_dark.py
AnkurPrabhu: I have done it in all the files right ?

AnkurPrabhu:
@neiljp I have split it based on ui. 
I have not split the message info view as i thought they are part of the message itself.
I Have fixed all the existing tests. 

@neiljp  can you guide me on what are the next steps or changes i should work on

zulipterminal/ui_tools/messages.py
```diff
@@ -91,16 +92,24 @@ def __init__(self, message: Message, model: "Model", last_message: Any) -> None:
...
-                )
+                self.recipients_names = ""
+                recipients = []
+                for recipient in self.message["display_recipient"]:
+                    if recipient["email"] != self.model.user_email:
+                        recipient_name = self.recipients_names = recipient["full_name"]
+                        if self.model.show_guest_suffix(user_id=recipient["id"]):
+                            recipient_name += " (guest)"
+                        recipients.append(recipient_name)
+                self.recipients_names = ", ".join(recipients)
```
mounilKshah, line 111: This works fine but you can also maintain the original python comprehension and only add the suffix to the `recipient_name` with a conditional, similar to how it's done above in the private message conditional block

mounilKshah (commented):
Thank you for updating the PR @AnkurPrabhu 
This works fine on my system. I think the commit titles and messages for all commits can be modified in accordance with the Zulip commit standards.

I had a few suggestions and have mentioned them in the review.

neiljp (commented):
@AnkurPrabhu Thanks for making the changes :+1:
(as noted in the stream, I started this yesterday, but had issues on my reviewing computer)

I've yet to test this on a live server with guest users - have you done so?

I've left some followup comments where you or I missed something.

In addition to the inline comments:
- Somewhere we discussed adding the `(guest)` string as a constant? (https://github.com/zulip/zulip-terminal/pull/1460#issuecomment-1925002202) The `ui_mappings.py` file seems like a reasonable place for this; it can be in it's own commit, and then we can squash it with the first UI commit that gets merged.
- It's good to see the user list changes in a separate commit, but could you separate the changes to each other UI part separately too? I expect they should be independent of each other, and that would make it much easier to see the changes associated with each, test them separately, reorder them, or work on them later - and the commit title/body can vary for each to explain the feature you are adding in that specific commit.
- Each commit title should be able to be relatively standalone, and describe what it does itself, not the PR overall; this might be just due to the current commits not being updated, but that also helps with review - it explains each commit as part of the set.

The user column looks like it is going to be the more complex part, so that commit might be better at the end. This may need to be a symbol, as for text like elsewhere it needs more space, and so might be dependent upon a dynamic larger column width if the window is wide enough, which someone looked into working on a while back.

PR #1470, opened 2024-02-14
Add support for copying raw message.

Sushmey:
Adding a key to copy raw message content since manually copying text in a terminal is pretty difficult due to the bars on the sides. Implemented using preexisting `copy_to_clipboard` function.

- [ ] Fully fixes #
- [x] Partially fixes issue #546
- [x] Is a follow-up to work in PR #1365
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

Screenshot 2024-02-14 at 7 45 35 PM

### DIFF
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -64,6 +64,7 @@
 |View current message in browser (from message information)|<kbd>v</kbd>|
 |Show/hide full rendered message (from message information)|<kbd>f</kbd>|
 |Show/hide full raw message (from message information)|<kbd>r</kbd>|
+|Copy message content to clipboard (from message information)|<kbd>c</kbd>|
 
 ## Stream list actions
 |Command|Key Combination|
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -557,6 +557,13 @@ class TestFullRawMsgView:
         assert self.full_raw_message.header.widget_list == msg_box.header
         assert self.full_raw_message.footer.widget_list == msg_box.footer
 
+        assert isinstance(self.full_raw_message.footer.widget_list[0], Text)
+        assert (
+            self.full_raw_message.footer.widget_list[0].text
+            == f"\nPress {keys_for_command('COPY_MESSAGE')}"
+            " to copy raw message content to clipboard"
+        )
+
     @pytest.mark.parametrize("key", keys_for_command("MSG_INFO"))
     def test_keypress_exit_popup(
         self, key: str, widget_size: Callable[[Widget], urwid_Size]
@@ -596,6 +603,21 @@ class TestFullRawMsgView:
             topic_links=OrderedDict(),
             message_links=OrderedDict(),
             time_mentions=list(),
+        )
+
+    @pytest.mark.parametrize(
+        "key",
+        {*keys_for_command("COPY_MESSAGE")},
+    )
+    def test_keypress_copy_full_raw_msg(
+        self, key: str, widget_size: Callable[[Widget], urwid_Size]
+    ) -> None:
+        size = widget_size(self.full_raw_message)
+
+        self.full_raw_message.keypress(size, key)
+
+        self.controller.copy_to_clipboard.assert_called_once_with(
+            self.full_raw_message.raw_message_content, "Message Content"
         )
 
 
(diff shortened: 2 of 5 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1939,14 +1939,21 @@ def __init__(
 
         # Get rendered message header and footer
         msg_box = MessageBox(message, controller.model, None)
+        msg_box.footer = [
```
neiljp, line 1942: This changes the existing behavior, since the footer stores footlinks and reactions in message boxes.

zulipterminal/ui_tools/views.py
```diff
@@ -1939,14 +1939,21 @@ def __init__(
...
+            (
+                urwid.Text(
+                    ("msg_bold", "\nPress [c] to copy message to clipboard"),
```
neiljp, line 1945: For specific keys, we always want to avoid hard-coding the key in the text, and rather lookup the key name via it's symbolic name: `COPY_MESSAGE`.

tests/ui_tools/test_popups.py
```diff
@@ -598,6 +598,23 @@ def test_keypress_show_msg_info(
             time_mentions=list(),
         )
 
+    @pytest.mark.parametrize(
+        "key",
+        {
+            *keys_for_command("COPY_MESSAGE"),
+        },
```
neiljp, line 605: nit: this fits on one line? It's probably only the extra comma(s) that's forcing black to spread it out.

zulipterminal/ui_tools/views.py
neiljp, line 1945: I'm not sure of the wording here, but including `content` in the text seems useful, particularly if we keep the footer and header as they are in `main`.

zulipterminal/ui_tools/views.py
```diff
@@ -1939,14 +1939,21 @@ def __init__(
...
+                    align="left",
+                )
+            )
+        ]
 
         # Get raw message content widget list
-        response = controller.model.fetch_raw_message_content(message["id"])
-
-        if response is None:
+        self.response = controller.model.fetch_raw_message_content(message["id"])
```
neiljp, line 1952: Since we're using this more widely now, it would be useful to give it a name which is clearer - so that we don't have to return here to look up what it means. That seems to be what you do when you use it on line ~1978?

You could also consider this change to be a refactor prep commit, since you've not updated `test_init`.

docs/hotkeys.md
```diff
@@ -64,6 +64,7 @@
 |View current message in browser (from message information)|<kbd>v</kbd>|
 |Show/hide full rendered message (from message information)|<kbd>f</kbd>|
 |Show/hide full raw message (from message information)|<kbd>r</kbd>|
+|Copy message content to clipboard|<kbd>c</kbd>|
```
neiljp, line 67: This doesn't have any `(from ...)` suffix like the others, which could imply that it works just when a message is selected.

neiljp (commented):
@Sushmey This looks to work well for the most part :+1:

I've left some mostly-minor feedback inline.

An alternative/follow-up to this would be to improve our popups more generally to include general & specific keys, as per #1002, so we could include `c` in the popup itself.

zulipterminal/ui_tools/views.py
Sushmey: You mean changing the variable name to something like `self.raw_message_content`? 

I'll create a different refactor PR changing the variable name and then rebase this one.

zulipterminal/ui_tools/views.py
Sushmey: Yes I noticed that but it's still a widget_list that is being returned and since originally footer was an empty list in the `FullRawMsgView` so the earlier assertion passes so didn't need to change anything in `test_init`.

Adding two new assertions now to check if the widget_list element is urwid.Text and to check if the text is what it's supposed to be.

zulipterminal/ui_tools/views.py
Sushmey: Yes that makes more sense! Making that change.

tests/ui_tools/test_popups.py
Sushmey: Agreed. Changed.

zulipterminal/ui_tools/views.py
Sushmey: How's "Press [c] to copy raw message content to clipboard" ?

docs/hotkeys.md
Sushmey: Missed that text. Adding `(from message information)`. Or should I add something like `(from raw message view)`? Although it might be confusing for new users since they might not know where raw message view is.

PR #1473, opened 2024-02-21
Fix ZT crash with an Index Error.

### LINKED ISSUE #1226: IndexError on starting (focus index vs message id)
This was discussed in #**zulip-terminal > IndexError while starting.**, and has been reported by:
- Aryan Shridhar
- @Dishti-Oberai 
- @rht 

This appears to be distinct from #326.

ZT is reported to crash before the UI loads.

The traceback ends with
```
IndexError: focus index is out of range: 1330877
```
(where the last number varies)

The bug currently occurs due to views.py line 119, on the assumption that:
- the model provides a current focused message id
- the focused message is set to the id
- the initializer for MessageView sets the listbox focus to that id...
- ...which is likely out of range, given ~30 messages in list, and message ids potentially being huge!

A temporary workaround is to remove that statement, but a fuller solution should likely consider how we handle/retain focus between narrows.

theViz343:
This PR fixes the Index Error mentioned in  #1226. The PR resolves this by setting the focus pointer to the message index instead of the message ID. Other functions depending on the focus pointer also treat it as the message index, so the change in this PR keeps its usage consistent.

### Outstanding aspect(s)    

- [ ] 

- [x] Discussed in **#zulip-terminal** in `IndexError while starting.`
- [x] Fully fixes #1226 
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -1454,14 +1454,14 @@ class TestModel:
         )
 
         # Setup mocks before calling get_messages
-        messages_successful_response["anchor"] = 0
         self.client.get_messages.return_value = messages_successful_response
         mocker.patch(MODULE + ".index_messages", return_value=index_all_messages)
 
         model = Model(self.controller)
+        messages_successful_response["anchor"] = 0
         model.get_messages(num_before=num_before, num_after=num_after, anchor=0)
         self.client.get_messages.return_value = messages_successful_response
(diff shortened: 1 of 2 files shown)


neiljp:
@theViz343 Just noting here as I mentioned in the stream, that this may fix the bug, but it would be more consistent to approach this the other way around to avoid the UI index being stored in the model.

Issue #1372, opened 2023-04-08
Message information popup actions (open, copy, ...) vs content (narrows, links, code, ...)


neiljp:
For a given message, various content is indexed and may be scrolled through when in the message information popup.

Depending upon the message/topic content, this can include one or more of:
- links in the topic (often linkifiers eg. #1372)
- external links (name + URL)
- media links (~attachments)
- narrow links (eg. #zulip-terminal>some topic)
- code snippets (see #1123 WIP)

For entries in each section, we likely want to support various actions. Right now we support scrolling amongst all the buttons that *could* be activated, but only some support an action, and it is not obvious that they do in advance:
- (most) narrow links will go to the indicated narrow
- media links (attachments) will offer to download and open the attachment

Ultimately we may wish
- every button to support 'copy'
- external links and topic links to be opened in a browser
- code snippets to support an 'open in playground'

Before we implement more entries it would be useful to
- [ ] Indicate if activating a focusable entry will do anything, and with what key (eg. `[<key>]`)
- [ ] Explore adding copying to each entry, including how to handle multiple actions on an entry

See previous discussion in [#zulip-terminal > Clarifying message actions](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Clarifying.20message.20actions)

PR #1478, opened 2024-03-28
[WIP] Add help text on how to interact with message links.

Sushmey:

- [x] Discussed in [Open external links with browser #T1440](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Open.20external.20links.20with.20browser.20.23T1440)
- [ ] Fully fixes #
- [x] Partially fixes issue #1372 
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

Screenshot 2024-03-28 at 11 38 11 PM

### DIFF
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -183,6 +183,43 @@ class TestPopUpView:
     @pytest.mark.parametrize(
         [
             "contents",
+            "expected_popup_width",
+            "expected_column_width",
+        ],
+        [
+            case(
+                [
+                    (
+                        "Viewing Actions",
+                        [
+                            ("Open in web browser", "[]"),
+                            ("Full rendered message", "[]"),
+                            ("Full raw message", "[]"),
+                        ],
+                    )
+                ],
+                25,
+                [21, 2],
+                id="heading_widget_with_row",
+            ),
+            case([("", "")], 17, [], id="widget_with_str"),
+        ],
+    )
+    def test_calculate_table_widths(
+        self,
+        contents: PopUpViewTableContent,
+        expected_popup_width: int,
+        expected_column_width: List[int],
+    ) -> None:
+        popup_width, column_width = self.pop_up_view.calculate_table_widths(
+            contents, len(self.title)
+        )
+        assert popup_width == expected_popup_width
+        assert column_width == expected_column_width
+
+    @pytest.mark.parametrize(
+        [
+            "contents",
             "expected_widget_type",
         ],
         [
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -4,7 +4,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple
 import pytest
 from pytest import param as case
 from pytest_mock import MockerFixture
-from urwid import Columns, Pile, Text, Widget
+from urwid import AttrWrap, Columns, Pile, Text, Widget
 
 from zulipterminal.api_types import Message
 from zulipterminal.config.keys import is_command_key, keys_for_command
@@ -24,6 +24,7 @@ from zulipterminal.ui_tools.views import (
     MsgInfoView,
     PopUpConfirmationView,
     PopUpView,
+    PopUpViewTableContent,
     StreamInfoView,
     StreamMembersView,
     UserInfoView,
@@ -178,6 +179,42 @@ class TestPopUpView:
         self.pop_up_view.keypress(size, navigation_key)
 
         self.super_keypress.assert_called_once_with(size, navigation_key)
+
+    @pytest.mark.parametrize(
+        [
+            "contents",
+            "expected_widget_type",
+        ],
+        [
+            case(
+                [
+                    (
+                        "Viewing Actions",
+                        [
+                            ("Open in web browser", "[]"),
+                            ("Full rendered message", "[]"),
+                            ("Full raw message", "[]"),
+                        ],
+                    )
+                ],
+                [Text, AttrWrap, AttrWrap, AttrWrap],
+                id="heading_widget_with_row",
+            ),
+            case([("", "")], [], id="widget_with_str"),
+        ],
+    )
+    def test_make_table_with_categories__instances(
+        self,
+        contents: PopUpViewTableContent,
+        expected_widget_type: List[Any],
+    ) -> None:
+        popup_width, column_widths = self.pop_up_view.calculate_table_widths(
+            contents, len(self.title)
+        )
+        widgets = self.pop_up_view.make_table_with_categories(contents, column_widths)
+        assert len(widgets) == len(expected_widget_type)
+        for widget, expected_type in zip(widgets, expected_widget_type):
+            assert isinstance(widget, expected_type)
 
 
 class TestAboutView:
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -203,6 +203,12 @@ class TestPopUpView:
                 id="heading_widget_with_row",
             ),
             case([("", "")], 17, [], id="widget_with_str"),
+            case(
+                [(("Message Links", "interact with links"), [])],
+                21,
+                [],
+                id="heading_widget_with_help_text",
+            ),
         ],
     )
     def test_calculate_table_widths(
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -201,6 +201,11 @@ class TestPopUpView:
                 id="heading_widget_with_row",
             ),
             case([("", "")], [], id="widget_with_str"),
+            case(
+                [(("Message Links", "interact with links"), [])],
+                [Columns],
+                id="heading_widget_with_help_text",
+            ),
         ],
     )
     def test_make_table_with_categories__instances(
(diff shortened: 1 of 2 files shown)


zulipterminal/ui_tools/views.py
```diff
@@ -1605,7 +1621,21 @@ def __init__(
...
         if message_links:
-            msg_info.append(("Message Links", []))
+            interact_with_link_keys = "[{}]".format(
+                ", ".join(map(str, keys_for_command("ENTER")))
+            )
+            msg_info.append(
+                (
+                    (
+                        "Message Links",
+                        f"Press {interact_with_link_keys} to interact with links",
```
neiljp, line 1631: This change is distinct from the change to the category feature that supports it - or to put it differently: can this specific change be included only with similar changes for the other categories which may appear?

This feature can also be applied to other popups, though we may want to discuss that first.

neiljp (commented):
@Sushmey I left some preliminary feedback. Please let me know here or on czo if you have points that I've not answered.

zulipterminal/ui_tools/views.py
```diff
@@ -1041,16 +1041,32 @@ def make_table_with_categories(
             if category:
                 if len(widgets) > 0:  # Separate categories with newline.
                     widgets.append(urwid.Text(""))
-                widgets.append(urwid.Text(("popup_category", category)))
+                if isinstance(category, tuple):
```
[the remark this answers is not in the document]
Sushmey: Yes ran into this while pushing but couldn't figure it out how to resolve it. Now I added the new case in the PopUpViewTableContent type so no more error :)

zulipterminal/ui_tools/views.py
```diff
@@ -1041,16 +1041,32 @@ def make_table_with_categories(
...
+                            dividechars=dividechars,
+                        )
+                    else:
+                        strip = urwid.Columns(
+                            [(column_widths[0], urwid.Text(label)), urwid.Text(data)],
+                            dividechars=dividechars,
+                        )
                     widgets.append(
                         urwid.AttrWrap(strip, None if index % 2 else "popup_contrast")
                     )
```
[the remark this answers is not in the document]
Sushmey: Yes updating the loop index is confusing. Figured out a better way to get what I expected.

No we don't need to wrap the popup_contrast. I did it because I was trying to make it visually discernible, updated it to work without the wrap.

We can have a discussion about the style.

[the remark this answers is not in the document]
neiljp: We don't necessarily need to think about a style in this first version, but at the time it struck me as an additional motivator for adjusting the loop.

neiljp:
I believe I covered your questions so far in the stream or here.

tests/ui_tools/test_popups.py
```diff
@@ -179,6 +180,44 @@ def test_keypress_navigation(
...
+                            ("Open in web browser", "[]"),
+                            ("Full rendered message", "[]"),
+                            ("Full raw message", "[]"),
+                        ],
+                    )
+                ],
+                [21, 24],
+                [Text, AttrWrap, AttrWrap, AttrWrap],
+            ),
+            ([("", "")], [21, 24], []),
```
neiljp, line 204: The `21` and `24` are the same in each test you have in this PR. Are they constants which must work with the test data in particular, or can these be constants in the test parameters/body, and simplify the test inputs?

tests/ui_tools/test_popups.py
```diff
@@ -179,6 +180,44 @@ def test_keypress_navigation(
 
         self.super_keypress.assert_called_once_with(size, navigation_key)
 
+    @pytest.mark.parametrize(
+        [
+            "contents",
+            "column_widths",
+            "expected_instance",
```
neiljp, line 187: `instance` is rather vague - what does this function return?

I believe each part of the `instance` is essentially a row, which we don't mention in the docstring and maybe in the code that uses it, but could be a useful refactor?

tests/ui_tools/test_popups.py
```diff
@@ -179,6 +180,44 @@ def test_keypress_navigation(
...
+                [Text, AttrWrap, AttrWrap, AttrWrap],
+            ),
+            ([("", "")], [21, 24], []),
+        ],
+        ids=[
+            "heading_widget_with_row",
+            "widget_with_str",
+        ],
+    )
+    def test_make_table_with_categories(
```
neiljp, line 213: Having this test is great, but it does focus on the urwid form of the output, so it may be worth adding a suffix to the test name to emphasize that, eg. `__instances` based on your naming elsewhere (which can be improved).

zulipterminal/ui_tools/views.py
neiljp: I'm not sure if we discussed this in the stream, but I was referring to having this in a different commit to the one which adds category heading support.

neiljp (commented):
@Sushmey It'd be great to get the category headings infrastructure merged, so I left some more feedback on what you have here :+1:

Please do reply if you have other issues with the typing.

neiljp:
Should the linked issue be #1372, not #1002?

tests/ui_tools/test_popups.py
Sushmey: No they are not constants, I had added them as constants for temporary purposes just to test if the test worked. Changed it so that it's not hardcoded now.

tests/ui_tools/test_popups.py
Sushmey: Changed the parameter name to `expected_widget_type` to be more descriptive since the docstring also mentions that the function is returning list of widgets.

Re second part, are you saying each widget of the list is essentially a row and that the docstring needs a refactor? I feel the current docstring is good enough, it does return a list of widgets that is used to make the table.

tests/ui_tools/test_popups.py
Sushmey: Adding an `__instances` suffix. Although I wonder if this implies there's a test that focuses on `make_table_with_categories` as a whole and not on some specific thing like instances.

neiljp: I agree that it could suggest another test exists, but also means it won't conflict with another later ;) If there are multiple tests for a function, my preferred form now is to separate the function and the intention behind the test.

I suggested `__instances` since that's what you were checking in the test, and used in the expected part of the test case name, as per another comment - so let's sync them up if you think the updated expected parameter name is better :)

tests/ui_tools/test_popups.py
neiljp: Using `widget` is consistent with the current docstring and terms used :+1:

The second point was my noting that the code might be more readable if we referred to the returned data as rows; widgets are any UI element, so are less specific.

neiljp (commented):
@Sushmey Glad to see you got the types sorted :+1: There seems to be a failing test? Maybe try `check-branch` locally, to see what commit is failing?

I replied to some previous comments and left a few more inline.

The PR could do with removing the exploratory parts, eg. commented out, and that would make it clearer what exactly is changing.

zulipterminal/ui_tools/views.py
```diff
@@ -1603,7 +1611,7 @@ def __init__(
         full_raw_message_keys = "[{}]".format(
             ", ".join(map(str, display_keys_for_command("FULL_RAW_MESSAGE")))
         )
-        msg_info = [
+        msg_info: PopUpViewTableContent = [
```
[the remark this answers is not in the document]
Sushmey: It doesn't work otherwise since `PopUpViewTableContent` also has types which doesn't support append so need an explicit assertion. 
I can't not explicitly type it as `PopUpViewTableContent` since Mypy assumes it to be the type of the first data assigned to it.

zulipterminal/ui_tools/views.py
```diff
@@ -1042,19 +1050,35 @@ def make_table_with_categories(
             if category:
                 if len(widgets) > 0:  # Separate categories with newline.
                     widgets.append(urwid.Text(""))
-                widgets.append(urwid.Text(("popup_category", category)))
+                if isinstance(category, tuple) and isinstance(content, list):
+                    content.append(category)
```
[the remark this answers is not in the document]
Sushmey: Thanks for pointing it out. There was no reason to put it as content. Changed it and removed the condition too.

tests/ui_tools/test_popups.py
```diff
@@ -179,6 +180,44 @@ def test_keypress_navigation(
...
+        self,
+        contents: PopUpViewTableContent,
+        expected_widget_type: List[Any],
+    ) -> None:
+        popup_width, column_widths = self.pop_up_view.calculate_table_widths(
+            contents, len(self.title)
+        )
+        widgets = self.pop_up_view.make_table_with_categories(contents, column_widths)
+        for index, widget in enumerate(widgets):
+            assert isinstance(widget, expected_widget_type[index])
```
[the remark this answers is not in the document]
Sushmey: zip is used to combine two iterables, right? Not sure how we can use that here.

tests/ui_tools/test_popups.py
Sushmey: I'e used `types` everywhere instead of instance because it seems more intuitive than `instances`. So let's change it to `__types` in the function name also?

neiljp: Yeah, that'd be good :+1:

tests/ui_tools/test_popups.py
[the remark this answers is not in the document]
neiljp: Yep, I already saw you changed it in the code :)

PR #1479, opened 2024-04-11
Improve hotkey rendering in documentation

### LINKED ISSUE #945: Improve key rendering in documentation
There are a few improvements we can make in the script which generates `docs/hotkeys.md`; some of the same changes could be added to the rendering of keys in the help menu too.

* [x] space-separated keys (page up/down) are rendered as separate keys in `hotkeys.md` - they should be one key
* [x] we may wish to capitalize or otherwise rename some keys, from those understood by urwid to how they should be rendered
* [ ] in particularly we should consider adding extra symbol text to some keys, eg. ←, ↑, →, ↓  (<kbd>left ←</kbd>?)
* [ ] we might discuss this, but it can be useful to show `P` as <kbd>shift</kbd>+<kbd>p</kbd> (or even <kbd>shift</kbd>+<kbd>P</kbd>?) [noted in #922]

Other ideas are welcome :)

**update** First two points fixed by d65cbee7406a3413946e78fb7fcff8ddc8875251 from #1479,

Niloth-p:

Improves the rendering of hotkeys in hotkeys.md.

- [x] Make lowercase alphabet keys display as uppercase to match their physical representation in a keyboard.
- [x] Added a prefix of Shift to uppercase alphabet keys
- [x] Use the new display_keys_for_command() helper function in keys.py
  - [x] Space-separated keys that were previously being rendered separately, are now rendered as a single key
  - [x] Special key names are capitalized
- [x] Added arrow symbols to the direction keys

### Outstanding aspect(s)
Tests are not included.

- [x] Fully fixes #945 

### How did you test this?

### Visual changes
image


Niloth-p:
@zulipbot add "area: hotkeys" "PR needs review"

tools/lint-hotkeys
```diff
@@ -104,7 +117,16 @@ def get_hotkeys_file_string() -> str:
...
                 [
-                    " + ".join([f"<kbd>{key}</kbd>" for key in key_combination.split()])
+                    " + ".join(
+                        [
+                            f"<kbd>Shift</kbd> + <kbd>{key}</kbd>"
+                            if len(key) == 1 and key.isupper()
+                            else f"<kbd>{key.capitalize()}</kbd>"
+                            if len(key) == 1
+                            else f"<kbd>{key + DIRECTION_TO_SYMBOL_MAP.get(key, '')}</kbd>"
+                            for key in key_combination.split()
```
neiljp, line 127: I replied separately in the stream, so we may not need this change as it stands.

However, when inline conditionals get large they get more difficult to read, so at a minimum I'd go with something like the following:
```suggestion
                            f"<kbd>Shift</kbd> + <kbd>{key}</kbd>"
                            if len(key) == 1 and key.isupper()
                            else (
                                f"<kbd>{key.capitalize()}</kbd>"
                                if len(key) == 1
                                else f"<kbd>{key + DIRECTION_TO_SYMBOL_MAP.get(key, '')}</kbd>"
                            )
                            for key in key_combination.split()
```

However, while it requires rewriting the comprehensions/loops, it may well be clearer to simply express it as regular `if` statements instead - particularly given one of the inline conditionals is itself a ` + `-separated string.

neiljp (commented):
@Niloth-p Thanks for this matching PR to #1476 :+1:

I replied about this in the stream, but left some comments inline too, which will apply depending on how we move forward with this.

tools/lint-hotkeys
```diff
@@ -69,7 +74,9 @@ def lint_hotkeys_file() -> None:
     else:
         print("No hotkeys linting errors")
         if not output_file_matches_string(hotkeys_file_string):
-            print(f"Run './tools/{SCRIPT_NAME}' to update {OUTPUT_FILE_NAME} file")
+            print(
+                f"Run './tools/{SCRIPT_NAME} --fix' to update {OUTPUT_FILE_NAME} file"
+            )
```
Niloth-p, line 79: The error hint was missing the '--fix' argument.

tools/lint-hotkeys
```diff
@@ -85,6 +92,27 @@ def generate_hotkeys_file() -> None:
...
+
+
+def format_key_combination(key_combination: str) -> str:
+    """
+    Format a hotkey combination for display
+    """
+    for symbol in ARROW_SYMBOLS:
+        if symbol in key_combination:
+            return f"<kbd>{key_combination}</kbd>"
+    return " + ".join([format_key(key) for key in key_combination.split()])
```
Niloth-p, line 113: Just added for an alternative option.
Also, there's the limitation that `format_key_combination()` does not work when the arrow key is used with modifier keys.

tools/lint-hotkeys
```diff
@@ -114,6 +142,10 @@ def get_hotkeys_file_string() -> str:
 
 
 def output_file_matches_string(hotkeys_file_string: str) -> bool:
+    """
+    Read the existing OUTPUT_FILE into a string,
+    and compare if it matches the input string
+    """
```
Niloth-p, line 148: This was the only function in the file missing docstrings.

zulipterminal/config/symbols.py
```diff
@@ -86,3 +86,8 @@
     bline=_MESSAGE_RECIPIENTS_BOTTOM,
     brcorner=_MESSAGE_RECIPIENTS_BOTTOM,
 )
+
+LEFT_ARROW = "←"  # LEFTWARDS ARROW, U+2190
+UP_ARROW = "↑"  # UPWARDS ARROW, U+2191
+RIGHT_ARROW = "→"  # RIGHTWARDS ARROW, U+2192
+DOWN_ARROW = "↓"  # DOWNWARDS ARROW, U+2193
```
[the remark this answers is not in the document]
Niloth-p: Just to clarify,
We're updating all direction keys (and just the direction keys) to be - text + space + symbol.
Direction keys that are used in keys.py, in hotkeys.md and the ones in core.py (popup headers).

zulipterminal/config/symbols.py
```diff
@@ -86,3 +86,9 @@
...
+ARROW_SYMBOLS = [LEFT_ARROW, UP_ARROW, RIGHT_ARROW, DOWN_ARROW]
```
Niloth-p, line 94: For convenient use in lint-hotkeys to check if the key_combination contains any of the arrow keys.

Niloth-p (commented):
Commit Flow:
1. Apply the display functions from keys.py.
2. Append direction key symbols - in UI and docs.
3. Update the scroll prompt in Popup headers.
4. Use Shift in docs (for demo). Created functions and refactored the formatting.

Also, some miscellaneous changes,
- Added a --fix argument in error message.
- Added a docstring to a function in lint-hotkeys, for full coverage.

The commit structure is temporary.
I've just submitted the changes, I'll need to re-order them once we choose the parts we want, and group them.

### Visual Changes

**Help Menu**
im2
**Popup Header**
im3

neiljp (commented):
@Niloth-p Thanks for splitting out the core changes - I just merged the first commit as d65cbee7406a3413946e78fb7fcff8ddc8875251 🎉 

Regarding the remainder, I left inline comments.

For reviewing, it's good to have separate changes in separate commits except for the earliest code demos. For now, let's restructure the remainder to have the capitalization first, symbols last (if left in for others to view), and any adjustment to the linting tool in one or more separate commits.

Leaving the symbols until last, which are more of an unknown, means we can drop the commit at the end without needing to adjust the intervening commits for conflicts.

Having just been able to test it, the use of the common text in the menu titles looks promising as a refactor, but less-so with the symbols with the text. Without the symbols it looks better, but the upper case text from the capitalization then dominates the menu titles? I do think that refactor would be a good improvement, if not using the new key casing yet for this reason. It could then lead onto the popup general key improvements.

zulipterminal/core.py
```diff
@@ -246,11 +255,11 @@ def exit_popup(self) -> None:
         self.loop.widget = self.view
 
     def show_help(self) -> None:
-        help_view = HelpView(self, "Help Menu (up/down scrolls)")
+        help_view = HelpView(self, f"Help Menu {SCROLL_PROMPT}")
         self.show_pop_up(help_view, "area:help")
 
     def show_markdown_help(self) -> None:
-        markdown_view = MarkdownHelpView(self, "Markdown Help Menu (up/down scrolls)")
+        markdown_view = MarkdownHelpView(self, "Markdown Help Menu " + SCROLL_PROMPT)
```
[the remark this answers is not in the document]
Niloth-p: > I've not tested this yet, but you updated the formatting of other lines but not this one?

It's the same formatting, just didn't use f-strings for this one deliberately because if I did, it exceeds the recommended line length. But, consistency is more important than minimizing LOC, I'll turn that back into an f-string.

PR #1483, opened 2024-04-15
Detect terminal text-based-browsers better and block them

### LINKED ISSUE #1475: Opening URLs can freeze terminal (X11 but no GUI browser?)
I've observed this behavior when:
- the `DISPLAY` variable is set (eg. when using `ssh -X`)
- the system fails to use a remote/gui web client

Currently our URL-handling function has an explicit check for `DISPLAY`, and avoids opening the URL if this is the case on Linux.

With `DISPLAY` set, for a terminal in a GUI or by `ssh -X`, we allow the python webbrowser module to find a browser. In my case it finds a *local* browser installed (links), which it attempts to run in the same terminal, which likely interacts badly with the urwid loop.

A possible way to reproduce this would be to run on a remote environment or one with no GUI browsers installed, and either connect using something like `ssh -X` (X forwarding), or explicitly set `DISPLAY` to trick the detection into thinking a GUI is present.

One possible solution is to manually stop and start the urwid loop, as occurs in #1394. However, currently I don't see a straightforward way using `webbrowser` to determine if it is a GUI or text web-browser. That means that a challenge with this approach, if it resolves the issue, is that for users within a GUI, this may leaves the ZT UI unusable until the browser returns/closes. This wasn't an issue for #1394, as the assumption is one is composing separately, and in most cases the editor temporarily replaces the ZT window.

Sushmey:
Opening links in browser for users using text-based browsers results in ZT being frozen indefinitely. This might be because of the loop to open link in browser clashing with urwid loop. 

This PR blocks out text-browsers better than what it used to earlier. It does so by temporarily unsetting TERM environ so text-browsers aren't detected.
This results in an exception which displays an error for text-browsers.

- [x] Discussed in **#zulip-terminal** in `topic` [Opening URLs can freeze terminal #T1475](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Opening.20URLs.20can.20freeze.20terminal.20.23T1475)
- [x] Fully fixes #1475 
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/core/test_core.py
+++ b/tests/core/test_core.py
@@ -427,6 +427,8 @@ class TestController:
     ) -> None:
         # Set DISPLAY environ to be able to run test in CI
         os.environ["DISPLAY"] = ":0"
+        # Set TERM environ to be able to check text-browsers
+        os.environ["TERM"] = "xterm-256color"
         mocked_report_success = mocker.patch(MODULE + ".Controller.report_success")
         mock_get = mocker.patch(MODULE + ".webbrowser.get")
         mock_open = mock_get.return_value.open
@@ -447,6 +449,7 @@ class TestController:
         self, mocker: MockerFixture, controller: Controller
     ) -> None:
         os.environ["DISPLAY"] = ":0"
+        os.environ["TERM"] = "xterm-256color"
         error = "No runnable browser found"
         mocked_report_error = mocker.patch(MODULE + ".Controller.report_error")
         mocker.patch(MODULE + ".webbrowser.get").side_effect = webbrowser.Error(error)
--- a/zulipterminal/core.py
+++ b/zulipterminal/core.py
@@ -410,6 +410,8 @@ class Controller:
         try:
             # Checks for a runnable browser in the system and returns
             # its browser controller, if found, else reports an error
+            term = os.environ.get("TERM")  # Saving the original value
+            del os.environ["TERM"]  # Temporarily deleting variable
             browser_controller = webbrowser.get()
             # Suppress stdout and stderr when opening browser
             with suppress_output():
@@ -424,6 +426,8 @@ class Controller:
                     [f"The link was successfully opened using {browser_name}"]
                 )
         except webbrowser.Error as e:
+            if isinstance(term, str):
+                os.environ["TERM"] = term  # resetting
             # Set a footer text if no runnable browser is located
             self.report_error([f"ERROR: {e}"])
 


zulipterminal/core.py
```diff
@@ -424,6 +426,8 @@ def open_in_browser(self, url: str) -> None:
                     [f"The link was successfully opened using {browser_name}"]
                 )
         except webbrowser.Error as e:
+            if isinstance(term, str):
+                os.environ["TERM"] = term  # resetting
```
neiljp, line 430: This certainly shouldn't be in the exceptional block. As it is now, TERM will not be reset if there is no error. If you want something to happen in every case, you could use `finally`.

The simplest solution would be to reset TERM as soon as the browser is determined. If so, this may be clearer with two try blocks, so you can isolate the 'only-terminal-browsers' error from others.

That is, unless the webbrowser module does other detection of this kind after that point, which could require the environment to be set for later operations?

zulipterminal/core.py
```diff
@@ -410,6 +410,8 @@ def open_in_browser(self, url: str) -> None:
         try:
             # Checks for a runnable browser in the system and returns
             # its browser controller, if found, else reports an error
+            term = os.environ.get("TERM")  # Saving the original value
+            del os.environ["TERM"]  # Temporarily deleting variable
```
neiljp, line 414: This looks like `pop`.

tests/core/test_core.py
```diff
@@ -427,6 +427,8 @@ def test_open_in_browser_success(
     ) -> None:
         # Set DISPLAY environ to be able to run test in CI
         os.environ["DISPLAY"] = ":0"
+        # Set TERM environ to be able to check text-browsers
+        os.environ["TERM"] = "xterm-256color"
```
neiljp, line 431: Your test changes make the tests pass, but don't test the expectations from the new behavior:
- is the environment variable unchanged?
- can we test specifically what happens if there is no 'valid' webbrowser found?

zulipterminal/core.py
Sushmey: Using `finally` is a very good idea! 
Re using 2 try blocks, we don't need that since `TERM` environ isn't used anywhere during or after unsetting it. Resetting it in `finally` seems better it seems like the cleaner approach.

tests/core/test_core.py
Sushmey: I'm a bit confused about writing the tests for this. The `TERM` environ doesn't really change, how do we test its transitory absence?
Also writing a test specifically for text_browser sounds good but again not sure what we can test. We can assert if `TERM` environ doesn't exist inos.environ but that's a value we are setting and not what we are fetching from the function.

Issue #515, opened 2020-02-25
Make help hints context-dependent


neiljp:
@shreyamalviya did some excellent work on this in #348, which I've rebased against master as of now as the latest commit at https://github.com/neiljp/zulip-terminal/tree/rebased-348-for-contextual-help

That should be able to act as a jumping off point to improve the hints.

It also provides some thoughts for improvements as to where other keys could work.

It would be useful to have this work transparently so that adding keys or context allows the hints to just keep working.

preetmishra:
@neiljp I pulled the https://github.com/neiljp/zulip-terminal/tree/rebased-348-for-contextual-help and played with it a bit. I was able to show `stream_list` keys whenever the user presses `q` to search streams.

It is a good jumping-off point though I am not sure in what context should we display keys from `general`, `navigation` and `searching` categories?

neiljp:
@preetmishra I think we could probably add some contexts, like you've indicated, without needing to achieve a complete solution straight away.

However `general` mostly applies everywhere, though ctrl-l is a duplicate (so confusing) and ? will just draw that character in edit mode (search boxes, compose boxes).

It is possible that we will want to separate the help description from the actual contexts to apply it to, and certainly that multiple contexts may apply to particular keys/help being shown.

That said, the help menu categories at the moment are not fixed, and we could expand them to make some things clearer - like `esc` does different things in different circumstances, but we just say 'go back' and assume it makes sense. That would have to balance making the help menu too wordy against their use in particular circumstances as hints being perhaps more descriptive?

preetmishra:
> It is possible that we will want to separate the help description from the actual contexts to apply it to, and certainly that multiple contexts may apply to particular keys/help being shown.

@neiljp I am not sure if I understand the separation bit properly. Could you give an example?

Regarding `esc` and other similar keys, we can append the context we're in to change its help text dynamically. For instance, if we're in a write box, we can say 'Go back from the write box'.

Should I proceed with adding some contexts which seem apparent right now?

zulipbot:
Hello @zulip/server-hotkeys members, this issue was labeled with the "area: hotkeys" label, so you may want to check it out!

PR #1484, opened 2024-04-18
Introduce help context

Niloth-p:

Refactors `set_footer_text()` of core.py
Refactors `lint_hotkeys` heavily, in preparation for contexts.
Introduces `context` to track the current focus.
Uses context to generate relevant footer hints.
Demonstrates a Contextual Help Menu and "parent contexts".

## Commit Grouping

<details>
<summary>Commits grouped by headings, and by another type
Includes references for motivation of some groups. </summary>

#### Footer updates:
- ui: Make the Help hotkey suggestion in the footer more conventional. (1)
- refactor: ui/core/boxes: Separate out function to reset footer text. (2)
- refactor: ui/core: Separate footer events from set_footer_text(). (2)
- ui: Add state variable to track live footer events. (2)

#### Rewrite Lint-hotkeys: general improvements (3)
- lint-hotkeys: Handle the doc file does not exist exception.
- refactor: lint-hotkeys: Remove the write_hotkeys_file() function.
- refactor: lint-hotkeys: Make the integer error flag boolean. (edited commit from #1518)
- refactor: lint-hotkeys: Rewrite function docstrings and comments.
- refactor: lint-hotkeys: Rename get_hotkeys_file_string().
- refactor: lint-hotkeys: Rename ambiguous keyword "action" to "batch".
- refactor: lint-hotkeys: Improve variable naming of generated dict.
- lint-hotkeys: Refactor the flow by creating new ENTRY_BY_CATEGORIES.
- refactor: lint-hotkeys: Delay generation of file string when linting.
- refactor: lint-hotkeys: Split out the help text linting function.
- refactor: lint-hotkeys: Split out the function that lints categories.
- lint-hotkeys: Restructure the error messages of lint_help_text().
- lint-hotkeys: Lint for typos in key_category values. (edited commit from #1518)

#### Refactor Lint-hotkeys: in preparation for linting contexts (3)

These commits share a common line at the bottom of their commit messages - "Part of adding generic variables in preparation for contexts."
Once reviewed, we could squash these commits.
- refactor: lint-hotkeys: Use help_group instead of HELP_CATEGORIES.
- refactor: lint-hotkeys: Create generic variable entries_by_group.
- refactor: lint-hotkeys: Create generic variables for key_category.
- refactor: lint-hotkeys: Replace 'category(ies)' with 'group(s)'.
- refactor: lint-hotkeys: Delete constant OUTPUT_FILE_NAME, compute it.
- refactor: lint-hotkeys: Replace usage of OUTPUT_FILE.

Inculcates the above commits.
- refactor: lint-hotkeys: Generalize the traversal of key group values.
- refactor: lint-hotkeys: Use bundled group values.

#### Adding contexts + linting:
- keys: Add a contexts field to Key Binding. (3)
- lint-hotkeys: Add argument to generate a keys file grouped by contexts. (3)

#### Use contexts in footer hint:
- keys: Use the help contexts in generating random help tips.
- keys: Include previously excluded random help hints.
- ui: Enable footer texts to display contextual help tips. (2)
- ui/keys: Display the context of the footer hint. (1)

#### Track context switching
- contexts/core/ui: Track the currently focused widget in the UI.
#### Contextual Help Menu
- core/ui/views/keys: Add a Contextual Help Menu. (4)
#### Parent Contexts
- keys/views: Add mapping of contexts to broader contexts they belong to. (3)(4)

---

### Another dimension of grouping of commits:
The commits/groups marked with the suffix number in brackets denotes commits related to the following groups:
(1) Footer Hint format
(2) commits related to set_footer_text()
(3) Linting commits
(4) Contextual Help Menu
(5) The contexts feature - all commits from the adding contexts part. The commits that do not depend on the contexts are obviously placed at the front, so all the commits after contexts have been introduced are all related to the contexts feature.

#### References: (corresponding to the numbered groups)
- (1) See CZO topic: `Improved UI design of footer hint`
- (2) [Motivation for refactoring set_footer_text](https://github.com/zulip/zulip-terminal/pull/1484#discussion_r1580184836)
- (3) [Motivation for lint-hotkeys refactoring commits](https://github.com/zulip/zulip-terminal/pull/1484#discussion_r1581245134), These commits also swallow the commits from #1518

</details>

## Outstanding Aspects:

<details>
<summary>Miscellaneous aspects</summary>

Feedback needed on all the naming changes. There are a lot.

I've rebased several times to get the order of commits just right. Hope they are sensible. Can freely be squashed after reviewing.

The contexts.py is added in a separate file, although we could put it inside another file.
The logic flow is not very smooth (several if-else blocks), due to the differences in cases. Any suggestions on making it smoother would be helpful.

Tests have not been included for the context tracking.
Guidance required on framing the tests and the setup.

We primarily need to:
- test the mapping of focus paths to widgets
- simulate focus on a widget, test context update

But, I'm having trouble setting up the scenarios.

The Contextual Help Menu is merely added to test the contexts feature. It needs a lot more work before it's ready for a merge.
- To view only the hints specific to a context, run ZT after setting HEAD to the penultimate commit
- To view all working hints for a specific context, run ZT with the entire PR.

</details>

---

<details>
<summary>FAQ and Assumptions</summary>

### FAQ:

1. Why bother this much with updating lint-hotkeys?

There was a feature in discussion - Rebinding/Remapping hotkeys, which require strong linting before users are allowed to customize their hotkeys.
Thus, refactoring lint-hotkeys is a necessity.

2. Why these many commits?

Since it's just a linting file, I was hoping we could get by with just the functionality, and not worry about the readability of the commits, as per [my question here](https://github.com/zulip/zulip-terminal/pull/1484#discussion_r1578483531). But, it turns out linting is quite important for later features, so we do need this to be readable.
And since all the changes are done to the same file, we need to do them step by step to ensure readability, unlike changes made to different files which can easily be squashed (eg: see the Parent Contexts commit (the last one) that introduces parent contexts, adds linting and contextual help menu in the same step).
Any ideas for squashing while retaining readability are quite welcome. Or if I am holding high standards for what I consider readable, criticism is also welcome.

3. Why all commits in the same PR?

The earlier commits are motivated by the requirements of the later commits, and their implementation strongly depends (in the case of lint-hotkeys refactoring) on the implementation of the later commits.
The later commits can only be built on top of the earlier commits, so even if there were separate PRs, they'd need to be stacked, and merged in order.

---

### Assumptions:

- We want to generate the contexts.md file just for the devs and not the users.
Several commits for lint-hotkeys operate under this assumption.
Was unable to add an edited .gitignore with the contexts.md entry.

- We want to start off with key_contexts instead of key_context.
Why not introduce this later? Because there are some obvious discrepancies, e.g., the toggle topics hint would only be able to mapped to either the stream list or the topic list.
See the context values for more such cases.

</details>

---

<details>
<summary>Follow-ups</summary>

### Follow-ups:

- [discuss] We currently allow the 'Tab' keypress to override any footer events. This can be controlled using the newly added variable that tracks live footer events.
- [discuss] Allow opening the Help Menu from even within popups and editors.
- [discuss] The design/formatting of the Contextual Help Menu
- [discuss] Do we want to rename the fields `key_category`/`key_context` in the key bindings?
- [discuss] I think that the testing function `test_commands_for_random_tips()` is quite overloaded. It is non-trivial and tests for several different things with very tight intersectional test cases:
	- the working of "excluded_from_random_tips"
	- usage of contexts
	- single vs multiple contexts
	- absence of context field
	- no tips in given context
	- the returned context value
- Condition checks for vaildity of specific commands

</details>

- [x] Discussed in **#zulip-terminal** in `Context Based Help Tips` and `Adding context to hotkeys`
- [ ] Fully fixes #
- [x] Partially fixes issue #515
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tools/lint-hotkeys
+++ b/tools/lint-hotkeys
@@ -91,7 +91,8 @@ def generate_hotkeys_file() -> None:
     """
     hotkeys_file_string = get_hotkeys_file_string()
     output_file_matches_string(hotkeys_file_string)
-    write_hotkeys_file(hotkeys_file_string)
+    with open(OUTPUT_FILE, "w") as hotkeys_file:
+        hotkeys_file.write(hotkeys_file_string)
     print(f"Hot Keys list saved in {OUTPUT_FILE}")
 
 
@@ -150,14 +151,6 @@ def read_help_categories() -> Dict[str, List[Tuple[str, List[str]]]]:
             (item["help_text"], display_keys_for_command(cmd))
         )
     return categories
-
-
-def write_hotkeys_file(hotkeys_file_string: str) -> None:
-    """
-    Write hotkeys_file_string variable once to OUTPUT_FILE
-    """
-    with open(OUTPUT_FILE, "w") as hotkeys_file:
-        hotkeys_file.write(hotkeys_file_string)
 
 
 if __name__ == "__main__":
--- a/tools/lint-hotkeys
+++ b/tools/lint-hotkeys
@@ -18,7 +18,7 @@ KEYS_FILE = (
     Path(__file__).resolve().parent.parent / "zulipterminal" / "config" / "keys.py"
 )
 KEYS_FILE_NAME = KEYS_FILE.name
-OUTPUT_FILE = Path(__file__).resolve().parent.parent / "docs" / "hotkeys.md"
+CATEGORY_OUTPUT_FILE = Path(__file__).resolve().parent.parent / "docs" / "hotkeys.md"
 SCRIPT_NAME = PurePath(__file__).name
 HELP_TEXT_STYLE = re.compile(r"^[a-zA-Z /()',&@#:_-]*$")
 
@@ -58,7 +58,7 @@ def lint_hotkeys_file() -> None:
     """
     Lint KEYS_FILE for valid key descriptions (help texts) and key categories,
     check for duplicate key combinations,
-    and compare with existing OUTPUT_FILE.
+    and compare with existing output file.
     """
     error_flag = False
 
@@ -71,14 +71,15 @@ def lint_hotkeys_file() -> None:
         )
     else:
         print("No hotkeys linting errors")
-        if not OUTPUT_FILE.exists():
+        output_file = CATEGORY_OUTPUT_FILE
+        if not output_file.exists():
             raise SystemExit(
-                f"Run './tools/{SCRIPT_NAME} --fix' to generate {OUTPUT_FILE.name} file"
+                f"Run './tools/{SCRIPT_NAME} --fix' to generate {output_file.name} file"
             )
         hotkeys_file_string = generate_hotkeys_file_string()
         if not output_file_matches_string(hotkeys_file_string):
             raise SystemExit(
-                f"Run './tools/{SCRIPT_NAME} --fix' to update {OUTPUT_FILE.name} file"
+                f"Run './tools/{SCRIPT_NAME} --fix' to update {output_file.name} file"
             )
 
 
@@ -152,19 +153,20 @@ def lint_help_groups() -> bool:
 
 def generate_hotkeys_file() -> None:
     """
-    Generate OUTPUT_FILE based on help text description and
+    Generate output file based on help text description and
     shortcut key combinations in KEYS_FILE, grouped by categories
     """
+    output_file = CATEGORY_OUTPUT_FILE
     hotkeys_file_string = generate_hotkeys_file_string()
     output_file_matches_string(hotkeys_file_string)
-    with open(OUTPUT_FILE, "w") as hotkeys_file:
+    with open(output_file, "w") as hotkeys_file:
         hotkeys_file.write(hotkeys_file_string)
-    print(f"Hot Keys list saved in {OUTPUT_FILE}")
+    print(f"Hot Keys list saved in {output_file.name}")
 
 
 def generate_hotkeys_file_string() -> str:
     """
-    Construct string in form for output to OUTPUT_FILE based on help text
+    Construct string in form for output to output file based on help text
     description and shortcut key combinations in KEYS_FILE
     """
     help_group = HELP_CATEGORIES
@@ -194,12 +196,13 @@ def generate_hotkeys_file_string() -> str:
 
 def output_file_matches_string(hotkeys_file_string: str) -> bool:
     """
-    Check if the OUTPUT_FILE exists and matches the generated hotkeys_file_string
+    Check if the output file exists and matches the generated hotkeys_file_string
     """
-    output_file_name = OUTPUT_FILE.name
+    output_file = CATEGORY_OUTPUT_FILE
+    output_file_name = output_file.name
     try:
-        with open(OUTPUT_FILE) as output_file:
-            content_is_identical = hotkeys_file_string == output_file.read()
+        with open(output_file) as existing_file:
+            content_is_identical = hotkeys_file_string == existing_file.read()
         if content_is_identical:
             print(
                 f"{output_file_name} file already in sync with config/{KEYS_FILE_NAME}"
@@ -221,7 +224,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--fix",
         action="store_true",
-        help=f"Generate {OUTPUT_FILE.name} file by extracting key description and key "
+        help=f"Generate {CATEGORY_OUTPUT_FILE.name} file by extracting key description and key "
         f"combination from config/{KEYS_FILE_NAME} file",
     )
     args = parser.parse_args()
--- a/tools/lint-hotkeys
+++ b/tools/lint-hotkeys
@@ -3,7 +3,7 @@ import argparse
 import re
 from collections import defaultdict
 from pathlib import Path, PurePath
-from typing import Dict, List, Tuple
+from typing import Dict, List, Tuple, Union
 
 from typing_extensions import Literal
 
@@ -38,9 +38,11 @@ def read_help_groups() -> Dict[str, List[Tuple[str, List[str]]]]:
     key_group: KeyGroup = "key_category"
     entries_by_group = defaultdict(list)
     for cmd, item in KEY_BINDINGS.items():
-        entries_by_group[item[key_group]].append(
-            (item["help_text"], display_keys_for_command(cmd))
-        )
+        groups = [item[key_group]]
+        for group in groups:
+            entries_by_group[group].append(
+                (item["help_text"], display_keys_for_command(cmd))
+            )
     return entries_by_group
 
 
@@ -138,10 +140,13 @@ def lint_help_groups() -> bool:
     # Lint for typos in key categories
     error_message = ""
     for key, binding in KEY_BINDINGS.items():
-        group_value = binding[key_group]
-        if group_value not in help_group:
-            error_message += f"  Invalid {group} '{group_value}' for key '{key}'.\n"
-            error_flag = True
+        group_values: Union[str, List[str]] = binding[key_group]
+        if isinstance(group_values, str):
+            group_values = [group_values]
+        for group_value in group_values:
+            if group_value not in help_group:
+                error_message += f"  Invalid {group} '{group_value}' for key '{key}'.\n"
+                error_flag = True
     if error_message:
         print(
             f"Choose a valid {group} value from:\n{', '.join(help_group.keys())}\n"
--- a/tools/lint-hotkeys
+++ b/tools/lint-hotkeys
@@ -151,12 +151,12 @@ def read_help_categories() -> Dict[str, List[Tuple[str, List[str]]]]:
     key: help category
     value: a list of help texts and key combinations, for each key binding in that help category
     """
-    categories = defaultdict(list)
+    entries_by_category = defaultdict(list)
     for cmd, item in KEY_BINDINGS.items():
-        categories[item["key_category"]].append(
+        entries_by_category[item["key_category"]].append(
             (item["help_text"], display_keys_for_command(cmd))
         )
-    return categories
+    return entries_by_category
 
 
 if __name__ == "__main__":
--- a/tools/lint-hotkeys
+++ b/tools/lint-hotkeys
@@ -82,15 +82,20 @@ def lint_help_text() -> bool:
     Lint each keybinding's help text / description for invalid characters
     """
     error_flag = False
+    error_message = ""
     for keybinding in KEY_BINDINGS.values():
         help_text = keybinding["help_text"]
         if not re.match(HELP_TEXT_STYLE, help_text):
             key_combinations = " / ".join(keybinding["keys"])
-            print(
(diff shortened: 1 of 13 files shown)


zulipterminal/config/keys.py
```diff
@@ -220,7 +220,7 @@ class KeyBinding(TypedDict):
         'key_context': 'stream_view',
     },
     'ALL_MESSAGES': {
-        'keys': ['a', 'esc'],
+        'keys': ['a'],
```
Niloth-p, line 223: This is a temporary change. To make the linting tests pass.
As both the `Esc` key mappings clash in the 'general' context.

I could add an exception instead too, if necessary later.
Or I could move the `ALL_MESSAGES` command to the "message_view" context instead until that bug is resolved.

tools/lint-hotkeys
Niloth-p, line 1: I know the changes to this file would be pretty difficult to read and verify.
Do I need to break it down into layers? 
Tested manually for the following cases:

[x] no hotkeys-by-context.md file
[x] mismatching existing hotkeys-by-context.md file
[x] matching existing hotkeys-by-context.md file

[x] no hotkeys.md file
[x] mismatching existing hotkeys.md file
[x] matching existing hotkeys.md file

[x] no arguments
[x] context argument
[x] fix argument
[x] both arguments

Niloth-p (commented):
There's a bunch of unfinished work here, and some significant changes to be made. I've mentioned the main ones in the CZO discussion.
I wanted to first ensure that I'm going in the right direction, so I've opened it up for review already.

neiljp (commented):
@Niloth-p There's a lot here!

I did give this a manual test, and many parts definitely seem to be working. I mainly moved around areas of the UI, and since some hotkeys (including just `?`) effectively show a new random hint, it allows some degree of sampling of them.

I've not looked at the hotkeys assignment doc, since `keys.py` works fairly well in the first instance, and I'd started reading that before you added the feature :)

I've only left inline comments until prior to the specific context setting. That's partly since I'm wondering if there may be an alternative approach rather than setting it explicitly in many places. For example, you've already set the context depending on a focus path in a few places instead, and for popups this could be a class variable that is accessed by generic popup show code?

One helpful change would be to address the 'cycling' issue sooner rather than later, but even before or part of that we can improve the design slightly. The original design was to indicate that `?` is general help, and show a random help next to it, but currently we have the help key right up next to `Help` and in parentheses rather than square brackets (which we now tend to use). There's also a colon rather than an indication that the general help is distinct from the hint. As an extension specific to this change, it'd be great to show the relevance of the hint - does it work in compose, stream list, selected message, etc. I could tell that it was displaying reasonable hints based on the context, but showing the context itself would be helpful for those less familiar :)

neiljp:
I know you have lots of ideas and are getting code out well, but I'd suggest being cautious regarding pushing too far in one direction. It can be better to have work in another area to work on in parallel, rather than needing to heavily prune/drop/adjust a single direction if work in that one area takes a different path.

zulipterminal/config/keys.py
```diff
@@ -34,376 +35,449 @@ class KeyBinding(TypedDict):
...
         'keys': ['q'],
         'help_text': 'Search Streams',
         'key_category': 'searching',
+        'key_context': 'general',
     },
     'SEARCH_TOPICS': {
         'keys': ['q'],
         'help_text': 'Search topics in a stream',
         'key_category': 'searching',
+        'key_context': 'topic_view',
```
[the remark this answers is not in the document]
Niloth-p: Yes, they both would actually belong in a general context, but need to apply a check (based on the state) to determine which one is applicable at the moment.

zulipterminal/config/keys.py
```diff
@@ -34,376 +35,449 @@ class KeyBinding(TypedDict):
...
         'keys': ['end', 'G'],
         'help_text': 'Go to bottom / Last message',
         'key_category': 'navigation',
+        'key_context': 'general',
     },
     'REPLY_MESSAGE': {
         'keys': ['r', 'enter'],
         'help_text': 'Reply to the current message',
         'key_category': 'msg_actions',
+        'key_context': 'message_view',
```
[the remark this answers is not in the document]
Niloth-p: True.
I hadn't considered just not using any suffix, was instead trying to choose a good suffix.
Good idea!

Yes, it does.

tools/lint-hotkeys
neiljp: It would definitely help to split this up, since this is multiple aspects:
- linting in any new ways, or relaxing linting as we move forward
- adding the new doc generation, if we end up going in that direction

However, as you say it would also be easier to verify, such as if you eg. refactor existing names first.

tools/lint-hotkeys
```diff
@@ -26,6 +26,23 @@ HELP_TEXT_STYLE = re.compile(r"^[a-zA-Z /()',&@#:_-]*$")
 KEYS_TO_EXCLUDE = ["q", "e", "m", "r", "Esc"]
 
 
+def read_help_categories() -> Dict[str, List[Tuple[str, List[str]]]]:
```
rsashank: NIT: I feel we can squash this and the previous commit. The previous commit seems redundant since you're refactoring the flow and removing the function anyways.

refactor: lint-hotkeys: Improve variable naming of generated dict.
lint-hotkeys: Refactor the flow by creating new ENTRY_BY_CATEGORIES.

tools/lint-hotkeys
```diff
@@ -93,6 +94,7 @@ def lint_hotkeys_file() -> None:
     error_flag |= lint_help_groups("category")
     error_flag |= lint_help_groups("context")
     error_flag |= lint_help_text()
+    error_flag |= lint_parent_contexts()
```
rsashank, line 97: Maybe
```python
error_flag |= (
    lint_help_groups("category") |
    lint_help_groups("context") |
    lint_help_text() |
    lint_parent_contexts()
)
```

rsashank (commented):
Wow, great PR @Niloth-p! It adds a much-needed functionality and is definitely the holy grail of your GSoC proposal. It’s fantastic to see it in action!

I left some comments, mostly minor changes and NITs. Functionally, it LGTM—I played around with it a bit and found that it recommends Esc to clear the search bar when you're in the users list, even if there's no search active. There might be a few outliers like that. Is it feasible to fix those on an individual scale? (That's the only one I’ve found so far, but I’ll let you know if I come across any others.)

tools/lint-hotkeys
Niloth-p: Ah, interesting, but we're not actually removing the function, it's just moved to the top to be placed before creating the new ENTRY_BY_CATEGORIES.

tools/lint-hotkeys
```diff
@@ -120,13 +120,19 @@ def get_hotkeys_file_string() -> str:
...
+        if content_is_identical:
+            print(
+                f"{OUTPUT_FILE_NAME} file already in sync with config/{KEYS_FILE_NAME}"
+            )
+            return True
+        else:
+            print(f"{OUTPUT_FILE_NAME} file not in sync with config/{KEYS_FILE_NAME}")
+            return False
+    except FileNotFoundError:
+        print(f"{OUTPUT_FILE_NAME} does not exist")
```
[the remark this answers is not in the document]
Niloth-p: Good spot again. Thanks for looking into the user-visible verbiage. Handled it separately now.

tools/lint-hotkeys
```diff
@@ -67,16 +67,16 @@ def lint_hotkeys_file() -> None:
             print(
                 f"Duplicate key combination for keys {duplicate_keys} for category ({HELP_CATEGORIES[action]}) detected"
             )
-            error_flag = 1
-    if error_flag == 1:
+            error_flag = True
+    if error_flag:
         print(f"Rerun this command after resolving errors in config/{KEYS_FILE_NAME}")
```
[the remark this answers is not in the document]
Niloth-p: Great suggestion! Done.

tools/lint-hotkeys
Niloth-p: Ah, yes. Done.

Niloth-p (commented):
Thank you for the testing it out and I appreciate the great suggestions, @rsashank!

> recommends Esc to clear the search bar when you're in the users list, even if there's no search active

Yes, good spot. Those need some extra logic to handle. There are at least 3 key bindings of that sort that need condition checks, and for all of them, I'd attached a comment `# TODO: condition check required` at the bottom of their key bindings, until we come up with a smooth solution for handling them.

But, what I failed to consider was that since the comment is just for the devs, from the user POV, the hints would just seem bad hints.
So, I've currently turned off those key bindings from being used in random hints until the condition checks are added. Hope that's a good way of handling it.

Please do share any other observations as you come across them :)

#### Updates to the code:
- Addressed Sashank's feedback, by adding the changes to the earlier commits and updating them gradually in sync with the changes in the later commits.
- Rebased onto main, and added context field for the new OPEN_EXTERNAL_EDITOR command.

zulipterminal/ui.py
```diff
@@ -130,15 +125,21 @@ def set_footer_text(
...
 
+    def set_footer_text_for_event(self, text: List[Any], style: str = "footer") -> None:
+        self.set_footer_text(text, style)
+
+    def set_footer_text_for_event_duration(
+        self, text: List[Any], duration: float, style: str = "footer"
+    ) -> None:
+        self.set_footer_text_for_event(text, style)
+        time.sleep(duration)
+        self.reset_footer_text()
```
neiljp, line 149: This function is not asynch, so blocks for that duration. This breaks existing behavior. For example, try editing an old message that you cannot - the UI just stops updating, before catching up.

We shouldn't make everything asynch, as that puts everything in a thread and makes it a lot more complex, which was one benefit to the existing approach. However, this needs resolving.

zulipterminal/ui.py
```diff
@@ -114,7 +114,11 @@ def get_random_help(self) -> List[Any]:
         return [
             "Help[?] ",
             ("footer_contrast", f" {random_command_display_keys} "),
-            f" {random_command['help_text']}",
+            f" ({tip_context}) ",
+            (
+                "footer_contrast",
+                f" {random_command['help_text']} ",
+            ),
```
neiljp, line 121: This shows the information quite well at this stage :+1:

Later, I can see we'll want to iterate on the layout - there's a similar issue as with the `Help [?]` point, but also that the context/keys/descriptions 'bounce around' spacing-wise.

neiljp (commented):
@Niloth-p I'm not sure if this was 34 commits before, but it certainly makes it challenging to review as it is - it can be intimidating just to see there are 34, at least ;) The commits aren't difficult to read in most cases, which helps :+1:

I started focusing on the early commits, since they have potential for simple refactoring too (and that bug?). However, as per your current PR intro you then have over 10 commits for tidying the linter, presumably from the other previous PR, before then extending that to work to prepare for the contextual help.

So, I've focused this review on those earlier commits, then mostly towards the end, since the early part may be merged sooner and the latter might be simplified. Getting some commits merged will make this easier to handle!

The approach you're using for setting context does seem to work, but I see I mentioned the class variables idea before. What do you think?

PR #1488, opened 2024-04-23
github: Add simple issue templates & configuration

neiljp:

This converts the basic issue filing system to use the GitHub template chooser.

The main feature at this point is to focus reporters on the scope of the issue:
- major bug (bug: crash)
- bug (bug)
- missing Zulip feature (missing feature)
- general feature/improvement (enhancement)

These should apply a corresponding default label to the issue, with the idea that this would provide an initial structure to fine-tune the labels, eg. a specific `parity` label to add to `missing feature`, which could also gain one of the `user` or `admin` sub-labels.

This already feels like an improvement over the current single blank 'template', though it may take time to see how it works in practice.

This takes some ideas from zulip/zulip#26000 and followups.

### Outstanding aspect(s)    

- How do these categories look like, if you were filing an issue?

- [x] Discussed in **#zulip-terminal** in `Issue types #T1488`
- [ ] Fully fixes #
- [x] Partially fixes issue #830 
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/.github/ISSUE_TEMPLATE/01_crash.md
+++ b/.github/ISSUE_TEMPLATE/01_crash.md
@@ -1,0 +1,39 @@
+---
+name: Bug report (major)
+about: A major bug which significantly impacts usability, eg. an unexpected quit, crash, freeze.
+labels: 'bug: crash',
+---
+
+### Bug description
+
+
+
+<!-- If discussed in #zulip-terminal or another channel on chat.zulip.org, paste link below: -->
+
+
+### How is the bug triggered?
+How can you reproduce the bug?
+1.
+
+
+### Does it produce a 'traceback' or 'exception'?
+<!-- Copy/paste it between the ``` lines below: -->
+```
+
+
+```
+
+### How are you running the application?
+Please include as many of the following as possible:
+- **Zulip-terminal version:**
+  eg. a specific version (0.7.0), or if running from `main` also ideally the git ref
+- **Zulip server version(s):**
+  eg. Zulip Cloud, the version you are running self-hosted, or the Zulip Community server (chat.zulip.org)
+- **Operating system (and version):**
+  eg. Debian Linux, Ubuntu Linux, macOS, WSL in Windows, Docker
+- **Python version (and implementation):**
+  eg. 3.8, 3.9, 3.10, ... (implementation is likely to be eg. CPython, or PyPy)
+
+If possible, please provide details from the `About` menu: (hotkey: Meta + ?)
+(this can provide some of the details above)
+
--- a/.github/ISSUE_TEMPLATE/02_bug.md
+++ b/.github/ISSUE_TEMPLATE/02_bug.md
@@ -1,0 +1,36 @@
+---
+name: Bug report
+about: A concrete bug report with steps to reproduce the behavior.
+labels: bug,
+---
+
+### Bug description
+
+
+
+<!-- If discussed in #zulip-terminal or another channel on chat.zulip.org, paste link below: -->
+
+
+### How is the bug triggered?
+How can you reproduce the bug?
+1.
+
+
+### What did you expect to happen?
+
+
+
+### How are you running the application?
+Please include as many of the following as possible:
+- **Zulip-terminal version:**
+  eg. a specific version (0.7.0), or if running from `main` also ideally the git ref
+- **Zulip server version(s):**
+  eg. Zulip Cloud, the version you are running self-hosted, or the Zulip Community server (chat.zulip.org)
+- **Operating system (and version):**
+  eg. Debian Linux, Ubuntu Linux, macOS, WSL in Windows, Docker
+- **Python version (and implementation):**
+  eg. 3.8, 3.9, 3.10, ... (implementation is likely to be eg. CPython, or PyPy)
+
+If possible, please provide details from the `About` menu: (hotkey: Meta + ?)
+(this can provide some of the details above)
+
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Report broader Zulip issues (beyond Zulip-Terminal)
+    about: Includes links to best-practices for filing bug reports, feature requests, security vulnerabilities and more.
+    url: https://github.com/zulip/zulip/issues/new/choose
+  - name: Not sure?
+    about: 'Chat with us, using any of the Zulip clients, in #zulip-terminal on the Developer Community server (chat.zulip.org).'
+    url: https://zulip.com/development-community
--- a/.github/ISSUE_TEMPLATE/03_parity.md
+++ b/.github/ISSUE_TEMPLATE/03_parity.md
@@ -1,0 +1,25 @@
+---
+name: Feature request (a missing known Zulip feature)
+about: A request for a feature which is missing, but present in another Zulip client (eg. Web/Desktop/mobile).
+labels: 'missing feature',
+---
+
+### Description of feature missing from another Zulip client
+<!-- Is this an entire missing feature? Or part of the feature? -->
+
+
+
+<!-- If discussed in #zulip-terminal or another channel on chat.zulip.org, paste link below: -->
+
+
+### When was this feature first available in Zulip?
+If you know:
+- **Zulip version:**
+  (eg. 2.1, 5.0, 8.0, ..., or 'Zulip Cloud today')
+- **Zulip feature level:**
+  (see https://zulip.com/api/changelog)
+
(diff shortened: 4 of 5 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Report broader Zulip issues (beyond Zulip-Terminal)
+    about: Includes best-practices for filing bug reports, feature requests, security vulnerabilities and more.
```
Niloth-p, line 4: nit: Would we want to use something other than just "best-practices"?
The linked page does contain links to best practices.
But, this description may make it sound like the best practices are available 1 level deeper instead of 2 levels deeper.
Which may not be obvious since no link in that issues page contains the explicit mention of "best-practices".

.github/ISSUE_TEMPLATE/02_bug.md
```diff
@@ -0,0 +1,30 @@
...
+### How are you running the application?
+Please include at least:
+- **Zulip-terminal version:**
+  eg. a specific version (0.7.0), or if running from `main` also ideally the git ref
+- **Zulip server version:**
+  eg. Zulip Cloud, or the version you are running self-hosted
+- **Operating system (and version):**
+  eg. Debian Linux, Ubuntu Linux, macOS, WSL in Windows
+
+If possible, please provide details from the `About` menu:
```
Niloth-p, line 29: Maybe we can also add the shortcut (Meta + ?) for the About Menu?

.github/ISSUE_TEMPLATE/02_bug.md
```diff
@@ -0,0 +1,30 @@
...
+### How is the bug triggered?
+How can you reproduce the bug?
+1.
+
+
+### What did you expect to happen?
+
+
+
+### How are you running the application?
```
Niloth-p, line 20: Would we also want to ask for the Python version and implementation?

.github/ISSUE_TEMPLATE/02_bug.md
```diff
@@ -0,0 +1,30 @@
...
+Please include at least:
+- **Zulip-terminal version:**
+  eg. a specific version (0.7.0), or if running from `main` also ideally the git ref
+- **Zulip server version:**
```
Niloth-p, line 24: I noticed that you chose to drop the checkbox list for Zulip server versions.
I'm not quite acquainted with the server versions yet.
But, would the same checkbox not apply for ZT?

Niloth-p (commented):
Looks pretty neat!
I was thinking that we needed issue templates too.
This will make a great improvement.

May I know why we choose to make no mention of CZO?
a) I understand we wouldn't want to lose bug reports getting submerged in CZO and that users don't need to have a CZO account, but can't we mention CZO ZT stream as an option if they're not sure about how to reproduce their bug?
b) I noticed zulip/zulip seems to have a separate template for issues discussed in CZO. But, yeah, there's no associated label and it only contains a hint/reminder to link a CZO message instead of a topic and to link back the created issue, but that could be useful too?

nit: Regarding the categories,
sometimes and for some users the border between a major and a minor bug may not be very distinct (despite the description).
But, I guess that margin of error would still be fine.

.github/ISSUE_TEMPLATE/config.yml
neiljp: Sure, how about:
```suggestion
    about: Includes guidance for filing bug reports, feature requests, security vulnerabilities and more.
```

.github/ISSUE_TEMPLATE/02_bug.md
neiljp: Yeah, sounds reasonable - though we'll need to update this once we know how we're going to express `Meta` consistently.
```suggestion
If possible, please provide details from the `About` menu (<kbd>Meta</kbd>+<kbd>?</kbd>):
```

.github/ISSUE_TEMPLATE/02_bug.md
neiljp: See the 'parity' file, which links to the API changelog for some server versions.

I will add chat.zulip.org as one of the Zulip server suggested version options.

neiljp: We could add each server version, but we've historically aimed to support from 2.1 onwards. That doesn't mean we support everything that web has, but we've not dropped support for older server versions. For that reason the list of server versions to select from wouldn't be 8/7/6/5(orless), but more like 8/7/6/5/4/3/2.1. That's only two more options, but it might make the list a little long, and even longer with new releases ;)

.github/ISSUE_TEMPLATE/02_bug.md
neiljp: Some of these are listed in the About menu, so it's difficult to know whether to include these in the top part or beneath that.

The reason I included these separately is
- in case users cannot access the about menu
- if we later add a 'copy' option to the about menu for this kind of reporting :)

The python details could probably go here, but may want to be freeform - maybe multiple versions have been tested?

.github/ISSUE_TEMPLATE/config.yml
Niloth-p: I meant something like "Includes links to best-practices".

.github/ISSUE_TEMPLATE/02_bug.md
Niloth-p: oh yeah, we could mention that these details can all be found in the About menu, at the starting of that section. Since one can't assume that, if they do not know.
Nice ideas! I didn't account for these possibilities.

neiljp:
@Niloth-p Does the fixup commit address your suggestions?

.github/ISSUE_TEMPLATE/02_bug.md
```diff
@@ -18,13 +21,16 @@ How can you reproduce the bug?
...
+- **Zulip server version(s):**
+  eg. Zulip Cloud, the version you are running self-hosted, or the Zulip Community server (chat.zulip.org)
 - **Operating system (and version):**
-  eg. Debian Linux, Ubuntu Linux, macOS, WSL in Windows
+  eg. Debian Linux, Ubuntu Linux, macOS, WSL in Windows, Docker
+- **Python version (and implementation):**
+  eg. 3.8, 3.9, 3.10, ... (implementation is likely to be eg. CPython, or PyPy)
 
-If possible, please provide details from the `About` menu:
+If possible, please provide details from the `About` menu: (Meta + ?)
```
Niloth-p, line 34: We might want to mention or clarify that the part in the brackets is the hotkey to access it.
It may not be obvious, as 'meta' and '?' are not just keys.

.github/ISSUE_TEMPLATE/01_crash.md
```diff
@@ -0,0 +1,39 @@
+---
+name: Bug report (major)
+about: A major bug which significantly impacts usability, eg. an unexpected quit, crash, freeze.
+labels: bug: crash,
```
Niloth-p, line 4: There might be a risk of a YAML syntax issue due to the colon? Maybe "bug: crash" instead?

Niloth-p (commented):
That's a lot more improvements than I was able to imagine :smiley:

.github/ISSUE_TEMPLATE/01_crash.md
```diff
@@ -8,6 +8,9 @@ labels: bug: crash,
 
 
 
+<!-- If discussed in #zulip-terminal or another channel on chat.zulip.org, paste link below: -->
```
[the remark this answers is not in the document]
neiljp: I did consider this, but I didn't want to get too technical for issue filing - ie. a link is a good starting point if not familiar with Zulip.

For ZT in particular, we could even simplify it back to something like 'name of topic', since we don't have a 'copy link to message' (or topic, or stream) as yet.

I'm inclined to leave this as a followup for now.

.github/ISSUE_TEMPLATE/01_crash.md
neiljp: Yeah, good catch. The GitHub parser may work around this, but yamllint does suggest that quotes are helpful, including for the name field that contains a `#` :)

.github/ISSUE_TEMPLATE/02_bug.md
neiljp: I dropped the `<kbd>` tags from the text as discussed previously, since they are only visible when previewed/submitted, so are a bit messy when filling out the original box.

Just added a `hotkey: ` prefix before the keys.

neiljp:
Pushed a few more minor changes. Will look at this again later, but I'd like to squash and merge this to get some real-life testing going, so we can iterate further if necessary.

PR #1491, opened 2024-04-27
Enhance the behaviour of `Esc` in the Emoji-picker popup

Niloth-p:

Previously, pressing Esc anytime when the emoji picker is in use will exit the popup.

Now, different behaviours are supported depending on the state.

**1. Escaping from a non-empty searchbox clears it** (This is useful since some other editors support using the `Ctrl l` shortcut to clear the editor, but there's none for the emoji-picker popup).
**2. Escaping from an empty searchbox closes the popup.**

**3. Escaping from the emoji list exits you out of the popup, if the selected emoji button was toggled after the last search** (Allow users to exit the popup as soon as they have made some changes (either newly selected or newly deselected)).
**4. Otherwise, it takes you back to the searchbox and retains the last search text.** (the user has not made changes or is still navigating in the emoji picker, so allow them to return to the search box without exiting the popup)

- [x] Discussed in **#zulip-terminal** in [`Enhance Escape key behaviour in Emoji-Picker popup`](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Enhance.20.60Escape.60.20key.20behaviour.20in.20Emoji-Picker.20popup.20.23T1491/near/1789095)
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/zulipterminal/ui_tools/views.py
+++ b/zulipterminal/ui_tools/views.py
@@ -2124,11 +2124,25 @@ class EmojiPickerView(PopUpView):
             self.emoji_search.set_caption(" ")
             self.controller.enter_editor_mode_with(self.emoji_search)
             return key
-        elif is_command_key("GO_BACK", key) or is_command_key("ADD_REACTION", key):
+        elif is_command_key("ADD_REACTION", key):
             for emoji_code, emoji_name in self.selected_emojis.items():
                 self.controller.model.toggle_message_reaction(self.message, emoji_name)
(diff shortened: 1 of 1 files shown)


neiljp:
#1492 reduced the need for point 1, but otherwise this is still under discussion in the stream.

I suspect we won't take this particular approach, but it is useful for discussion.

PR #1492, opened 2024-05-04
boxes/views: Adopt urwid_readline for all editors.

Niloth-p:
### What does this PR do, and why?
Adopts urwid_readline.ReadlineEdit for all the editors in Zullip Terminal.
The panel search boxes are the only remaining editors not using urwid_readline.ReadlineEdit.

### External discussion & connections
- [x] Discussed in **#zulip-terminal** in [`Readline shortcuts in search boxes`](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Readline.20shortcuts.20in.20search.3F)
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Builds upon previous unmerged work in PR #
- [ ] Is a follow-up to work in PR #
- [ ] Requires merge of PR #
- [x] Merge will enable work on #1484 

### How did you test this?
- [x] Manually - Behavioral changes
- [ ] Manually - Visual changes
- [ ] Adapting existing automated tests
- [ ] Adding automated tests for new behavior (or missing tests)
- [ ] Existing automated tests should already cover this (*only a refactor of tested code*)

### Self-review checklist for each commit
- [x] It is a [minimal coherent idea](https://github.com/zulip/zulip-terminal#structuring-commits---speeding-up-reviews-merging--development)
- [x] It has a commit summary following the [documented style](https://github.com/zulip/zulip-terminal#structuring-commits---speeding-up-reviews-merging--development) (title & body)
- [x] It has a commit summary describing the  motivation and reasoning for the change
- [x] It individually passes linting and tests
- [ ] It contains test additions for any new behavior
- [ ] It flows clearly from a previous branch commit, and/or prepares for the next commit

PR #1494, opened 2024-05-09
Improve the handling of editor actions as a help category

Niloth-p:
Improves the usage of urwid_readline hotkey commands.

Now that all editors support readline shortcuts after the merge of #1492,
all the readline shortcuts have been grouped into their own category.

Add missing key combinations, commands.
Re-categorize, re-order the keybindings.
Improve the help texts.

- [x] Discussed in **#zulip-terminal** in [`Re-categorization of hotkeys`](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Re-Categorization.20of.20Hotkeys/near/1793560) and [`Readline shortcuts in search`](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Readline.20shortcuts.20in.20search.3F.20.23T1492/near/1794294)`
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [x] Is a follow-up to work in PR #1492 
- [ ] Requires merge of PR #
- [x] Merge will enable work on #1498

### How did you test this?

### Visual changes (Outdated)
image

### DIFF
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -87,12 +87,12 @@
 ## Editor: Navigation
 |Command|Key Combination|
 | :--- | :---: |
-|Jump to the beginning of line|<kbd>Ctrl</kbd> + <kbd>a</kbd> / <kbd>Home</kbd>|
-|Jump to the end of line|<kbd>Ctrl</kbd> + <kbd>e</kbd> / <kbd>End</kbd>|
-|Jump backward one word|<kbd>Meta</kbd> + <kbd>b</kbd> / <kbd>Shift</kbd> + <kbd>Left</kbd>|
-|Jump forward one word|<kbd>Meta</kbd> + <kbd>f</kbd> / <kbd>Shift</kbd> + <kbd>Right</kbd>|
-|Jump to the previous line|<kbd>Up</kbd> / <kbd>Ctrl</kbd> + <kbd>p</kbd>|
-|Jump to the next line|<kbd>Down</kbd> / <kbd>Ctrl</kbd> + <kbd>n</kbd>|
+|Start of line|<kbd>Ctrl</kbd> + <kbd>a</kbd> / <kbd>Home</kbd>|
+|End of line|<kbd>Ctrl</kbd> + <kbd>e</kbd> / <kbd>End</kbd>|
+|Start of current or previous word|<kbd>Meta</kbd> + <kbd>b</kbd> / <kbd>Shift</kbd> + <kbd>Left</kbd>|
+|Start of next word|<kbd>Meta</kbd> + <kbd>f</kbd> / <kbd>Shift</kbd> + <kbd>Right</kbd>|
+|Previous line|<kbd>Up</kbd> / <kbd>Ctrl</kbd> + <kbd>p</kbd>|
+|Next line|<kbd>Down</kbd> / <kbd>Ctrl</kbd> + <kbd>n</kbd>|
 
 ## Editor: Text Manipulation
 |Command|Key Combination|
@@ -105,6 +105,6 @@
 |Cut backwards to the start of the current word|<kbd>Ctrl</kbd> + <kbd>w</kbd> / <kbd>Meta</kbd> + <kbd>Backspace</kbd>|
 |Cut the current line|<kbd>Meta</kbd> + <kbd>x</kbd>|
 |Paste last cut section|<kbd>Ctrl</kbd> + <kbd>y</kbd>|
-|Delete previous character (to left)|<kbd>Ctrl</kbd> + <kbd>h</kbd>|
-|Transpose characters|<kbd>Ctrl</kbd> + <kbd>t</kbd>|
+|Delete previous character|<kbd>Ctrl</kbd> + <kbd>h</kbd>|
+|Swap with previous character|<kbd>Ctrl</kbd> + <kbd>t</kbd>|
 
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -97,14 +97,14 @@
 ## Editor: Text Manipulation
 |Command|Key Combination|
 | :--- | :---: |
-|Delete previous character (to left)|<kbd>Ctrl</kbd> + <kbd>h</kbd>|
-|Transpose characters|<kbd>Ctrl</kbd> + <kbd>t</kbd>|
+|Undo last action|<kbd>Ctrl</kbd> + <kbd>_</kbd>|
+|Clear text box|<kbd>Ctrl</kbd> + <kbd>l</kbd>|
 |Cut forwards to the end of the line|<kbd>Ctrl</kbd> + <kbd>k</kbd>|
 |Cut backwards to the start of the line|<kbd>Ctrl</kbd> + <kbd>u</kbd>|
 |Cut forwards to the end of the current word|<kbd>Meta</kbd> + <kbd>d</kbd>|
 |Cut backwards to the start of the current word|<kbd>Ctrl</kbd> + <kbd>w</kbd> / <kbd>Meta</kbd> + <kbd>Backspace</kbd>|
 |Cut the current line|<kbd>Meta</kbd> + <kbd>x</kbd>|
 |Paste last cut section|<kbd>Ctrl</kbd> + <kbd>y</kbd>|
-|Undo last action|<kbd>Ctrl</kbd> + <kbd>_</kbd>|
-|Clear text box|<kbd>Ctrl</kbd> + <kbd>l</kbd>|
+|Delete previous character (to left)|<kbd>Ctrl</kbd> + <kbd>h</kbd>|
+|Transpose characters|<kbd>Ctrl</kbd> + <kbd>t</kbd>|
 
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -83,10 +83,20 @@
 |Autocomplete @mentions, #stream_names, :emoji: and topics|<kbd>Ctrl</kbd> + <kbd>f</kbd>|
 |Cycle through autocomplete suggestions in reverse|<kbd>Ctrl</kbd> + <kbd>r</kbd>|
 |Narrow to compose box message recipient|<kbd>Meta</kbd> + <kbd>.</kbd>|
+
+## Editor: Navigation
+|Command|Key Combination|
+| :--- | :---: |
 |Jump to the beginning of line|<kbd>Ctrl</kbd> + <kbd>a</kbd> / <kbd>Home</kbd>|
 |Jump to the end of line|<kbd>Ctrl</kbd> + <kbd>e</kbd> / <kbd>End</kbd>|
 |Jump backward one word|<kbd>Meta</kbd> + <kbd>b</kbd> / <kbd>Shift</kbd> + <kbd>Left</kbd>|
 |Jump forward one word|<kbd>Meta</kbd> + <kbd>f</kbd> / <kbd>Shift</kbd> + <kbd>Right</kbd>|
+|Jump to the previous line|<kbd>Up</kbd> / <kbd>Ctrl</kbd> + <kbd>p</kbd>|
+|Jump to the next line|<kbd>Down</kbd> / <kbd>Ctrl</kbd> + <kbd>n</kbd>|
+
+## Editor: Text Manipulation
+|Command|Key Combination|
+| :--- | :---: |
 |Delete previous character (to left)|<kbd>Ctrl</kbd> + <kbd>h</kbd>|
 |Transpose characters|<kbd>Ctrl</kbd> + <kbd>t</kbd>|
 |Cut forwards to the end of the line|<kbd>Ctrl</kbd> + <kbd>k</kbd>|
@@ -96,7 +106,5 @@
 |Cut the current line|<kbd>Meta</kbd> + <kbd>x</kbd>|
 |Paste last cut section|<kbd>Ctrl</kbd> + <kbd>y</kbd>|
 |Undo last action|<kbd>Ctrl</kbd> + <kbd>_</kbd>|
-|Jump to the previous line|<kbd>Up</kbd> / <kbd>Ctrl</kbd> + <kbd>p</kbd>|
-|Jump to the next line|<kbd>Down</kbd> / <kbd>Ctrl</kbd> + <kbd>n</kbd>|
-|Clear compose box|<kbd>Ctrl</kbd> + <kbd>l</kbd>|
+|Clear text box|<kbd>Ctrl</kbd> + <kbd>l</kbd>|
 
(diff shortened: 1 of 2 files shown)


zulipterminal/config/keys.py
```diff
@@ -323,77 +323,77 @@ class KeyBinding(TypedDict):
...
         'help_text': 'Jump to the previous line',
-        'key_category': 'msg_compose',
+        'key_category': 'editor',
     },
     'NEXT_LINE': {
         'keys': ['down', 'ctrl n'],
         'help_text': 'Jump to the next line',
-        'key_category': 'msg_compose',
+        'key_category': 'editor',
     },
```
neiljp, line 392: These appear to be applicable in all the cases, which I was a little surprised by - perhaps at some point I assumed that some text boxes didn't wrap, or that we had to pass an explicit multiline parameter.

I remember occasional strange behavior with using these keys in the past, and now these are the widgets used for the other input boxes it looks like they may be a problem there too. This is a separate issue to discuss or file, but I'd suggest exploring the behavior with long text - we or urwid-readline may have a bug there.

zulipterminal/config/keys.py
```diff
@@ -323,77 +323,77 @@ class KeyBinding(TypedDict):
...
         'help_text': 'Paste last cut section',
-        'key_category': 'msg_compose',
+        'key_category': 'editor',
     },
     'UNDO_LAST_ACTION': {
         'keys': ['ctrl _'],
         'help_text': 'Undo last action',
-        'key_category': 'msg_compose',
+        'key_category': 'editor',
     },
```
neiljp, line 382: What do you think about having two categories for editing, ie. one for navigation, and one for cutting/pasting/manipulation/undo? I think undo should be higher, if not top, of the latter set of ones in any case, since it's pretty useful! (even undoing `ctrl l`).

zulipterminal/config/keys.py
```diff
@@ -415,6 +415,7 @@ class KeyBinding(TypedDict):
     "msg_actions": "Message actions",
     "stream_list": "Stream list actions",
     "msg_compose": "Composing a Message",
+    "editor": "Editor actions",
```
neiljp, line 418: This may change if we split the category, but I'm not sure if editor/input/text/entry are better words. I remember reading about this somewhere on czo semi-recently, but can't seem to dig out a reference.

The web app doesn't have a similar category.

neiljp (commented):
@Niloth-p Thanks for the followup on #1492 :+1:

This looks mostly OK, though I left a few notes inline.

One aspect to consider for these keys is that they do not derive from any direct use in the code of the commands, and we depend on there being a correlation between them and the readline library. The list is exposed via the library, but we don't make use of it - see eg. https://github.com/rr-/urwid_readline/issues/8

We could potentially do a check to ensure these are available and map correctly, or build some elements at runtime.

It may also be worth somehow tagging the COMMANDs as unused - even if just with a suffix - since then using them in the repo would make it obvious that they are in fact being used :)

zulipterminal/config/keys.py
Niloth-p: I tried searching CZO.
I could only come up with "Editor Navigation" and "Text Operations", on my own.
Should I keep trying or proceed with these for now?

neiljp: Generally if there are other changes to be made and you're unsure of a name or text (or something else that doesn't affect the changes to be made too much), it's easier to make that change in an overall update to the PR with a name you select, and highlight that you're unsure in eg. a self-review within the PR (on the line, if you want) before marking it for review. Then the other changes are already addressed, and it's possible the name you select will be fine - or it can be either updated quickly before merging by a committer, or discussed first.

However, since I already was thinking about it...

Since the name here refers to the same type of component, I often prefer to keep the same prefix, rather than use a phrase that can either have the word that refers to the component move around, or be different. That can then be a phrase like `Prefix using part 2`, or eg. `Prefix - part 1` and `Prefix - part 2`. The latter form may be clearer here, and in terms of the choice of textbox/editor/... itself, we can always update that name later; generally I just aim to be consistent.

The library doesn't support all the features, but https://www.man7.org/linux/man-pages/man3/readline.3.html#EDITING_COMMANDS provides some ideas for naming.

So for the parts we could have
- navigation or movement (based on the man page)
- manipulation? modification?

neiljp (commented):
@Niloth-p I left the feedback here since you were asking about changes specific to the code.

When you split some commands into a new category, their existing description didn't make sense - at least in terms of the 'compose' case I highlighted - without being part of the same change, so it belongs in the same commit. Was that what you meant by the "compose box" change?

The other changes you highlighted in the channel that aren't yet in this PR are incremental improvements to the wording. They are fine in a separate commit in this PR.

I responded to most other changes in the topic, but for the `ctrl l` case which I hadn't yet done so, let's just keep this consistent with the category section name?

Niloth-p:
### Visual Changes
Before re-ordering: Screenshot from 2024-05-12 17-32-03
After re-ordering: Screenshot from 2024-05-12 17-27-43
Our shortcuts missing in Readline's keymap: 
Screenshot from 2024-05-12 22-45-46
On successful linting: 
Screenshot from 2024-05-12 22-46-30
Linting errors:
Screenshot from 2024-05-12 22-47-02

Niloth-p:
### Updated Help Menu
Screenshot from 2024-05-14 02-03-42
Screenshot from 2024-05-14 02-04-03
### Updated Hotkeys.md
Screenshot from 2024-05-14 02-04-27

docs/hotkeys.md
```diff
@@ -83,16 +83,16 @@
 |Autocomplete @mentions, #stream_names, :emoji: and topics|<kbd>Ctrl</kbd> + <kbd>f</kbd>|
```
neiljp, line 83: Thanks for finding these!

I understand what this means, but for future readers of the log, it would be useful to clarify in what way these are missing, in the commit text.

For example, where did you find this information?

zulipterminal/config/keys.py
```diff
@@ -370,6 +370,11 @@ class KeyBinding(TypedDict):
         'help_text': 'Cut backwards to the start of the current word',
         'key_category': 'msg_compose',
     },
+    'CUT_WHOLE_LINE': {
+        'keys': ['meta x'],
+        'help_text': 'Cut the current line',
+        'key_category': 'msg_compose',
+    },
```
neiljp: We might end up squashing this and the previous commit, since they're both about missing documented keys, but this is clearer as a separate commit at this point :+1:

Unless I'm missing something, the title suggests that we're adding a command, not documenting an existing command.

docs/hotkeys.md
```diff
@@ -83,10 +83,20 @@
...
 |Cycle through autocomplete suggestions in reverse|<kbd>Ctrl</kbd> + <kbd>r</kbd>|
 |Narrow to compose box message recipient|<kbd>Meta</kbd> + <kbd>.</kbd>|
+
```
neiljp, line 86: In this commit, it'd be helpful to clarify where the keys in the new categories come *from*. The code suggests that (via the change in category tags), but it's helpful if the commit text (or even title) mentions that.

Note that if a commit is merged to `main`, you can reference it in the commit text here, ie. where all the editors became readline-capable.

zulipterminal/config/keys.py
```diff
@@ -323,82 +323,82 @@ class KeyBinding(TypedDict):
...
         'keys': ['down', 'ctrl n'],
         'help_text': 'Jump to the next line',
-        'key_category': 'msg_compose',
+        'key_category': 'editor_navigation',
     },
     'CLEAR_MESSAGE': {
         'keys': ['ctrl l'],
-        'help_text': 'Clear compose box',
-        'key_category': 'msg_compose',
+        'help_text': 'Clear text box',
```
neiljp: This change connects with why I'd mention the category move/split in the commit text, not just the new categories. All the other changes are category changes, this is necessary due to it being a move.

zulipterminal/config/keys.py
```diff
@@ -287,6 +287,16 @@ class KeyBinding(TypedDict):
...
+    'FULL_RENDERED_MESSAGE': {
+        'keys': ['f'],
+        'help_text': 'Show/hide full rendered message (from message information)',
+        'key_category': 'msg_actions',
+    },
+    'FULL_RAW_MESSAGE': {
+        'keys': ['r'],
+        'help_text': 'Show/hide full raw message (from message information)',
+        'key_category': 'msg_actions',
+    },
```
neiljp, line 299: I'm not averse to this part of the commit, but this isn't related to reordering the keys, based on the output in `hotkeys.md` being unchanged? That is, this seems like an independent refactor.

docs/hotkeys.md
```diff
@@ -97,14 +97,14 @@
...
 |Cut forwards to the end of the line|<kbd>Ctrl</kbd> + <kbd>k</kbd>|
 |Cut backwards to the start of the line|<kbd>Ctrl</kbd> + <kbd>u</kbd>|
 |Cut forwards to the end of the current word|<kbd>Meta</kbd> + <kbd>d</kbd>|
 |Cut backwards to the start of the current word|<kbd>Ctrl</kbd> + <kbd>w</kbd> / <kbd>Meta</kbd> + <kbd>Backspace</kbd>|
 |Cut the current line|<kbd>Meta</kbd> + <kbd>x</kbd>|
 |Paste last cut section|<kbd>Ctrl</kbd> + <kbd>y</kbd>|
-|Undo last action|<kbd>Ctrl</kbd> + <kbd>_</kbd>|
-|Clear text box|<kbd>Ctrl</kbd> + <kbd>l</kbd>|
+|Delete previous character (to left)|<kbd>Ctrl</kbd> + <kbd>h</kbd>|
+|Transpose characters|<kbd>Ctrl</kbd> + <kbd>t</kbd>|
```
neiljp: We discussed this reordering, and I agree with it, but I'd look to see a motivation for the change expressed in the commit text.

The current commit title/text looks much more broader than I'd expect, given this change to `hotkeys.md`, which is also the impact I'd expect on the help menu.

zulipterminal/config/keys.py
neiljp: Good to see this present, I think it helps :+1:

zulipterminal/config/keys.py
```diff
@@ -332,22 +332,22 @@ class KeyBinding(TypedDict):
...
     'END_OF_LINE': {
         'keys': ['ctrl e', 'end'],
-        'help_text': 'Jump to the end of line',
+        'help_text': 'Jump to the end of the current line',
         'key_category': 'editor_navigation',
     },
     'ONE_WORD_BACKWARD': {
         'keys': ['meta b', 'shift left'],
-        'help_text': 'Jump backward one word',
+        'help_text': 'Jump backward to the start of current or previous word',
```
neiljp, line 345: I'm not entirely convinced of all of these - they're clearer wording in some cases, but also longer. More compact and concise would make these clearer.

For the navigation text, perhaps we can simplify using the fact that they all start with 'Jump'? That common phrasing suggests a simplification.

The other reason to keep these shorter is that they make the help menu wider or start to wrap, though currently there is one command that is even longer than these.

neiljp (commented):
@Niloth-p This looks like being close to merging, and the improvements since the last iteration look good, including those you found out from elsewhere :+1:

There are a few spots where we could adjust things further, or that we could simplify, as per the notes I left inline - which for some reason appear to have been posted as a review separately already.

It's not important to handle this now, but one aspect I'd considered for the future was to handle the readline commands automatically, so setting the 'keys' lists via the dict(s) in urwid-readline. I don't expect we'd need to keep updating them, but it would ensure that they didn't get reset in the help, and were kept synchronized.

neiljp:
@zormit I started the original reviewing here, but would welcome any feedback you have :)

docs/hotkeys.md
zormit: It seems this is resolved? 0d1babd now explains that they keys are coming from `urwid_readline`.

zulipterminal/config/keys.py
zormit: This also seems to be resolved.

Niloth-p:
@zormit The wording for the final commit was not particularly discussed anywhere beyond the feedback here. So, they're not yet decided.

The tagging of commands has been done as part of #1498 after Neil's recommendation in the feedback for this PR.

The suggestion of programmatic synchronization was made recently, after #1498 was opened, so has not been attempted yet.

rsashank:
LGTM. I've caught up with the previous comments left, and everything looks in order to me. Removing the buddy review tag.

I apologize for the delay in buddy reviews; I was unavailable due to exams when the buddy review tag was added. I'll be quicker with reviews now on.

zormit:
So from my point of view, to finalize this PR you'll need to
- [ ] make sure the wording is final (this requires a review by/discussion with @neiljp )
- [ ] update the last commit message (I'm imagining a description how you came up with the changes and why)

Niloth-p:
Just added a commit description to the last commit, as per feedback.
@zormit This good?

zulipterminal/config/keys.py
neiljp: I posted about this in `#zulip-terminal>Up/Down bug (urwid-readline?)`, to ensure we keep track of that.

docs/hotkeys.md
```diff
@@ -105,6 +105,6 @@
 |Cut backwards to the start of the current word|<kbd>Ctrl</kbd> + <kbd>w</kbd> / <kbd>Meta</kbd> + <kbd>Backspace</kbd>|
 |Cut the current line|<kbd>Meta</kbd> + <kbd>x</kbd>|
 |Paste last cut section|<kbd>Ctrl</kbd> + <kbd>y</kbd>|
-|Delete previous character (to left)|<kbd>Ctrl</kbd> + <kbd>h</kbd>|
-|Transpose characters|<kbd>Ctrl</kbd> + <kbd>t</kbd>|
+|Delete character behind cursor|<kbd>Ctrl</kbd> + <kbd>h</kbd>|
+|Swap with previous character|<kbd>Ctrl</kbd> + <kbd>t</kbd>|
```
neiljp, line 109: My only nit with this change is that the other changes are more consistent with using 'previous', so maybe we just drop the `(to left)`.

neiljp (commented):
I've left a few comments, but I really don't want to dwell on minor details too much and instead get this merged soon.

Thanks to @Niloth-p and for everyone's reviews - I'll likely make a few minor text changes and merge a bit later, but am busy for a bit now.

neiljp:
@Niloth-p I just pushed the textual corrections I mentioned, and again after rebasing against a PR I just pushed to allow long URLs in the commit text body, and am merging now - thanks again! :tada:

Please do try a `git range-diff` against your local branch to see the specific changes; we can discuss that in the project stream or private stream, if that would be helpful more generally.

PR #1495, opened 2024-05-11
tests: boxes: Add tests for saving drafts.

Niloth-p:
Add 2 test functions:
1. Draft message to a channel
2. Draft message to a private narrow (DM and PM huddle)

Test for an existing saved draft for each of the test cases.

Test for:
- a function call to display the save-draft-confirmation popup, in case of existing saved draft
- the created draft object
- return of the key without creating a draft, in case of invalid recipients

This does not test the `_tidy_valid_recipients_and_notify_invalid_ones` or the `session_draft_message` functions. They're mocked. Tests for them would need to be added separately.

- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [x] Builds upon previous unmerged work in PR #950 
- [ ] Requires merge of PR #

### How did you test this?

- [ ] It has a commit summary describing the motivation and reasoning for the change

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -39,21 +39,33 @@ MODULE = "zulipterminal.ui_tools.boxes"
 WRITEBOX = MODULE + ".WriteBox"
 
 
-def composition_factory(
-    type: str,
-    to: Union[str, List[int]] = "",
-    subject: Optional[str] = None,
-    read_by_sender: bool = True,
-) -> Dict[str, object]:
-    data = {
-        "type": type,
-        "content": f"Random {type} message",
-        "read_by_sender": read_by_sender,
-        "to": to,
-    }
-    if subject:
-        data["subject"] = subject
-    return data
+def composition_factory(type: str, group: bool = False) -> Composition:
+    read_by_sender: bool = True
+    content = "Random {} message"
+    if type == "private":
+        if group:
+            to = [5140, 5180]
+        else:
+            to = [5140]
+        content = content.format(type)
+        return PrivateComposition(
+            {
+                "type": "private",
+                "content": content,
+                "to": to,
+                "read_by_sender": read_by_sender,
+            }
+        )
+    else:
+        return StreamComposition(
+            {
+                "type": "stream",
+                "content": content,
+                "subject": "Topic",
+                "to": "Current stream",
+                "read_by_sender": read_by_sender,
+            }
+        )
 
 
 @dataclass
@@ -149,27 +161,9 @@ class TestWriteBox:
     @pytest.fixture(
         params=[
             saved_draft_factory(),
-            saved_draft_factory(
-                StreamComposition(
-                    **composition_factory(  # type: ignore[typeddict-item]
-                        type="stream", to="Current stream", subject="Topic"
-                    )
-                )
-            ),
-            saved_draft_factory(
-                PrivateComposition(
-                    **composition_factory(  # type: ignore[typeddict-item]
-                        type="private", to=[5179]
-                    )
-                )
-            ),
-            saved_draft_factory(
-                PrivateComposition(
-                    **composition_factory(  # type: ignore[typeddict-item]
-                        type="group", to=[5140, 5179]
-                    )
-                )
-            ),
+            saved_draft_factory(composition_factory(type="stream")),
+            saved_draft_factory(composition_factory(type="private")),
+            saved_draft_factory(composition_factory(type="private", group=True)),
         ],
         ids=[
             "no_saved_draft_exists",
@@ -183,10 +177,9 @@ class TestWriteBox:
 
     @pytest.fixture(
         params=[
-            composition_factory("private", to=[5140]),
-            composition_factory("private", to=[5140, 5180]),
+            composition_factory("private"),
+            composition_factory("private", group=True),
         ],
-        ids=["private_draft", "group_draft"],
     )
     def private_draft_composition(
         self, request: FixtureRequest
@@ -194,7 +187,7 @@ class TestWriteBox:
         return request.param
 
     @pytest.fixture(
-        params=[composition_factory("stream", to="Another stream", subject="Topic")],
+        params=[composition_factory("stream")],
     )
     def stream_draft_composition(
         self, request: FixtureRequest
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -2,7 +2,7 @@ import datetime
 from collections import OrderedDict
 from dataclasses import dataclass
 from functools import reduce
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import pytest
 from pytest import FixtureRequest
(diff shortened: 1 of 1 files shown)


tests/ui_tools/test_boxes.py
```diff
@@ -103,6 +105,124 @@ def test_not_calling_typing_method_without_recipients(
...
+            is_draft_saved_in_current_session
+        )
+        write_box.compose_box_status = "open_with_private"
+        write_box.recipient_user_ids = draft_composition["to"]
+
+        size = widget_size(write_box)
+        return_value = write_box.keypress(size, key)
+
+        if not are_valid_recipients:
+            assert key == return_value
```
zormit, line 214: For my understanding: why are we testing the key here? How is that dependent on the validity of the recipients? It seems to me (from a cursory glance) that `keypress` always returns its `key` argument.

zormit (commented):
Generally this looks fine to me. I left some comments on things that came to mind. I don't know enough about the codebase yet to give an approval, but given that the test is green and it looks to me like it's testing something useful, I probably would approve it if I had to :P 

hope that helps :)

tests/ui_tools/test_boxes.py
```diff
@@ -103,6 +105,124 @@ def test_not_calling_typing_method_without_recipients(
...
+        "is_draft_saved_in_current_session",
+        [True, False],
+        ids=["saved_draft_exists", "no_saved_draft_exists"],
+    )
+    def test_keypress_SAVE_DRAFT_MESSAGE_private(
+        self,
+        key: str,
+        mocker: MockerFixture,
+        draft_composition: PrivateComposition,
+        are_valid_recipients: bool,
```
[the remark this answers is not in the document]
Niloth-p: Went with this [Stackoverflow answer](https://stackoverflow.com/a/22397978).
Thanks for the info!

tests/ui_tools/test_boxes.py
```diff
@@ -103,6 +105,124 @@ def test_not_calling_typing_method_without_recipients(
 
         assert not write_box.model.send_typing_status_by_user_ids.called
 
+    @pytest.mark.parametrize("key", keys_for_command("SAVE_AS_DRAFT"))
+    @pytest.mark.parametrize(
+        "is_draft_saved_in_current_session",
+        [True, False],
+        ids=["saved_draft_exists", "no_saved_draft_exists"],
+    )
+    def test_keypress_SAVE_DRAFT_MESSAGE_stream(
```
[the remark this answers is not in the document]
Niloth-p: yes, I had pulled that from the unmerged PR, and had failed to notice it and update it. Thanks for the pointer.

neiljp (commented):
I'll take another look later, but wanted to get this comment out that I'd drafted previously, before mentoring had started.

Niloth-p:
Updated to add the `mocked_saved_drafts` list. Took that outside both the test functions to reduce redundancy.
Also switched from using 'huddle' to 'dm_group' in the ids and message.

tests/ui_tools/test_boxes.py
```diff
@@ -103,6 +106,144 @@ def test_not_calling_typing_method_without_recipients(
...
+        )
+        write_box.compose_box_status = "open_with_private"
+        write_box.recipient_user_ids = draft_composition["to"]
+
+        size = widget_size(write_box)
+        write_box.keypress(size, key)
+
+        if not is_every_recipient_valid:
+            assert not write_box.model.save_draft.called
+            assert not write_box.view.controller.save_draft_confirmation_popup.called
```
rsashank, line 237: NIT: Should we prefer using? 
```
write_box.model.save_draft.assert_not_called()
write_box.view.controller.save_draft_confirmation_popup.assert_not_called()
```

tests/ui_tools/test_boxes.py
```diff
@@ -103,6 +106,144 @@ def test_not_calling_typing_method_without_recipients(
...
+        else:
+            assert write_box.model.session_draft_message.called
+            if draft_saved_in_current_session is not None:
```
rsashank, line 240: NIT: Can this simply be `if draft_saved_in_current_session`?

rsashank:
LGTM! Great work @Niloth-p 👍

Niloth-p:
Thank you for catching those details, @rsashank! Great review, that was very helpful.
Updated the requested changes.

tests/ui_tools/test_boxes.py
neiljp: These are equivalent - or should be!

However, the great thing about using `assert` and the core `called` style values are:
- pytest can dig through levels of code using `assert`, if they fail
- `called` and similar results are values, not assertions in a function

The most useful aspect here though, is that this form can be more easily extracted into parametrize blocks.

tests/ui_tools/test_boxes.py
```diff
@@ -103,6 +106,144 @@ def test_not_calling_typing_method_without_recipients(
...
+        write_box.compose_box_status = "open_with_private"
+        write_box.recipient_user_ids = draft_composition["to"]
+
+        size = widget_size(write_box)
+        write_box.keypress(size, key)
+
+        if not is_every_recipient_valid:
+            write_box.model.save_draft.assert_not_called()
+            write_box.view.controller.save_draft_confirmation_popup.assert_not_called()
+        else:
```
neiljp, line 238: To simplify this, since the test has a nested `if`, I'm wondering if it would be cleaner to extract this first part into a separate test?

That avoids a parametrize over is_every_recipient_valid.

neiljp (commented):
@Niloth-p I was just mid-review yesterday, so just posting the notes I made and a few more, since I note you updated this already :)

tests/ui_tools/test_boxes.py
rsashank: Interesting! Thanks for sharing this :)

My bad, you had to update it again @Niloth-p 😅

tests/ui_tools/test_boxes.py
rsashank: Hmm, the parametrization part makes sense. Should we stick to using .called from now on?

zormit:
@Niloth-p I have tried simplifying and typing your code a bit in https://github.com/zormit/zulip-terminal/tree/1495-save-draft/pr (713bf505, in case I change the branch). What do you think?

I'm still not fully happy with all the layers of indirection, probably it can be simplified even more, but this might give you an idea. I have to stop for now, so I'm sending you this idea, please take it if you like it.

Niloth-p:
@zormit Thank you lots for making those changes yourself! That does look much better.

Sorry for the delay in replying back, I'd absorbed your commits right away, but hadn't managed to type out a reply here.

If you find this interesting, please feel free to push your own PR with other improvements and add me as a co-author, that could help get it merged too, as I've done several iterations on this previously, and have been struggling to identify good code and bad code here.

So, no, I'm not really clear on how we can further reduce the indirection, my apologies. I can think of several ways to change the tests, but I have no idea what improves it and what worsens it. If someone is able to guide me with 1 liners like the commit messages, I'll try my best to implement it and get back.

Being new to the complexities of testing, when I come across best practices and guidance, I'm not sure to what extent I should be following them, that I realise I might have overdone it here to the point of complexity, and I'm still not clear enough on all the tradeoffs, so please do let me know how we can proceed from here.

Also, I'm not sure how many commits this is supposed to be. I was initially under the impression this should be a single commit, since it's just tests, and we wouldn't want to be tracking the additions incrementally. But, after the suggestion from @neiljp to break down the refactoring commits of lint-hotkeys, I've been wondering whether we should break this down too. Do we need to?

I've added some newer test cases. But I'm not sure if we want to be that thorough or if it bulks up the testing time unnecessarily.

Niloth-p:
Github will report the tests as failed, because I've kept my addition separate from Moritz's commits, but they'll pass if they're squashed. So, please ignore them.

PR #1496, opened 2024-05-14
Standardize hotkey help text

Niloth-p:
Improves the conformity of help descriptions.

### Outstanding aspect(s)
[ ] Usage of 'the' needs to be made uniform too. This has not yet been addressed.

- [x] Discussed in **#zulip-terminal** inside `re-categorize` [here](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Re-Categorization.20of.20Hotkeys/near/1793558) and [here](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Re-Categorization.20of.20Hotkeys/near/1797363)
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -154,7 +154,7 @@ KEY_BINDINGS: Dict[str, KeyBinding] = {
     },
     'ADD_REACTION': {
         'keys': [':'],
-        'help_text': 'Show/hide emoji picker popup for current message',
+        'help_text': 'Show/hide emoji picker for current message',
         'key_category': 'msg_actions',
     },
     'STREAM_NARROW': {
@@ -240,7 +240,7 @@ KEY_BINDINGS: Dict[str, KeyBinding] = {
     },
     'SEARCH_EMOJIS': {
         'keys': ['p'],
-        'help_text': 'Search emojis from emoji picker popup',
+        'help_text': 'Search emojis from emoji picker',
         'excluded_from_random_tips': True,
         'key_category': 'searching',
     },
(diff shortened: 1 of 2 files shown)


zulipterminal/config/keys.py
```diff
@@ -154,7 +154,7 @@ class KeyBinding(TypedDict):
     },
     'ADD_REACTION': {
         'keys': [':'],
-        'help_text': 'Show/hide Emoji picker popup for current message',
+        'help_text': 'Show/hide Emoji Picker popup for current message',
```
zormit, line 157: I'm not sure on this one, given that "Emoji Picker" is not the title used for the popup (which is different to the other cases above).

My personal preference would be to use "Emoji-picker" instead of "Emoji picker popup" and "Emoji-picker popup", but any of these three could work.

zormit (approved):
Generally looks good to me and makes things indeed more consistent. Thanks for working on this. I left one comment where my personal opinion is a bit different, but it's on the level of [bike-shedding](https://en.wikipedia.org/wiki/Law_of_triviality) so I certainly won't insist on it.

My approval means that I don't need to look again at this after this one thing is resolved, but of course I'll leave the final decision to @neiljp.

zulipterminal/config/keys.py
Niloth-p: Yes, that's a tricky one indeed.
Yep, we could remove the 'popup' word. Good one.
Googling shows a usage preference of not using a hyphen. So, I went with that.
And yes, the popup does not have a header.

Niloth-p:
Just updated the "Emoji Picker Popup" mentions.
commit 1 (case consistency): Lower cased the emoji picker.
commit 2 (update wording): Removed popup.
Updated both commit messages.

rsashank:
LGTM - seems pretty clear :)

neiljp:
@Niloth-p Thanks for this again - looking forward to more iterations along these lines and the contextual work! :tada:

PR #1497, opened 2024-05-14
Break down the generalized ENTER command

Niloth-p:
Breaks down the broader ENTER command into more specific commands.
- ACTIVATE_BUTTON
- EXECUTE_SEARCH
- NEW_LINE

This would enable using more specific help texts, categories and context.
And clarifies the purposes of the key better.

Updated the usage of each command in the UI tools.
Tests and docs updated.

The NEW_LINE command would also allow us to create a new user setting, where `ENTER` key and `Meta`+`Enter` can be switched between entering new lines and sending messages.

### Outstanding aspect(s)
- Need to consider the placement of the new commands within their categories in the Help Menu.
- `Ctrl`+`M` can work as `Enter` too. This has not yet been addressed.

- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -24,7 +24,7 @@
 |Scroll up|<kbd>PgUp</kbd> / <kbd>K</kbd>|
 |Scroll down|<kbd>PgDn</kbd> / <kbd>J</kbd>|
 |Go to bottom / Last message|<kbd>End</kbd> / <kbd>G</kbd>|
-|Trigger the selected entry|<kbd>Enter</kbd>|
+|Trigger the selected entry|<kbd>Enter</kbd> / <kbd>Space</kbd>|
 |Narrow to all messages|<kbd>a</kbd> / <kbd>Esc</kbd>|
 |Narrow to all direct messages|<kbd>P</kbd>|
 |Narrow to all starred messages|<kbd>f</kbd>|
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -24,13 +24,13 @@
 |Scroll up|<kbd>PgUp</kbd> / <kbd>K</kbd>|
 |Scroll down|<kbd>PgDn</kbd> / <kbd>J</kbd>|
 |Go to bottom / Last message|<kbd>End</kbd> / <kbd>G</kbd>|
+|Trigger the selected entry|<kbd>Enter</kbd>|
 |Narrow to all messages|<kbd>a</kbd> / <kbd>Esc</kbd>|
 |Narrow to all direct messages|<kbd>P</kbd>|
 |Narrow to all starred messages|<kbd>f</kbd>|
 |Narrow to messages in which you're mentioned|<kbd>#</kbd>|
 |Next unread topic|<kbd>n</kbd>|
 |Next unread direct message|<kbd>p</kbd>|
-|Perform current action|<kbd>Enter</kbd>|
 
 ## Searching
 |Command|Key Combination|
@@ -40,6 +40,7 @@
 |Search streams|<kbd>q</kbd>|
 |Search topics in a stream|<kbd>q</kbd>|
 |Search emojis from emoji picker|<kbd>p</kbd>|
+|Submit search and browse results|<kbd>Enter</kbd>|
 
 ## Message actions
 |Command|Key Combination|
@@ -83,6 +84,7 @@
 |Autocomplete @mentions, #stream_names, :emoji: and topics|<kbd>Ctrl</kbd> + <kbd>f</kbd>|
 |Cycle through autocomplete suggestions in reverse|<kbd>Ctrl</kbd> + <kbd>r</kbd>|
 |Narrow to compose box message recipient|<kbd>Meta</kbd> + <kbd>.</kbd>|
+|Insert new line|<kbd>Enter</kbd>|
 
 ## Editor: Navigation
 |Command|Key Combination|
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1961,7 +1961,9 @@ class TestMessageBox:
         assert isinstance(footlinks, expected_instance)
 
     @pytest.mark.parametrize(
-        "key", keys_for_command("ENTER"), ids=lambda param: f"left_click-key:{param}"
+        "key",
+        keys_for_command("ACTIVATE_BUTTON"),
+        ids=lambda param: f"left_click-key:{param}",
     )
     def test_mouse_event_left_click(
         self, mocker, msg_box, key, widget_size, compose_box_is_open
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -7,7 +7,7 @@ from bs4 import BeautifulSoup
 from pytest import param as case
 from urwid import Columns, Divider, Padding, Text
 
-from zulipterminal.config.keys import keys_for_command
+from zulipterminal.config.keys import keys_for_command, primary_key_for_command
 from zulipterminal.config.symbols import (
     ALL_MESSAGES_MARKER,
     DIRECT_MESSAGE_MARKER,
@@ -1968,6 +1968,7 @@ class TestMessageBox:
     def test_mouse_event_left_click(
         self, mocker, msg_box, key, widget_size, compose_box_is_open
     ):
+        expected_keypress = primary_key_for_command("ACTIVATE_BUTTON")
         size = widget_size(msg_box)
         col = 1
         row = 1
@@ -1981,4 +1982,4 @@ class TestMessageBox:
         if compose_box_is_open:
             msg_box.keypress.assert_not_called()
         else:
-            msg_box.keypress.assert_called_once_with(size, key)
+            msg_box.keypress.assert_called_once_with(size, expected_keypress)
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -6,6 +6,7 @@ from typing import Dict, List
 
 from typing_extensions import NotRequired, TypedDict
 from urwid.command_map import (
+    ACTIVATE,
     CURSOR_DOWN,
     CURSOR_LEFT,
     CURSOR_MAX_RIGHT,
@@ -92,7 +93,7 @@ KEY_BINDINGS: Dict[str, KeyBinding] = {
         'key_category': 'navigation',
     },
     'ACTIVATE_BUTTON': {
-        'keys': ['enter'],
+        'keys': ['enter', ' '],
         'help_text': 'Trigger the selected entry',
         'key_category': 'navigation',
     },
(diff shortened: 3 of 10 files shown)


zormit (approved):
Looks good to me. I think it achieves the goal of preparing for better contextual help.

A downside of restructuring it like this, is that the code now could accidentally get stale as no one is stopped from using for example `ACTIVATE_BUTTON` when actually in a context where `EXECUTE_SEARCH` would be the "correct Enter" variant. A follow-up change could be to not have a single `KEY_BINDINGS`, but one per category/context. But that's not meant as a request, just a thought that came up. Or add the category in the `keys_for_command` calls.

Regarding your outstanding items:

> Need to consider the placement of the new commands within their categories in the Help Menu.

How would you imagine the placement? Is it related to any of the other open PRs?

> Ctrl+M can work as Enter too. This has not yet been addressed.

How did you mean to address this? The fact that ^M works "as Enter" (or rather CR) has to do with [ASCII control characters](https://en.wikipedia.org/wiki/Control_character). My understanding is that it's determined on a _system_ or maybe terminal emulator level. My opinion is that, as an _application_, we should  not worry about this. It might even be impossible to tell the difference at this level. As for documenting this, I wouldn't do that either for similar reasons (it's nothing people usually _need_, and if people know about this, it's rather more widely applicable than just zulip-terminal)

Niloth-p:
Thank you for addressing the outstanding items @zormit

> How would you imagine the placement? Is it related to any of the other open PRs?

It's something that Neil suggested in the feedback of #1494. That we could re-order these help entries in the help menu, based on their significance / frequency of use, which is fairly subjective.
Hence, the question to confirm whether the order I've used here is fine, as this was not previously discussed in the stream.
(the order in the help menu is the same as the order in which the key bindings are listed in keys.py)

> How did you mean to address this?

I didn't get till there. Just mentioning that it exists, for the sake of completeness.

> My opinion is that, as an application, we should not worry about this.

Great then.
I wasn't quite aware about the details behind Ctrl M, and wasn't sure if I should spend any time looking into it either.
So, thanks for the explanation!

> A follow-up change could be to not have a single KEY_BINDINGS, but one per category/context. Or add the category in the keys_for_command calls.

Ah, great point, thank you for the suggestions!
We should probably have a discussion finalizing the approach and the order of merging, as this connects to a bunch of open PRs and discussed work related to categories/contexts.

zormit:
@Niloth-p Can you please rebase this to the latest `main`? (It basically needs a regeneration of `docs/hotkeys.md` due to recent merges)

Niloth-p:
Linting and tests are passing on my fork [here](https://github.com/Niloth-p/zulip-terminal/actions/runs/9346473251).

zulipterminal/config/keys.py
```diff
@@ -400,6 +405,11 @@ class KeyBinding(TypedDict):
         'help_text': 'Swap with previous character',
         'key_category': 'editor_text_manipulation',
     },
+    'NEW_LINE': {
+        'keys': ['enter'],
+        'help_text': 'Insert new line',
+        'key_category': 'msg_compose',
+    },
```
neiljp, line 412: In some ways this comes across like one of those obvious ones like `backspace`, though I agree that it could depend on 'enter to send' - so this is fine for now, and balances out the explicit calling out of `enter` for other purposes elsewhere in the UI.

However, we may want to add a comment here about those points, or possibly connect it to the urwid-readline 'suffix' work, since that is where the multi-line key comes from.

neiljp (commented):
@Niloth-p It'll be useful to have these split :+1:

I left a few points inline. From a code perspective this is fairly straightforward, though the ordering/category looks a bit strange.

@zormit, you wrote:
> A downside of restructuring it like this, is that the code now could accidentally get stale as no one is stopped from using for example ACTIVATE_BUTTON when actually in a context where EXECUTE_SEARCH would be the "correct Enter" variant. A follow-up change could be to not have a single KEY_BINDINGS, but one per category/context. But that's not meant as a request, just a thought that came up. Or add the category in the keys_for_command calls.

I'm not sure I follow this; do you mean where `enter` ends up being present for multiple actions, it could be unclear what should occur if two are present on the same UI object by mistake, or something like that?

zulipterminal/config/keys.py
```diff
@@ -29,6 +29,11 @@ class KeyBinding(TypedDict):
     # Key that is displayed in the UI is determined by the method
     # primary_key_for_command. (Currently the first key in the list)
 
+    'ACTIVATE_BUTTON': {
+        'keys': ['enter'],
+        'help_text': 'Trigger the selected entry',
+        'key_category': 'general',
+    },
```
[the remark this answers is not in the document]
Niloth-p: Re. UI Navigation category,
I can't particularly recall discussing "UI Navigation".
Do you mean moving the narrowing commands in #1516?
We haven't used the term "UI Navigation" there yet, just extracting the narrowing commands from it. I'll update the category name.
I think we could do them independently, and in any order? As it would make sense to have ACTIVATE_BUTTON in navigation category even before it's split.
Please let me know if you have more in mind, as I don't think we've discussed this before.

> space works on buttons

I totally missed that one.
Thank you for that clear guidance. That really helped.
I've added the space key, updated the display function.
I have tied it to the Urwid mapping even though we do not add any new hotkeys at the moment, for consistency and safety.

> document the behavior of enter in the user list to reply/compose to a user. That's distinct from this split, and likely depends on a user-list/panel category.

Yes, I've added that as part of [this commit](https://github.com/zulip/zulip-terminal/pull/1515/commits/625344fa29b22474f0a221b3c41cfa2ade01dabd) in #1515.

tests/ui_tools/test_messages.py
```diff
@@ -1968,6 +1968,7 @@ def test_footlinks_limit(self, maximum_footlinks, expected_instance):
     def test_mouse_event_left_click(
         self, mocker, msg_box, key, widget_size, compose_box_is_open
     ):
+        expected_keypress = primary_key_for_command("ACTIVATE_BUTTON")
```
Niloth-p, line 1971: Needed to edit this test as mouse_event always calls the `primary_key_for_command("ACTIVATE_BUTTON")` and does not use the passed key.

zulipterminal/config/keys.py
```diff
@@ -487,6 +492,9 @@ def display_key_for_urwid_key(urwid_key: str) -> str:
     """
     Returns a displayable user-centric format of the urwid key.
     """
+    if urwid_key == " ":
```
Niloth-p, line 495: Did not add this as an entry to the `URWID_KEY_TO_DISPLAY_KEY_MAPPING` as we do not want to affect the behavior of space when it is used as a separator between words.

zulipterminal/config/keys.py
Niloth-p: Thank you for the pointer!
I've added a comment.
Let me know if it needs to be edited.

zulipterminal/config/keys.py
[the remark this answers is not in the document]
Niloth-p: Re. UI Navigation category,
I can't particularly recall discussing "UI Navigation".
Do you mean moving the narrowing commands in #1516?
We haven't used the term "UI Navigation" there yet, just extracting the narrowing commands from it. I'll update the category name.
I think we could do them independently, and in any order? As it would make sense to have ACTIVATE_BUTTON in navigation category even before it's split.
Please let me know if you have more in mind, as I don't think we've discussed this before.

> space works on buttons

I totally missed that one.
Thank you for that clear guidance. That really helped.
I've added the space key, updated the display function.
I have tied it to the Urwid mapping even though we do not add any new hotkeys at the moment, for consistency and safety.

> document the behavior of enter in the user list to reply/compose to a user. That's distinct from this split, and likely depends on a user-list/panel category.

Yes, I've added that as part of [this commit](https://github.com/zulip/zulip-terminal/pull/1515/commits/625344fa29b22474f0a221b3c41cfa2ade01dabd) in #1515.

Niloth-p (commented):
@neiljp Thank you for the detailed review and the great explanations!

I hope I've addressed all your suggestions with this update.

> I'm not sure I follow this; do you mean where enter ends up being present for multiple actions, it could be unclear what should occur if two are present on the same UI object by mistake, or something like that?

This is my understanding of what @zormit meant,
If we add a new button, and support keypress for it, then the code there can use `is_command_key("EXECUTE_SEARCH")` instead of `is_command_key("ACTIVATE_BUTTON")`, and achieve the same functionality.
As in the developer could add a mismatched command that may still work because both of them map to Enter. But, this error would create wrong contextual hints.

neiljp (commented):
@Niloth-p It looks like most of my points were accounted for already, while you were updating :)

Re the point about potentially incorrect use of Commands that are essentially equivalent in terms of keys, I agree this is a possible challenge. However, the aim is to maintain semantic naming, a little like we do with theme styles which may have the same content but are still named differently. We can examine this again when the contextual part becomes finalized.

Niloth-p:
Squashed the commits.
Kept the addition of 'space' in a separate commit as recommended, and newly wrote its commit message.

docs/hotkeys.md
```diff
@@ -24,7 +24,7 @@
 |Scroll up|<kbd>PgUp</kbd> / <kbd>K</kbd>|
 |Scroll down|<kbd>PgDn</kbd> / <kbd>J</kbd>|
 |Go to bottom / Last message|<kbd>End</kbd> / <kbd>G</kbd>|
-|Trigger the selected entry|<kbd>Enter</kbd>|
+|Trigger the selected entry|<kbd>Enter</kbd> / |
```
neiljp, line 27: It looks like we need a special case to handle the documented keys here?

zulipterminal/config/keys.py
```diff
@@ -244,16 +249,16 @@ class KeyBinding(TypedDict):
         'excluded_from_random_tips': True,
         'key_category': 'searching',
     },
+    'EXECUTE_SEARCH': {
+        'keys': ['enter'],
+        'help_text': 'Execute the search and go to search results',
```
neiljp, line 255: Nit: We can address this on merging, but 'execute' is likely not the best user-facing language.

Perhaps "Submit search and browse results"?

I don't have the web app to compare to right now (as we discussed in the call).

neiljp:
@Niloth-p To clarify, this looks good to go except for space-handling in the hotkeys doc translation, so that looks like it'll need to be addressed if we're going to include that in this PR.

docs/hotkeys.md
Niloth-p: Oops with the squashing. Guess I didn't check the diff, my apologies.

zulipterminal/config/keys.py
Niloth-p: I checked the [list here](https://zulip.com/help/keyboard-shortcuts). They don't seem to use an equivalent for the search op. But, I did find an equivalent for ACTIVATE_BUTTON - "Use Enter to engage with elements".

I have updated the EXECUTE_SEARCH description to match your suggestion, and left the ACTIVATE_BUTTON as-is.

neiljp:
@Niloth-p Thanks for the adjustment - good to know it was the simple change you had previously that already fixed this and was just dropped by mistake :+1:

Merging now :tada:

Issue #678, opened 2020-06-08
make zuliprc a dot-file


alfonsrv:
Following the configuration of other terminal-configuration files, the default `zuliprc` file should be called `.zuliprc` without having to explicitly specify it using the `-c` switch.

Jellby:
or `~/.config/zulip/zuliprc`

neiljp:
This was discussed on chat.zulip.org in #**backend>storing client settings** in May, among other aspects.

From my perspective a dotfile would be good, but standardizing on `~/.config/zulip/zuliprc` would be better (or a good addition).

Aspects to consider: (possibly to update later)
* Should we support both `.zuliprc` and the xdg path? Which takes priority?
* What name should we use within `.config`? We should check with the main project.
* We could use the (tiny) `xdg` library to handle the `~/.config` aspect, though it's python3.6+ (we use 3.5+ currently)
* We should probably separate zuliprc files and one config file, to allow updating the connection details and config separately

neiljp:
Another benefit of some kind of `.zuliprc/` rather than a simpler file approach, is that this provides a namespace that ZT can use, including treating the filenames (or names without eg. a .zuliprc extension) as names for each server connection.

Since py3.5 is now EOL, this opens up the use of `xdg` after we migrate to py3.6.

zee-bit:
@neiljp I too think `~/.config/zulip/zuliprc` would be better approach than having a `.zuliprc`, mainly because having just one sub-directory instead of a number of dot files(for different profiles) and/or dot directories will reduce clutter when "hidden" files are displayed in GUI, and ofcourse pushing all user-configuration files to a dedicated sub-directory is always an attractive solution. :)


- Thus, I believe supporting just the `xdg` path would be better, because of the above reason. Having a `.zuliprc/` directory would still mean one extra clutter in the home directory?
- I am not completely sure what do you mean by "main project"(`zulip/zulip` maybe?), but we could use `~/.config/zulip-term/zuliprc`, hinting the name of the PyPi package that we currently go with?
- Since v3.5 is probably EOL, their shouldn't be any issue with any of these points.
- This seems to contradict my first point, but does sound like a better way to handle zuliprc. We could place all the config details in `~/.config/zulip-term/<config-file>` this would be common for all organizations. And, all the user-profile for organizations could be placed in `.zuliprc/` as you said, but deciding a specific name format should be the main challenge in my opinion. :)

neiljp:
[...]

PR #1503, opened 2024-05-21
Migration of zuliprc location to a config directory

Niloth-p:

Adds the following features:

- Support recognizing zuliprc from multiple select locations. 
  - `~/zuliprc`
  - the user's downloads directory
  - the user's config directory

- Move zuliprc creation into user's config directory.
The new path would look like `.config/zulip-terminal/zuliprc`.

- Create realm-wise zuliprc files in the configuration directory.
Each realm gets its own directory.
The new directory structure would look like:
```
<user's config directory>/zulip-terminal/
  org1-name/zuliprc
  org2-name/zuliprc
  zuliprc
```

- Add new positional argument - prefix of realm name to get its configuration.
New command examples:
`zulip-term zuli`
`zulip-term "full organization name"`

- Add -o argument to list realms with stored zuliprc.
```
$zulip-term -o
The following organizations are available:
  Zulip Community
  Another Organization
```

- Add -n argument to login to a realm, and create config.
```
$zulip-term -n
Please enter your credentials to login into your Zulip organization.
```

<details>
<summary>What does this PR do, and why? (expired)</summary>

Moves config location to 'XDG_CONFIG/zulip/'.
Makes zuliprc files dot files - .zuliprc.
Use independent org.zuliprc files for each organization.

Does not yet separate the zterm config from the api config.
Does not yet update the documents (README.md, FAQ.md, etc).
New tests not added.

The code is not yet meant for review.
This is only a draft to play around with the configuration in order to finalize what behavior is desired.
</details>

- [x] Discussed in **#zulip-terminal** in `Moving config`
- [ ] Fully fixes #
- [x] Partially fixes issue #678
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -285,7 +285,10 @@ def get_api_key(realm_url: str) -> Optional[Tuple[str, str, str]]:
     return None
 
 
-def fetch_zuliprc(zuliprc_path: str) -> None:
+def login_and_save(zuliprc_path: str) -> None:
+    """
+    Prompts the user for their login credentials and saves them to a zuliprc file.
+    """
     print(
         f"{in_color('red', f'zuliprc file was not found at {zuliprc_path}')}"
         f"\nPlease enter your credentials to login into your Zulip organization."
@@ -347,7 +350,7 @@ def parse_zuliprc(zuliprc_str: str) -> Dict[str, SettingData]:
     zuliprc_path = path.expanduser(zuliprc_str)
     while not path.exists(zuliprc_path):
         try:
-            fetch_zuliprc(zuliprc_path)
+            login_and_save(zuliprc_path)
         # Invalid user inputs (e.g. pressing arrow keys) may cause ValueError
         except (OSError, ValueError):
             # Remove zuliprc file if created.
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -11,6 +11,7 @@ import sys
 import traceback
 from enum import Enum
 from os import path, remove
+from pathlib import Path
 from typing import Dict, List, NamedTuple, Optional, Tuple
 
 import requests
@@ -28,6 +29,9 @@ from zulipterminal.core import Controller
 from zulipterminal.model import ServerConnectionFailure
 from zulipterminal.platform_code import detected_platform, detected_python_in_full
 from zulipterminal.version import ZT_VERSION
+
+
+HOME_PATH_ZULIPRC = str(Path.home() / "zuliprc")
 
 
 class ConfigSource(Enum):
@@ -136,7 +140,7 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
         "-c",
         action="store",
         help="config file downloaded from your zulip "
-        "organization (default: ~/zuliprc)",
+        f"organization (default: {HOME_PATH_ZULIPRC})",
     )
     parser.add_argument(
         "--theme",
@@ -312,7 +316,7 @@ def login_and_save(zuliprc_path: Optional[str]) -> str:
     # for adding "/api"
     realm_url = realm_url.rstrip("/")
     login_data = get_api_key(realm_url)
-    zuliprc_path = zuliprc_path or path.expanduser("~/zuliprc")
+    zuliprc_path = zuliprc_path or path.expanduser(HOME_PATH_ZULIPRC)
 
     while login_data is None:
         print(in_color("red", "\nIncorrect Email(or Username) or Password!\n"))
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -346,7 +346,12 @@ def _write_zuliprc(
         return f"{ex.__class__.__name__}: zuliprc could not be created at {to_path}"
 
 
-def parse_zuliprc(zuliprc_str: str) -> Dict[str, SettingData]:
+def resolve_to_valid_path(zuliprc_str: str) -> str:
+    """
+    Returns the path to a valid zuliprc file.
+    If none are found or the path provided is invalid, prompts the user to login
+    and returns the path to the created zuliprc file.
+    """
     zuliprc_path = path.expanduser(zuliprc_str)
     while not path.exists(zuliprc_path):
         try:
@@ -360,7 +365,10 @@ def parse_zuliprc(zuliprc_str: str) -> Dict[str, SettingData]:
         except EOFError:
             # Assume that the user pressed Ctrl+D and continue the loop
             print("\n")
+    return zuliprc_path
 
+
+def parse_zuliprc(zuliprc_path: str) -> Dict[str, SettingData]:
     mode = os.stat(zuliprc_path).st_mode
     is_readable_by_group_or_others = mode & (stat.S_IRWXG | stat.S_IRWXO)
 
@@ -458,6 +466,7 @@ def main(options: Optional[List[str]] = None) -> None:
         zuliprc_path = args.config_file
     else:
         zuliprc_path = "~/zuliprc"
+    zuliprc_path = resolve_to_valid_path(zuliprc_path)
 
     print(
         "Detected:"
--- a/zulipterminal/cli/run.py
+++ b/zulipterminal/cli/run.py
@@ -27,11 +27,16 @@ from zulipterminal.config.themes import (
 )
 from zulipterminal.core import Controller
 from zulipterminal.model import ServerConnectionFailure
-from zulipterminal.platform_code import detected_platform, detected_python_in_full
+from zulipterminal.platform_code import (
+    detected_platform,
+    detected_python_in_full,
+    xdg_config_home,
+)
 from zulipterminal.version import ZT_VERSION
 
 
 HOME_PATH_ZULIPRC = str(Path.home() / "zuliprc")
+CONFIG_PATH = str(xdg_config_home() / "zulip-terminal")
 
 
 class ConfigSource(Enum):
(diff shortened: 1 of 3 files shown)


zulipterminal/cli/run.py
```diff
@@ -49,7 +49,7 @@ class SettingData(NamedTuple):
 
 DOWNLOADED_PATH_ZULIPRC = str(downloads_file_path() / "zuliprc")
 HOME_PATH_ZULIPRC = str(Path.home() / "zuliprc")
-CONFIG_PATH_ZULIPRC = str(config_file_path() / "zulip-terminal" / "zuliprc")
```
Niloth-p: Changed names here.
Felt CONFIG_PATH_ZULIPRC was a more consistent name for the single zuliprc in config.
Feel ZULIP_CONFIG_PATH would be better when pointing to the directory. Could use CONFIG_PATH_ZULIP too.
Would be open to rebasing any name changes if the noise of this change is more important than the consistency of the name in the initial commits.

zulipterminal/cli/run.py
```diff
@@ -343,13 +349,19 @@ def _write_zuliprc(
...
         DOWNLOADED_PATH_ZULIPRC,
         HOME_PATH_ZULIPRC,
     ]
     valid_locations = [
         location for location in zuliprc_locations if Path(location).exists()
     ]
+    valid_locations.extend(
+        str(file) for file in Path(ZULIP_CONFIG_PATH).glob("zuliprc")
+    )
+    valid_locations.extend(
```
Niloth-p, line 385: The current implementation explicitly does not use `Path(ZULIP_CONFIG_PATH).glob("**/zuliprc")`, to avoid the situation (in later commits) where the realm has multiple zuliprc files inside it (in folders or as loose file). This implementation will only recognize the outermost zuliprc within a realm's directory, irrespective of the subdirectories' contents.

zulipterminal/cli/run.py
```diff
@@ -448,6 +448,30 @@ def list_themes() -> str:
...
+        matching_dirs_string = "\n".join(str(dir) for dir in matching_dirs)
+        exit_with_error(
+            f"Found multiple organizations starting with '{realm_name_prefix}':\n"
+            f"{matching_dirs_string}\n"
+        )
+    elif len(matching_dirs) == 0:
+        exit_with_error(
+            f"Could not find any organizations starting with '{realm_name_prefix}'"
+        )
+    matching_files = list(matching_dirs[0].glob("zuliprc"))
```
Niloth-p, line 509: If we used path.glob(**/zuliprc) in the previous commit, this is where we would need to add another case where len(matching_files) > 1.

zulipterminal/cli/run.py
```diff
@@ -448,6 +448,30 @@ def list_themes() -> str:
     )
 
 
+def path_to_realm_zuliprc(realm_name_prefix: str) -> str:
+    if len(realm_name_prefix) < 3:
```
Niloth-p, line 492: Assumption: There are no zulip realms whose name property contains 2 or fewer characters.
We could remove this case.

zulipterminal/cli/run.py
```diff
@@ -387,15 +398,21 @@ def check_for_default_zuliprc() -> str:
...
+            if zuliprc_str == ""
+            else path.expanduser(zuliprc_str)
+        )
+        if not new_realm
+        else ""
     )
     while zuliprc_path == "" or not path.exists(zuliprc_path):
         try:
-            zuliprc_path = fetch_zuliprc(zuliprc_path)
+            zuliprc_path = fetch_zuliprc(zuliprc_path, new_realm)
```
Niloth-p, line 415: Re. naming,
used 'realm' for all internal references,
and 'org' / 'organization' for all references visible to the user.

Niloth-p (commented):
Updated the PR description, to describe the new features.
Each item in that feature list is its own commit (in the same order), with preparatory commits in between (mentioned for easier reviewing).

I have not yet added tests in this branch.
The ones I had were too tightly tied to the functions that were introduced.
Planning to re-write them to directly test main([]) instead.

#### Reference: Testing cases:
Feel free to come up with any missing / edge cases.
- default call (without positional argument)
	- loose zuliprc file in config
	- zuliprc files in config inside realms directories
		- one realm
		- multiple realms
	- mix of both loose zuliprc and inside realms in config
	- several zuliprc files found error
	- no zuliprc in any locations
- positional argument
	- too few characters (<3)
	- no org found
	- several matching orgs
	- one matching org, but zuliprc is inside subdirectory
	- one matching org
	- full name instead of prefix
	- use along with --config-file option & -c options
- login newly into a few servers, test the created directory structure
- n option:
	- if configuration for that realm already exists
	- if a directory with that realm name exists, but it does not have a zuliprc file inside it
- o option:
	- no orgs found
	- several orgs found

zulipterminal/cli/run.py
```diff
@@ -322,11 +371,48 @@ def _write_zuliprc(
...
+    count = len(valid_locations)
+    if count > 1:
+        valid_locations_string = "\n".join(valid_locations)
+        exit_with_error(
+            f"Found multiple zuliprc files at:\n{valid_locations_string}\n"
+            "Please retry by specifying the path to your target zuliprc file."
+        )
+    elif count == 1:
+        return valid_locations[0]
+    return ""
```
rsashank, line 398: This implementation seems great! However, should we consider giving priority to `HOME_PATH_ZULIPRC` if multiple configuration files are found, since that was previously the location of zuliprc?

If users want to use a different configuration file instead of the default, they can specify a specific target zuliprc file to override the default. We can notify the config file used while loading ZT, similar to how we do with the settings.

Alternatively, instead of exiting with an error, we could include an index and prompt the user to choose which organization they would like to log into.

For example:

```shell
Choose which organization you'd like to log in to (Enter the number):
1: Zulip Community: chat.zulip.org
2: CTF Chat: ctf.zulipchat.com
3: Example: example.zulip.com
```

zulipterminal/cli/run.py
```diff
@@ -392,6 +478,40 @@ def list_themes() -> str:
...
+
+
+def path_to_realm_zuliprc(realm_name_prefix: str) -> str:
+    if len(realm_name_prefix) < 3:
+        exit_with_error("Organization name prefix must be at least 3 characters long")
+    matching_dirs = [
+        dir
+        for dir in Path(ZULIP_CONFIG_PATH).glob("*/")
+        if dir.name.startswith(realm_name_prefix)
+    ]
```
rsashank, line 498: Can we make this case sensitive?

Maybe:

```python
    matching_dirs = [
        dir
        for dir in Path(ZULIP_CONFIG_PATH).glob("*/")
        if dir.name.startswith(realm_name_prefix)
        or dir.name.lower().startswith(realm_name_prefix)
    ]
```

zulipterminal/cli/run.py
```diff
@@ -392,6 +478,40 @@ def list_themes() -> str:
...
+    if len(matching_dirs) > 1:
+        matching_dirs_string = "\n".join(str(dir) for dir in matching_dirs)
+        exit_with_error(
+            f"Found multiple organizations starting with '{realm_name_prefix}':\n"
+            f"{matching_dirs_string}\n"
+        )
```
rsashank, line 504: I'm not sure how feasible this is, but should we consider prompting the user to select which directory to log into from the matched directories?

zulipterminal/cli/run.py
```diff
@@ -392,6 +478,40 @@ def list_themes() -> str:
...
+    elif len(matching_dirs) == 0:
+        exit_with_error(
+            f"Could not find any organizations starting with '{realm_name_prefix}'"
+        )
```
rsashank, line 508: Maybe we could suggest using zulip-term -o to show the list of available organizations. Alternatively, we could display the list along with the error message?

rsashank (commented):
Great work, @Niloth-p! I manually tested it, and it works well. I added a few comments in some areas where I think it would be nice to consider changes, but those are more opinion-based.

This looks very useful!

rsashank:
Thanks for updating this, @Niloth-p! I'll leave it for the mentor and maintainers to review. I'm removing the buddy review label now.

zulipterminal/cli/run.py
```diff
@@ -452,6 +452,30 @@ def list_themes() -> str:
     )
 
 
+def path_to_realm_zuliprc(realm_name_prefix: str) -> str:
```
neiljp, line 496: I'll wait to see this updated for the new version first, but I expect this and later commits would be the mapping of aliases to zuliprcs instead.

zulipterminal/cli/run.py
```diff
@@ -305,8 +311,13 @@ def fetch_zuliprc(zuliprc_path: str) -> str:
...
         else "any of the following locations:\n  " + "\n  ".join(supported_locations)
     )
+    missing_zuliprc_text = (
+        ""
+        if new_realm
+        else f"{in_color('red', f'zuliprc file was not found at {locations_checked}')}"
+    )
     print(
-        f"{in_color('red', f'zuliprc file was not found at {locations_checked}')}"
+        f"{missing_zuliprc_text}"
```
neiljp, line 320: This will likely be simplified with the other changes I suggested?

zulipterminal/cli/run.py
```diff
@@ -157,6 +157,12 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
         action="store_true",
         help="list all the organizations that you have configurations for, and exit",
     )
+    parser.add_argument(
+        "-n",
+        "--new-organization",
+        action="store_true",
+        help="login to a new organization",
+    )
```
neiljp, line 165: It doesn't quite fit within the remit of the zuliprc movement work, but it'd be good to have `-c` work with creating a zuliprc file from username/password, which I don't think it does right now?

neiljp (commented):
@Niloth-p I hope this review helps get you back on track with this work :+1:

The discussion in the channel seems to have helped too :)

PR #1507, opened 2024-05-29
messages: Add tests for transform_content.

rsashank:
This update PR tests for the transform_content class method for `self.message_links` and `self.time_mentions`. The addition of these tests addresses feedback from issue #1455, which pointed out the lack of tests for these aspects. 

@neiljp said:
> It'd be good to see some tests for what ends up in self.code_snippets, though we don't seem to have tests for the others.

- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

- [] It contains test additions for any new behavior

### DIFF
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1,10 +1,12 @@
 from collections import OrderedDict, defaultdict
 from datetime import date
+from typing import Any, Dict, List, Tuple
 
 import pytest
 import pytz
 from bs4 import BeautifulSoup
 from pytest import param as case
+from pytest_mock import MockerFixture
 from urwid import Columns, Divider, Padding, Text
 
 from zulipterminal.config.keys import keys_for_command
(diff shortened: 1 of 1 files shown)


tests/ui_tools/test_messages.py
```diff
@@ -1593,6 +1593,96 @@ def test_transform_content(self, mocker, raw_html, expected_content):
...
+            ),
+            (
+                """
+                <p><a href="https://foo.com">https://foo.com</a></p>
+                """,
+                {"https://foo.com": ("https://foo.com", 1, False)},
+            ),
+            (
+                """
+                <p><a href="#narrow/stream/206-zulip-terminal/topic/announce">https://chat.zulip.zulip/#narrow/stream/206-zulip-terminal/topic/announce</a></p>
```
Niloth-p, line 1619: Why not chat.zulip.org? xD

tests/ui_tools/test_messages.py
```diff
@@ -1593,6 +1593,96 @@ def test_transform_content(self, mocker, raw_html, expected_content):
         rendered_text = Text(content)
         assert rendered_text.text == expected_content
 
+    @pytest.mark.parametrize(
+        "raw_html, expected_message_links",
+        [
+            (
+                """
+                <p><a href="https://github.com/zulip/zulip-terminal/pull/1">https://github.com/zulip/zulip-terminal/pull/1</a></p>
```
Niloth-p, line 1601: Should we have some link where the text is not the same as the link?
One of the first two?

tests/ui_tools/test_messages.py
```diff
@@ -1593,6 +1593,96 @@ def test_transform_content(self, mocker, raw_html, expected_content):
...
+                        "https://chat.zulip.zulip/#narrow/stream/206-zulip-terminal/topic/announce",
+                        1,
+                        True,
+                    )
+                },
+            ),
+        ],
+    )
+    def test_transform_content_message_links(
+        self, mocker, raw_html, expected_message_links
```
Niloth-p, line 1632: mocker can be removed.

Niloth-p (commented):
Same as the other PR,
- type annotations
- test ids

Nice work with the noqa exceptions :wink:

Nicely done!
I haven't looked into the scope and history behind this PR yet.
But the implementation lgtm.

tests/ui_tools/test_messages.py
rsashank: `SERVER_URL = "https://chat.zulip.zulip"` is defined on line 28 of `test_messages.py`, and I used it when calling `transform_content` in the test. Could be changed though, if needed :)

tests/ui_tools/test_messages.py
rsashank: Done :)

rsashank:
Thanks for helping out with the noqa exceptions and for the detailed review! :) @Niloth-p

mounilKshah (commented):
Thanks for working on this @rsashank 
The test cases considered for each of the functions seem appropriate and enough for testing the respective aspects of the `transform_content` function.

tests/ui_tools/test_messages.py
```diff
@@ -1585,14 +1587,120 @@ def test_keypress_EDIT_MESSAGE(
             # fmt: on
         ],
     )
-    def test_transform_content(self, mocker, raw_html, expected_content):
+    def test_transform_content(self, raw_html: str, expected_content: Any) -> None:
```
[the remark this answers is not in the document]
neiljp: This file is not currently type-checked, though it would be good to do so (there's an issue for it). Adding these right now are more like comments, which is fine but won't be tested or maintained, which is the downside - though if they keep passing manual review then they should help when we do apply mypy to these files.

Given the introduction of the similar transform_content tests, it would make sense to rename this test to define what set of cases it handles, so I'd consider it reasonable to update nearby lines like the types.

tests/ui_tools/test_messages.py
```diff
@@ -1585,14 +1587,120 @@ def test_keypress_EDIT_MESSAGE(
...
+                """,
+                {"https://github.com/zulip/zulip-terminal": ("test", 1, True)},
+            ),
+        ],
+        ids=[
+            "github_pr",
+            "external_link_no_footlink",
+            "internal_link",
+            "github_link_with_text",
+        ],
```
neiljp, line 1643: Let's have these ids inline.

Also the first is a github_pr (link) but it'd be useful to clarify what form it takes, ie. the link matches the text in that first case.

tests/ui_tools/test_messages.py
```diff
@@ -1585,14 +1587,120 @@ def test_keypress_EDIT_MESSAGE(
...
+            ),
+            (
+                """
+                <p><a href="https://foo.com">https://foo.com</a></p>
+                """,
+                {"https://foo.com": ("https://foo.com", 1, False)},
+            ),
+            (
+                """
+                <p><a href="#narrow/stream/206-zulip-terminal/topic/announce">https://chat.zulip.zulip/#narrow/stream/206-zulip-terminal/topic/announce</a></p>
```
neiljp, line 1621: Are these test cases all from real example messages, to indicate this is what we actually receive?

tests/ui_tools/test_messages.py
```diff
@@ -1585,14 +1587,120 @@ def test_keypress_EDIT_MESSAGE(
...
+                """,  # noqa: E501
+                [
+                    (
+                        "Fri, Aug 15 1947, 1:17 (IST)",
+                        "Original text was 1947-08-15T01:17:00+05:30",
+                    )
+                ],
+            ),
+        ],
+        ids=["test_date", "distant_future_date", "past_date"],
```
neiljp, line 1689: I'd suggest inlining these too.

neiljp (commented):
@rsashank This looks like an improvement to me :+1:

In addition to my points inline, I'd consider the case of no links/mentions, and multiple ones.

Did you see an improvement in numeric coverage compared to this rebased against `main`?

Historically we tested transform_content separately since it became a wrapper around soup2markup for quotes. These tests *could* be considered more complex tests for soup2markup, but transform_content seems fine too - the current soup2markup test is a simple rendering test only. One readability confusion in that regard is that all these tests are not near the soup2markup test (L676). 

This PR has an interesting connection to the muted user work; there I mentioned that we likely need to parse the message to get these fields in any case, which could suggest we do the transform completely separately (earlier) than the header calculation. We currently remake the entire messagebox, I recall, but storing the content earlier may make updating only the content/author easier for muting/unmuting events.

tests/ui_tools/test_messages.py
rsashank: `github_pr_link_matches_text`?

rsashank:
Updated this PR @neiljp! Thanks for the feedback

PR #1511, opened 2024-06-05
Fix tests failing on WSL systems.

rsashank:

Addressed issues with file path separators `\\` and GUI return codes specific to WSL environments.

Fixes the failing of these tests on WSL environments, these have been failing for a while now but I just got used to ignoring them, decided to look into them today and fix. 😓 

image

### Outstanding aspect(s)    

- [ ] 

- [x] Discussed in **#zulip-terminal** in [WSL specific test fails](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/WSL.20specific.20test.20fails/near/1817087)
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -548,7 +548,7 @@ def test_download_media(
         ("Linux", True, True, "xdg-open", "/path/to/media"),
         ("MacOS", True, True, "open", "/path/to/media"),
         ("WSL", True, True, "explorer.exe", "\\path\\to\\media"),
-        ("UnknownOS", True, False, "unknown-tool", "/path/to/media"),
+        ("UnknownOS", True, False, "invalid", "/path/to/media"),
     ],
     ids=[
         "Linux_os_user",
@@ -572,9 +572,13 @@ def test_process_media(
         MODULE + ".download_media", return_value=media_path
     )
     mocked_open_media = mocker.patch(MODULE + ".open_media")
-    mocker.patch(MODULE + ".PLATFORM", platform)
-    mocker.patch("zulipterminal.platform_code.PLATFORM", platform)
-    mocker.patch("zulipterminal.core.Controller.show_media_confirmation_popup")
+
+    # Mocking the PLATFORM variable in both platform_code and the main module
+    # to ensure consistency during tests that rely on platform detection.
+    mocker.patch("zulipterminal.platform_code" + ".PLATFORM", platform)
+    mocker.patch(
+        "zulipterminal.platform_code" + ".process_media_tool", return_value=tool
+    )
 
     process_media(controller, link)
 
@@ -603,28 +607,49 @@ def test_process_media_empty_url(
 
 
 @pytest.mark.parametrize(
-    "returncode, error",
+    "platform, returncode, tool, error",
     [
-        (0, []),
-        (
+        case("Linux", 0, "xdg-open", [], id="Linux:success"),
+        case("MacOS", 0, "open", [], id="MacOS:success"),
+        case("WSL", 1, "explorer.exe", [], id="WSL:success"),
+        case(
+            "Linux",
             1,
+            "xdg-open",
             [
                 " The tool ",
                 ("footer_contrast", "xdg-open"),
                 " did not run successfully" ". Exited with ",
                 ("footer_contrast", "1"),
             ],
+            id="Linux:error",
         ),
+        case(
+            "MacOS",
+            1,
+            "open",
+            [
+                " The tool ",
+                ("footer_contrast", "open"),
+                " did not run successfully" ". Exited with ",
+                ("footer_contrast", "1"),
+            ],
+            id="Mac:error",
+        ),
+        # NOTE: explorer.exe (WSL) always returns a non-zero exit code (1)
+        # so we do not test for it.
     ],
 )
 def test_open_media(
     mocker: MockerFixture,
+    platform: str,
     returncode: int,
+    tool: str,
     error: List[Any],
-    tool: str = "xdg-open",
     media_path: str = "/tmp/zt-somerandomtext-image.png",
 ) -> None:
     mocked_run = mocker.patch(MODULE + ".subprocess.run")
+    mocker.patch("zulipterminal.platform_code" + ".PLATFORM", platform)
     mocked_run.return_value.returncode = returncode
     controller = mocker.Mock()
 
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -573,6 +573,7 @@ def test_process_media(
     )
     mocked_open_media = mocker.patch(MODULE + ".open_media")
     mocker.patch(MODULE + ".PLATFORM", platform)
+    mocker.patch("zulipterminal.platform_code.PLATFORM", platform)
     mocker.patch("zulipterminal.core.Controller.show_media_confirmation_popup")
 
     process_media(controller, link)
--- a/tests/platform_code/test_platform_code.py
+++ b/tests/platform_code/test_platform_code.py
@@ -6,7 +6,7 @@ from zulipterminal.platform_code import (
     SupportedPlatforms,
     normalized_file_path,
     notify,
-    successful_GUI_return_code,
+    validate_GUI_exit_status,
 )
 
 
@@ -76,20 +76,25 @@ def test_notify_quotes(
 
 
 @pytest.mark.parametrize(
-    "platform, expected_return_code",
+    "platform, return_code, expected_return_value",
     [
-        ("Linux", 0),
-        ("MacOS", 0),
-        ("WSL", 1),
+        ("Linux", 0, "success"),
+        ("MacOS", 0, "success"),
+        ("WSL", 1, "success"),
+        ("Linux", 1, "failure"),
+        ("MacOS", 1, "failure"),
+        # NOTE: explorer.exe (WSL) always returns a non-zero exit code (1),
+        # so we do not test for it.
     ],
 )
-def test_successful_GUI_return_code(
+def test_validate_GUI_exit_status(
     mocker: MockerFixture,
     platform: SupportedPlatforms,
-    expected_return_code: int,
+    return_code: int,
+    expected_return_value: str,
 ) -> None:
     mocker.patch(MODULE + ".PLATFORM", platform)
-    assert successful_GUI_return_code() == expected_return_code
+    assert validate_GUI_exit_status(return_code) == expected_return_value
 
 
 @pytest.mark.parametrize(
--- a/zulipterminal/platform_code.py
+++ b/zulipterminal/platform_code.py
@@ -89,18 +89,19 @@ def notify(title: str, text: str) -> str:
     return ""
 
 
-def successful_GUI_return_code() -> int:  # noqa: N802 (allow upper case)
+def validate_GUI_exit_status(  # noqa: N802 (allow upper case)
+    exit_status: int,
+) -> Literal["success", "failure"]:
     """
     Returns success return code for GUI commands, which are OS specific.
     """
-    # WSL uses GUI return code as 1. Refer below link to know more:
-    # https://stackoverflow.com/questions/52423031/
-    # why-does-opening-an-explorer-window-and-selecting-a-file-through-pythons-subpro/
-    # 52423798#52423798
+    # NOTE: WSL (explorer.exe) always returns a non-zero exit code (1) for GUI commands.
+    # This is a known issue. Therefore, we consider it a success if the exit code is 1.
+    # For more information, see: https://github.com/microsoft/WSL/issues/6565
     if PLATFORM == "WSL":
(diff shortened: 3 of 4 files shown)


tests/helper/test_helper.py
```diff
@@ -539,6 +540,9 @@ def test_download_media(
     ) = media_path
     controller = mocker.Mock()
 
+    if PLATFORM == "WSL":
+        media_path = media_path.replace("/", "\\")
+
```
rsashank, line 545: I'm not sure if I can import PLATFORM into tests, but this seemed like the only way to address the difference in file paths in WSL.

tests/helper/test_helper.py
```diff
@@ -604,14 +608,14 @@ def test_process_media_empty_url(
...
+            0 if PLATFORM == "WSL" else 1,
             [
                 " The tool ",
                 ("footer_contrast", "xdg-open"),
                 " did not run successfully" ". Exited with ",
-                ("footer_contrast", "1"),
+                ("footer_contrast", "0" if PLATFORM == "WSL" else "1"),
             ],
         ),
     ],
```
rsashank, line 621: This test didn't consider the different return codes in WSL, as given in `platform_code.py` line 92.

```python
def successful_GUI_return_code() -> int:  # noqa: N802 (allow upper case)
    """
    Returns success return code for GUI commands, which are OS specific.
    """
    # WSL uses GUI return code as 1. Refer below link to know more:
    # https://stackoverflow.com/questions/52423031/
    # why-does-opening-an-explorer-window-and-selecting-a-file-through-pythons-subpro/
    # 52423798#52423798
    if PLATFORM == "WSL":
        return 1

    return 0```

**Comment by Niloth-p** [comment]
Nice work! Thanks for addressing those.
The motivation and code LGTM.
As discussed, this is not something that I can run and test though, as I don't use WSL.

**Comment by neiljp** [comment]
@rsashank I started a review, but will leave that pending until I hear back to confirm the implementation details I noted in #zulip-terminal.

File: tests/helper/test_helper.py (hunk 1 above)
Comment (rsashank) [reply tests/helper/test_helper.py:545]: To further clarify, based on a discussion on CZO, the issue stems from the `normalized_path` function used in the `download_media` method. The `media_path` assertion doesn't account for the WSL-specific normalized path.

File: tests/helper/test_helper.py (hunk 2 above)
Comment (rsashank) [reply tests/helper/test_helper.py:621]: `explorer.exe` seems to always return a non-zero exit code, regardless of whether the task is successful.

More information could be found on this [issue](https://github.com/microsoft/WSL/issues/6565).

File: tests/helper/test_helper.py (hunk 1 above)
Comment (neiljp) [reply tests/helper/test_helper.py:545]: For a conditional in a test, the cleaner way to handle this would normally be to parametrize the test (over PLATFORM), though note you'd want a third path for the `expected_return_value` or similar output parameter, since it doesn't correspond to `url` or `media_path` in every case (though the latter may not be representative - see below re confirming).

The alternative to this conditional/parametrizing solution, which may also be clearer in the long term, is related to how (as we confirmed in the channel) the path returned is the path used for *opening* the temporary file, not the path used for *saving* it. So it would be clearer to refactor the code before fixing the tests, such that the `normalized_file_path` occurs outside of this function, and this function would be platform-independent to test.

**Review by neiljp — COMMENTED** [review]
@rsashank Thanks for finding these problems and working on solutions - we don't have WSL in CI (GitHub PR tests), so some of these issues are not always obvious, and not everything can be tested in CI in any case.

Following the discussion in the channel, I've adjusted some comments, though the same general gist of my suggestions remains the same, ie. that refactoring slightly first would be preferable, and may allow further subsequent refactoring which would make everything a lot easier to read.

Due to limited resources for testing on each platform, in general its better if we can isolate as much into platform-code as we can, during times while people are available to test it manually. That applies to this code in `helper`, as well as anywhere else we explicitly rely upon PLATFORM. That allows the platform-dependent code to be manually-tested well, and the interface between that and the application to be as clean as possible, as well as tests like these to be less fragile :)

File: tests/helper/test_helper.py (hunk 1 above)
Comment (rsashank) [reply tests/helper/test_helper.py]: Moved the `normalized_file_path` to be called in `process_media` 👍 
 Clear explanation, thanks!

File: tests/helper/test_helper.py (hunk 2 above)
Comment (rsashank) [reply tests/helper/test_helper.py]: > we can refactor the function you quoted, such that it returns some value, depending upon a passed-in return code. This would move the comparison, and hence these explicit error codes, into a platform-dependent location.

Implemented this approach, thanks for the suggestion! Also will look into refactoring some `open_media` logic to a platform-specific location in another PR.

File: tests/platform_code/test_platform_code.py [hunk 3]
```diff
@@ -76,20 +76,31 @@ def test_notify_quotes(
...
     [
-        ("Linux", 0),
-        ("MacOS", 0),
-        ("WSL", 1),
+        ("Linux", 0, "success"),
+        ("MacOS", 0, "success"),
+        ("WSL", 1, "success"),
+        ("Linux", 1, "failure"),
+        ("MacOS", 1, "failure"),
+        # NOTE: WSL always returns a non-zero exit code (1), so we do not test for it.
```
neiljp, line 86: As per another comment, we should test for this, even if it results in the same response - we're testing the function, not the system behavior, in this test.

neiljp (commented):
@rsashank I hoped to merge this fully but ended up leaving a few comments aimed primarily at the second commit - I've picked off the first commit and merged it already, after making the change I indicated in the comment re which item to patch.

Hopefully that first commit reduces the failing tests you're finding regularly, and the other commit will be quick to follow :)

tests/helper/test_helper.py
```diff
@@ -573,6 +573,7 @@ def test_process_media(
     )
     mocked_open_media = mocker.patch(MODULE + ".open_media")
     mocker.patch(MODULE + ".PLATFORM", platform)
+    mocker.patch("zulipterminal.platform_code.PLATFORM", platform)
```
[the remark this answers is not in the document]
rsashank: While testing locally in my WSL environment, the platform detection incorrectly identifies the platform as WSL when either one of the patches is missing. This causes the tests to enter the if PLATFORM == 'WSL' condition and change the path.

So, I had to mock the PLATFORM value in both the platform_code and helper.

I'm not entirely sure why this is happening, but patching both seems to ensure the consistency for all platforms that are being tested.
image

zulipterminal/platform_code.py
```diff
@@ -89,18 +89,19 @@ def notify(title: str, text: str) -> str:
...
-    # WSL uses GUI return code as 1. Refer below link to know more:
-    # https://stackoverflow.com/questions/52423031/
-    # why-does-opening-an-explorer-window-and-selecting-a-file-through-pythons-subpro/
-    # 52423798#52423798
+    # NOTE: WSL always returns a non-zero exit code (1) for GUI commands.
+    # This is a known issue. Therefore, we consider it a success if the exit code is 1.
+    # For more information, see: https://github.com/microsoft/WSL/issues/6565
     if PLATFORM == "WSL":
-        return 1
+        return "success" if exit_status == 1 else "failure"
```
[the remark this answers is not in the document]
rsashank: Thank you for pointing this out. I've removed the mention of "failure" handling.

The linked issue specifically refers to explorer.exe, and since validate_GUI_exit_status only deals with explorer.exe, I believe it should be fine to document here. What do you say?

zulipterminal/platform_code.py
```diff
@@ -89,18 +89,19 @@ def notify(title: str, text: str) -> str:
     return ""
 
 
-def successful_GUI_return_code() -> int:  # noqa: N802 (allow upper case)
+def validate_GUI_exit_status(  # noqa: N802 (allow upper case)
+    exit_status: int,
+) -> str:
```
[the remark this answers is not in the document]
rsashank: Got it!

tests/platform_code/test_platform_code.py
```diff
@@ -76,20 +76,31 @@ def test_notify_quotes(
...
+    ],
+    ids=[
+        "Linux_0_return_code",
+        "MacOS_0_return_code",
+        "WSL_1_return_code",
+        "Linux_1_return_code",
+        "MacOS_1_return_code",
     ],
```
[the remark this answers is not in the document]
rsashank: Got it.

tests/helper/test_helper.py
```diff
@@ -603,28 +603,53 @@ def test_process_media_empty_url(
...
     mocker: MockerFixture,
+    platform: str,
     returncode: int,
+    tool: str,
     error: List[Any],
-    tool: str = "xdg-open",
     media_path: str = "/tmp/zt-somerandomtext-image.png",
 ) -> None:
     mocked_run = mocker.patch(MODULE + ".subprocess.run")
+    mocker.patch("zulipterminal.platform_code.PLATFORM", platform)
```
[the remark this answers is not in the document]
rsashank: Doesn't seem to work on my local WSL environment - trying to figure out why :) 

I need to have both patched.

tests/helper/test_helper.py
```diff
@@ -603,28 +603,53 @@ def test_process_media_empty_url(
...
+            1,
+            "open",
+            [
+                " The tool ",
+                ("footer_contrast", "open"),
+                " did not run successfully" ". Exited with ",
+                ("footer_contrast", "1"),
+            ],
+        ),
+        # NOTE: WSL always returns a non-zero exit code (1), so we do not test for it.
```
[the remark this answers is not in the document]
rsashank: True, explorer.exe (WSL). Works?

tests/platform_code/test_platform_code.py
rsashank: I removed the failure handling entirely since it's considered an abnormality that might be addressed in the future. `explorer.exe` seems to be intended to return 1 for failure and 0 for success, so returning 0 for a failure didn't make sense. 

This could be updated if WSL fixes this, what do you think?

tests/helper/test_helper.py
```diff
@@ -603,28 +603,53 @@ def test_process_media_empty_url(
...
+    ],
+    ids=[
+        "Linux_os_user",
+        "Mac_os_user",
+        "WSL_os_user",
+        "Linux_os_error",
+        "Mac_os_error",
```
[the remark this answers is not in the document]
rsashank: Got it!

rsashank:
Updated this PR, thanks for the feedback @neiljp!

tests/helper/test_helper.py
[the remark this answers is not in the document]
neiljp: Thanks for this feedback! This seems likely since we still use PLATFORM in both locations for this :/

This appears to be the only outstanding change in this commit after a rebase, since I merged the first commit previously, so the commit text is currently inaccurate at that point. It'd also be useful to group and comment the two lines, if they appear to be necessary to pass the tests, as well as update the commit.

This is a 'test bugfix' commit, specific to WSL, essentially?

tests/helper/test_helper.py
```diff
@@ -603,28 +603,49 @@ def test_process_media_empty_url(
...
         ),
+        case(
+            "MacOS",
+            1,
+            "open",
+            [
+                " The tool ",
+                ("footer_contrast", "open"),
+                " did not run successfully" ". Exited with ",
+                ("footer_contrast", "1"),
```
neiljp, line 631: Reading this again, a reasonable PR after this would be to not output the error code, but provide more detailed output somewhere for followup later, a little like we do with the crash or threading errors. That, or try and diagnose further why the tool did not run properly.

We could also possibly provide a 'ran successfully' message in the footer for opening, but I've not looked at the code flow for that, and whether it would eg. overwrite other messages too rapidly?

neiljp (commented):
@rsashank I'd like to get this merged to firm up the WSL behavior - while someone is testing it :)

Regards to the platform isolation, placing the remaining PLATFORM-dependent functionality in platform_code.py would be preferable and ease testing overall, and could be another refactor to explore in addition to the followup I mentioned in a comment.

tests/helper/test_helper.py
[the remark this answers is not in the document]
rsashank: Grouped them together and added a comment, also changed commit message.

Yes, it's a test bugfix specific to WSL.

tests/helper/test_helper.py
rsashank: Understood 👍 

Wouldn't diagnosing why the tool did not run properly be platform dependent though?

rsashank:
@neiljp  I've updated this PR—thank you!

I added another commit to move the platform-based conditions in the helper function to platform_code.py.

Can you elaborate on this a bit?
> Reading this again, a reasonable PR after this would be to not output the error code, but provide more detailed output somewhere for followup later, a little like we do with the crash or threading errors. That, or try and diagnose further why the tool did not run properly.

PR #1514, opened 2024-06-06
Break down the GO_BACK command into separate commands

Niloth-p:
Breaks GO_BACK command into the following commands:
- CLEAR_SEARCH (clears search box and search list of current panel)
- EXIT_COMPOSE (clears compose box state and closes it)
- EXIT_POPUP (closes the currently open popup)

This would enable using more specific help texts, categories and context.
And clarifies the purposes of the key better for users and developers.

I've kept these as separate commits for the sake of readability. Introducing each command in its own commit, then finally removing GO_BACK command.
I could break #1497 into commits similar to this one for readability, if necessary.
I've kept this as a separate PR as it's an independent change, even though the motivation is shared.

- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [x] Is a follow-up to work in PR #1497
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -46,12 +46,6 @@ KEY_BINDINGS: Dict[str, KeyBinding] = {
         'help_text': 'Show/hide About Menu',
         'key_category': 'general',
     },
-    'GO_BACK': {
-        'keys': ['esc'],
-        'help_text': 'Go back',
-        'excluded_from_random_tips': False,
-        'key_category': 'general',
-    },
     'OPEN_DRAFT': {
         'keys': ['d'],
         'help_text': 'Open draft message saved in this session',
@@ -186,6 +180,11 @@ KEY_BINDINGS: Dict[str, KeyBinding] = {
     'NARROW_MESSAGE_RECIPIENT': {
         'keys': ['meta .'],
         'help_text': 'Narrow to compose box message recipient',
+        'key_category': 'msg_compose',
+    },
+    'EXIT_COMPOSE': {
+        'keys': ['esc'],
+        'help_text': 'Exit message compose box',
         'key_category': 'msg_compose',
     },
     'TOGGLE_NARROW': {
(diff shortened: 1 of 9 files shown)


zulipterminal/config/keys.py
```diff
@@ -172,6 +171,11 @@ class KeyBinding(TypedDict):
         'help_text': 'Narrow to compose box message recipient',
         'key_category': 'msg_compose',
     },
+    'EXIT_COMPOSE': {
+        'keys': ['esc'],
+        'help_text': 'Stop composing a message and close text box',
+        'key_category': 'msg_compose',
+    },
```
rsashank, line 178: NIT: Would it be better to keep this simple and just write it as `Exit message compose box`?

zulipterminal/config/keys.py
```diff
@@ -244,6 +248,11 @@ class KeyBinding(TypedDict):
         'excluded_from_random_tips': True,
         'key_category': 'searching',
     },
+    'CLEAR_SEARCH': {
+        'keys': ['esc'],
+        'help_text': 'Clear search in current panel',
+        'key_category': 'searching',
+    },
```
rsashank, line 255: Since we're aiming at contextual help messages, should we be using the word `panel`?  It feels more like code terminology and might not be useful as a help message.

rsashank:
LGTM! The intention seems clear since we're aiming for contextual help.

Just a note: Either this PR or #1442 will need to handle conflicts in a rebase, depending on which gets merged first.

zulipterminal/config/keys.py
Niloth-p: Hmm.
Found mentions of 'panel' in the user facing documentation.
What other word do you think would fit better in this context?

Niloth-p: I found mentions of 'panel' in the user-facing documentation,
and as we discussed, it's hard to come up with suitable alternatives,
so I've left it as 'panel'.

zormit (approved):
This is pretty nicely structure and looks good to me. Thank you!

neiljp:
@Niloth-p I agree with @zormit - this is clear and easy to read :+1:

The changes I pushed back with were only structural/text (the code overall is identical):
- squash the last commit with the smaller of the two later commits, to avoid a state where we have a menu entry that does nothing
- adjust some commit text (these change the UI, so aren't strictly refactors
- added some file names in the commit titles

PR #1515, opened 2024-06-07
Add new help categories to remove parentheses (context) from help texts

Niloth-p:
Adds the following new help categories to remove the corresponding context-suffixes (in parentheses) from help texts:
- Stream information ("from stream information" suffixes)
- Message information ("from message information" suffixes)
- User list actions ("from user list" suffixes)

That removes parantheses from all the help texts.

Each of the categories are added in their own commit, for readability.
A new command entry `NARROW_TO_USER_PM` is also added to the user list actions.

### Outstanding aspect(s)
This PR is about removing the contexts mentioned in parentheses as suffixes. And it removes all of them.
But, it also introduces new parentheses for the help category names itself.

I went with this because
1. There is not much repetition of the content in the parentheses
2. Length constraints apply to help texts, they don't wrap. But, help category names do wrap.
3. They add some needed clarity, on when/where the help category's hotkeys will work.

Feedback welcome!

- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    
The information popup categories:
image - info popups
The user list actions category:
image - user list actions

### DIFF
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -314,16 +314,16 @@ KEY_BINDINGS: Dict[str, KeyBinding] = {
     },
     'STREAM_MEMBERS': {
         'keys': ['m'],
-        'help_text': 'Show/hide stream members (from stream information)',
+        'help_text': 'Show/hide stream members',
         'excluded_from_random_tips': True,
-        'key_category': 'stream_list',
+        'key_category': 'stream_info',
     },
     'COPY_STREAM_EMAIL': {
         'keys': ['c'],
         'help_text':
-            'Copy stream email to clipboard (from stream information)',
+            'Copy stream email to clipboard',
         'excluded_from_random_tips': True,
-        'key_category': 'stream_list',
+        'key_category': 'stream_info',
     },
     'REDRAW': {
         'keys': ['ctrl l'],
@@ -456,6 +456,10 @@ HELP_CATEGORIES = {
     "compose_box": "Writing a message",
     "editor_navigation": "Editor: Navigation",
     "editor_text_manipulation": "Editor: Text Manipulation",
+    "stream_info": (
+        f"Stream information (press {KEY_BINDINGS['STREAM_INFO']['keys'][0]}"
+        f" to view info of a stream)"
+    ),
 }
 
 ZT_TO_URWID_CMD_MAPPING = {
--- a/zulipterminal/config/keys.py
+++ b/zulipterminal/config/keys.py
@@ -337,8 +337,16 @@ KEY_BINDINGS: Dict[str, KeyBinding] = {
     },
     'USER_INFO': {
         'keys': ['i'],
-        'help_text': 'Show/hide user information (from users list)',
-        'key_category': 'general',
+        'help_text': 'Show/hide user information',
+        'key_category': 'user_list',
+    },
+    'NARROW_TO_USER_PM': {
+        # Added to clarify functionality of button activation,
+        # as opposed to opening user profile or other effects.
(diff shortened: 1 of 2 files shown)


rsashank (commented):
LGTM. Seems clear :)

zulipterminal/config/keys.py
```diff
@@ -317,8 +317,13 @@ class KeyBinding(TypedDict):
     },
     'USER_INFO': {
         'keys': ['i'],
-        'help_text': 'Show/hide user information (from users list)',
-        'key_category': 'general',
+        'help_text': 'Show/hide user information',
+        'key_category': 'user_list',
+    },
+    'NARROW_TO_USER_PM': {
```
zormit, line 323: This one is a bit tricky because we don't actually use this specific key, but if I understand it correctly the functionality to narrow to a DM happens through the fact that the user list is composed of `UserButton`s, which is a `TopButton`, which maps `ENTER` to `activate` [here](https://github.com/zormit/zulip-terminal/blob/e5e9010850ab4c678890686e3707d7e3e842385e/zulipterminal/ui_tools/buttons.py#L123).

I would suggest to leave a comment here to explain this in fewer words than I do, in order to prevent confusing future readers of this code. Maybe something like "this key is only for documenting the functionality, but the functionality happens through a generic ENTER"?

zormit (approved):
Generally looks good & clear to me!

I have left one PR comment of something that could be clarified by adding a comment in the code. Apart from that all is approved, so I don't mind if the contributor fixes this or the maintainer adds the comment when merging.

Niloth-p:
Rebased onto main, added a comment as per feedback.

zulipterminal/config/keys.py
zormit: @Niloth-p it seems you have now added a comment for `USER_INFO`, which might make sense as well to some extent, but that one is actually used to retrieve its key, e.g. in `zulipterminal/ui_tools/buttons.py`.

Whereas `NARROW_TO_USER_PM` is not used anywhere and just lives here for the documentation/help-text.

Niloth-p: Haha, thanks for the catch. I can't believe I accidentally wrote the comment for the wrong command.

> which might make sense as well to some extent

No, the comment was intended for this command, I wrote it in another text editor and I pasted it into the wrong command. The help text mentions that it clarifies that it does not open the user profile info, and I pasted it for the user info command lol.

Thank you for the save!

Niloth-p:
Rebased and pushed.

We just have 2 keys for the 2 categories 'Stream message info actions' and 'User list actions'. As discussed, those make it tiny categories. So,
- we could either discuss the alternative options that I previously suggested in the CZO topic,
- or leave this PR until more hotkeys are added to these categories,
- or we could push these changes but put these tiny categories at the bottom of the help menu.

neiljp:
@Niloth-p Merging this now; thanks for the earlier review zormit :+1:

My main concern is the location of the resulting extra categories (and their size, as discussed before), but I think this does read more cleanly :)

zulipterminal/config/keys.py
```diff
@@ -460,6 +460,10 @@ class KeyBinding(TypedDict):
         f"Stream information (press {KEY_BINDINGS['STREAM_INFO']['keys'][0]}"
         f" to view info of a stream)"
     ),
+    "msg_info": (
+        f"Message information (press {KEY_BINDINGS['MSG_INFO']['keys'][0]}"
+        f" to view info of a message)"
+    ),
```
neiljp, line 475: Ah, I meant to point out here that the f-string should not be necessary for the second string :)

PR #1516, opened 2024-06-07
Group narrowing commands and improve descriptions

Niloth-p:
- Groups commands that switch narrows
- Improves their help descriptions

### Outstanding aspect(s)
Feedback on the following are needed:
- name of the chosen help category
- order of the commands within the narrowing category
- help descriptions

- [x] Discussed in **#zulip-terminal** in [`re-categorization`](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Re-Categorization.20of.20Hotkeys/near/1793697)
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes
image

### DIFF
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -32,8 +32,8 @@
 | :--- | :---: |
 |Narrow to the stream of the current message|<kbd>s</kbd>|
 |Narrow to the topic of the current message|<kbd>S</kbd>|
-|Narrow to compose box message recipient|<kbd>Meta</kbd> + <kbd>.</kbd>|
 |Narrow to a topic/direct-chat, or stream/all-direct-messages|<kbd>z</kbd>|
+|Narrow to compose box message recipient|<kbd>Meta</kbd> + <kbd>.</kbd>|
 |Narrow to all messages|<kbd>a</kbd> / <kbd>Esc</kbd>|
 |Narrow to all direct messages|<kbd>P</kbd>|
 |Narrow to all starred messages|<kbd>f</kbd>|
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -30,14 +30,14 @@
 ## Switching Messages View
 |Command|Key Combination|
 | :--- | :---: |
-|Narrow to the stream of the current message|<kbd>s</kbd>|
-|Narrow to the topic of the current message|<kbd>S</kbd>|
-|Narrow to a topic/direct-chat, or stream/all-direct-messages|<kbd>z</kbd>|
-|Narrow to compose box message recipient|<kbd>Meta</kbd> + <kbd>.</kbd>|
-|Narrow to all messages|<kbd>a</kbd> / <kbd>Esc</kbd>|
-|Narrow to all direct messages|<kbd>P</kbd>|
-|Narrow to all starred messages|<kbd>f</kbd>|
-|Narrow to messages in which you're mentioned|<kbd>#</kbd>|
+|View the stream of the current message|<kbd>s</kbd>|
+|View the topic of the current message|<kbd>S</kbd>|
+|Zoom in/out the message's conversation context|<kbd>z</kbd>|
+|Switch message view to the compose box target|<kbd>Meta</kbd> + <kbd>.</kbd>|
+|View all messages|<kbd>a</kbd> / <kbd>Esc</kbd>|
+|View all direct messages|<kbd>P</kbd>|
+|View all starred messages|<kbd>f</kbd>|
+|View all messages in which you're mentioned|<kbd>#</kbd>|
 |Next unread topic|<kbd>n</kbd>|
 |Next unread direct message|<kbd>p</kbd>|
 
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -26,6 +26,14 @@
 |Scroll down|<kbd>PgDn</kbd> / <kbd>J</kbd>|
 |Go to bottom / Last message|<kbd>End</kbd> / <kbd>G</kbd>|
 |Trigger the selected entry|<kbd>Enter</kbd> / <kbd>Space</kbd>|
+
+## Switching Messages View
+|Command|Key Combination|
+| :--- | :---: |
+|Narrow to the stream of the current message|<kbd>s</kbd>|
+|Narrow to the topic of the current message|<kbd>S</kbd>|
+|Narrow to compose box message recipient|<kbd>Meta</kbd> + <kbd>.</kbd>|
+|Narrow to a topic/direct-chat, or stream/all-direct-messages|<kbd>z</kbd>|
 |Narrow to all messages|<kbd>a</kbd> / <kbd>Esc</kbd>|
 |Narrow to all direct messages|<kbd>P</kbd>|
 |Narrow to all starred messages|<kbd>f</kbd>|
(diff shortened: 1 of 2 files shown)


docs/hotkeys.md
```diff
@@ -34,11 +34,11 @@
...
-|Narrow to a topic/direct-chat, or stream/all-direct-messages|<kbd>z</kbd>|
-|Narrow to all messages|<kbd>a</kbd> / <kbd>Esc</kbd>|
-|Narrow to all direct messages|<kbd>P</kbd>|
-|Narrow to all starred messages|<kbd>f</kbd>|
-|Narrow to messages in which you're mentioned|<kbd>#</kbd>|
+|Zoom in/out the message's conversation context|<kbd>z</kbd>|
+|Go to your combined feed|<kbd>a</kbd> / <kbd>Esc</kbd>|
+|Go to your direct message feed|<kbd>P</kbd>|
+|Go to your starred messages|<kbd>f</kbd>|
+|Go to messages in which you're mentioned|<kbd>#</kbd>|
```
rsashank, line 41: I'm not entirely sure about the language we should use here, but should we consider using "Go to" instead of "Narrow" for lines 34-36 as well? To me, "Go to" seems clearer than "Narrow." However, this is just my opinion, so we'll need to wait for feedback from Neil and the mentors.

The rewording of the zoom in/out feature seems great, though!

rsashank:
LGTM 👍

docs/hotkeys.md
Niloth-p: Thanks for the review!

Ah, good point! Yes, I was considering the same, but I wasn't sure because those narrowing commands kind of use the current context and don't switch to something completely different or a named narrow, like the menu buttons do. But, it's probably just me.
And that's also why line 41 doesn't feel like a great fit either, because it's not a named narrow.
Thank you for sharing your thoughts!

Zoom in/out credits: Zulip webapp

neiljp: For context, historically - at least in ZT - we've had confusion over communicating 'a narrow' (message view; noun) and 'narrowing' (switching to a message view; verb).

Most of this likely comes from the phrasing in the code, and also how the API defines a narrow as a filter of the messages you can see: https://zulip.com/api/construct-narrow.
(note we currently use an old API style for narrows, which needs to be refactored to support the documented style which uses a dict)

You can also see this in our user-facing tutorial: https://github.com/zulip/zulip-terminal/blob/main/docs/getting-started.md#Narrowing

So, there's potentially some small follow-up we could do to clarify this for end-users, if we're switching to 'message view' for the noun form.

Regarding the specific verb prefix selection (go-to vs narrow):
- ideally it would be as concise as possible, as well as consistent
- some of the contextual grouping we have was originally intended to make it easier to understand based on the category titles, but since the descriptions can appear without them, then we can't assume their presence and simplify that much
- 'go' is used a lot in the web app descriptions for 'narrowing' (https://zulip.com/help/keyboard-shortcuts)
- the web app or docs actually don't seem to use narrow for either form? (outside of API)
- narrowing can change keyboard focus, but we currently use 'go' for the keyboard movement, so that could be confusing when really it's specifically changing the message view

neiljp: Re `z`, this is a simple toggleable hotkey in ZT, which we introduced before the web app switched to using just `s` for doing *almost* the same. In web it doesn't apply to DMs, so it's not really a 'safe' toggle between views - and I'm not sure if `shift s` (`S`) still works. I'm inclined to suggest Toggle could be a good phrase here.

'Zoom in/out' is definitely cleaner, and relates to why we chose `z` in the first place, but the challenge with the phrasing here has always been how to cleanly describe the different contexts. 'Conversation' seems to be the term used for 'most zoomed in' view at the moment, but DM feed and channel feeds are not quite as clear cut, but might be useful?

(for the reasons above, `z` should be right next to `s`/`S` in the help)

neiljp (commented):
@Niloth-p Thanks for this thoughtful work :+1:
(and @rsashank's review :+1)

Following up from the changes to eg. have `Help Menu` and `About Menu` capitalized similarly, we could also consider doing similar changes to parts of the UI like `All Messages` and `Direct Messages`. These could be simple capitalization, or possibly subsequently styled versions, to highlight that they refer to specific defined items.

As per most of this work, the challenge is picking the correct terms and phrases, and using them consistently. That'll likely be easier as we go on, and we may want to discuss those specific terms, including eg. for referring to popups/menus/...

Niloth-p:
I've made the following changes:

- [x] Re-ordered to `s`, `S`, `z`, `Meta .`
- [x] User-facing category name: "Switch message view" -> "Switching messages view"
- [x] Internal category name: "narrow" -> "narrowing"
- [x] The previous PR version used "Go to", this version uses "View" and "View all". Please refer to the commit description for details.

Follow-ups in separate commits / PRs:
- All Messages -> Combined Feed everywhere
	- All DMs -> your DM feed or equivalent
- Replace "narrow" with "message view" or others in user-facing documentation like the user tutorial.
- Consider casefix changes to parts of the UI like - All Messages and Direct Messages
I didn't do this as part of the 2nd commit because it makes sense grammatically, without marking them as proper nouns. We could do that when we start using "Feed" terminology?
- Add a new key binding mentioning that 'Enter' activates narrowing buttons
I don't see this help text properly fitting in any category that we have at the moment, and general is just too broad.
We could instead not add this to a category, but just add it to a context, following the discussions in CZO topics `Restructure Key Bindings?` and `Supporting Help Multi-categories #T1519`. 
We have the ACTIVATE_BUTTON key binding at the moment.
- I've used a mention of 'PM' in the help text. I know that we ideally want to use DMs and key it in with ZFL, but hopefully we can handle that as a follow-up PR too, as we're currently using PM everywhere in the code base.
- Return to `n` & `p` later after fine-tuning the View/Narrow/Navigate nuances.

Please let me know if I've missed anything or misunderstood something.

zormit (approved):
To me this looks very fine. I won't think about wording too much/too deeply, I think it's improving in this PR and that's the most important thing to me :) Neil had already a look, so I'll leave the details to him.

Niloth-p:
Note: The "Narrow to compose box message recipient" command that we just assigned to category "compose_box" in #T1520 is now re-assigned to category "narrowing".
Comparing the 2 categories, I think the "narrowing" category is a better fit, as the compose box relation can be covered with the contextual help when added.

neiljp:
@Niloth-p Thanks for the rebase :+1: I agree about the narrowing category, certainly until we consider multi-categories (if we do).

It would seem that n/p would also be best in this same category for now? That is, compared to others that will be present in the short term?

The `meta .` change has issues:
- it changes to *any* conversation, not just DMs/PMs
- if it was DMs/PMs only, the UI itself should all be expressed as 'Direct' by now
- 'Open PM' reads strangely to me, but given the above it can be changed in any case

I believe those were the two points I had noticed and was starting to work through before you pushed again this evening :)

Niloth-p:
1. Grouped 'p' and 'n' into the 'narrowing' category.

2. Oops, thank you!
I've rephrased it to "Switch message view to the compose box target". I felt 'recipient' could be made a bit more user-friendly.

3. I noticed that I had somehow had missed the re-ordering changes while rebasing on merges, at some point of time. So, I created a separate commit for that and kept it in the middle, so that it can be squashed into either the first or the last, if necessary.

4. Updated the commit descriptions.

neiljp:
@Niloth-p Thanks for the update 👍 I partially reversed the mentions text, but that's a marginal language change, and otherwise I just tweaked the commit text.

The middle commit is small, but helps keep the reordering clear, since the diff is less clear otherwise, so I've left it separate :+1:

Merging this now, thanks all! :tada:

neiljp:
I've added the follow-up points you noted above as a summary into the bottom of #1524 to keep them tracked, as well as a few other points.

PR #1520, opened 2024-06-17
Create new help categories related to opening compose box

Niloth-p:
Groups key bindings from message actions into the new categories:
- Compose: New Message
- Compose: Replies

And in the 2nd commit, suggests renaming the "Composing a message" category to "Writing a message" category, to improve clarity, in contrast with the newly added categories.

- [x] Discussed in **#zulip-terminal** in `Re-Categorization of Hotkeys`
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #
- [x] Merge will enable work on #1484

### How did you test this?

### Visual Changes
image

### DIFF
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -81,7 +81,7 @@
 |New message to a stream|<kbd>c</kbd>|
 |New message to a person or group of people|<kbd>x</kbd>|
 
-## Composing a Message
+## Writing a message
 |Command|Key Combination|
 | :--- | :---: |
 |Cycle through recipient and content boxes|<kbd>Tab</kbd>|
--- a/docs/hotkeys.md
+++ b/docs/hotkeys.md
@@ -8,7 +8,6 @@
 |Show/hide Help Menu|<kbd>?</kbd>|
 |Show/hide Markdown Help Menu|<kbd>Meta</kbd> + <kbd>m</kbd>|
 |Show/hide About Menu|<kbd>Meta</kbd> + <kbd>?</kbd>|
-|Open draft message saved in this session|<kbd>d</kbd>|
 |Copy information from About Menu to clipboard|<kbd>c</kbd>|
 |Redraw screen|<kbd>Ctrl</kbd> + <kbd>l</kbd>|
 |Quit|<kbd>Ctrl</kbd> + <kbd>c</kbd>|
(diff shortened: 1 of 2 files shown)


zormit (approved):
This is nice, thank you. I'm looking forward to having a reduced list of "message actions" (which was too long).

Niloth-p:
Rebased, resolving merge conflicts and updating INSERT_NEW_LINE's category along with the rest in the 2nd commit.

neiljp:
@Niloth-p Thanks for sticking with these improvements - I agree they are definitely helping :+1:

Following on from my summary in #1524, my main concern here is that the categories are small (3 or 4 entries), plus the draft option doesn't entirely fit as a 'new message' (which would make it 2, and find somewhere to put the draft entry?). While it potentially makes it easier to find some entries when there are very many categories, the vertical spacing starts to take over. Do you agree?

I think we'll still gain from splitting these commands out into one section, and the renaming of the other section certainly remains worthwhile too. 

Having just merged the GO_BACK splitting PR (#1514), it also prompts me to think whether the exiting of compose could be included here (too), as a 'Starting/Exiting compose' sort of category. However, that introduces multi-category commands again, and could make naming the single category more challenging - this isn't as important as getting the splitting and renaming done and merged.

Niloth-p:
Updates:
- Merged the 2 new categories.
- Rebased onto main, and updated the recently newly added EXIT_COMPOSE.

Things to check:
- The naming of the newly formed category - both internally and externally
- Is the ordering of help keys in the new category satisfactory?

Niloth-p:
> While it potentially makes it easier to find some entries when there are very many categories, the vertical spacing starts to take over. Do you agree?

Regarding my opinion,
The spacing would definitely dominate in smaller screens. 3 is quite a small category indeed.

But, I like smaller help categories in general for any application, but that's probably because I try to memorize them (and the related commands) instead of just referring to a single command.
So, tighter categories help in that aspect, instead of me having to break them down into further tiny categories in my own mind, it's already done for me. So, the tightness/fit of entries into categories would be more preferred over the size of categories. 
But, that's probably me, I have no idea what users majorly prefer.

zormit:
I think small categories can be nice as long as the titles are really really clear.

My reason would be a bit different than yours: If I'm looking for a specific command I can just skim the headlines and hopefully quickly find the right category. In that category it then should also be quite fast to identify the command I'd be looking for, because there are not that many present.

neiljp:
@Niloth-p Thanks for this and your slight update :+1: It's definitely nuanced between the two but let's move on and get this merged for now, and we can always return to examine the category splits later, particularly if we have the section jump keys :)

PR #1532, opened 2024-07-11
Add DM panel to the left side view (rebased #1416)

neiljp:

This is an update on the current work in #1416.

Currently this only includes a few small fixups over #1416, that may be included or not:
- Make the new stream messages button not selectable (cannot receive focus)
- Adds a simple `>` prefix for the new Stream messages button as `STREAM_MESSAGE_MARKER`
- Make the DM messages button not selectable (cannot receive focus) [for now]

The way this is currently structured, this could be merged in two chunks (note that the summary of each chunk is not on a commit-wise basis):
- up to and including my `symbols/buttons` commit
  - adds the data for the total stream messages into a new (non-selectable) button
  - changes the left panel structure to have a simple dividing line from the top-left view buttons, rather than a `---Streams---` or `---Topics---` title
  - clarifies the context after that change by specifying 'Search topics' and 'Search streams' rather than the previous 'Search' in the panel search box
- that plus the rest of the commits
  - Moves the Direct messages button into a section with its own solid line separating it
  - Adds lots of model support for DM data, and updating it
  - Adjusts hotkeys so that from the left panel, `S` shows the view after the first PR chunk (above), with `P` switching to a view with the `P` view fully expanded
  - The fully expanded `P` view shows conversations with unreads, and user status
    (group conversations are included, unlike in the user list)
  - position in the DM list (and stream list) is preserved on switching between them

An additional UI 'chunk' could be to do the first bullet of the last chunk distinct from the data and view changes, but this change is currently combined with one of the later commits in the second chunk. That last chunk could also be split into data and UI updates, potentially.

### Feedback sought
Manual testing and feedback on:
1. the first chunk (eg. `git rebase -i` and 'edit' on the 'symbols/buttons' commit)
  - This is the anticipated UI direction to enable the full PR to merge
  - how does it feel during use?
  - does stream/topic switching look good?
  - how does it compare to `main` for you?
2. the full PR
  - This is the current plan for the first iteration of showing DM conversations
  - See previous conversations for motivation for this design:
    - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/Private.20messages.20sidepanel.20.23T1416
    - https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/DM.20Side.20Panel
  - Note the outstanding (known) aspects below!
  - How does it look and work for you? Is anything clearly broken? any other improvements?

### Outstanding aspect(s)    

- [ ] `P` for DM feed functionality is currently removed, even from message context; it would be useful to retain this feature generally? [discuss]
  - [ ] Even the `P` DM feed button is not usable right now, for simplicity
- [ ] From the user list, `P` changes the left panel view, but doesn't re-focus; that's clearly buggy, but should it do something like that? [discuss]
- [ ] `S` now works differently in left panel (new view-change) vs message context (zoom-in); is that OK?
- [ ] While position in the stream list, and in any connected topic list, is preserved upon switching to/from the DM list, if the topic list was shown previously then it instead shows the stream list, which seems inconsistent? (eg. streams -`t`-> topics -`P`-> DMs -`S`-> streams)
- [ ] Sort user names in DM list like web app?
- [ ] Ensure DM list status symbols are updated
- [ ] Use 'huddle' symbols for conversations with >2 people (eg. ∷)

Note that not all of these are necessary for a full merge and could belong in follow-ups, but some relate to what seems like buggy, inconsistent or reduced behavior.

- [x] Discussed in **#zulip-terminal** in `Private messages sidepanel #T1416` and `DM Side Panel`
- [ ] Fully fixes #
- [x] Partially fixes issue #1197 
- [x] Builds upon previous unmerged work in PR #1416 
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -869,6 +869,15 @@ def clean_custom_profile_data_fixture() -> List[CustomProfileData]:
 
 
 @pytest.fixture
+def sorted_recent_dms_fixture() -> List[Dict[str, Any]]:
+    return [
+        {"max_message_id": 4, "user_ids": []},
+        {"max_message_id": 3, "user_ids": [2]},
+        {"max_message_id": 2, "user_ids": [1]},
+    ]
+
+
+@pytest.fixture
 def initial_data(
     logged_on_user: Dict[str, Any],
     users_fixture: List[Dict[str, Any]],
@@ -1050,6 +1059,11 @@ def initial_data(
         "zulip_feature_level": MINIMUM_SUPPORTED_SERVER_VERSION[1],
         "starred_messages": [1117554, 1117558, 1117574],
         "custom_profile_fields": custom_profile_fields_fixture,
+        "recent_private_conversations": [
+            {"max_message_id": 4, "user_ids": []},
+            {"max_message_id": 2, "user_ids": [1]},
+            {"max_message_id": 3, "user_ids": [2]},
+        ],
     }
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1469,6 +1469,7 @@ def classified_unread_counts() -> Dict[str, Any]:
     return {
         "all_msg": 12,
         "all_pms": 8,
+        "all_stream_msg": 4,
         "unread_topics": {
             (1000, "Some general unread topic"): 3,
             (99, "Some private unread topic"): 1,
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -288,6 +288,7 @@ def test_sort_unread_topics(
             [["Some general stream", "Some general unread topic"]],
             {
                 "all_msg": 8,
+                "all_stream_msg": 0,
                 "streams": {99: 1},
                 "unread_topics": {(99, "Some private unread topic"): 1},
                 "all_mentions": 0,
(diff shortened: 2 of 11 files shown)


Niloth-p:
> P for DM feed functionality is currently removed, even from message context; it would be useful to retain this feature generally? [discuss]

I don't see why we'd need to retain the feed.

> Sort user names in DM list like web app?

Yep, I agree, that's a required feature.

Narrowing: Keep the panel switching hotkeys separate from narrowing, and not couple it with the panel switching hotkeys.
	- Reason: Loss of unread info when switching between the panels

> S now works differently in left panel (new view-change) vs message context (zoom-in); is that OK?

I am not sure that's intuitive.

> From the user list, P changes the left panel view, but doesn't re-focus; that's clearly buggy, but should it do something like that? [discuss]

Well, re-focusing could help.

> Ensure DM list status symbols are updated

By status, do you mean the presence status? Is that not currently in sync?

### Features to add / fix / address:

- [ ] Fix tests and linting.
- [ ] Adding new tests.

- [ ] Currently only fetches recent DMs, as was previously discussed in CZO and the PR - I have not personally tested this.
- [ ] Refactor the DMButton code, as per Vishwesh's last comment on his PR.
- [ ] Sort user names in DM list like web app
- [ ] The DM button unread counts are not updated. The total DM counts get updated, but not for each DM.
- [ ] DMs with self are not handled properly. Sets user as None.
- [ ] Opening stream panel with `S` must open the previously active list (stream / topic).
	Focus is currently retained on the stream list, must also retain focus on the topic list.

#### Optionally, after discussion:
- [ ] Switch focus to the DM Panel on pressing P
- [ ] Pressing P/S again must fold/collapse the open panel, and/or expand the other panel
	Or alternatively, hide the `[P]` or `[S]` suggestion on the buttons when the respective panel is open.
- [ ] The `S` key behavior needs to be adjusted. Currently,
	- Opens the stream panel only from the left panel.
	- Topic narrows from the middle column. But it shows the narrow as "# <stream_name> topic narrow".
	- Does not work from the user list.
- [ ] Ensure DM list status symbols are updated
- [ ] Use 'huddle' symbols for conversations with >2 people (eg. ∷)
- [ ] Handling the no DM case smoothly.
	We don't currently handle the no messages case / empty narrows either.

The empty space in the DM Panel can be dominating, when there aren't enough DMs, as discussed previously on a call. The Stream Panel being folded to the bottom can be easy to miss until one gets used to the UI.

Issue #1001, opened 2021-04-20
Add support for muted users


WesleyAC:
We've finished merging the backend for muting users in zulip/zulip#16915, and are discussing the UI in zulip/zulip#18222.

This is being implemented in the webapp (zulip/zulip#16915), and mobile (zulip/zulip#4678), and it would be excellent for ZT to support it as well.

zulip/zulip-mobile#4655 may be helpful for figuring out what implementing this looks like, it's unlikely to be very tricky.

neiljp:
@WesleyAC Thanks for opening this - I was just considering doing so myself :+1:

This is projected for Zulip 4.0?

WesleyAC:
Yep, the plan is for this to be in 4.0 :)

Subhasish-Behera:
The mobile implementation seems to be [zulip/zulip-mobile#4678](https://github.com/zulip/zulip-mobile/issues/4678) and not [zulip/zulip#4678](https://github.com/zulip/zulip/issues/4678) as mentioned above.

PR #1546, opened 2024-09-20
[WIP] Support muted users

rsashank:
Adds support to handle muted users

TODO: Add support to mute and unmute users on ZT

### Outstanding aspect(s)    

- [ ] 

- [ ] Fully fixes #
- [ ] Partially fixes issue #1001
- [ ] Builds upon previous unmerged work in PR #1425
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -903,6 +903,7 @@ def initial_data(
             }
         ],
         "result": "success",
+        "muted_users": {},
         "queue_id": "1522420755:786",
         "realm_users": users_fixture,
         "cross_realm_bots": [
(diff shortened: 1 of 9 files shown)


neiljp (commented):
@rsashank Thanks for taking this PR forward :+1:

Fixing the co-authored field will help this be more obvious which commits are yours, but it's worth noting what you've added too - eg tests in some cases :)

Most of the previous comments appear to have been addressed, and the feature broadly works. The remaining aspects are missing muted-users features (eg. see the API page), and that the updating of muting status in the UI relies upon some unrelated or indirect calls.

For the message update, I expect it is relying upon `update_message_list_status_markers`, which is indirectly updating the entire message list - including this message content. That makes for a rather confusing control flow right now, if so, and it would be good to make each part explicit.

For outstanding muted users features, I think we could build on that in a follow-up PR, most notably for
- compressing multiple lines of message by the same muted user into one different message (does web do this?)
- hiding 1:1 DM conversations (various views)
- hiding in DM autocomplete
- possibly hiding in message content mention autocomplete (web allows `@` only?)

PR #1552, opened 2024-10-24
Recipient headers for specific text searches

### LINKED ISSUE #1550: Search results should show recipient headers for every message
### Bug description

When searching (for text, our only option right now), it is possible to end up with results that show messages adjacently under the same heading (recipients, conversation).

This can be confusing, since these messages may be in the same conversation, but there may be messages between them that do not meet the search criteria (filter text), and it appears that the messages simply follow one another.

This was partially inspired by `# design > Date indicators in search results`.

### How is the bug triggered?
How can you reproduce the bug?

Currently:
1. Go to All messages view in ZT
2. Search for the text (eg. /, 'date indicators', Enter)
3. (Possibly scroll up/down to look for examples)
4. Observe multiple conversations with adjacent messages under the same header line

Repeating in the web app shows each message with its own equivalent of our header line

A few examples with this search text are:
- `# api design > trimming date_joined` (two messages on 27 & 31 July 2024, separated by others)
- `# design > Date indicators in search results` (two time-adjacent messages)

In the web client, this appears to be resolved by showing the recipient (conversation) header for every message in a search result.

While extra lines are at a premium in a terminal, we should show extra header lines a similar way, ie. showing the message recipient line above each message. 

Note that it should not display the recipient part of the header in conversation views (specific DMs, topics), just like now, since the header information is in the part above that already.

We should also ensure the message line with the status/author/date, particularly the date, should always be shown - that is the default with the current code, except for two messages written to the same conversation within a narrow time window by the same sender. We should ensure the date appears in *every* message, when such messages are split by a recipient header, once this feature is implemented.

- **Zulip-terminal version:** `main` in September 2024
- **Zulip server version(s):** chat.zulip.org
- **Operating system (and version):** Debian
- **Python version (and implementation):** CPython 3.9

### Related future follow-ups

- To save space, we might be able to 'fold' conversations where eg. the search text is in a topic name. That would avoid showing *every* matching message, along with the title bar for each.
- Highlighting search terms in search results.

u7327620:
Previously, when text searching, messages appear adjacent under the same heading, this can be confusing since they appear to be sequential but there may be messages in-between that don't meet the search criteria. PR adds logic to add recipient headers if the search isn't a specific dm or a topic narrow

### Outstanding aspect(s)    

- [ ] Uncertain if automated tests are covering the feature or not (new to pytest) 
- [ ] Related follow-ups aren't fixed in this PR ... should they be?

- [x] Fully fixes #1550 
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### Visual changes    

Before: swappy-20241024-191318
After: swappy-20241024-191427

### DIFF
--- a/zulipterminal/ui_tools/messages.py
+++ b/zulipterminal/ui_tools/messages.py
@@ -121,6 +121,14 @@ class MessageBox(urwid.Pile):
         if len(self.model.narrow) == 2 and self.model.narrow[1][0] == "topic":
             return False
 
+        if self.model.is_search_narrow():
+            if len(self.model.narrow) == 1:
+                return True
+            if len(self.model.narrow) == 2:
+                return self.message["type"] != "private"
+            if len(self.model.narrow) == 3:
+                return self.model.narrow[1][0] != "topic"
+
         last_msg = self.last_message
         if self.message["type"] == "stream":
             return not (


neiljp:
@u7327620 Thanks for working on this :+1: This does appear to work from manual testing, but it seems like it could be simpler, and it's unclear why the extra logic is necessary. Including a test for the new logic would demonstrate this clearly.

I've not seen you around recently, so I'm marking this as a completion candidate.

u7327620:
Hello Neil! I looked through my notes:

Originally I had `if self.model.is_search_narrow(): return True` but it didn't exclude headers when searching private message chains or topic-narrows. In addition, tests (didn't write down which ones) failed due to the added headers. The added logic is meant to exclude these cases.

PR #1556, opened 2024-11-18
Preparation for deprecating reactions' user metadata

Niloth-p:
- Prepares for the deprecation of user metadata from messages' reactions and reaction events
    - Supports both `user_id` (ZFL>=2) and `user["id"]` for messages' reactions (with added test cases)
    - Supports both `user_id` (ZFL>=2) and `user["user_id"]` for reaction events (with added test cases)
    - Introduces a common `get_user_id_from_reaction()` for handling the different possible fields
    - Switches to using `_all_users_by_id["user_id"]` to remove dependency on `reaction["user"]["full_name"]
    - Improves the `ReactionEvent` schema to support both versions
- Fixes a bug in reaction event handling (remove op), and adds tests covering the missing cases
- Refactors the reactions' functions a bit before the above changes

- [x] Discussed in **#zulip-terminal** in `reaction metadata`
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -1849,9 +1849,20 @@ class Model:
                 )
             else:
                 for reaction in message["reactions"]:
-                    # Since Who reacted is not displayed,
-                    # remove the first one encountered
-                    if reaction["emoji_code"] == event["emoji_code"]:
+                    reaction_user_id = (
+                        reaction["user"]["id"]
+                        if "user" in reaction
+                        else reaction["user_id"]
+                    )
+                    event_user_id = (
+                        event["user"]["user_id"]
+                        if event.get("user") and isinstance(event["user"], dict)
+                        else event["user_id"]
+                    )
+                    if (
+                        reaction["emoji_code"] == event["emoji_code"]
+                        and reaction_user_id == event_user_id
+                    ):
                         message["reactions"].remove(reaction)
                         break
 
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -48,6 +48,7 @@ from zulipterminal.api_types import (
     MessagesFlagChange,
     PrivateComposition,
     PrivateMessageUpdateRequest,
+    ReactionEvent,
     RealmEmojiData,
     RealmUser,
     StreamComposition,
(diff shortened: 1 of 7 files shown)


mounilKshah (commented):
Thanks for working on this @Niloth-p 
At first glance, the `get_user_id_from_reaction()` seemed unnecessary to me but on closer look, multiple usages of the same logic make the function a necessity. The coverage of cases is more vast than the impact that I had expected. Overall, this worked as expected across all the use cases that I tried for. 
I think the commit structure can be changed. You can squash the initial commits which work towards simplifying/switching to the new API response. Then the next commit can be the patch providing the new function of `get_user_id_from_reaction()` and its test function. Then you may change all implementations of the older response and add the function calls wherever necessary.
Apart from this, I don't have any other changes to suggest for this PR.

zulipterminal/model.py
```diff
@@ -1849,9 +1849,20 @@ def _handle_reaction_event(self, event: Event) -> None:
...
+                        event["user"]["user_id"]
+                        if event.get("user") and isinstance(event["user"], dict)
+                        else event["user_id"]
+                    )
+                    if (
+                        reaction["emoji_code"] == event["emoji_code"]
+                        and reaction_user_id == event_user_id
+                    ):
                         message["reactions"].remove(reaction)
                         break
```
timabbott, line 1874: Feature level 3.0 is old enough that we could drop support for it. But I guess this is OK.

timabbott:
I did a fairly careful read, and this seems like a good PR to me, and I think it makes sense to merge and fix-forward from here as needed. It's sadly a bit of wasted effort -- we could just process `user_id` given no other clients support older versions than 5.0 -- but doing that can be left for future work, as it's not a problem for this code path to support ancient versions, just extra code we don't need.

Niloth-p:
Thank you!
True, I did consider dropping the older version support, but during our GSoC calls, Neil has explicitly expressed how he prefers retaining support for the older versions unlike all the other clients, so this was done for his satisfaction. We've had a few discussions on adding a uniform framework for version-specific features / ZFL-specific handling, so knowing that, I didn't want to be the person who drops the support for older versions.

Issue #1502, opened 2024-05-20
pass command instead of password in zuliprc?


dkwo:
My zuliprc contains a [key] entry, which I guess is derived from the password.
Would it be possible (or maybe already is? I did not find documentation) to have a pass comand instead, like e.g. in mbsync or similar, so the key entry would look more like "PassCmd "/usr/bin/pass name@domain"?

neiljp:
Thanks for the inquiry! Is this a blocker to you being able to use Zulip or Zulip-terminal?

I wasn't familiar with `mbsync`, and it took me a while to figure out what you meant by a 'pass command', since I wasn't familiar with that tool either :)

My understanding of your query is that you're wanting to store the password-like detail centrally in a `pass` password store, and extract it on-demand upon starting the application?

Currently `zuliprc` (-style) files generated by the Zulip web app have the form of:
```
[api]
email=...  # login (email or username)
key=...    # key
site=...   # server
```
Zulip-terminal uses the python zulip wrapper library, and further details on the structure of this kind of zuliprc file is at:
  https://github.com/zulip/python-zulip-api/tree/main/zulip (README.md)

As a client we could implement this feature individually, since we could support arbitrary features/extensions in a zuliprc file, a little like we do right now for our current options. However, if we added such an option, we may want to coordinate with the larger community to coordinate on this.

The Zulip server supports other login methods that we have yet to incorporate, which are typically browser-based, and mostly about connecting to a central authority (see eg. https://zulip.com/help/configure-authentication-methods). My understanding is that your request is specifically for a local authentication feature; the reason I mention this is I'm wondering if the deferral of the authentication to a command might be a common factor to both if we implemented these.

dkwo:
Thanks for looking at this, and sorry for being cryptic. You got me right.
I'm mostly concerned with local auth, but indeed a pass cmd could also deal with tokens.
My concern is that I don't want to store key material in the config (zuliprc), so that it can be checked in git with other dotfiles.

To keep the analogy with mbsync, the token case (for oauth) has `PassCmd "~/bin/mutt_oauth2.py ${XDG_DATA_HOME}/oauth-tokens/gmail.tokens"`.

dkwo:
To be unambiguous: the `pass cmd` option shuold let us use `pass`, `gopass` or whatever else we can use from the terminal.

neiljp:
@dkwo To confirm the original intent here:
[...]

PR #1564, opened 2025-03-22
Secure api_key in zuliprc file using encryption

Gopinath-Mahendiran:
##This PR enhances security by encrypting the api_key stored in the zuliprc file. It ensures that sensitive credentials are not stored in plaintext, reducing the risk of exposure. A decryption mechanism is also included to seamlessly retrieve the api_key when needed.

Outstanding aspect(s)

	•	Ensure compatibility with existing zuliprc files containing plaintext api_key.
	•	Add documentation for users upgrading to the encrypted format.

 Discussed in #zulip-terminal in topic
[ x] Partially fixes issue https://github.com/zulip/zulip-terminal/issues/1502

How did you test this?

[ x] Manually - Verified that api_key is encrypted when stored.
[ x] Manually - Confirmed api_key decryption works correctly.

###How this behaves
When a user logs in directly from the terminal without downloading the zuliprc file from the web application, the encryption process will be handled internally, ensuring a seamless and secure experience.

However, if the zuliprc file is downloaded from the web application, encryption must be performed manually using the following command:
`zulip-term --encrypt <file_name>`

This command encrypts the api_key stored in the file, enhancing security before use.

Regardless of whether the user logs in with the default zuliprc file or a custom configuration file, decryption will be handled automatically within the application. This ensures a smooth authentication process without requiring any manual intervention.

### DIFF
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1,5 +1,6 @@
 from collections import OrderedDict, defaultdict
 from datetime import date
+from unittest.mock import patch
 
 import pytest
 import pytz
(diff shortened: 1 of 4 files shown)


neiljp:
@Gopinath-Mahendiran I replied in the issue, due to some confusion over the specification for the expected fix? I also moved your Zulip message into a new topic to make it easier to see the dedicated discussion for the issue and any PRs that result including this one.

Gopinath-Mahendiran:
@neiljp  I understand the problem and will work towards it, making the necessary changes.

Additionally, I’d like to clarify whether the key stored by the `/usr/bin/key_command` should be password-protected or not. Also, should the **key-command** be a system-wide command or specific to the application?

PR #1580, opened 2025-04-08
Fix #1506 : Improve handling of narrows (views) with no messages

### LINKED ISSUE #1506: Improve handling of narrows (views) with no messages
### Bug description

When selecting a new narrow (view) to show, via a hotkey, a button in either side panel, or using the message search (via `/`), the message focus changes to a message in the resulting view.

When the message list for that narrow/view is **empty**, ie. no messages are found, there is no visible message to gain focus, and the UI behaves strangely.

This can occur for many narrows/views, including:
- streams (channels)
  - #607 
  - #259
- searches which result in no results
  - #580
- direct messages with a specific user (eg. if not messaged them before)
  - #1480
- starred messages
  - ([#zulip-terminal > keyboard movement stops with no messages starred section](https://chat.zulip.org/#narrow/stream/206-zulip-terminal/topic/keyboard.20movement.20stops.20with.20no.20messages.20starred.20section))
- topics (difficult, but possible via exiting compose to a new topic)
- mentions (if never mentioned?)
- all direct messages (?)

This results in
- no obvious focused UI element (#607)
- the 'Current message recipients' not updating (remaining that of the previously focused message)
- the message narrow potentially not updating (#1480, #259)

It is possible to navigate out of this state by moving sideways and down (#580), or using one of the 'global' hotkeys such as `w` or `q`, P/#/f (though *not* `a`). However, these are workarounds.

This has been discussed variously in #zulip-terminal, and in at least the above linked issues, as well as early work on it in #278.

Certainly:
- Empty the current message recipients content, or hide that section entirely (as per another issue)
- Ensure the narrow title is properly updated (eg. starred, etc.)

In addition:

1) Unlike the web app, we expect the focus to move somewhere after a new narrow is requested, so a placeholder in the message list may be the easiest short-term solution to use.

This was worked on historically in #278 as a fix for #259, where a 'fake message' was used as a placeholder. However, as demonstrated by the more complex logic we now support for searching/filtering the side panels, it should be reasonable to support a custom element for 'no results found' to be shown instead.

Even so, that solution would need to support various cases where messages 'arrive' in that empty narrow, and the placeholder is replaced by them.

2) Alternatively, focus could be retained in the previous location, if the resulting narrow is found to be empty. This would be convenient for message searches, where an empty set of results might be reasonably expected by a user to leave the focus in the search box to adjust the search.

3) Some combination of (1) and (2), since
- a placeholder shows that the narrow is empty (no messages)
- a placeholder can be focused, allowing easy movement horizontally across the app (left/right), such as from left panel to right panel.

### Notes

- I'm closing many of the linked issues, since they are essentially duplicates, ie. examples of the symptoms from the underlying problem.
- This could unblock other features, such as #1049

sahith-ch:
This PR improves handling of narrows with no messages by implementing a placeholder component

- [ x] Fully fixes #1506
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

- [ x] Manually - Visual changes

- [ x] It individually passes linting and tests
- [x ] It flows clearly from a previous branch commit, and/or prepares for the next commit

### Visual changes    

image

### DIFF
--- a/zulipterminal/model.py
+++ b/zulipterminal/model.py
@@ -83,6 +83,7 @@ from zulipterminal.helper import (
 from zulipterminal.platform_code import notify
 from zulipterminal.ui_tools.utils import create_msg_box_list
 
+from zulipterminal.ui_tools.messages import PlaceholderMessageBox
 
 class ServerConnectionFailure(Exception):
     pass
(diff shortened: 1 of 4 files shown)


neiljp:
@Niloth-p Do you have the bandwidth to give some thoughts on this, compared to your approach in #1543 ?

PR #1610, opened 2026-01-28
Improve support for todo widgets

### LINKED ISSUE #1575: Improve support for todo widgets
Initial work in https://github.com/zulip/zulip-terminal/pull/1549 by @rsashank provided

- parsing of todo data
- rendering into a styled message
- updating of the styled message upon changes in the server (events)

This issue is intended to split out todo-related follow-up work, part of which was originally tracked in #986.

Outstanding elements include
- [ ] Enable responding to tasks, ie. toggling the 'check mark' or 'done' status
- [ ] Allow editing the title in your own todo list
- [ ] Allow adding tasks to todo lists
- [ ] Limiting the editing of todo lists in the client (widget messages (content) may not be edited)

yummyPancake2607:
## Improve support for todo widgets

Fixes #1575

---

## Goal of the work

The todo widget in **zulip-terminal** relies on widget submessages to represent user actions such as toggling, adding, and renaming items.  
This PR improves support for todo widgets by completing missing UI interactions and ensuring todo actions are sent correctly to the server.

---

## What was missing or broken

Before this change:

- Some todo widget actions were not fully supported in the UI.
- Widget submessages were not consistently sent via the correct Zulip endpoint.
- Users could not reliably add or rename todo items through dedicated popups.

---

## What this PR changes

This PR is intentionally split into **small, focused commits**:

### 1) Send widget submessages via the correct endpoint
- Widget submessages are now sent using `POST /api/v1/submessage`.

**Result:** todo actions are correctly delivered to the server.

---

### 2) Support toggling todo tasks
- Added UI support for toggling (striking/unstriking) todo items.

**Result:** users can reliably mark tasks as complete or incomplete.

---

### 3) Add todo text input popup
- Introduced a popup UI for adding new todo items.

**Result:** users can add todo items interactively from the terminal UI.

---

### 4) Add todo rename popup
- Added a popup UI for renaming/editing existing todo items.

**Result:** todo items can be edited without leaving the widget.

---

### 5) Support adding todo tasks
- Wired add-todo UI actions to widget submessages.

**Result:** newly added todo items persist and sync correctly.

---

## User-visible behavior

When focused on a todo widget:

- **Toggle item:** press `1`–`9`
- **Add item:** press `a`, type the text, press `Enter`
- **Rename/edit item:** press `t`, type the new text, press `Enter`

---

## Screenshots

**Add todo popup**  
Add todo popup

**Rename / edit popup**  
Rename todo popup

---

## How to test

- Open a message containing a todo widget
- Toggle items using number keys
- Add a new item using `a`
- Rename an item using `t`

### DIFF
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1992,3 +1992,47 @@ class TestMessageBox:
         self.model.send_widget_submessage.assert_called_once_with(
             101, {"type": "new_task_list_title", "title": "New"}
         )
+
+    def test_todo_widget_add_shows_popup_and_sends_submessage(self, mocker):
+        mocker.patch.object(MessageBox, "main_view")
+        self.model.send_widget_submessage = mocker.MagicMock(return_value=True)
+        self.model.controller.maximum_popup_dimensions.return_value = (80, 24)
+
+        message = dict(
+            id=102,
+            type="stream",
+            display_recipient="general",
+            stream_id=5,
+            subject="todo",
+            sender_id=42,
+            sender_email="foo@zulip.com",
+            sender_full_name="Foo",
+            content="<p>todo</p>",
+            submessages=[
+                {
+                    "msg_type": "widget",
+                    "sender_id": 42,
+                    "content": (
+                        '{"widget_type":"todo","extra_data":{"tasks":[{"task":"A"}]}}'
+                    ),
+                }
+            ],
+        )
+        msg_box = MessageBox(message, self.model, None)
+
+        msg_box.keypress((80, 24), "a")
+        popup = self.model.controller.show_pop_up.call_args[0][0]
+
+        popup._edit.set_edit_text("Task: Desc")
+        popup.keypress((80, 24), "enter")
+
+        self.model.send_widget_submessage.assert_called_once_with(
+            102,
+            {
+                "type": "new_task",
+                "task": "Task",
+                "desc": "Desc",
+                "key": 1,
+                "completed": False,
+            },
+        )
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1954,3 +1954,41 @@ class TestMessageBox:
         self.model.send_widget_submessage.assert_called_once_with(
             100, {"type": "strike", "key": "0,canned"}
         )
+
+    def test_todo_widget_rename_shows_popup_and_sends_submessage(self, mocker):
+        mocker.patch.object(MessageBox, "main_view")
+        self.model.user_id = 42
+        self.model.send_widget_submessage = mocker.MagicMock(return_value=True)
+        self.model.controller.maximum_popup_dimensions.return_value = (80, 24)
+
+        message = dict(
+            id=101,
+            type="stream",
+            display_recipient="general",
+            stream_id=5,
+            subject="todo",
+            sender_id=42,
+            sender_email="foo@zulip.com",
+            sender_full_name="Foo",
+            content="<p>todo</p>",
+            submessages=[
+                {
+                    "msg_type": "widget",
+                    "sender_id": 42,
+                    "content": (
+                        '{"widget_type":"todo","extra_data":{"task_list_title":"Old","tasks":[]}}'
+                    ),
+                }
+            ],
+        )
+        msg_box = MessageBox(message, self.model, None)
+
+        msg_box.keypress((80, 24), "t")
+        popup = self.model.controller.show_pop_up.call_args[0][0]
+
+        popup._edit.set_edit_text("New")
+        popup.keypress((80, 24), "enter")
+
+        self.model.send_widget_submessage.assert_called_once_with(
+            101, {"type": "new_task_list_title", "title": "New"}
+        )
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1922,3 +1922,35 @@ class TestMessageBox:
             msg_box.keypress.assert_not_called()
         else:
             msg_box.keypress.assert_called_once_with(size, expected_keypress)
+
+    def test_todo_widget_strike_sends_submessage(self, mocker):
+        mocker.patch.object(MessageBox, "main_view")
+        self.model.send_widget_submessage = mocker.MagicMock(return_value=True)
+
+        message = dict(
+            id=100,
+            type="stream",
+            display_recipient="general",
+            stream_id=5,
+            subject="todo",
+            sender_id=42,
+            sender_email="foo@zulip.com",
+            sender_full_name="Foo",
+            content="<p>todo</p>",
+            submessages=[
+                {
+                    "msg_type": "widget",
+                    "sender_id": 42,
+                    "content": (
+                        '{"widget_type":"todo","extra_data":{"tasks":[{"task":"A"}]}}'
+                    ),
+                }
+            ],
+        )
+        msg_box = MessageBox(message, self.model, None)
+
+        msg_box.keypress((80, 24), "1")
+
+        self.model.send_widget_submessage.assert_called_once_with(
+            100, {"type": "strike", "key": "0,canned"}
+        )
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1922,3 +1922,195 @@ class TestMessageBox:
             msg_box.keypress.assert_not_called()
         else:
             msg_box.keypress.assert_called_once_with(size, expected_keypress)
+
+    def test_keypress_TODO_STRIKE(self, mocker, widget_size):
+        mocker.patch.object(MessageBox, "main_view")
+        self.model.send_widget_submessage = mocker.MagicMock(return_value=True)
+
+        message = dict(
+            id=100,
+            type="stream",
+            display_recipient="general",
+            stream_id=5,
+            subject="todo",
+            sender_id=42,
+            sender_email="foo@zulip.com",
+            sender_full_name="Foo",
+            content="<p>todo</p>",
+            submessages=[
+                {
+                    "msg_type": "widget",
+                    "sender_id": 42,
+                    "content": (
+                        '{"widget_type":"todo","extra_data":{"tasks":[{"task":"A"}]}}'
+                    ),
+                }
+            ],
+        )
+        msg_box = MessageBox(message, self.model, None)
+        size = widget_size(msg_box)
+
+        msg_box.keypress(size, "1")
+
+        self.model.send_widget_submessage.assert_called_once_with(
+            100, {"type": "strike", "key": "0,canned"}
+        )
+
+    @pytest.mark.parametrize("key", keys_for_command("TODO_RENAME"))
+    def test_keypress_TODO_RENAME(self, mocker, key, widget_size):
+        mocker.patch.object(MessageBox, "main_view")
+        self.model.user_id = 42
+        self.model.send_widget_submessage = mocker.MagicMock(return_value=True)
+        self.model.controller.maximum_popup_dimensions.return_value = (80, 24)
+
+        message = dict(
+            id=101,
+            type="stream",
+            display_recipient="general",
+            stream_id=5,
+            subject="todo",
+            sender_id=42,
+            sender_email="foo@zulip.com",
+            sender_full_name="Foo",
+            content="<p>todo</p>",
+            submessages=[
+                {
+                    "msg_type": "widget",
+                    "sender_id": 42,
+                    "content": (
+                        '{"widget_type":"todo","extra_data":{"task_list_title":"Old","tasks":[]}}'
+                    ),
+                }
+            ],
+        )
+        msg_box = MessageBox(message, self.model, None)
+        size = widget_size(msg_box)
+
+        msg_box.keypress(size, key)
+        popup = self.model.controller.show_pop_up.call_args[0][0]
+
+        popup._edit.set_edit_text("New")
+        popup.keypress((80, 24), "enter")
+
+        self.model.send_widget_submessage.assert_called_once_with(
+            101, {"type": "new_task_list_title", "title": "New"}
+        )
+
+    @pytest.mark.parametrize("key", keys_for_command("TODO_ADD"))
+    def test_keypress_TODO_ADD(self, mocker, key, widget_size):
+        mocker.patch.object(MessageBox, "main_view")
+        self.model.send_widget_submessage = mocker.MagicMock(return_value=True)
+        self.model.controller.maximum_popup_dimensions.return_value = (80, 24)
+
+        message = dict(
+            id=102,
+            type="stream",
+            display_recipient="general",
+            stream_id=5,
+            subject="todo",
+            sender_id=42,
+            sender_email="foo@zulip.com",
+            sender_full_name="Foo",
+            content="<p>todo</p>",
+            submessages=[
+                {
+                    "msg_type": "widget",
+                    "sender_id": 42,
+                    "content": (
+                        '{"widget_type":"todo","extra_data":{"tasks":[{"task":"A"}]}}'
+                    ),
+                }
+            ],
+        )
+        msg_box = MessageBox(message, self.model, None)
+        size = widget_size(msg_box)
+
+        msg_box.keypress(size, key)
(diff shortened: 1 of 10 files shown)


mounilKshah:
Hi @yummyPancake2607 Thank you for taking this up!
I rebased my local repo to this PR and tried the functionality. It worked fine. I am yet to dive deeper into the code and not sure if can it be done using some existing methods or if there is a need to place a POST request, then we shall have it that way.

zulipterminal/model.py
```diff
@@ -1900,6 +1900,37 @@ def _handle_submessage_event(self, event: Event) -> None:
...
+        try:
+            # Zulip expects widget submessages on POST /api/v1/submessage.
+            if hasattr(self.client, "do_api_query"):
+                response = self.client.do_api_query(
+                    request, "/api/v1/submessage", method="POST"
+                )
+            else:
+                response = self.client.call_endpoint(
+                    "submessage", method="POST", request=request
+                )
```
neiljp: Why do you have both of these? Which is more appropriate?

zulipterminal/model.py
```diff
@@ -1900,6 +1900,37 @@ def _handle_submessage_event(self, event: Event) -> None:
...
+        except Exception as e:
+            self.controller.report_error([f" {e}"])
+            return False
```
neiljp: What is the source of this exception?

This is also a very general exception, which is good to avoid.

zulipterminal/model.py
```diff
@@ -1900,6 +1900,37 @@ def _handle_submessage_event(self, event: Event) -> None:
...
+
+        if response.get("result") != "success":
+            msg = response.get("msg") or "Request failed."
+            self.controller.report_error([f" Submessage failed: {msg}"])
+            return False
+
+        return True
```
neiljp: It would be good to have this in the same format as that in other similar functions.

zulipterminal/model.py
```diff
@@ -1900,6 +1900,37 @@ def _handle_submessage_event(self, event: Event) -> None:
             self.index["messages"][message_id] = message
             self._update_rendered_view(message_id)
 
+    def send_widget_submessage(
```
neiljp, line 1909: Comment in the commit description where you found how this is the correct approach, and if you changed anything compared to that.

This also requires a test.

Rchie121 (approved):

De-ASI-INTERFACE (approved):

De-ASI-INTERFACE (approved):

De-ASI-INTERFACE (approved):

De-ASI-INTERFACE (approved):

Rchie121 (approved):

Rchie121 (approved):

Rchie121 (approved):

De-ASI-INTERFACE (approved):

zulipterminal/model.py
```diff
@@ -1887,7 +1887,7 @@ def _handle_submessage_event(self, event: Event) -> None:
         message_id = event["message_id"]
         if message_id in self.index["messages"]:
             message = self.index["messages"][message_id]
-            message["submessages"].append(
+            message.setdefault("submessages", []).append(
```
neiljp, line 1896: Why is this necessary?

Rchie121 (approved):

tests/ui_tools/test_messages.py
```diff
@@ -1922,3 +1922,35 @@ def test_mouse_event_left_click(
             msg_box.keypress.assert_not_called()
         else:
             msg_box.keypress.assert_called_once_with(size, expected_keypress)
+
+    def test_todo_widget_strike_sends_submessage(self, mocker):
```
neiljp: Let's name this similar to the other keypress tests.

tests/ui_tools/test_messages.py
```diff
@@ -1922,3 +1922,35 @@ def test_mouse_event_left_click(
...
+                {
+                    "msg_type": "widget",
+                    "sender_id": 42,
+                    "content": (
+                        '{"widget_type":"todo","extra_data":{"tasks":[{"task":"A"}]}}'
+                    ),
+                }
+            ],
+        )
+        msg_box = MessageBox(message, self.model, None)
```
neiljp, line 1950: Is the existing msg_box fixture usable instead?

zulipterminal/ui_tools/messages.py
```diff
@@ -965,6 +966,29 @@ def mouse_event(
...
+            self.message.get("submessages")
+            and find_widget_type(self.message.get("submessages", [])) == "todo"
+        ):
+            digit_key = None
+            if key:
+                m = re.search(r"([1-9])$", key.strip())
+                if m:
+                    digit_key = m.group(1)
+
+            if digit_key in {str(i) for i in range(1, 10)}:
```
neiljp: This would be better if using `keys_for_command`, similar to the other conditional below for other keys.

The submessage check can then be moved further down.

zulipterminal/ui_tools/messages.py
```diff
@@ -965,6 +966,29 @@ def mouse_event(
...
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
+        if (
+            self.message.get("submessages")
+            and find_widget_type(self.message.get("submessages", [])) == "todo"
+        ):
+            digit_key = None
+            if key:
+                m = re.search(r"([1-9])$", key.strip())
+                if m:
+                    digit_key = m.group(1)
```
neiljp: This is not necessary when using the `keys_for_command` approach I mentioned elsewhere.

zulipterminal/ui_tools/messages.py
```diff
@@ -965,6 +966,29 @@ def mouse_event(
         return super().mouse_event(size, event, button, col, row, focus)
 
     def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
+        if (
+            self.message.get("submessages")
+            and find_widget_type(self.message.get("submessages", [])) == "todo"
+        ):
+            digit_key = None
+            if key:
+                m = re.search(r"([1-9])$", key.strip())
```
neiljp: I'd expect a todo could have more than 9 items. We don't have to solve that right away, but can we show a warning in some way, if that's the case?

zulipterminal/ui_tools/messages.py
```diff
@@ -989,6 +990,36 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
                     self.model.controller.report_error([" No such to-do item."])
                 return None
 
+            if key == "t":
```
neiljp: This should use the same approach as I suggested in the previous commit, ie. not use a hard-coded key.

zulipterminal/ui_tools/messages.py
```diff
@@ -989,6 +990,36 @@ def keypress(self, size: urwid_Size, key: str) -> Optional[str]:
...
+
+                popup = TodoTextInputPopup(
+                    self.model.controller,
+                    title="Edit to-do title",
+                    prompt="Title:",
+                    initial_text=current_title,
+                    on_submit=on_title_submit,
+                    footer_text="Enter to save • Esc to cancel",
+                )
+                self.model.controller.show_pop_up(popup, "area:msg")
```
neiljp, line 1027: Note where other show_pop_up calls occur. This should work the same way.

We may change this in future, but this is the current design.

tests/ui_tools/test_messages.py
```diff
@@ -1954,3 +1954,41 @@ def test_todo_widget_strike_sends_submessage(self, mocker):
         self.model.send_widget_submessage.assert_called_once_with(
             100, {"type": "strike", "key": "0,canned"}
         )
+
+    def test_todo_widget_rename_shows_popup_and_sends_submessage(self, mocker):
```
neiljp: Again, test should be renamed.

zulipterminal/ui_tools/messages.py
```diff
@@ -965,6 +966,29 @@ def mouse_event(
...
+                _, tasks = process_todo_widget(self.message.get("submessages", []))
+                task_ids = list(tasks.keys())
+                idx = int(digit_key) - 1
+                if idx < len(task_ids):
+                    self.model.send_widget_submessage(
+                        self.message["id"],
+                        {"type": "strike", "key": task_ids[idx]},
+                    )
+                else:
+                    self.model.controller.report_error([" No such to-do item."])
```
neiljp, line 995: This would benefit from a test, or if done carefully a combined test with parametrize.

zulipterminal/ui_tools/todo_input.py
neiljp, line 1: This class would be better inside the file with other popup views. Those should be extracted into a dedicated popup file in future, but it's best to keep them together for now.

I'd also suggest using the PopUpView class to avoid duplicating effort.

neiljp (commented):
@yummyPancake2607 Thanks for working on this 👍

As someone else said in the channel, this appears to work fine 👍 

My comments are mainly about ensuring the new code follows similar practices to the existing code. Let us know if you have questions in the review comments, or generally in the channel topic.

zulipterminal/model.py
yummyPancake2607: I have both here intentionally to handle the different client implementations available in the project.

`do_api_query()` is used when it is available because it lets me explicitly send the request to `/api/v1/submessage`, which is the endpoint required for widget submessages. The `call_endpoint()` fallback is there for clients where `do_api_query()` is not exposed, while still allowing the same request to be made through the higher-level client interface.

I wanted `send_widget_submessage()` to work with both interfaces rather than assuming that `do_api_query()` is always available. Since this method is responsible specifically for sending widget submessages to the `/api/v1/submessage` endpoint, I kept the explicit endpoint call as the primary path and `call_endpoint()` as the fallback.

zulipterminal/model.py
yummyPancake2607: The exception is coming from the API request itself in `send_widget_submessage()`, specifically from `do_api_query()` or `call_endpoint()` when making the POST request to `/api/v1/submessage`.

I used the exception handling here so that if the request itself fails, the error is reported through `controller.report_error()` and `send_widget_submessage()` returns `False` instead of letting the exception propagate to the widget action.

zulipterminal/model.py
yummyPancake2607: Yes, that makes sense. I added the explicit `"Submessage failed: {msg}"` message to make the widget submessage failure clear, but I agree that this should follow the same error-response format used by the other similar API functions in `model.py`. I'll update it to match the existing pattern.

zulipterminal/model.py
yummyPancake2607: I found the approach from the existing message-handling implementation in `model.py`, and adapted it for widget submessages. The main change I made was using the `submessage` endpoint and handling the response through `send_widget_submessage()`.

I'll also add a test covering the widget submessage behavior.

zulipterminal/model.py
yummyPancake2607: The `setdefault()` was intended to handle the case where the message does not already have a `submessages` key. Since the existing message structure already initializes `submessages`, this is not necessary here. I'll remove the redundant `setdefault()` and use the existing `submessages` list directly.

tests/ui_tools/test_messages.py
yummyPancake2607: Sure, I'll rename it to follow the naming convention used by the other keypress tests in this file.

tests/ui_tools/test_messages.py
yummyPancake2607: I created a separate `MessageBox` here because this test needs a specifically constructed message containing the todo widget data that is used to trigger the strike action. The existing `msg_box` fixture provides the standard message-box setup, but it doesn't provide the message structure required for this particular widget interaction.

I used the explicit `MessageBox(message, self.model, None)` construction so that the test controls the exact message being tested and verifies the submessage sent for that todo widget action.

zulipterminal/ui_tools/messages.py
yummyPancake2607: I added the digit check directly because the todo widget uses the number keys to identify the task being toggled. I agree that using `keys_for_command` is more consistent with the existing key handling in this method.

I'll move the submessage-specific check further down and use `keys_for_command` for the key detection, following the pattern of the other conditional below.

zulipterminal/ui_tools/messages.py
yummyPancake2607: Yes, this regex was only being used to extract the digit from the key. With the `keys_for_command` approach, that parsing is no longer necessary. I'll remove it and use the command key directly, following the existing keypress handling pattern.

zulipterminal/ui_tools/messages.py
yummyPancake2607: The current implementation only allows todo items to be selected through the `1`–`9` keys, so items beyond the ninth cannot currently be toggled using this interaction.

I agree that we should make this limitation visible rather than silently ignoring the additional items. I'll add a warning when the todo contains more than 9 items, while keeping the current `1`–`9` interaction unchanged for now.

zulipterminal/ui_tools/messages.py
yummyPancake2607: The `"t"` check was added because `t` is currently the key used for the todo action. I agree that this should go through `keys_for_command` instead, so that it follows the existing key-binding approach and doesn't depend on a hard-coded key. I'll update this to use the configured command key.

zulipterminal/ui_tools/messages.py
yummyPancake2607: I used `"area:msg"` explicitly here because the todo title editor is opened from the message area and I wanted the popup to be displayed there. I understand that the existing `show_pop_up` calls follow the current popup-handling design instead of specifying the area this way.

I'll update this to follow the same pattern as the other `show_pop_up` calls, while keeping the existing popup behavior unchanged.

zulipterminal/model.py
neiljp: You seem to have simplified this, even with the defense above.

It now looks like what I would expect, since one of these calls the other in the python api wrapper, and have been present for a long time. We use only one type of these elsewhere in the code, where a specific api method is not defined.

zulipterminal/model.py
neiljp: This looks simpler now, and good for now. However, I do agree with your observations:

I do agree that the API functions can raise exceptions, though we haven't observed that in practice so far. However, when we implement that, we should do so across all such calls, so it would be better to not include it here for one case. We'd want to limit the exceptions we handle if we can in any case. If you have ideas for implementing that, it'd be great to investigate, maybe with a draft PR and discussing in the channel :+1:

For the more-specific error reporting you had before for submessages, note how that varies from the general report-error function(s). That suggests we could implement the 'where the error came from' as part of all that error reporting, and reduce the need to handle what you had before in each operation. That would be fine in an earlier commit, or more simply in a separate PR.

zulipterminal/model.py
neiljp: See another comment. I do think we could include a little more information, depending what kinds of errors the server may respond with eg if they do include the type of api call that failed, it's less necessary. You may want to take a look in the server code for that.

zulipterminal/model.py
neiljp: I wasn't asking how you decided the function was a good thing to do or how, I was referring how you find the appropriate data to send to the server.

Thanks for adding the test, I'll look later.

neiljp:
I've replied to some points, and will do more later.

I thought you added the changes into one commit at the end, but it looks like you have added everything into one commit? Do you know how to safely split them up again? The Zulip git guide may have information, look online, but also ask on chat.zulip.org in a relevant channel if you need further help.

yummyPancake2607:
@neiljp Thanks for the feedback! I've addressed the comments:

* Simplified the API call to use the existing `call_endpoint()` pattern.
* Removed the broad `except Exception` and aligned error handling with existing API functions.
* Verified the data flow for the parameters sent to `/api/v1/submessage`.
* Added/updated the relevant tests.
* Split the changes into logical commits and removed unrelated changes.

All tests, Ruff, and mypy are passing.

yummyPancake2607:
@neiljp Kindly review it once. I have implemented what you have told me.

yummyPancake2607:
@neiljp can you review it once. I have implemented what you have told me

Issue #1582, opened 2025-04-16
Using `n` should move between topics by recency


neiljp:
When cycling through unread topics using `n`, this is not always sorted by the order in the topic list, ie. by most recent new message to a topic.

For discussion on related issues, see the topic in `# zulip-terminal` starting `Improve algorithm for 'next unread topic' (n) #T1333 #T1356`.

This behavior should be noticeable compared to when using the web app.

Using 'Explore' mode may be useful to ensure you don't keep marking messages lots of messages read during manual testing.

sahith-ch:
hey @neiljp can i work on this issue

Gopinath-Mahendiran:
@sahith-ch , are you working on this?

sahith-ch:
Yes am working on it

K-Preetham-Reddy:
@zulipbot claim

zulipbot:
Welcome to Zulip, @K-Preetham-Reddy! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-terminal/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

zulipbot:
@K-Preetham-Reddy You have been unassigned from this issue because you have not made any updates for over 14 days. Please feel free to reclaim the issue if you decide to pick up again. Thanks!

lavanyayadawad30-lab:
Hi, I’d like to work on this issue.
Please let me know if it’s okay to proceed.

yummyPancake2607:
I want to work on this issue. If no one is working please assign this issue to me

@zulipbot claim

zulipbot:
@yummyPancake2607 It looks like we've already sent you a collaboration invite at https://github.com/zulip/zulip-terminal/invitations, but you haven't accepted it yet. Please accept the invite, and then try to claim this issue again. Thanks!

yummyPancake2607:
@zulipbot claim

[...]

PR #1617, opened 2026-02-15
Improve ordering for next unread topic (n)

BhushanVidhate-27:
Issue : #1582 

Fixes the order when cycling through unread topics with n so it matches the topic list (most recent first), consistent with the web app.
Related: #T1333, #T1356.
### _Problem_

Pressing n (next unread topic) did not always follow the order of the topic list. Within a stream, unread topics were ordered alphabetically by topic name instead of by most recent activity. The web app (and the left-panel topic list) use recency order.

### sort_unread_topics 1

- Added an optional topic_order_by_stream argument.
- When provided, topics within each stream are ordered by their position in that list (index 0 = most recent).
- When omitted, behavior is unchanged (stream order + alphabetical by topic name).

### next_unread_topic_from_message_id 2

- Builds topic_order_by_stream for every stream that has unread topics (or contains the current topic when it’s not in the unread list) by calling topics_in_stream(stream_id).
- Passes this into sort_unread_topics so the sequence used for n is: left-panel stream order, then within each stream the same order as the topic list (most recent first).

### _Testing_

- Helper: New test test_sort_unread_topics_by_topic_list_order in checks that with topic_order_by_stream the result order follows the given topic list.
- Regression: Existing test_sort_unread_topics and model tests for next_unread_topic_from_message_id still pass (no topic_order_by_stream → same as before).
- Manual check: use Explore mode, have unread messages in several topics, and press n repeatedly; order should match the topic list and the web app.

### DIFF
--- a/tests/helper/test_helper.py
+++ b/tests/helper/test_helper.py
@@ -280,6 +280,23 @@ def test_sort_unread_topics(
     assert sort_unread_topics(unread_topics, stream_list) == expected_value
 
 
+def test_sort_unread_topics_by_topic_list_order() -> None:
+    """With topic_order_by_stream, order follows topic list (most recent first)."""
+    unread_topics = {
+        (1, "topic1"): 1,
+        (1, "topic2"): 1,
+        (1, "topic3"): 1,
+    }
+    stream_list = [1]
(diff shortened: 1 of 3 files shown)


Niloth-p:
Please read the contribution guide. Especially the commit discipline page.

PR #1618, opened 2026-02-20
tests: Add assertion for model.server_url in test_init

pushkarscripts:
Adds the missing assertion for `model.server_url` in `test_init`, resolving the existing `FIXME` comment.

Also extends coverage with a parametrized test covering different `base_url` scheme and `netloc` combinations to ensure correct normalization behavior.

- [x] Discussed in **#zulip-terminal** in `topic` [PR: Add test for model.server_url in test_init](https://chat.zulip.org/#narrow/channel/206-zulip-terminal/topic/PR.3A.20Add.20test.20for.20model.2Eserver_url.20in.20test_init/with/2393792)
- [ ] Fully fixes #
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -89,7 +89,7 @@ class TestModel:
         assert model.user_full_name == user_profile["full_name"]
         assert model.user_email == user_profile["email"]
         assert model.server_name == initial_data["realm_name"]
-        # FIXME Add test here for model.server_url
+        assert model.server_url == ":///"
         model._update_users_data_from_initial_data.assert_called_once_with()
         assert model.users == []
         self.classify_unread_counts.assert_called_once_with(model)
@@ -117,6 +117,28 @@ class TestModel:
         assert model.active_emoji_data["joker"]["type"] == "realm_emoji"
         # zulip_extra_emoji replaces all other emoji types for 'zulip' emoji.
         assert model.active_emoji_data["zulip"]["type"] == "zulip_extra_emoji"
+
+    @pytest.mark.parametrize(
+        "base_url, expected_server_url",
+        [
+            ("chat.zulip.zulip", ":///"),
+            ("https://chat.zulip.org", "https://chat.zulip.org/"),
+            ("http://localhost:9991", "http://localhost:9991/"),
+        ],
+    )
+    def test_server_url_generation(
+        self, mocker, initial_data, user_profile, base_url, expected_server_url
+    ):
+        self.client.base_url = base_url
+        mocker.patch(MODEL + ".get_messages", return_value="")
+        self.client.register.return_value = initial_data
+        self.client.get_profile.return_value = user_profile
+        mocker.patch(MODEL + "._update_users_data_from_initial_data")
+        mocker.patch(MODULE + ".classify_unread_counts")
+
+        model = Model(self.controller)
+
+        assert model.server_url == expected_server_url
 
     @pytest.mark.parametrize(
         "sptn, expected_sptn_value",


tests/model/test_model.py
```diff
@@ -89,7 +89,7 @@ def test_init(
         assert model.user_full_name == user_profile["full_name"]
         assert model.user_email == user_profile["email"]
         assert model.server_name == initial_data["realm_name"]
-        # FIXME Add test here for model.server_url
+        assert model.server_url == ":///"
```
clado, line 92: Looking at [the code that generates `server_url`](https://github.com/zulip/zulip-terminal/blob/6a799870eccc00d612e25ff881d18f4ff66d92fa/zulipterminal/model.py#L123-L125) it looks like you've covered the case where `scheme` and `netlock` are both `None`. What about if `scheme` and `netlock` assume different values?

(It sounds like the best route for this might involve a parameterized test + a bit more work, which is perhaps why the original author didn't feel that he could write the test in the time he had to merge the PR. :) )

pushkarscripts: > Looking at [the code that generates `server_url`](https://github.com/zulip/zulip-terminal/blob/6a799870eccc00d612e25ff881d18f4ff66d92fa/zulipterminal/model.py#L123-L125) it looks like you've covered the case where `scheme` and `netlock` are both `None`. What about if `scheme` and `netlock` assume different values?
> 
> (It sounds like the best route for this might involve a parameterized test + a bit more work, which is perhaps why the original author didn't feel that he could write the test in the time he had to merge the PR. :) )

Good point. 

I’ll extend the test to cover different `scheme` and `netloc` cases and update it with a parametrized version.

tests/model/test_model.py
```diff
@@ -118,6 +118,28 @@ def test_init(
...
+        [
+            ("chat.zulip.zulip", ":///"),
+            ("https://chat.zulip.org", "https://chat.zulip.org/"),
+            ("http://localhost:9991", "http://localhost:9991/"),
+        ],
+    )
+    def test_server_url_generation(
+        self, mocker, initial_data, user_profile, base_url, expected_server_url
+    ):
+        self.client.base_url = base_url
```
clado, line 132: Huh! This is not the code that I expected you to write! I really thought that the `mocker.patch('urllib.parse.urlparse')` was not going to pass through to urllib (shows you how much pytest I know lol).

I think sometimes in unit tests we do want to fully mock external libraries like these so that we're not testing their behavior while running our tests (otherwise then it would be an integration test?). However I might be misremembering and certainly don't feel confident making a definitive statement on how this codebase would prefer to proceed. :)

clado (approved):
lgtm! thanks for your work on improving it!! :)
maybe @neiljp you want to take a look, especially as the original writer of the `fixme`?

clado:
wait, quick addendum (I am "commit undisciplined" lol) but it may be a good idea to turn your two commits into one commit, since they're pretty much the same idea? (And, since I'd argue that it's only the second commit that truly resolved the `fixme` that was removed in the first commit.)

The Zulip documentation discussing [squashing commits](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html#squashing-commits) in this manner, but perhaps not in enough detail to follow if you've never done an interactive rebase before... Totally unsure whether that describes you or not, but want to encourage you to ask for help if you need it! :saluting_face:

pushkarscripts:
> wait, quick addendum (I am "commit undisciplined" lol) but it may be a good idea to turn your two commits into one commit, since they're pretty much the same idea? (And, since I'd argue that it's only the second commit that truly resolved the `fixme` that was removed in the first commit.)
> 
> The Zulip documentation discussing [squashing commits](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html#squashing-commits) in this manner, but perhaps not in enough detail to follow if you've never done an interactive rebase before... Totally unsure whether that describes you or not, but want to encourage you to ask for help if you need it! 🫡

Thanks for the suggestion. 
That makes sense since the second commit completes the fix. 

I’ll squash them into a single commit 👍

PR #1622, opened 2026-02-27
Always show recipient headers in search results

### LINKED ISSUE #1550: Search results should show recipient headers for every message

clado:

This PR updates the representation of messages in search results to more closely match the web UI.

Previously, search results in the terminal would omit recipient headers for messages in the same stream & topic. This is great for saving space, but presenting search results this way can sow confusion by silently dropping messages that may have provided important context in between the messages that match the search results.

The web UI solves this issue by always showing recipient messages when searching. This PR (hopefully) mimics that behavior. :)

Fixes #1550

### Outstanding aspect(s)
- [ ] question on approach when searching within topic narrow:

The original issue states:

> Note that it should not display the recipient part of the header in conversation views (specific DMs, topics), just like now, since the header information is in the part above that already.

I want to make sure that I'm understanding this correctly. As written, my understanding is that "in conversation views" excludes "search narrows for searching _within_ a conversation view", as these would presumably be search views, not conversation views. However, am I misunderstanding this? (Or is there an implicit "part of the header in [searches of] conversation views" phrase that is missing?)

Of note: given the space constraints on the terminal I personally think that it might be a good idea to remove the recipient header for searches that are very clearly constrained to a topic/DM (I have not yet taken the time to fully investigate the technical feasibility of this, though in a moment of doubt I partially implemented it before reverting to my current code change + understanding of the issue... :sweat_smile: ) I double checked what the web UI does, and it looks like the web UI does add the recipient header for all search results, [even when constrained to a single topic](https://chat.zulip.org/#narrow/channel/7-test-here/topic/arie/search/supercalifragilisticexpialidocious):

screenshot of zulip web ui showing a search for 'supercalifragilisticexpialidocious' in the arie topic of the #test here stream. all results have recipient headers

- [x] Fully fixes #1550 
- [ ] Partially fixes issue #
- [ ] Requires merge of PR #

### How did you test this?

- [x] It individually passes linting and tests (it looks like official ci only ran against the final commit, but I can confirm that running `pytest` against the interstitial commit locally passes without any new test failures)

### Visual changes
took a single before/after of a search constrained to a specific topic, as that seems the most germane to my outstanding question(s), but please ask for more/smaller screenshots & I'll generate them!!

| Before | After |
| --- | --- |
| before screenshot for a search for 'supercalifragilisticexpialidocious' in the arie topic of #test here. current message recipients are announced in the topic narrow then only once in the search results | after screenshot for a search for 'supercalifragilisticexpialodocious' in the arie topic of #test here. current message recipients are announced in the topic narrow then once for each search result |

[the same search results](https://chat.zulip.org/#narrow/channel/7-test-here/topic/arie/search/supercalifragilisticexpialidocious)  in the web ui

sincere gratitude to those who worked on #1552, which this PR is a continuation/redux of <3

### DIFF
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -903,34 +903,39 @@ class TestMessageBox:
         assert isinstance(view_components[0][2], Divider)
 
     @pytest.mark.parametrize(
-        "msg_narrow, msg_type, assert_header_bar, assert_search_bar",
+        "msg_narrow, searching, msg_type, assert_header_bar, assert_search_bar",
         [
             (
                 [],
+                False,
                 0,
                 f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR} ",
                 f" {ALL_MESSAGES_MARKER} All messages ",
             ),
             (
                 [],
+                False,
                 1,
                 f" {DIRECT_MESSAGE_MARKER} You and ",
                 f" {ALL_MESSAGES_MARKER} All messages ",
             ),
             (
                 [],
+                False,
                 2,
                 f" {DIRECT_MESSAGE_MARKER} You and ",
                 f" {ALL_MESSAGES_MARKER} All messages ",
             ),
             (
                 [["stream", "PTEST"]],
+                False,
                 0,
                 f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR} ",
                 ("bar", ("s#bd6", f" {STREAM_MARKER_PUBLIC} PTEST ")),
             ),
             (
                 [["stream", "PTEST"], ["topic", "b"]],
+                False,
                 0,
                 f" {STREAM_MARKER_PUBLIC} PTEST {STREAM_TOPIC_SEPARATOR}",
                 (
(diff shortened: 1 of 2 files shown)


clado:
I realized in reviewing #1618 that my assumption about what went wrong in the tests such that the first commit is needed was likely incorrect and that the commit message should likely be amended. That said, the patch as a whole remains necessary for the tests to pass.
Still waiting on (product) review, but as I said on the attached issue I will be unable to personally act on review feedback until April. :)

PR #1624, opened 2026-03-05
about_popup: Add current terminal size to Detected Environment section

### LINKED ISSUE #1536: Add current terminal window size to About popup
### Description of suggested feature

It would be useful during reporting issues to know the size of the current terminal.

This is already available at runtime via urwid, since the popup code relies on it to determine which width (and maximum width) to use.

This entry should appear at the bottom of the 'Detected Environment' section of the 'About' popup.

Since the terminal size can often be changed while a popup is open, we will likely want to call this 'Current terminal size', though this can be adjusted later during a PR or after merging.

kevinraj23:
### What does this PR do?

Adds the current terminal window size to the "Detected Environment"
section of the About popup.

The terminal dimensions are obtained using `self.loop.screen.get_cols_rows()` in `core.py` and displayed as
"{cols} x {rows}" (for example "80 x 24").

### How was this tested?

- Ran the existing test suite
- Updated tests in `tests/ui_tools/test_popups.py`
- Verified the value appears in the About popup

Fixes #1536

### DIFF
--- a/tests/ui_tools/test_popups.py
+++ b/tests/ui_tools/test_popups.py
@@ -170,9 +170,9 @@ class TestPopUpView:
         mocker.patch(
             MODULE + ".is_command_key",
             side_effect=(
-                lambda command, key: False
-                if command == self.command
-                else is_command_key(command, key)
+                lambda command, key: (
+                    False if command == self.command else is_command_key(command, key)
+                )
             ),
         )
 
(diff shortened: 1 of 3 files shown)


neiljp:
@kevinraj23 Thanks for working on this 👍 There are many PRs for this issue, I'm closing this in favor of another.

PR #1630, opened 2026-03-31
Add type hints to tests/ui_tools/test_messages.py

### LINKED ISSUE #1337: Add type hints to `test_messages.py`

khairien:
Adds type hints to all test functions and fixtures in 
tests/ui_tools/test_messages.py, as per requested in #1337.
Also removed the file from mypy exclusion list in pyproject.toml

- [x] Fully fixes #1337

### How did you test this?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -128,7 +128,6 @@ ignore_missing_imports = true
 module = [
     "tests.model.test_model",
     "tests.ui.test_ui_tools",
-    "tests.ui_tools.test_messages",
 ]
 disallow_untyped_defs = false
 disallow_incomplete_defs = false
--- a/tests/ui_tools/test_messages.py
+++ b/tests/ui_tools/test_messages.py
@@ -1,11 +1,13 @@
 from collections import OrderedDict, defaultdict
 from datetime import date
+from typing import Any
 from unittest.mock import patch
 
 import pytest
 import pytz
 from bs4 import BeautifulSoup
 from pytest import param as case
+from pytest_mock import MockerFixture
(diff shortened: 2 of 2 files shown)


khairien:
Hi @neiljp, I've addressed all CI checks and the PR is ready for review. Would appreciate any feedback!

PR #1633, opened 2026-04-10
views: Preserve focus and search on external pin events

### LINKED ISSUE #1487: UI bugs from external pinning/unpinning stream events
#411 (support stream pinning events) was effectively completed by #695, which also integrated existing stream muting events and updating into a UI in the stream popup, where they could be modified from ZT.

However, #411 was left open, to reflect ongoing discussion regarding outstanding issues.

One aspect of that was expanded into #747, to ensure that the stream popup data was synchronized with server data, if the latter changed while the popup was open.

The other was to ensure the behavior of the UI was reasonable when stream (pinning) events occurred externally to ZT, in particular when the focus is in the left panel, which is the focus of this issue.

### Apparent Bugs
These are all from the focus/selection starting in the stream or topic list, before then updating a pinning status from another client (web or elsewhere)
1. In stream list:
   - Move focus on a given stream, eg. 5 streams down from the top
   - **pinning event occurs**
   - Focus jumps to top of stream list
2. In topic list, with previous stream list focus:
   - Move focus on a given stream, eg. 5 streams down from the top
   - Use `t` to enter topic list
   - Move focus on an entry eg. 6 down in the topic list
   - **pinning event occurs**
   - Nothing happens in topic list (as expected)
   - Use `t` to return to stream list
   - focus in stream list is at the top, not 5 down
   - Moving down to the previous stream and using `t`, the topic focus is retained (as expected)
3. In stream list, during search:
   - Use `q` to start a stream search
   - type some text to search for a stream name
   - hit enter
   - focus is in limited list of streams
   - **pinning event occurs**
   - stream search is reset, focus is on top of stream list
4. Combination of 2 and 3: search for subset of streams, use `t`
   - **pinning event occurs**
   - Nothing happens in topic list (as expected)
   - Use `t` to return to stream list
   - stream search is reset
   - focus in stream list is at the top, not 5 down
   - Moving down to the previous stream and using `t`, the topic focus is retained (as expected)
5. In (stream) panel search box:
   - **pinning event occurs**
   - Traceback overlaid on screen:
    ```
    line 346 in update_streams
    with self.search_lock:
    AttributeError: 'StreamsView' object has no attribute 'search_lock'
    ```
   - Search is no longer usable without a restart

In all cases, the expected behavior is:
- the stream list updates to adjust the (un)pinned stream position
- focus remains on the same item
- stream search is not interrupted
- search doesn't cause a traceback :)

It's possible the last issue will need more adjusting, and similar issues may be in the user list or elsewhere.

maciej103:

When a stream pin/unpin event arrives from an external client, 
update_stream_view() rebuilt StreamsView from scratch. This reset 
focus to the top of the list and cleared any active search. If a 
pin event occurred while the search box was active, it also caused 
an AttributeError due to search_lock not yet being initialized.

This fixes the bugs in #1487 by saving and restoring focus position 
and search text around the rebuild, and by moving search_lock 
initialization earlier in StreamsView.__init__().

- [x] Fully fixes #1487

### How did you test this?

### DIFF
--- a/tests/ui/test_ui_tools.py
+++ b/tests/ui/test_ui_tools.py
@@ -1221,6 +1221,77 @@ class TestLeftColumnView:
             ]
         )
 
+    def test_update_stream_view_restores_focus_by_stream_id(self, mocker):
+        mocker.patch(VIEWS + ".LeftColumnView.streams_view")
+        mocker.patch(VIEWS + ".LeftColumnView.menu_view")
+        mocker.patch(VIEWS + ".LeftColumnView.show_stream_view")
+        left_col_view = LeftColumnView(self.view)
+        left_col_view.is_in_topic_view = False
+
+        # Simulate focus on a stream widget with stream_id=42 at index 3.
+        focused_widget = mocker.Mock(stream_id=42)
+        self.view.stream_w = mocker.Mock()
+        self.view.stream_w.log.get_focus.return_value = (focused_widget, 3)
+        self.view.stream_w.stream_search_box.edit_text = ""
+
+        # After rebuild, stream_id=42 moved to index 1 (e.g. another stream
+        # was pinned and pushed to the top).
+        new_widgets = [
+            mocker.Mock(stream_id=99),
+            mocker.Mock(stream_id=42),
+            mocker.Mock(stream_id=7),
+        ]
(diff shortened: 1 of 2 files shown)


maciej103:
Hi! I'm a new contributor. All 22 CI checks are passing and I've verified the fix manually against chat.zulip.org. Would appreciate any feedback or review when you get a chance.

PR #1635, opened 2026-04-11
model: Use "direct" type for DMs on servers at feature level 174+

### LINKED ISSUE #1388: Use "direct" as the message `type` when sending direct/private messages.
Zulip servers (at API feature level 174) support "direct" as the preferred message `type` for direct/private messages when sending messages and sending typing notifications; see https://github.com/zulip/zulip/pull/25154. 

For servers at this feature level, when sending direct messages ([`POST /messages`](https://zulip.com/api/send-message#parameter-type)), Zulip terminal should switch to using the new preferred "direct" value for direct messages.

For typing notifications, the default value was switched from "private" to "direct", so no changes need to be made by clients if the default is being used for direct/private message typing notifications.

See [this #api design conversation](https://chat.zulip.org/#narrow/stream/378-api-design/topic/message.20type.20parameter.2C.20.2324960/near/1549164) for more context and discussion on the `type` parameter.

From that chat.zulip.org conversation related specifically to Zulip terminal:

> @_**Neil Pilgrim (neiljp)|2406** [said](https://chat.zulip.org/#narrow/stream/378-api-design/topic/message.20type.20parameter.2C.20.2324960/near/1552522):
>
> From a ZT perspective, the post and get valid strings could be left different, or the same and loosened in advance. However, I'm thinking it may be cleaner to have a type which encompasses both, so it may make sense to hold off until the capability is introduced, given 7.0 will be backwards-compatible.

IlijevskiM:
Zulip servers at API feature level 174 (Zulip 7.0) introduced "direct" as the preferred value for the `type` parameter when sending direct messages via POST /messages. Older servers require "private" and do not accept "direct".

This change:
- Widens `PrivateComposition.type` to `Literal["private", "direct"]`
- Adds `DIRECT_MESSAGE_TYPE_FEATURE_LEVEL = 174` in `model.py`
- Gates the type value in `send_private_message()` on the feature level
- Parametrizes the existing test to cover both code paths

Fixes #1388

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -12,6 +12,7 @@ from zulip import Client, ZulipError
 from zulipterminal.config.symbols import STREAM_TOPIC_SEPARATOR
 from zulipterminal.helper import initial_index, powerset
 from zulipterminal.model import (
+    DIRECT_MESSAGE_TYPE_FEATURE_LEVEL,
     MAX_MESSAGE_LENGTH,
     MAX_STREAM_NAME_LENGTH,
     MAX_TOPIC_NAME_LENGTH,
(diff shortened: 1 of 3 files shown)


emilysutherlin:
@Niloth-p do you want to take a look?

PR #1648, opened 2026-07-10
compose: Limit stream box width to max stream name length

### LINKED ISSUE #905: Minimize length of stream box during compose based on available space & maximum stream length

harrish243:
Fixes #905
The Previous One:
Screenshot 2026-07-09 094651

Previously, the stream input widget in the compose header used urwid's default weight-based sizing, causing it to take up half of the available terminal width regardless of stream name lengths.

The Updated One :
Screenshot 2026-07-09 150724

Now we calculate the maximum stream name length from the user's subscriptions and allocate only that width (plus 1 for cursor padding) to the stream box, allowing the topic field to use all remaining space.

### DIFF
--- a/tests/ui_tools/test_boxes.py
+++ b/tests/ui_tools/test_boxes.py
@@ -53,6 +53,7 @@ class TestWriteBox:
         users_fixture: List[Dict[str, Any]],
         user_groups_fixture: List[Dict[str, Any]],
         streams_fixture: List[Dict[str, Any]],
+        stream_dict: Dict[int, Dict[str, Any]],
         unicode_emojis: "OrderedDict[str, Dict[str, Any]]",
         user_dict: Dict[str, MinimalUserData],
     ) -> WriteBox:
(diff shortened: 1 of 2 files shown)


neiljp:
@harrish243 Thanks for the contribution!

The functionality looks reasonable, but catching the TypeError in the code to fix a test is not appropriate. The test should be updated instead.

harrish243:
@neiljp As you have told i have updated

neiljp:
Thanks @harrish243, this looks good :+1:

harrish243:
Thanks @neiljp ! Appreciate the feedback and guidance throughout the PR! 🙌

PR #1654, opened 2026-08-13
Avoid ValueError when peer_remove references unknown subscriber.

### LINKED ISSUE #1220: Event traceback from peer_remove event
I've only seen this once, so this is likely an uncommon situation. This caused the popup which occurs when an event handler function doesn't handle all exceptions, from an error on line 1241 of models.py.

We likely want to check that a user removed from our known users in a stream is present before removing them, or else handle the exception instead - basically limit the exception from leaking out and giving the popup which occurs in these situations.

This is very likely a small change, but shouldn't really occur unless we miss a peer_add event for some reason, or else there is some timing issue, though it is possible there is something else going on. For that reason we likely want a comment in the code. It may also be worth exploring if the web app or mobile modifies this data in another situation which we may be missing.

Due to the challenge of debugging this behavior, we definitely want to add/extend tests for this case.

Mo200227:
Fixes #1220

The `peer_remove` event handler assumed the user being removed was
always present in the stream's subscriber list. If the corresponding
`peer_add` event was missed for some reason, this caused a ValueError
and the exception popup described in the issue.

This adds a membership check before removing, and a test covering
the case where the user_id is not present in subscribers.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -4187,6 +4187,27 @@ class TestModel:
         new_subscribers = model.stream_dict[stream_ids[0]]["subscribers"]
         assert new_subscribers == expected_subscribers
 
+    def test__handle_subscription_event_peer_remove_user_not_subscribed(
+        self, model, mocker, stream_dict
+    ):
+        # Regression test for #1220: peer_remove for a user_id that is
+        # not present in subscribers (e.g. because we missed the
+        # corresponding peer_add event) should not raise ValueError.
+        event = {
+            "type": "subscription",
+            "op": "peer_remove",
+            "stream_ids": [2],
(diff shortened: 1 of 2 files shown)


neiljp:
@Mo200227 Thanks for this contribution :+1:

I'm going to merge this as it is, since it fits with the current surrounding code.

In future, we could consider various changes in multiple tests, though other internal refactors may be worthwhile considering first. Examples could be removing any unnecessary use of mocker, add asserts to explicitly confirm assumptions (eg. 999 not being present in subscribers here), and ultimately remove the use of fixed numbers in the tests instead of relying upon the default data from elsewhere in test fixtures. The latter we've done using test parameters with default values, but a fixture approach may be also be relevant.

